PR #13, opened 2025-11-07
feat(cloudflare): Worker, Assets, R2, KV

john-royal:


alchemy-effect/src/cloudflare/context.ts
```diff
@@ -0,0 +1,56 @@
...
+import * as Layer from "effect/Layer";
+
+export class CloudflareContext extends Context.Tag("Cloudflare.Context")<
+  CloudflareContext,
+  {
+    env: unknown;
+    ctx: ExecutionContext;
+  }
+>() {
+  static readonly getEnvKey = Effect.fnUntraced(function* <T>(key: string) {
```
sam-goodwin: i don't think we should ever have methods because of tree-shaking

alchemy-effect/src/cloudflare/index.ts
```diff
@@ -1,4 +1,38 @@
...
+import { CloudflareAccountId, CloudflareApi } from "./api.ts";
+import { assetsProvider } from "./assets.provider.ts";
+import * as KV from "./kv/index.ts";
+import * as R2 from "./r2/index.ts";
+import * as Worker from "./worker/index.ts";
+
+export * as Assets from "./assets.fetch.ts";
+export * as KV from "./kv/index.ts";
+export * as R2 from "./r2/index.ts";
+export * as Worker from "./worker/index.ts";
```
sam-goodwin: these won't tree-shake?

PR #15, opened 2025-11-12
feat(test): add test utility

sam-goodwin:
```ts
test(
  "create, update, delete table",
  Effect.gen(function* () {
    const dynamodb = yield* DynamoDB.DynamoDBClient;

    class Table extends DynamoDB.Table("Table", {
      tableName: "test",
      items: type<{ id: string }>,
      attributes: {
        id: S.String,
      },
      partitionKey: "id",
    }) {}

    const stack = yield* apply(Table);

    const actualTable = yield* dynamodb.describeTable({
      TableName: stack.Table.tableName,
    });
    expect(actualTable.Table?.TableArn).toEqual(stack.Table.tableArn);

    yield* destroy();
  }).pipe(Effect.provide(AWS.live)),
);
```

### DIFF
--- a/alchemy-effect/src/apply.ts
+++ b/alchemy-effect/src/apply.ts
@@ -1,4 +1,3 @@
-import * as Console from "effect/Console";
 import * as Context from "effect/Context";
 import * as Effect from "effect/Effect";
 import * as Option from "effect/Option";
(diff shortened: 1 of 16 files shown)


alchemy-effect/src/apply.ts
```diff
@@ -267,7 +297,14 @@ export const apply = <P extends Plan, Err, Req>(
                     news: node.news,
                     bindings: bindingOutputs,
                     session: scopedSession,
+                    ...(node.action === "update"
+                      ? {
+                          output: node.output,
+                          olds: node.olds,
+                        }
```
sam-goodwin: @john-royal i fixed the update missing output/olds here

PR #48, opened 2026-04-12
fix(Http): lazy-import platform-node to avoid crash on Bun

yovanoc:
## Problem

`Http.ts` has static top-level imports of **both** `@effect/platform-bun` and `@effect/platform-node`:

```ts
import * as BunHttpServerPlatform from "@effect/platform-bun/BunHttpServer";
import * as NodeHttpServerPlatform from "@effect/platform-node/NodeHttpServer";
```

Because `Http.ts` is re-exported from the package barrel (`index.ts` → `export * as Http from "./Http.ts"`), **every import from `alchemy-effect` crashes at module-load time** on whichever runtime doesn't have its peer installed:

- On **Bun** (no `@effect/platform-node`): `Cannot find module '@effect/platform-node/NodeHttpServer'`
- On **Node** (no `@effect/platform-bun`): `Cannot find module '@effect/platform-bun/BunHttpServer'`

Fixed by using a dynamic import in the Layer


alchemy-effect/src/Cloudflare/Container.ts
```diff
@@ -746,12 +746,12 @@ ${
   runtime === "bun"
     ? `
 import { BunServices } from "@effect/platform-bun";
-import { BunHttpServer } from "alchemy-effect/Http";
+import { BunHttpServer } from "alchemy-effect/HttpBun";
```
sam-goodwin: I think we need to update `exports` in `package.json`?

sam-goodwin:
Thanks for catching this bug. I am thinking we could also just do a dynamic import instead. That way we avoid additional exports `HttpBun` and `HttpNode`? Or we could make them `Http/Bun` and `Http/Node`?

PR #50, opened 2026-04-12
Discriminate on redacted

czxtm:


alchemy-effect/src/Cloudflare/Workers/Worker.ts
```diff
@@ -998,12 +998,19 @@ ${[
         if (news.env) {
           for (const [key, value] of Object.entries(news.env)) {
```
sam-goodwin, line 999: Do we need to update the `env` type in WorkerProps to allow `string | Redacted<string>`?

czxtm: addressed in other PR

PR #57, opened 2026-04-14
feat(cloudflare): add Astro resource

Makisuo:
## Summary
- New `Cloudflare.Astro` resource that composes `Command` + `Worker`, mirroring the `StaticSite` pattern with Astro-specific defaults (`bun astro build`, `outdir: "dist"`, auto-detected `output` mode from `astro.config.*`).
- For `output: "server"`, points the Worker entrypoint at the `@astrojs/cloudflare` adapter output (`dist/_worker.js/index.js`) and augments the build command to write a `.assetsignore` excluding `_worker.js`, `_worker.js/**`, `_worker.js.map`, and `_routes.json`, so the server bundle isn't re-uploaded as public static assets.
- For `output: "static"`, generates a no-op placeholder Worker at `.alchemy/local/worker.js` (matches the legacy `Website` behavior) since `WorkerProps.main` is required.
- Re-exported from `Cloudflare/Website/index.ts`, so it's reachable as `Cloudflare.Astro`.

## Non-goals
- No port of the old `plugin.ts` (Astro-side `@astrojs/cloudflare` runtime override).
- No package-manager-runner auto-detection; defaults to `bun astro build`, override via `command`.
- No dedicated tests in this PR — `StaticSite` / `Vite` also ship without them and a real Astro test would need a fixture project plus Cloudflare credentials.

## Test plan
- [x] `bun tsc -b` passes
- [x] `bun run generate:docs` produces `alchemy-effect/docs/Cloudflare/Website/Astro.md`
- [x] Generated `.assetsignore` shell command verified against a temp directory
- [ ] Manual smoke test: switch `website/alchemy.run.ts` to `Cloudflare.Astro(...)` and deploy

### DIFF
--- a/alchemy-effect/src/Cloudflare/Website/Astro.ts
+++ b/alchemy-effect/src/Cloudflare/Website/Astro.ts
@@ -1,0 +1,217 @@
+import * as NodeFs from "node:fs";
+import * as NodePath from "node:path";
+import * as Effect from "effect/Effect";
+import { Command, type CommandProps } from "../../Build/Command.ts";
+import type { InputProps } from "../../Input.ts";
+import * as Namespace from "../../Namespace.ts";
+import type { AssetsConfig } from "../Workers/Assets.ts";
+import {
+  Worker,
+  type WorkerAssetsConfig,
+  type WorkerBindingProps,
+  type WorkerProps,
+} from "../Workers/Worker.ts";
+
+export interface AstroProps<Bindings extends WorkerBindingProps = {}>
+  extends
+    Omit<WorkerProps<Bindings, WorkerAssetsConfig>, "assets" | "main">,
+    Omit<CommandProps, "command" | "outdir" | "env"> {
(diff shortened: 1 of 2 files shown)


Mkassabov:
Just a heads up this might take us a while to get to.
We really really really don't want to do framework specific resource in alchemy-effect because they were kind of a pain in async-alchemy.

In an ideal world everything would go through the vite resource; the problem with that is certain vite-based frameworks (e.g. Nuxt, Astro) wrap vite which causes issues with our new vite resource.

We are still trying to figure out if we want Command + Worker or if we want to modify vite or if we want to do something seperate.

Nothing actionable on your end; just want to make sure you don't think we're ignoring this. We appreciate the contribution 🤍!

Makisuo:
Ah makes sense, yeah totally get you. Will just close this for now then and leave those kind of providers, if I use them in my own projects!

PR #59, opened 2026-04-15
fix: normalize VpcEndpoint.State casing so wait helpers terminate

marbemac:
AWS returns VpcEndpoint.State in lowercase (e.g., "available", "deleted", ...), but @distilled.cloud/aws/ec2 types it as PascalCase. As a result, equality checks in waitForVpcEndpointAvailable and waitForVpcEndpointDeleted never matched, causing create/delete operations to loop until the 3-minute retry ceiling and fail with a timeout—even though the endpoint was ready seconds after the API call.

### DIFF
--- a/packages/alchemy/src/AWS/EC2/VpcEndpoint.ts
+++ b/packages/alchemy/src/AWS/EC2/VpcEndpoint.ts
@@ -595,11 +595,15 @@ const waitForVpcEndpointAvailable = (
       return yield* new VpcEndpointNotFound({ vpcEndpointId });
     }
 
-    if (ep.State === "Available") {
+    // AWS returns State in lowercase on the wire ("available", "failed", …)
+    // even though the SDK types it as PascalCase. Normalize before compare.
+    const state = ep.State?.toLowerCase();
+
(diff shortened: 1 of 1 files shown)


Mkassabov:
we should change the distilled SDK to return the correct types instead of calling to lowercase here.

Mkassabov:
kicked off a release for distilled with this fix:
https://github.com/alchemy-run/distilled/actions/runs/24534217774

Mkassabov:
closing in favor of a distilled based fix here: https://github.com/alchemy-run/alchemy-effect/pull/63

thanks for pointing out the issue and that something was wrong with the SDK!

Mkassabov:
@marbemac added you as a co-author on https://github.com/alchemy-run/alchemy-effect/commit/f1fd20b7c8755629640781abb326f21076766cbf btw, since its essentially just your change

marbemac:
thanks for the fast turnaround!

Mkassabov:
> thanks for the fast turnaround!

np! its in the latest release (`bun i alchemy@2.0.0-beta.4`) if you want to use it.

PR #60, opened 2026-04-15
feat: S3 state backend with user-supplied Stack state override

marbemac:
Tested with our setup and seems to work well!


packages/alchemy/src/State/S3State.ts
```diff
@@ -0,0 +1,378 @@
...
+
+const retryTransient = <A, E, R>(effect: Effect.Effect<A, E, R>) =>
+  effect.pipe(
+    Effect.retry({
+      while: (e) => TRANSIENT_TAGS.has((e as any)?._tag ?? ""),
+      schedule: Schedule.exponential(100).pipe(
+        Schedule.both(Schedule.recurs(5)),
+      ),
+    }),
+  );
```
sam-goodwin: I am pretty sure this is handled by distilled's default retry policy (unless something is broken there )

https://github.com/alchemy-run/distilled/blob/main/packages/aws/src/retry.ts#L81-L111

packages/alchemy/src/State/S3State.ts
```diff
@@ -0,0 +1,378 @@
...
+
+const bodyToString = (body: S3.GetObjectOutput["Body"]) =>
+  body === undefined
+    ? Effect.succeed("")
+    : (body as Stream.Stream<Uint8Array, Error, never>).pipe(
+        Stream.decodeText({ encoding: "utf-8" }),
+        Stream.runFold(
+          () => "",
+          (acc: string, chunk: string) => acc + chunk,
+        ),
```
sam-goodwin: `Stream.mkString` should do the same?

packages/alchemy/src/State/S3State.ts
```diff
@@ -0,0 +1,378 @@
...
+              Prefix: fullPrefix,
+              Delimiter: "/",
+              ContinuationToken,
+            }).pipe(retryTransient);
+            for (const cp of res.CommonPrefixes ?? []) {
+              const name = stripCommonPrefix(fullPrefix, cp.Prefix);
+              if (name) names.add(name);
+            }
+            ContinuationToken = res.NextContinuationToken;
+          } while (ContinuationToken);
```
sam-goodwin: You can use distilled's pagination streams
```ts
yield* listObjectsV2.pages({
  Bucket: bucketName,
  ..
)
```
This will give you a `Stream` that auto paginates lazily

sam-goodwin (changes requested):
Thanks for putting this together! Looks like it could be simplified a bunch by leaning more on distilled's internals.

marbemac:
On it!

marbemac:
@sam-goodwin done - caveat I haven't run the tests because I don't have a localstack account.

packages/alchemy/src/State/S3State.ts
sam-goodwin, line 1: I think this file belongs in `AWS/S3`

```ts
AWS.S3.state({ .. })
```

packages/alchemy/src/State/S3State.ts
```diff
@@ -0,0 +1,358 @@
...
+          id: v.LogicalId,
+          type: v.Type,
+          props: v.Props,
+          attr: v.Attributes,
+        };
+      }
+      return v;
+    },
+    2,
+  );
```
sam-goodwin: This already exists in the local file system state. We should pull that out into a re-usable State/Serde.ts file and share it

packages/alchemy/src/State/S3State.ts
```diff
@@ -0,0 +1,358 @@
...
+  });
+
+/**
+ * If the S3 prefix is empty but the local `.alchemy/state/` directory is
+ * not, log a one-shot warning suggesting the user may want to lift their
+ * existing state before deploying. All services are picked up via
+ * `serviceOption` so the check silently no-ops in environments where
+ * `DotAlchemy`, `FileSystem`, or `Path` aren't in scope (e.g. tests).
+ */
+const warnOnStaleLocalState = (bucketName: string, prefix: string) =>
```
sam-goodwin: Do we need this? It is perfectly valid to have multiple stacks in the same project using different state stores.

packages/alchemy/src/State/S3State.ts
```diff
@@ -0,0 +1,358 @@
...
+ *   "my-stack",
+ *   {
+ *     providers: AWS.providers(),
+ *     state: S3State({ bucketName: "my-infra-state" }),
+ *   },
+ *   Effect.gen(function* () { ... }),
+ * );
+ * ```
+ */
+export const S3State = (
```
sam-goodwin: You should look at src/AWS/Bootstrap.ts and bin/alchemy.ts#bootstrapCommand

It creates a S3 bucket and tags it. We do a search via tags to see if a state bucket already exists and use it if it does. I think the bucket should be auto-created by default: `AWS.S3.state()` will create a bucket per AWS account (and maybe per AWS region?)

sam-goodwin (changes requested):

marbemac:
Alrighty @sam-goodwin I think I addressed most of it, lmk! There is some duplication of logic now re tagged-bucket discover-or-create patterns in bootstrap.ts and state.ts - I left it alone so-as not to touch bootstrap.ts, but I could refactor that if you like.

marbemac:
Also, still haven't/can't run the tests so I have no idea if they actually pass :).

Mkassabov:
closing this cause its quite outdated. We weren't in a good place to deal with it and it got neglected, we'll make our own AWS state store soon. Or feel free to open an updated PR

PR #64, opened 2026-04-17
feat(cloudflare): secret store resources

Mkassabov:


examples/cloudflare-secrets-store/src/Api.ts
```diff
@@ -0,0 +1,42 @@
...
+        if (request.url === "/secret") {
+          const value = yield* secret
+            .get()
+            .pipe(
+              Effect.catchTag("SecretError", (err) =>
+                Effect.succeed(`Error: ${err.message}`),
+              ),
+            );
+          const masked = value.slice(0, 4) + "****";
+          return HttpServerResponse.text(`Secret (masked): ${masked}`);
```
sam-goodwin: You're gonna return the error message as if it's the secret

packages/alchemy/src/Cloudflare/SecretsStore/Secret.ts
```diff
@@ -0,0 +1,225 @@
...
+        stables: ["secretId", "secretName", "storeId", "accountId"],
+        diff: Effect.fn(function* ({ id, olds = {} as any, news, output }) {
+          if (!isResolved(news)) return undefined;
+          const oldStoreId = output?.storeId ?? olds.store?.storeId;
+          const newStoreId = news.store.storeId;
+          const oldName = output?.secretName ?? resolveName(id, olds.name);
+          const newName = resolveName(id, news.name);
+          if (oldStoreId !== newStoreId || oldName !== newName) {
+            return { action: "replace" } as const;
+          }
```
sam-goodwin, line 115: The name can't be updated?

packages/alchemy/src/Cloudflare/SecretsStore/Secret.ts
```diff
@@ -0,0 +1,225 @@
...
+          const name = resolveName(id, news.name);
+          const scopes = resolveScopes(news.scopes);
+          const response = yield* createStoreSecret({
+            accountId: news.store.accountId,
+            storeId: news.store.storeId,
+            body: [
+              {
+                name,
+                scopes,
+                value: Redacted.value(news.value),
```
sam-goodwin, line 132: Should the `value` property be `Redacted<string>` so the user is required to pass a Redacted. We can add additional envelope encryption for Redacted values in the state store

packages/alchemy/src/Cloudflare/SecretsStore/Secret.ts
Mkassabov: its not in the api. so I assume no

packages/alchemy/src/Cloudflare/SecretsStore/Secret.ts
Mkassabov: it is, and I think its best it stay that way. most of the time you put something in the secret store I imagine it is a secret.

```
  /**
   * The secret value. Treated as redacted and never logged.
   */
  value: Redacted.Redacted<string>;
```
the example uses Random.text which is `Output<Redacted<string>, never>`

examples/cloudflare-secrets-store/src/Api.ts
Mkassabov: fixed; returns a 500 code now

examples/cloudflare-secrets-store/src/Api.ts
```diff
@@ -0,0 +1,47 @@
+import * as Cloudflare from "alchemy/Cloudflare";
+import * as Effect from "effect/Effect";
+import * as Layer from "effect/Layer";
+import { HttpServerRequest } from "effect/unstable/http/HttpServerRequest";
+import * as HttpServerResponse from "effect/unstable/http/HttpServerResponse";
+import { ApiKey } from "./Secret.ts";
```
sam-goodwin: File should be ApiKey.ts

examples/cloudflare-secrets-store/src/Api.ts
```diff
@@ -0,0 +1,47 @@
...
+
+export default class Api extends Cloudflare.Worker<Api>()(
+  "Api",
+  {
+    main: import.meta.path,
+    compatibility: {
+      flags: ["nodejs_compat"],
+      date: "2026-03-17",
+    },
```
sam-goodwin: This can be removed

examples/cloudflare-secrets-store/src/Api.ts
```diff
@@ -0,0 +1,47 @@
...
+            ),
+          );
+        }
+
+        return HttpServerResponse.text(
+          "Hello from Cloudflare Secrets Store example!",
+        );
+      }),
+    };
+  }).pipe(Effect.provide(Layer.mergeAll(Cloudflare.SecretBindingLive))),
```
sam-goodwin: Merge all is not needed

packages/alchemy/src/Cloudflare/SecretsStore/Secret.ts
```diff
@@ -0,0 +1,225 @@
...
+   */
+  scopes?: string[];
+  /**
+   * Optional free-form description.
+   */
+  comment?: string;
+};
+
+export type StoreSecret = Resource<
+  "Cloudflare.SecretsStore.Secret",
```
sam-goodwin: Should this be called `Secret`?

packages/alchemy/src/Cloudflare/SecretsStore/Secret.ts
sam-goodwin: Oh I got it mixed up. Redacted.value extracts the value. Good. That's right.

packages/alchemy/src/Cloudflare/SecretsStore/Secret.ts
```diff
@@ -0,0 +1,225 @@
...
+          }).pipe(
+            Effect.catchTag("StoreNotFound", () =>
+              Effect.succeed({
+                result: [],
+                resultInfo: {},
+              } as any),
+            ),
+          );
+          const match = listed.result.find((s: any) => s.name === name);
+          if (!match) return undefined;
```
sam-goodwin: Can this paginate

packages/alchemy/src/Cloudflare/SecretsStore/SecretBinding.ts
```diff
@@ -0,0 +1,91 @@
...
+export interface SecretClient {
+  /**
+   * Effect that resolves to the raw Cloudflare `SecretsStoreSecret` binding.
+   */
+  raw: Effect.Effect<runtime.SecretsStoreSecret, never, WorkerEnvironment>;
+  /**
+   * Read the current value of the secret.
+   */
+  get(): Effect.Effect<string, SecretError, WorkerEnvironment>;
+}
```
sam-goodwin: Maybe make it an Effect so you can just `yield* apiKey` instead of `yield* apiKey.get()`? Support both.

packages/alchemy/src/Cloudflare/SecretsStore/SecretsStore.ts
```diff
@@ -0,0 +1,127 @@
...
+ * exists it is treated as account-level infrastructure that outlives
+ * any single stack.
+ *
+ * @section Creating a Store
+ * @example Basic Secrets Store (adopts existing or creates one)
+ * ```typescript
+ * const store = yield* Cloudflare.SecretsStore("MyStore");
+ * ```
+ *
+ * @example Adopt a specific named store
```
sam-goodwin, line 38: "Adopt"? What does that mean

packages/alchemy/src/Cloudflare/SecretsStore/SecretsStore.ts
```diff
@@ -0,0 +1,127 @@
...
+        }),
+        update: Effect.fn(function* ({ output }) {
+          return output;
+        }),
+        delete: Effect.fn(function* () {
+          // Intentional no-op. Cloudflare only allows one Secrets Store per
+          // account and deleting it permanently destroys all secrets inside.
+          // The store is treated as shared, account-level infrastructure that
+          // should never be torn down by a single stack.
+        }),
```
sam-goodwin, line 90: This looks bad. Is this copied from async alchemy? If so, is it still true

sam-goodwin (changes requested):

packages/alchemy/src/Cloudflare/SecretsStore/SecretsStore.ts
Mkassabov: This is true I tried it. It's horrific

Mkassabov: Like the comment is written by Claude but deleting the secret store deletes your secrets and makes a new one with a new id

examples/cloudflare-secrets-store/src/Api.ts
Mkassabov: you are wrong; both the flag and the date are needed without it they error.

sam-goodwin: Are you on latest?

packages/alchemy/src/Cloudflare/SecretsStore/Secret.ts
Mkassabov: not worth it, will always be length 1.

examples/cloudflare-secrets-store/src/Api.ts
sam-goodwin: https://github.com/alchemy-run/alchemy-effect/commit/deda381d8519b1a60ad010be224a8b3ba8e41e44
https://github.com/alchemy-run/alchemy-effect/commit/ff780f3187a74be99cc50d2a51b60f7bd1759ed7

packages/alchemy/src/Cloudflare/SecretsStore/SecretsStore.ts
Mkassabov: ehhh its some schizo shit RE trying to adopt the existing store but since its only 1 we can just grab it. figured I'd leave it in case they ever mademultiple secret stores supported, but for now its just clutter. removed it.

packages/alchemy/src/Cloudflare/SecretsStore/Secret.ts
sam-goodwin: because max secrets is `100`? It takes no effort to paginate here. Stream.find. If they ever expand beyond 100 secrets, then our code will continue to work. If the API supports pagination, we should paginate

packages/alchemy/src/Cloudflare/SecretsStore/SecretsStore.ts
sam-goodwin: Is it even worth having a SecretsStore resource if we're only allowed one and it's created automatically?

packages/alchemy/src/Cloudflare/SecretsStore/Secret.ts
Mkassabov: oh I thought this was secret store, yeah will paginate; am tired

packages/alchemy/src/Cloudflare/SecretsStore/SecretsStore.ts
Mkassabov: yes because if somebody else removes it, it will have a different ID and you can just know your secrets are gone

packages/alchemy/src/Cloudflare/SecretsStore/Secret.ts
sam-goodwin: sleep!

packages/alchemy/src/Cloudflare/SecretsStore/Secret.ts
Mkassabov: `packages\alchemy\src\Cloudflare\Container\ContainerApplication.ts` exports 
```ts
export type Secret = NonNullable<
  Containers.CreateContainerApplicationRequest["configuration"]["secrets"]
>[number];
```

so the types would collide. Both distilled and worker types call this a secret store secret, so we should properly keep it as is.

but also `ContainerApplication` is horrific; it exports so many types that are incredibly generic and so easy to collide with IMO.

examples/cloudflare-secrets-store/src/Api.ts
Mkassabov: didn't realize this branch was that old; you're right

packages/alchemy/src/Cloudflare/SecretsStore/SecretBinding.ts
Mkassabov: done

Mkassabov: not sure the best way of doing this to support both yield* and yield* .get; what cluade did works; is there a better example of how to do this?

packages/alchemy/src/Cloudflare/SecretsStore/Secret.ts
sam-goodwin: Namespace the container application ones

PR #69, opened 2026-04-18
feat(Cloudflare): add Queue, QueueBinding, and QueueConsumer resources

yovanoc:
## Summary

Adds complete Cloudflare Queue support with three new resources:

- **`Queue`** -- Create/update/delete queues via the Cloudflare API
- **`QueueBinding`** -- Bind a queue as a producer to a Worker (`send`/`sendBatch`)
- **`QueueConsumer`** -- Register a Worker as a consumer of a queue

All three follow the existing Resource pattern (`R2Bucket`, `KVNamespace`, `D1Database`) with full lifecycle management (create/update/delete/read/diff).

## Files

| File | Description |
|------|-------------|
| `Queue/Queue.ts` | Queue resource with provider (create/update/delete/read/diff) |
| `Queue/QueueBinding.ts` | Producer binding with `QueueSender` interface |
| `Queue/QueueConsumer.ts` | Consumer resource -- registers a Worker as queue consumer |
| `Queue/index.ts` | Barrel export |
| `Cloudflare/index.ts` | Added Queue export |
| `Cloudflare/Providers.ts` | Registered Queue, QueueBinding, QueueConsumer providers |

## Usage

```typescript
import * as Cloudflare from "alchemy/Cloudflare";
import * as Effect from "effect/Effect";

export default Alchemy.Stack("MyApp", {
  providers: Cloudflare.providers(),
}, Effect.gen(function* () {
  const queue = yield* Cloudflare.Queue("MyQueue", {
    name: "my-app-queue",
  });

  const worker = yield* Cloudflare.Worker("Worker", {
    main: "./src/worker.ts",
    bindings: { MY_QUEUE: queue },
  });

  // Register the worker as a consumer
  yield* Cloudflare.QueueConsumer("MyConsumer", {
    queueId: queue.queueId,
    scriptName: "my-worker",
    settings: {
      batchSize: 10,
      maxRetries: 3,
      maxWaitTimeMs: 5000,
    },
  });
}));
```

## Notes

- The `QueueConsumer` resource uses `@distilled.cloud/cloudflare/queues` APIs (`createConsumer`, `updateConsumer`, `deleteConsumer`, etc.)
- Idempotent create: if a consumer already exists for the same script, it looks it up instead of failing
- Queue/script changes trigger replacement; settings changes trigger update
- TypeScript: 0 errors

PR #71, opened 2026-04-19
fix(Cloudflare): provide WorkerEnvironment to Workflow body, not wrapper

### LINKED ISSUE #70: Cloudflare Workflow: WorkerEnvironment not provided to effect body (Service not found at runtime)
## Symptom

Deploying a `Cloudflare.Workflow` succeeds. Invoking an instance (`env.MY_WORKFLOW.create(...)`) crashes at runtime:

> `Service not found: Cloudflare.Workers.WorkerEnvironment (defined at virtual_alchemy-entry.js:<line>:<col>)`

This is 100% reproducible for any workflow that transitively requires `WorkerEnvironment` (which is essentially any workflow that reads bindings). Deploys on `alchemy@2.0.0-beta.x` + `effect@4.0.0-beta.48`.

## Root cause

`packages/alchemy/src/Cloudflare/Workers/Workflow.ts` lines 325-332 provide `WorkerEnvironment` to the **wrapping** `Effect.succeed(body)` instead of to `body` itself:

```ts
make: (env: unknown) =>
  Effect.succeed(body).pipe(
    Effect.provideContext(services),
    Effect.provideService(WorkerEnvironment, env as Record<string, any>),
  ),
```

`body` is `Effect<R, never, WorkflowRunServices>` (requires `WorkerEnvironment`), but `Effect.succeed(body)` produces `Effect<typeof body, never, never>` — it has no such requirement. `provideService` on this outer effect is a no-op. When `WorkflowBridge.run()` in `Rpc.ts` later runs the inner `body`, `WorkerEnvironment` is missing.

Compare `DurableObjectNamespace.ts` / `makeDurableObjectBridge` — those layer services onto the correct inner effect, which is why DOs and Containers work fine.

## Fix

Provide `WorkerEnvironment` to `body` directly, then wrap in `Effect.succeed`:

```ts
make: (env: unknown) =>
  Effect.succeed(
    body.pipe(
      Effect.provideService(WorkerEnvironment, env as Record<string, any>),
    ),
  ).pipe(Effect.provideContext(services)),
```

PR incoming with this change.

## Reproducer

Minimal workflow that reads a binding from env:

```ts
const MyWorkflow = Cloudflare.Workflow<{ input: string }>()("MyWorkflow", {
  run: Effect.gen(function* () {
    const env = yield* Cloudflare.WorkerEnvironment  // <-- crashes here at runtime
    // ...
  }),
})
```

Deploy, invoke `env.MY_WORKFLOW.create({ params: { input: "test" } })`, observe the error in Cloudflare dashboard / `wrangler tail`.

## Workaround

Patch `node_modules/alchemy/src/Cloudflare/Workers/Workflow.ts` post-install with the diff above. I'm shipping this as an idempotent script in my project until the fix lands.

yovanoc:
Fixes #70.

## Problem

`provideService(WorkerEnvironment, env)` was being applied to the outer `Effect.succeed(body)` rather than to `body` itself. Since `Effect.succeed(body)` produces an effect with `R = never`, the `provideService` call is a no-op — `WorkerEnvironment` is never actually provided to `body`.

When `WorkflowBridge.run()` in `Rpc.ts` later runs `body`, `WorkerEnvironment` is missing and any workflow instance crashes at runtime with:

> `Service not found: Cloudflare.Workers.WorkerEnvironment`

## Fix

Apply `provideService` to `body` before wrapping in `Effect.succeed`. This matches the working pattern used by `makeDurableObjectBridge` (which is why DOs and Containers are unaffected by this bug).

```diff
 make: (env: unknown) =>
-  Effect.succeed(body).pipe(
-    Effect.provideContext(services),
-    Effect.provideService(WorkerEnvironment, env as Record<string, any>),
-  ),
+  Effect.succeed(
+    body.pipe(
+      Effect.provideService(WorkerEnvironment, env as Record<string, any>),
+    ),
+  ).pipe(Effect.provideContext(services)),
```

## Testing

Verified in a real Cloudflare Workers deploy: workflows with `yield* Cloudflare.WorkerEnvironment` now progress past `step.do()` where they previously crashed immediately on invocation.

### DIFF
--- a/examples/cloudflare-worker/src/NotifyWorkflow.ts
+++ b/examples/cloudflare-worker/src/NotifyWorkflow.ts
@@ -8,15 +8,48 @@ export default class NotifyWorkflow extends Cloudflare.Workflow<NotifyWorkflow>(
     const rooms = yield* Room;
 
     return Effect.gen(function* () {
+      // Regression guard for https://github.com/alchemy-run/alchemy-effect/pull/71
+      //
+      // Accessing `Cloudflare.WorkerEnvironment` inside the workflow body
+      // previously crashed with `Service not found: Cloudflare.Workers.WorkerEnvironment`
+      // because `provideService(WorkerEnvironment, env)` was applied to the
+      // outer `Effect.succeed(body)` wrapper (a no-op) instead of `body`
+      // itself in `Workflow.ts`. Keeping this yield + the KV roundtrip below
+      // ensures the integ test catches any future regression of that fix.
+      const env = yield* Cloudflare.WorkerEnvironment;
       const event = yield* Cloudflare.WorkflowEvent;
       const { roomId, message } = event.payload as {
         roomId: string;
         message: string;
       };
 
+      // Exercise an env binding from inside a workflow step — the real-world
+      // pattern users follow (`env.<binding>.put(...)` / `.get(...)` etc).
+      const stored = yield* Cloudflare.task(
+        "kv-roundtrip",
+        Effect.tryPromise({
+          try: async () => {
+            const key = `workflow:smoke:${roomId}`;
+            await env.KV.put(key, message);
+            const got = await env.KV.get(key);
+            if (got !== message) {
+              throw new Error(
+                `KV roundtrip mismatch: expected "${message}", got "${got ?? "null"}"`,
+              );
+            }
+            return got;
+          },
+          catch: (cause) =>
+            cause instanceof Error ? cause : new Error(String(cause)),
+        }).pipe(Effect.orDie),
+      );
+
       const processed = yield* Cloudflare.task(
         "process",
-        Effect.succeed({ text: `Processed: ${message}`, ts: Date.now() }),
+        Effect.succeed({
+          text: `Processed: ${stored}`,
+          ts: Date.now(),
+        }),
       );
 
       const room = rooms.getByName(roomId);
--- a/examples/cloudflare-worker/test/integ.test.ts
+++ b/examples/cloudflare-worker/test/integ.test.ts
@@ -14,3 +14,65 @@ test(
     expect(url).toBeString();
   }),
 );
+
+/**
+ * Regression guard for https://github.com/alchemy-run/alchemy-effect/pull/71
+ *
+ * `NotifyWorkflow` accesses `Cloudflare.WorkerEnvironment` inside its body and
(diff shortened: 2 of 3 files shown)


sam-goodwin:
nit: please title your PRs `fix(Cloudflare)`, we only scope changes to the Cloud provider

sam-goodwin:
We should update examples/cloudflare-worker/src/NotifyWorkflow.ts to use a binding so that it can be used as a smoke test that would have caught this case

yovanoc:
Both addressed in [`fd47680`](https://github.com/yovanoc/alchemy-effect-fork/commit/fd47680):

**1. PR title**: renamed to `fix(Cloudflare): provide WorkerEnvironment to Workflow body, not wrapper`. Did the same sweep across my other open PRs (#69, #73, #75) so they all stick to `(Cloudflare)`.

**2. NotifyWorkflow smoke test**:

- `examples/cloudflare-worker/src/NotifyWorkflow.ts` now yields `Cloudflare.WorkerEnvironment` **inside the body** (the exact call site that crashed with `Service not found: Cloudflare.Workers.WorkerEnvironment` pre-fix) and adds a `kv-roundtrip` step that hits `env.KV.put` / `env.KV.get` — mirroring how users actually access bindings from a workflow step.
- `examples/cloudflare-worker/test/integ.test.ts` now starts the workflow via `POST /workflow/start/<roomId>`, polls `/workflow/status/<id>` for up to 60s, and asserts the instance reaches `complete` with no `error`. If the runtime fix is reverted, this test times out or surfaces the `errored` status — explicit regression guard.

**3. Type-system half of the same bug** (bundled into this PR since it's the user-facing side of the fix):

`WorkflowRunServices` was `WorkflowEvent | WorkflowStep` — it did not include `WorkerEnvironment`. The runtime fix in this PR provides the service via `body.pipe(Effect.provideService(WorkerEnvironment, env))`, but the type signature of the body forbade the very `yield* WorkerEnvironment` that the runtime fix enables. I found this while writing the smoke test: the new KV-roundtrip step failed to compile because the body's inferred service set leaked `WorkerEnvironment`.

Fix is a one-liner — extend the union:

```ts
export type WorkflowRunServices =
  | WorkflowEvent
  | WorkflowStep
  | WorkerEnvironment;
```

Happy to split that into its own PR if you'd prefer, but it seemed cleaner to land runtime + type together since either alone leaves users with a broken experience (runtime-only: TS error on every `env.` access; type-only: runtime `Service not found`).

Issue #72, opened 2026-04-19
Container app deploys without DO namespace linkage due to circular Output dependency in bindContainer


yovanoc:
## Summary

When binding a `Cloudflare.Container` to a `Cloudflare.DurableObjectNamespace` via `Cloudflare.Container.bind` inside a DO class body, the container application is created in Cloudflare's control plane **without its Durable Object namespace linkage**. Runtime consequence:

> `Error: There is no container application assigned to this Durable Object namespace`

The container is endlessly restarted by `StartContainer.ts`'s exponential retry loop, silently masking the real error.

## Environment

- `alchemy@2.0.0-beta.9` (reproduces on fork `yovanoc/alchemy-effect-fork@2282ff3`, same code on upstream `main`)
- `effect@4.0.0-beta.48`
- `@cloudflare/workers-types` latest
- Bun runtime, macOS darwin
- Pattern: canonical Container Layer — class + `.make()` in separate files, DO imports class

## Reproduction

Canonical pattern from the Container docstring:

```ts
// src/containers/device.ts
export class DeviceContainer extends Cloudflare.Container<DeviceContainer>()("DeviceContainerApp", {
  main: "./src/containers/device/server/server.ts",
  runtime: "bun",
  instanceType: "dev",
}) {}

// src/containers/device-do.ts
export default class DeviceContainerDO extends Cloudflare.DurableObjectNamespace<DeviceContainerDO>()(
  "DeviceContainer",
  Effect.gen(function* () {
    const container = yield* Cloudflare.Container.bind(DeviceContainer);
    return Effect.gen(function* () {
      const started = yield* Cloudflare.start(container);
      const port = yield* started.getTcpPort(3000);
      return {
        fetch: Effect.gen(function* () {
          const request = yield* HttpServerRequest;
          return yield* port.fetch(request);
        }),
      };
    });
  }),
) {}
```

Deploy succeeds cleanly. Any invocation of the DO that reaches `Cloudflare.start(container)` triggers the runtime error above.

## Evidence

Alchemy state file `.alchemy/state/<app>/<stage>/DeviceContainerApp.json`:

```json
{
  "attr": {
    "applicationId": "a03a9b73-d707-42bb-b301-1f61457d943a",
    "durableObjects": null          // ← should contain { namespaceId: "..." }
  },
  "bindings": [
    {
      "sid": "DeviceContainer",
      "data": {
        "durableObjects": {}         // ← should contain { namespaceId: "<resolved>" }
[...]

PR #73, opened 2026-04-19
fix(Cloudflare): fail loudly when DO namespaceId is unresolved (refs #72)

yovanoc:
## Summary

Minimum-viable fix for #72. Turns a silent deploy-time failure into a
loud one so users can diagnose the underlying circular-Output bug in
`bindContainer` without having to read framework internals.

## Problem (from #72)

`bindContainer` calls `container.bind` with `{ durableObjects: { namespaceId: namespace.namespaceId } }`, where `namespace.namespaceId` is an `Output<string>` that only resolves once the Worker is applied. But the Container is **upstream** of the Worker, so when `Output.evaluate(bindings)` runs at Container-create time (`Apply.ts:451`), `namespaceId` is `undefined`.

`getDurableObjects` in `ContainerApplication.ts` then returns a truthy `{ namespaceId: undefined }` (because the outer object is truthy), which is sent to the Cloudflare API. The container app gets created with **no** DO linkage, and the only user-visible symptom at runtime is:

> `no container application assigned to this Durable Object namespace`

This also survives `--force` deploys because the recreate guard uses `deepEqual` on the malformed value.

## Fix

In `getDurableObjects`, validate `namespaceId` is a non-empty string. If not, `Effect.die` with a diagnostic pointing at the real cause and the tracking issue.

```diff
-        if (dos.length === 1) {
-          return Effect.succeed(dos[0]);
-        }
-        return Effect.die(
-          new Error(
-            `A Container can only be bound to one Durable Object namespace...`,
-          ),
-        );
+        if (dos.length > 1) {
+          return Effect.die(new Error(`A Container can only be bound to...`));
+        }
+        const only = dos[0]!;
+        if (typeof only.namespaceId !== "string" || only.namespaceId.length === 0) {
+          return Effect.die(
+            new Error(
+              `Container binding has an unresolved Durable Object namespaceId...`
+              + ` See https://github.com/alchemy-run/alchemy-effect/issues/72`,
+            ),
+          );
+        }
+        return Effect.succeed(only);
```

## What this does NOT fix

This does not resolve the circular Output dependency itself. Users hitting this will still be blocked — but they will know **why**, instead of chasing a runtime error with no deploy-time signal.

The three structural fix options discussed in #72 (post-Worker re-evaluation pass, two-phase binding, or decoupling namespace creation from Worker apply) are larger changes and benefit from maintainer direction on preferred approach. Happy to follow up with a PR on whichever direction is preferred.

## Test plan

- `tsc -b` on `packages/alchemy`: clean.
- Reproduced the original silent failure against a real CF account (see #72 repro).
- With this patch applied locally: deploy fails early with the new diagnostic instead of deploying broken state.

## Checklist

- [x] Draft because maintainer sign-off wanted on direction before widening scope.
- [x] Linked to #72.
- [ ] No tests added — `getDurableObjects` is a closure inside `Provider.make`; happy to refactor for testability if desired.

### DIFF
--- a/packages/alchemy/src/Cloudflare/Container/ContainerApplication.ts
+++ b/packages/alchemy/src/Cloudflare/Container/ContainerApplication.ts
@@ -1,4 +1,5 @@
 import * as Containers from "@distilled.cloud/cloudflare/containers";
+import * as Data from "effect/Data";
 import * as Effect from "effect/Effect";
 import * as FileSystem from "effect/FileSystem";
 import * as Option from "effect/Option";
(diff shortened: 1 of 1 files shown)


packages/alchemy/src/Cloudflare/Container/ContainerApplication.ts
```diff
@@ -868,14 +868,32 @@ await Effect.runPromise(serverEffect).catch((err) => {
         if (dos.length === 0) {
           return Effect.succeed(undefined);
         }
-        if (dos.length === 1) {
-          return Effect.succeed(dos[0]);
+        if (dos.length > 1) {
+          return Effect.die(
+            new Error(
+              `A Container can only be bound to one Durable Object namespace. Found ${dos.length} namespaces in bindings: ${bindings.map((b) => b.data.durableObjects?.namespaceId).join(", ")}`,
+            ),
```
sam-goodwin: Should we start defining Data.TaggedError for every error, even when used in Effect.die?

yovanoc: good idea! you want me to do that in the PR ?

Mkassabov: > Should we start defining Data.TaggedError for every error, even when used in Effect.die

I think so, feels like good practice.

I like what @john-royal had me do with auth stuff where you don't die you just swallow everything in a generic error with a Schema.defect containing the real error.

then its on the harness itself to decide if its die worthy.
so maybe we have resource return a `CriticalError` or something that the Core is responsible for turning into a die call. Might also reduce how often AIs use .die when writing resources (not that this is really a problem rn.)

yovanoc: Done in [`3754e06`](https://github.com/yovanoc/alchemy-effect-fork/commit/3754e06), both `Effect.die(new Error(...))` call sites in `getDurableObjects` now use purpose-built `Data.TaggedError` classes:

- `ContainerMultipleDurableObjectNamespacesError` — `{ count, namespaceIds }`
- `ContainerUnresolvedDurableObjectNamespaceIdError` — `{ received }`

Both have a computed `message` getter so the text-form diagnostic is preserved (and stays visible if the die bubbles up into a fiber runtime log) while still being inspectable via `_tag` / structured fields from callers or tests. Follows the existing module-local pattern used by `Container.ts` (`ContainerError`), `Util/assert-never.ts` (`UnexpectedValueError`) and `Plan.ts` (`DeleteResourceHasDownstreamDependencies`, `UnsatisfiedResourceCycle`).

@Mkassabov — really like the `CriticalError` / `Schema.defect` direction. Seems orthogonal to this PR's scope though: the guard here was already a `.die` before these edits, and I didn't want to fan-out into a provider-wide convention change on a minimum-viable fix for #72. Happy to follow up with a dedicated PR (or issue) that pilots the harness-level pattern on the Container resources + a couple of adjacent providers if you or @sam-goodwin want to scope it — probably worth discussing the exact shape of `CriticalError` (single tag with `Schema.defect`? tagged union per-category?) before touching call-sites.

Also pinged separately on #72 re: sam's example-repo question — waiting on his pointer before I build the standalone repro.

Mkassabov: @yovanoc 
> Seems orthogonal to this PR's scope though

100%, just brought it up for the sake of future discussion.

yovanoc:
Draft PR #84 adds a minimal example (`examples/cloudflare-worker-paired-do`) reproducing the silent failure this PR converts into a loud `Effect.die`. Uses the paired-DO + split-runtime pattern (Container + DurableObjectNamespace as separate resources with mismatched LogicalIds).

yovanoc:
Standalone repro repo ready: https://github.com/yovanoc/alchemy-effect-container-cycle-repro

Contains two minimal variants:
- **Variant A** (canonical Sandbox/Agent pattern): deploys cleanly, durableObjects populated
- **Variant B** (paired-resource pattern): demonstrates the silent failure this PR converts into a loud Effect.die

Both variants are ready to  against a real CF account for verification.

sam-goodwin:
Please open an issue if this is still a problem. We're pretty far behind now.

PR #75, opened 2026-04-19
fix(Cloudflare): install external deps in container Dockerfile

yovanoc:
## Summary

When a Container is declared with `external: ["impit", "tough-cookie"]`, the bundler correctly excludes these packages from the bundle. However, the generated Dockerfile only copies `index.mjs` into the image — it does **NOT** run `bun add` to install the external packages.

At container runtime (Firecracker VM), Bun encounters bare `import` statements for the external packages and fails to resolve the module because it isn't in `node_modules`. The container process **crashes before any Effect code runs**, producing zero stdout and causing `port.fetch()` to hang indefinitely with `outcome: canceled`.

## Fix

`buildFinalDockerfile` now accepts an `external: string[]` parameter. When external packages exist and runtime is "bun", it injects a `RUN bun add <deps>` step between `WORKDIR` and `COPY`. Both call sites (`computeImageHash` and `buildAndPushImage`) now pass `props.external`.

### Example generated Dockerfile

```dockerfile
FROM oven/bun:1

WORKDIR /app
RUN bun add impit tough-cookie

COPY index.mjs /app/index.mjs
ENTRYPOINT ["bun", "/app/index.mjs"]
```

## Verification

Deployed to Cloudflare Workers with a DeviceContainer (Bun runtime). After the fix:
- Container image contains `node_modules/impit/` and `node_modules/tough-cookie/`
- `package.json` and `bun.lock` are generated in the image
- Container process no longer crashes on startup (no zero-stdout issue)
- `port.fetch()` connects successfully

### DIFF
--- a/packages/alchemy/src/Cloudflare/Container/ContainerApplication.ts
+++ b/packages/alchemy/src/Cloudflare/Container/ContainerApplication.ts
@@ -103,6 +103,32 @@ export interface ContainerApplicationProps extends PlatformProps {
    */
   runtime?: "bun" | "node";
   /**
+   * Module specifiers that Rolldown should mark as external when bundling
+   * the container entrypoint. The matching packages are installed inside the
+   * image via the runtime's package manager (`bun add` for `runtime: "bun"`,
+   * `npm install` for `runtime: "node"`) before the entrypoint runs.
+   *
+   * Use this for native dependencies that must not be bundled (e.g. `sharp`,
+   * `impit`) or for packages that intentionally ship in the base image.
+   *
+   * Install inside the image is controlled by {@link autoInstallExternals}
+   * (default `true`); set it to `false` if your custom `dockerfile` already
+   * installs these packages and you want to avoid the redundant step.
+   */
+  external?: string[];
+  /**
+   * Whether to auto-install the packages listed in {@link external} inside
+   * the container image (via `bun add` or `npm install`) before running the
+   * entrypoint.
+   *
+   * @default true
+   *
+   * Set to `false` when your custom `dockerfile` already installs these
+   * packages (for example, via a base image that pre-installs `sharp`), to
+   * avoid the redundant install step.
+   */
+  autoInstallExternals?: boolean;
+  /**
    * Human-readable application name. If omitted, Alchemy derives a deterministic
    * physical name from the stack, stage, and logical ID.
    */
@@ -418,7 +444,12 @@ export const ContainerProvider = () =>
           isExternal: props.isExternal,
         });
 
-        const finalDockerfile = buildFinalDockerfile(props.dockerfile, runtime, props.external);
+        const finalDockerfile = buildFinalDockerfile(
+          props.dockerfile,
+          runtime,
+          props.external,
+          props.autoInstallExternals,
+        );
         const imageHash = (yield* sha256Object({
           bundleHash,
           dockerfile: finalDockerfile,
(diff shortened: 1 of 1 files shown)


packages/alchemy/src/Cloudflare/Container/ContainerApplication.ts
```diff
@@ -561,15 +561,21 @@ await Effect.runPromise(serverEffect).catch((err) => {
...
+        external: string[] = [],
       ): string => {
         const base =
           userDockerfile?.trim() ??
           (runtime === "bun" ? "FROM oven/bun:1" : "FROM node:22-slim");
         const runtimeBin = runtime === "bun" ? "bun" : "node";
+        const installStep =
+          external.length > 0 && runtime === "bun"
+            ? `RUN bun add ${external.join(" ")}`
+            : "";
```
sam-goodwin: implement `runtime === "node"` too

sam-goodwin:
I like this idea, but is it too opinionated? What if the image already has these dependencies installed?

I've been thinking that it would be cool to provide a way to programmatically contribute to the Dockerfile so that Layers can install their own dependencies (for example) 

```ts
yield* Docker.RUN("bun add my-dep")
```

This could then be packaged into utilities (like a package manager-agnostic "install" utilitiy)
```ts
yield* install(["my-dep"]
```

Thoughts?

packages/alchemy/src/Cloudflare/Container/ContainerApplication.ts
yovanoc: Addressed both in [`e8ec40d`](https://github.com/yovanoc/alchemy-effect-fork/commit/e8ec40d):

### 1. `runtime === "node"` support

Install step now dispatches on runtime:
```ts
const installCmd = runtime === "bun" ? "bun add" : "npm install";
const installStep =
  autoInstallExternals && external.length > 0
    ? `RUN ${installCmd} ${external.join(" ")}`
    : "";
```

So for `runtime: "node"` with `external: ["sharp"]`, the generated Dockerfile is:
```dockerfile
FROM node:22-slim

WORKDIR /app
RUN npm install sharp

COPY index.mjs /app/index.mjs
ENTRYPOINT ["node", "/app/index.mjs"]
```

### 2. "too opinionated? what if the image already has these deps installed?"

Fair concern. Added `autoInstallExternals?: boolean` (default `true`) on `ContainerApplicationProps`:

```ts
Cloudflare.Container(...)({
  runtime: "node",
  external: ["sharp"],
  // My base image already has sharp installed → skip the RUN step:
  dockerfile: "FROM my-org/node-with-sharp:1",
  autoInstallExternals: false,
});
```

Also properly typed `external?: string[]` on `ContainerApplicationProps` — the previous commit on this PR accessed `props.external` but the field was never declared on the type (pre-existing TS2551 on lines 421 and 607). That's now fixed.

### 3. Re: the longer-term `Docker.RUN` / `install([...])` API

Love that direction. The `autoInstallExternals` flag is intentionally narrow — just an opt-out escape hatch until the layer-contribution API lands. Once something like
```ts
yield* Docker.RUN("bun add sharp")
// or
yield* install(["sharp"])
```
exists, both `autoInstallExternals` *and* the current bundled-in `bun add <external>` step could be removed: users either declare an external and use `install([...])` explicitly, or they omit it entirely. Happy to help scope that follow-up once this lands — feels like it wants its own issue/PR rather than bundling with this fix.

PR #84, opened 2026-04-22
docs(Cloudflare): add paired-DO container example reproducing #72

yovanoc:
Minimal repro for #72.

Pattern: Container + paired `DurableObjectNamespace` as separate resources, container runtime in a split file (`container-runtime/server.ts` via `main:`). Deploy succeeds, state shows `durableObjects: null`, runtime: `no container application assigned to this Durable Object namespace`.

Also documents the mandatory LogicalId mismatch (`MyContainerApp` vs `MyContainer`) to dodge `Diff.ts` last-write-wins `sid` dedup.

Complements #73 (fail-loud via `Effect.die`). Does not fix the underlying circular `Output<T>`.

`tsc --noEmit` clean. Not deployed — README has the repro recipe.


sam-goodwin:
Please open an issue if this is still a problem. We're pretty far behind now.

PR #85, opened 2026-04-22
chore: update Container examples and container docs for enableInternet

yovanoc:
## Problem

CF Containers default to blocked public egress. When `StartContainer.ts` calls `container.start()` with no options:

- `cfeth0` only gets private addresses (10.0.0.1/24 + fd00::11) — every `dns.resolve` / `fetch` / TCP connect hangs indefinitely (25s hard-timeout).
- Firecracker VMs are hermetic and don't inherit the Worker's env, so containers can't read Worker secrets via `Bun.env.XXX`.

## Fix

- Opt in to public egress by setting `ALCHEMY_CONTAINER_ENABLE_INTERNET=1` in the Worker env (or as a secret). Off by default to match CF's default.
- Log a one-line hint at container start when the flag is absent, so users hitting the DNS-hang symptom can discover the fix in logs instead of debugging Firecracker networking blind.
- Forward string-valued Worker env entries into `ContainerStartupOptions.env` via `Effect.serviceOption(WorkerEnvironment)`.

## Verification

Confirmed via a `/debug/connectivity` probe inside a bun container:
- **Without flag**: every outbound test hung 25s (hard-timeout). Log line emitted.
- **With `ALCHEMY_CONTAINER_ENABLE_INTERNET=1`**: DNS resolves in <30ms, TCP connect to 1.1.1.1 completes in <20ms.

## Notes

- Opt-in keeps parity with CF's platform default (blocked egress) and avoids surprising users with opened network access.
- Only string values are forwarded (DO/KV/R2 bindings are objects and can't cross the FFI boundary anyway).
- The discoverability log is `Effect.logInfo` (not `logWarning`) since locked-down egress is a valid deployment choice.


packages/alchemy/src/Cloudflare/Container/StartContainer.ts
```diff
@@ -21,7 +22,33 @@ export const start = Effect.fnUntraced(function* <
...
+    // env (or as a secret). We log a one-line hint on first start when the
+    // flag is absent, so users hitting the DNS-hang symptom can find the
+    // fix in logs instead of debugging Firecracker networking blind.
+    const enableInternet = env.ALCHEMY_CONTAINER_ENABLE_INTERNET === "1";
+    if (!enableInternet) {
+      yield* Effect.logInfo(
+        "Container starting with public egress DISABLED. Set ALCHEMY_CONTAINER_ENABLE_INTERNET=1 in the Worker env to enable internet access (cfeth0 has only private addresses by default).",
+      );
+    }
+    yield* container.start({ enableInternet, env });
```
sam-goodwin: WHy do this with an environment variable? 

Let's update this function to accept those arguments so the user can just pass an arg
```ts
export const start = Effect.fnUntraced(function* <
  Shape extends Container,
  Req = never,
>(containerEff: Effect.Effect<Shape, never, Req | DurableObjectState>) {
```

yovanoc:
@sam-goodwin Good call — much cleaner API. Updated:

**Before (implicit env reading):**
```typescript
const container = yield* Cloudflare.start(sandbox);
// secretly reads ALCHEMY_CONTAINER_ENABLE_INTERNET from WorkerEnvironment
```

**After (explicit options):**
```typescript
const container = yield* Cloudflare.start(sandbox, { enableInternet: true });
```

Changes:
- Removed `WorkerEnvironment` service dependency from `StartContainer.ts`
- `start()` now accepts optional second arg `{ enableInternet?: boolean; env?: Record<string, string> }`
- Updated `Agent.ts` example + all docs / test types to use the new signature
- When `enableInternet` is absent/false, the log hint now says "Pass { enableInternet: true }" instead of referencing the env var

Pushed as a fixup commit on the same branch.

packages/alchemy/src/Cloudflare/Container/StartContainer.ts
```diff
@@ -15,13 +15,29 @@ import { type Container, ContainerError } from "./Container.ts";
 export const start = Effect.fnUntraced(function* <
   Shape extends Container,
   Req = never,
->(containerEff: Effect.Effect<Shape, never, Req | DurableObjectState>) {
+>(
+  containerEff: Effect.Effect<Shape, never, Req | DurableObjectState>,
+  options?: {
+    enableInternet?: boolean;
+    env?: Record<string, string>;
+  },
```
sam-goodwin: Should we use the same type in the underlying Container .start function?

yovanoc:
@sam-goodwin Done, didn't saw that type, thanks!

packages/alchemy/src/Cloudflare/Container/StartContainer.ts
```diff
@@ -7,21 +7,34 @@ import * as HttpServerRequest from "effect/unstable/http/HttpServerRequest";
...
+    // CF Containers default to BLOCKED public egress. Containers get only
+    // private addresses (10.0.0.1/24 + fd00::11) on cfeth0 and every DNS
+    // lookup / TCP connect / fetch hangs indefinitely unless enableInternet
+    // is explicitly set.
+    const enableInternet = options?.enableInternet ?? false;
+    if (!enableInternet) {
+      yield* Effect.logInfo(
+        "Container starting with public egress DISABLED. Pass { enableInternet: true } to enable internet access (cfeth0 has only private addresses by default).",
+      );
+    }
```
sam-goodwin: Why are we dying here? This is not an error. It's a choice.

yovanoc:
Removed the log, just passes options through now.

yovanoc:
Superseded by #341 which landed the `Cloudflare.start(options)` change in a narrower scope. The remaining diff here (doc updates, test fixes, Worker env forwarding) can be rebased and reopened as a follow-up if still needed.

PR #96, opened 2026-04-26
fix(cloudflare): infer worker service bindings

### LINKED ISSUE #95: Cloudflare Vite Worker bindings should infer Worker service bindings as Fetcher
I hit this while trying to use a Cloudflare Vite app with a separate auth Worker, eg better-auth running in its own `Cloudflare.Worker`
**Setup is roughly**
```typescript
const authWorker = yield* AuthWorker
const app = yield* Cloudflare.Vite("App", {
  bindings: {
    AUTH: authWorker, // error
  },
})
type Env = Cloudflare.InferEnv<typeof app>
//    ^?
// type Env = {
//   [x: string]: Fetcher | R2Bucket | D1Database | KVNamespace<string> | Queue<unknown> | DurableObjectNamespace<any>
//   ASSETS: Fetcher
// }
//
// Problem: The key is not literal and the value is not narrowed
```

**Expected:**
- AUTH should be accepted as a Worker service binding and Env["AUTH"] should narrow to Cloudflare Fetcher

**Actual:**
- AUTH: authWorker is rejected because Worker is not part of WorkerBindingResource
- Also InferEnv does not have a Worker case, so instead of a concrete AUTH: Fetcher you only get the broad fallback/index binding union, eg Fetcher | R2 | D1 | KV | Queue | DO, which is not useful for calling env.AUTH.fetch(...)
- This is separate from effect-native Worker to Worker RPC via bindWorker
- For this use case I do not need typed RPC, just the standard Cloudflare service binding so env.AUTH.fetch(...) is typed correctly

nr1brolyfan:
Fixes #95 

## Summary
- allow Cloudflare Worker resources as Vite/plain Worker service bindings
- infer Worker bindings as Cloudflare Fetcher in InferEnv
- add a type fixture covering the Vite + auth Worker setup

### DIFF
--- a/packages/alchemy/src/Cloudflare/Workers/InferEnv.ts
+++ b/packages/alchemy/src/Cloudflare/Workers/InferEnv.ts
@@ -10,9 +10,15 @@ export type InferEnv<W> = W extends
   | Worker<infer Bindings>
   | Effect.Effect<Worker<infer Bindings>, any, any>
   ? {
-      [K in keyof Bindings]: GetBindingType<UnwrapEffect<Bindings[K]>>;
+      [K in keyof Bindings]: GetBindingType<UnwrapBinding<Bindings[K]>>;
     }
   : never;
+
+type UnwrapBinding<T> = T extends {
+  asEffect(): Effect.Effect<infer A, any, any>;
+}
+  ? A
+  : UnwrapEffect<T>;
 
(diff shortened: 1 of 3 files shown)


sam-goodwin:
Tried pulling this down and got in weird git state. Going to fix in a separate PR. Thanks for the change though!

john-royal:
Fixed in #310

PR #103, opened 2026-04-27
feat(ECS): add CapacityProvider resource

skusez:
## Summary

Adds an ASG-backed `AWS.ECS.CapacityProvider` resource so capacity providers can be defined alongside an `AWS.ECS.Cluster` instead of bridging to CDK or the AWS CLI.

- Resource + provider in `packages/alchemy/src/AWS/ECS/CapacityProvider.ts` covering create / read / diff (replace on `name` or `autoScalingGroupArn` change) / update (managed scaling, termination protection, draining, tags) / idempotent delete
- Adoption: when [`AdoptPolicy`](https://github.com/alchemy-run/alchemy-effect/blob/main/packages/alchemy/src/AdoptPolicy.ts) is provided and a provider with the same name already exists, it is adopted (and re-tagged with internal tags if needed). Without `AdoptPolicy`, the create dies with a helpful message — matching the documented intent
- Registered in `AWS/ECS/index.ts` and `AWS/Providers.ts`
- Smoke test gated on `TEST_ASG_ARN` (AutoScaling has no test infra in this repo yet, so the test takes a pre-provisioned ASG ARN rather than inventing the LaunchTemplate/ASG fixture)

`ManagedInstancesProvider` is intentionally out of scope — only `autoScalingGroupArn` is supported.

### Drive-by fixes (`scripts/audit-service.ts`)

- Resolve resource paths under `packages/alchemy/...` to match the current monorepo layout (previously hardcoded `alchemy/...`, which no longer exists at the repo root)
- Add `cloudfront` and `cloudwatch` to the service-name → directory map so the script case-folds them to `CloudFront`/`CloudWatch` instead of the wrong `Cloudfront`/`Cloudwatch`

## Test plan

- [x] `bun tsc -b` passes
- [ ] Smoke test passes locally with `TEST_ASG_ARN=<arn> bun vitest CapacityProvider`
- [ ] `bun scripts/audit-service.ts ecs` runs without "directory not found" errors


sam-goodwin:
Looks great. Do you want to merge it or are you still working on testing?

skusez:
Thanks! I'm all done here if you're happy with it

PR #106, opened 2026-04-28
feat(CloudFront): CachePolicy, OriginRequestPolicy, ResponseHeadersPolicy, PublicKey, KeyGroup

skusez:
Adds five CloudFront resources, modeled after `OriginAccessControl` / `Distribution`:

- `CachePolicy`
- `OriginRequestPolicy`
- `ResponseHeadersPolicy`
- `PublicKey`
- `KeyGroup`

Tests: CRUD test per resource + a `Distribution.smoke.test.ts` that wires the three policies and a `KeyGroup` into a Distribution over a private S3 origin. All gated on `ALCHEMY_RUN_LIVE_AWS_WEBSITE_TESTS=true`.

`bun tsc -b` clean.


packages/alchemy/src/AWS/CloudFront/PublicKey.ts
```diff
@@ -0,0 +1,243 @@
...
+     */
+    publicKeyId: string;
+    /**
+     * Name of the public key.
+     */
+    name: string;
+    /**
+     * PEM-encoded public key body.
+     */
+    encodedKey: string;
```
sam-goodwin, line 47: This should probably be a Redacted, always use Redacted<string> for sensitive values

skusez: Fixed this one up

packages/alchemy/src/AWS/CloudFront/CachePolicy.ts
```diff
@@ -0,0 +1,271 @@
...
+                          ),
+                        ),
+                  ),
+                ),
+              ),
+            );
+          if (!created.CachePolicy?.Id) {
+            return yield* Effect.fail(
+              new Error("createCachePolicy returned no identifier"),
+            );
```
sam-goodwin, line 226: This should probably be an Effect.die. We don't support errors thrown from lifecycle operations yet. Must die

packages/alchemy/src/AWS/CloudFront/CachePolicy.ts
```diff
@@ -0,0 +1,271 @@
...
+                            CachePolicyConfig: existing.config,
+                          },
+                          ETag: existing.etag,
+                          Location: undefined,
+                        })
+                      : Effect.fail(
+                          new Error(
+                            `Cache policy '${name}' already exists but could not be recovered`,
+                          ),
+                        ),
```
sam-goodwin, line 218: Die. 

One thing we do in AWS resources is add tags on create so we can know if we created it. We invoice stack, stage and logical id. 

Look at some other resources and Tags.ts. 

When creating, if it already exists, check the tags and adopt of it has the expected tags otherwise die (unless AdoptionPolicy is set to adopt). 

Only do this if the create api accepts tags. Because only then can we guarantee it atomically.

skusez: Took a look at this happy to add it, but want to flag a wrinkle first. Of the 5 resources in this PR, none of the create APIs accept tags atomically: CreateCachePolicy, CreateOriginRequestPolicy, CreateResponseHeadersPolicy, CreatePublicKey, and CreateKeyGroup all take only their respective *Config struct. The only CloudFront create APIs that bundle tags are CreateDistributionWithTags and CreateStreamingDistributionWithTags. There's a generic TagResource API but it'd be a separate call after create, so it wouldn't be atomic. The existing OriginAccessControl resource is in the same boat and just does name-based adoption without a tag check, so I modeled these 5 off that. Happy to add a non-atomic create-then-tag flow with a hasAlchemyTags adoption check anyway if you'd prefer just wanted to confirm

sam-goodwin: If it's not atomic then it becomes a mitigation. Still helpful but doesn't prevent the brick. 

I'm also not super concerned, we can add it later if it becomes a problem.

PR #113, opened 2026-04-29
feat(planetscale): Add planetscale resources

lucasthevenet:
Adds six Planetscale resources, based on the resources from v1:

- Database
- Branch
- Role and DefaultRole for Postgres
- Password for MySQL

Tests pending due to a module resolution issue on [the sdk side](https://github.com/alchemy-run/distilled/pull/199)


sam-goodwin:
Sweet! Looking awesome. Curious, did you refer to .vendor/alchemy's v1 planetscale resources or just build from first principles?

packages/alchemy/src/Planetscale/Database.ts
```diff
@@ -0,0 +1,551 @@
...
+   * The database cluster size. Required.
+   * @see https://planetscale.com/docs/concepts/cluster-size
+   */
+  clusterSize: PlanetScaleClusterSize;
+
+  /**
+   * Engine kind. Cannot be changed after creation.
+   * @default "mysql"
+   */
+  kind?: "mysql" | "postgresql";
```
Mkassabov: This was a mistake in async alchemy. MySQL and postgres should be split into 2 different resources.

Mkassabov (changes requested):

lucasthevenet:
> Sweet! Looking awesome. Curious, did you refer to .vendor/alchemy's v1 planetscale resources or just build from first principles?

Hi! Yes I used .vendor/alchemy as a reference to build the new resources here.

packages/alchemy/src/Planetscale/Database.ts
lucasthevenet: Yeah I agree I does seem better to split them. I've gone ahead and added the PostgreSQLDatabase (Maybe should be PostgresDatabase now that I think about it) and MySQLDatabase resources to replace the previous Database resource.

.vendor/alchemy
Mkassabov, line 1: this doesn't look right? why are there changes in the submodule?

packages/alchemy/src/Planetscale/AuthProvider.ts
```diff
@@ -0,0 +1,293 @@
...
+    label: "Environment Variables",
+    hint: "PLANETSCALE_API_TOKEN_ID + PLANETSCALE_API_TOKEN + PLANETSCALE_ORGANIZATION",
+  },
+  {
+    value: "stored",
+    label: "Service Token",
+    hint: "enter service token interactively, stored in ~/.alchemy/credentials",
+  },
+];
+
```
Mkassabov, line 41: don't expect it to be done in this PR but can you please add a comment here
`//todo(pear): add planetscale oauth`

that way I know to handle it later. thanks!

packages/alchemy/src/Planetscale/AuthProvider.ts
```diff
@@ -0,0 +1,293 @@
...
+}
+
+/**
+ * Layer that registers the PlanetScale {@link AuthProvider} into the
+ * {@link AuthProviders} registry when built. Include this in the
+ * PlanetScale `providers()` layer so `alchemy login` can discover it.
+ *
+ * Supported methods:
+ * - `env`: reads `PLANETSCALE_API_TOKEN_ID`/`PLANETSCALE_API_TOKEN`/`PLANETSCALE_ORGANIZATION`
+ *   (or legacy `PLANETSCALE_SERVICE_TOKEN_ID`/`PLANETSCALE_SERVICE_TOKEN`/`PLANETSCALE_ORG_ID`).
```
Mkassabov: get rid of the legacy way of doing this please. just `PLANETSCALE_API_TOKEN_ID`/`PLANETSCALE_API_TOKEN`/`PLANETSCALE_ORGANIZATION`

packages/alchemy/src/Planetscale/AuthProvider.ts
```diff
@@ -0,0 +1,293 @@
...
+        // Primary env vars
+        const tokenId =
+          (yield* getEnvRedacted("PLANETSCALE_API_TOKEN_ID")) ??
+          (yield* getEnvRedacted("PLANETSCALE_SERVICE_TOKEN_ID"));
+        const token =
+          (yield* getEnvRedacted("PLANETSCALE_API_TOKEN")) ??
+          (yield* getEnvRedacted("PLANETSCALE_SERVICE_TOKEN"));
+        const organization =
+          (yield* getEnvRedacted("PLANETSCALE_ORGANIZATION")) ??
+          (yield* getEnvRedacted("PLANETSCALE_ORG_ID"));
```
Mkassabov: use `getEnvRedactedRequired`

Mkassabov (commented):
can we organize resources into postgres and mysql (and maybe put them in a namespace) so its easier to reason about which resources make sense for a given use case

lucasthevenet:
> can we organize resources into postgres and mysql (and maybe put them in a namespace) so its easier to reason about which resources make sense for a given use case

do you mean grouping them like this?:

```typescript
 import * as Planetscale from "alchemy/Planetscale";
 
const db = yield* Planetscale.MySQL.Database("MyDb", {
   clusterSize: "PS_10",
 });

 const pwd = yield* Planetscale.MySQL.Password("Reader", {
   database: db,
   role: "reader",
 });
``` 
or just separate them by path like `alchemy/Planetscale`, `alchemy/Planetscale/PostgreSQL`, and `alchemy/Planetscale/MySQL`

sam-goodwin:
Maybe Planetscale.MySQLDatabase?

lucasthevenet:
Yeah that makes sense, but I think @Mkassabov was talking more about a way to hint that certain resources are meant only for Postgres or only for MySQL (e.g. Password -> only MySQL, Role and DefaultRole-> only Postgres, Branch -> both).

So I guess in this case we could just do Planetscale.MySQLPassword, to make it as obvious as possible.

examples/planetscale-mysql/alchemy.run.ts
```diff
@@ -0,0 +1,44 @@
...
+  Effect.gen(function* () {
+    const database = yield* Planetscale.MySQLDatabase("Database", {
+      clusterSize: "PS_10",
+    }).pipe(adopt());
+    const branch = yield* Planetscale.Branch("Branch", {
+      parentBranch: database.defaultBranch,
+      database,
+      isProduction: false,
+    }).pipe(adopt());
+    const password = yield* Planetscale.MySQLPassword("WorkerPassword", {
```
sam-goodwin: Should we have MySQLBranch to align?

examples/planetscale-mysql/alchemy.run.ts
```diff
@@ -0,0 +1,44 @@
...
+export default Alchemy.Stack(
+  "PlanetScaleMySQLExample",
+  {
+    providers: Layer.mergeAll(Cloudflare.providers(), Planetscale.providers()),
+    state: Cloudflare.state(),
+  },
+  Effect.gen(function* () {
+    const database = yield* Planetscale.MySQLDatabase("Database", {
+      clusterSize: "PS_10",
+    }).pipe(adopt());
```
sam-goodwin: why `adopt()`? User can pass `--adopt` if you want it to be adopted

packages/alchemy/src/Planetscale/Database.ts
```diff
@@ -0,0 +1,110 @@
...
+
+/**
+ * Region selector for a PlanetScale Database.
+ */
+export interface DatabaseRegion {
+  /**
+   * Region slug (e.g. `"us-east"`, `"eu-west"`, `"gcp-us-central1"`).
+   * @see https://planetscale.com/docs/concepts/regions#available-regions
+   */
+  slug: string;
```
sam-goodwin: `"us-east" | "eu-west" | "gcp-us-central1" | (string &{})`

examples/planetscale-mysql/alchemy.run.ts
lucasthevenet: Yeah I was actually thinking about this last night. Even at the branch level, planetscale has completely different behaviours for mysql and postgresql

examples/planetscale-mysql/alchemy.run.ts
lucasthevenet: this example was outdated, I just pushed a new cloudflare-planetscale-drizzle example that also aligns better with what the current neon example does.

sam-goodwin:
Sorry if I messed anything up. I had a coding agent add the effect test fixtures

examples/planetscale-mysql/alchemy.run.ts
sam-goodwin: Awesome. That's right. I had a coding agent do the same, didn't realized you were actively working on it. Sorry if I messed anything up

examples/cloudflare-planetscale-postgres-drizzle/src/Api.ts
```diff
@@ -0,0 +1,92 @@
...
+export default class Api extends Cloudflare.Worker<Api>()(
+  "Api",
+  {
+    main: import.meta.filename,
+  },
+  Effect.gen(function* () {
+    const conn = yield* Cloudflare.Hyperdrive.bind(Hyperdrive);
+    const db = yield* Drizzle.postgres(conn.connectionString, {
+      relations,
+    });
```
Mkassabov, line 20: can you also throw in a pscale mysql + hyperdrive example.
can just be another db in the current example

packages/alchemy/src/Planetscale/Branch.ts
```diff
@@ -0,0 +1,499 @@
...
+
+          // If parentBranch is a Branch resource, wait for it to be ready
+          // before we touch the child branch.
+          if (news.parentBranch && typeof news.parentBranch !== "string") {
+            yield* waitForBranchReady(
+              organization,
+              databaseName,
+              parentBranchName,
+            );
+          }
```
Mkassabov: is this needed?
If the parent branch isn't ready the parent branch resource shouldn't be completed which means the child branch resource doesn't have its inputs.

So how can this get hit?

CC @sam-goodwin

packages/alchemy/src/Planetscale/Branch.ts
```diff
@@ -0,0 +1,499 @@
...
+            return yield* Effect.fail(
+              new PlanetscaleConflict({
+                message:
+                  `Planetscale branch "${current.name}" (database "${databaseName}") has kind ` +
+                  `"${current.kind}" but this resource is a ${opts.engineLabel}. ` +
+                  `Use the matching ${current.kind === "mysql" ? "MySQLBranch" : "PostgresBranch"} ` +
+                  `resource instead.`,
+              }),
+            );
+          }
```
Mkassabov, line 355: does this also surface as a type issue?

IMO the only way to run into this should be if you make a branch and instead of passing in the planetscale db resource you pass in a string that points to one (implying that alchemy isn't managing your pscale db)

packages/alchemy/src/Planetscale/Branch.ts
```diff
@@ -0,0 +1,499 @@
...
+              branch: current.name,
+              new_name: desiredBranchName,
+            });
+          }
+
+          const branchName = current.name;
+
+          // Sync production status before branch settings that depend on it.
+          if (current.production !== news.isProduction) {
+            yield* waitForBranchReady(organization, databaseName, branchName);
```
Mkassabov: can't we just call `waitForBranchReady` before all of these syncs?

packages/alchemy/src/Planetscale/Branch.ts
```diff
@@ -0,0 +1,499 @@
...
+              region: current.region.slug,
+            });
+            if (current.cluster_name !== sanitized) {
+              if (!news.isProduction) {
+                return yield* Effect.fail(
+                  new PlanetscaleConflict({
+                    message: `Cannot change cluster size of non-production branch "${branchName}".`,
+                  }),
+                );
+              }
```
Mkassabov: is there a reason we are specifying this error here rather than in coming from distilled?
if it comes from the pscale api we use their language and don't need to make a change if they change behavior

packages/alchemy/src/Planetscale/Branch.ts
```diff
@@ -0,0 +1,499 @@
...
+
+        delete: Effect.fn(function* ({ output }: any) {
+          yield* deleteBranch({
+            organization: output.organization,
+            database: output.database,
+            branch: output.name,
+          }).pipe(
+            Effect.catchTag("NotFound", () => Effect.void),
+            // might throw UnprocessableEntity when branch is the default branch for the database
+            Effect.catchTag("UnprocessableEntity", () => Effect.void),
```
Mkassabov: this feel quite dangerous to catch since it could indicate other issues, does it come with a text error explaining this so we can check `UnprocessableEntity` + that text?

packages/alchemy/src/Planetscale/MySQL/MySQLMigrations.ts
```diff
@@ -0,0 +1,237 @@
...
+        ttl: 3600,
+      });
+      if (created.plain_text == null) {
+        return yield* Effect.fail(
+          new MySQLMigrationError({
+            message:
+              "PlanetScale did not return a MySQL password for migrations.",
+          }),
+        );
+      }
```
Mkassabov: when does this actually occur?

packages/alchemy/src/Planetscale/MySQL/MySQLMigrations.ts
```diff
@@ -0,0 +1,237 @@
...
+    use,
+    (password) =>
+      ops
+        .deletePassword({
+          organization: target.organization,
+          database: target.database,
+          branch: target.branch,
+          id: password.id,
+        })
+        .pipe(Effect.catch(() => Effect.void)),
```
Mkassabov: this can fail and thus risks orphaned passwords with migration access which is quite dangerous

packages/alchemy/src/Planetscale/MySQL/MySQLDatabase.ts
```diff
@@ -0,0 +1,426 @@
...
+              htmlUrl: data.html_url,
+              region: { slug: data.region.slug },
+              migrationsDir: output?.migrationsDir ?? olds?.migrationsDir,
+              migrationsTable: output?.migrationsTable ?? olds?.migrationsTable,
+              migrationsHashes: output?.migrationsHashes ?? {},
+              importHashes: output?.importHashes ?? {},
+              // Kind validation is deferred to reconcile, where the mismatch is
+              // surfaced as a typed `PlanetscaleConflict`. Read's job is just
+              // existence + ownership routing.
+              kind: (data.kind === "mysql" ? "mysql" : data.kind) as "mysql",
```
Mkassabov: @sam-goodwin is this correct behavior?

packages/alchemy/src/Planetscale/MySQL/MySQLPassword.ts
```diff
@@ -0,0 +1,356 @@
...
+              replica: news.replica,
+              ttl: news.ttl,
+              cidrs: news.cidrs,
+            });
+            if (created.plain_text == null) {
+              return yield* Effect.die(
+                "Planetscale createPassword response did not include plain_text. The password cannot be recovered.",
+              );
+            }
+            plaintext = toRedacted(created.plain_text);
```
Mkassabov: why is toRedacted needed? won't this always be a plain_text password since it comes from api.
Also we should look into redacting on the distilled side

packages/alchemy/src/Planetscale/Postgres/PostgresDatabase.ts
```diff
@@ -0,0 +1,372 @@
...
+  /**
+   * PostgreSQL major version. Defaults to the latest available major
+   * version.
+   */
+  majorVersion?: string;
+
+  /**
+   * CPU architecture. Cannot be changed after creation.
+   */
+  arch?: "x86" | "arm";
```
Mkassabov, line 49: document the default

packages/alchemy/src/Planetscale/Postgres/PostgresDatabase.ts
```diff
@@ -0,0 +1,372 @@
...
+              migrationsDir: output?.migrationsDir ?? olds?.migrationsDir,
+              migrationsTable: output?.migrationsTable ?? olds?.migrationsTable,
+              migrationsHashes: output?.migrationsHashes ?? {},
+              importHashes: output?.importHashes ?? {},
+              // Kind validation is deferred to reconcile, where the
+              // mismatch is surfaced as a typed `PlanetscaleConflict`.
+              // Read's job is just existence + ownership routing.
+              kind: (data.kind === "postgresql"
+                ? "postgresql"
+                : data.kind) as "postgresql",
```
Mkassabov: same thing as in MysqlDatabase. is this the right place to do this? CC @sam-goodwin

packages/alchemy/src/Planetscale/Postgres/PostgresDatabase.ts
```diff
@@ -0,0 +1,372 @@
...
+              clusterSize: output?.clusterSize ?? "",
+              arch: output?.arch ?? "x86",
```
Mkassabov: is it really possible for planetscale not return arch here or is that because they have poor typing for mysql/postgres.

might prefer this actually be `!` if thats the case

packages/alchemy/src/Planetscale/AuthProvider.ts
```diff
@@ -0,0 +1,281 @@
...
+        );
+
+    const prettyPrint = (profileName: string, config: PlanetscaleAuthConfig) =>
+      resolveCredentials(profileName, config).pipe(
+        Effect.tap((creds) => {
+          const sourceStr = creds.source.details
+            ? `${creds.source.type} - ${creds.source.details}`
+            : creds.source.type;
+          return Effect.all([
+            Console.log(`  tokenId: ${creds.tokenId}`),
```
Mkassabov: can we redact tokenId too please

packages/alchemy/src/Planetscale/ClusterSize.ts
```diff
@@ -0,0 +1,53 @@
...
+  region?: string;
+}): string {
+  if (input.kind === "postgresql" && !input.size.match(/(AWS|GCP)/)) {
+    // Not all AWS regions start with "aws-", but all GCP regions start with "gcp-".
+    const provider = input.region?.startsWith("gcp") ? "GCP" : "AWS";
+    const arch = (input.arch ?? "x86").toUpperCase();
+    return `${input.size}_${provider}_${arch}`;
+  }
+  return input.size;
+}
```
Mkassabov: would prefer these not be unified across mysql/postgres and each get their own.
this means the user can't truly specify a different string if planetscale either makes a custom one for them or changes their format.

Mkassabov:
by the looks of this the planetscale safe migrations feature won't play well with alchemy's migrations.
You would need to see if safe migrations is enabled on production and then if it is, when migrating production specify a reference branch to migrate into main (e.g. a staging branch). probably out of scope here since it touches cross stack stuff. but a `//todo` comment would be nice

examples/cloudflare-planetscale-postgres-drizzle/src/Api.ts
lucasthevenet: I actually didn't know you could use Hyperdrive with MySQL lol, will do!

packages/alchemy/src/Planetscale/MySQL/MySQLDatabase.ts
sam-goodwin: I don't understand, why can't we retrieve this in read? read's job is to refresh as many properties as it can.

packages/alchemy/src/Sql/SqlFile.ts
sam-goodwin, line 1: Is this duplicated with neon? Do we need a src/Sql/ folder to share stuff like this?

packages/alchemy/src/Planetscale/SqlFile.ts
```diff
@@ -0,0 +1,95 @@
...
+import * as FileSystem from "effect/FileSystem";
+import * as Path from "effect/Path";
+
+export interface SqlFile {
+  id: string;
+  sql: string;
+  hash: string;
+}
+
+export const DEFAULT_MIGRATIONS_TABLE = "planetscale_migrations";
```
sam-goodwin: Is this actually what we want? Is this what other migrators do? Why does it need planetscale in the name?

Mkassabov: usually its something like `__migrations` or `__drizzle_migrations` or `_prisma_migrations` I'd be in favor of `__alchemy_migrations` but IDC too much, as long as it says migrations and isn't offensive i think its fine

packages/alchemy/src/Planetscale/MySQL/MySQLDatabase.ts
Mkassabov: well this is the mysql resource, so we know it should always be mysql.
the planetscale api treats mysql/postgresql as the same thing so that particular api can theoretically return `postgresql`.

only way for that to happen would be for alchemy to deploy a mysql, then somebody manually removes that and creates a new one with the same name using postgres - which is undefined behavior for sure.

I don't know how we should handle that case

packages/alchemy/src/Planetscale/MySQL/MySQLPassword.ts
lucasthevenet: I think I added this because Planetscale Distilled API says plain_text can be redacted or also string. Same thing happens when we create a PostgresRole. 
image

packages/alchemy/src/Planetscale/MySQL/MySQLMigrations.ts
lucasthevenet: It doesn't, we should fix the distilled schema to make it always a string when returned by createPassword. 
image
In the meantime I'll remove this and use !

packages/alchemy/src/Planetscale/MySQL/MySQLMigrations.ts
lucasthevenet: what do you propose we do in this case? passwords and roles in migrations are created with a ttl so they would just eventually expire if deletion fails, but I get that is not ideal either.

Mkassabov:
oh damn claude squashed everything doing a rebase :(

packages/alchemy/src/Planetscale/MySQL/MySQLMigrations.ts
Mkassabov: Hardened the cleanup path in [`0c035758`](https://github.com/alchemy-run/alchemy-effect/pull/113/commits/0c035758) (mirrored the same change in `PostgresMigrations.ts` since it had the identical pattern):

- `NotFound` is now treated as success — already deleted, nothing to do.
- Other failures retry with exponential backoff (500ms base, up to 5 attempts).
- TTL shortened from `3600s` → `600s` to bound the worst-case orphan window (migrations take seconds, 10 minutes is plenty of headroom).
- If cleanup still fails after retries, we log a warning with `organization` / `database` / `branch` / `id` so an operator can clean up manually, and let the parent migration succeed (the actual migration work already finished — failing here would be punishing a successful run for a transient hiccup).

-claude

Mkassabov: idk maybe there is some way we can use alchemy here and shove it in the state store as a resource that needs to be deleted. maybe a problem for another day

CC @sam-goodwin

packages/alchemy/src/Planetscale/MySQL/MySQLDatabase.ts
Mkassabov: Took Sam's read here. The cast was a lie — if the cloud actually held the other engine, read claimed `kind: "mysql"` while reconcile silently caught the mismatch a few hundred ms later. That's the wrong place to surface it.

Fixed in [`b91c9cf8`](https://github.com/alchemy-run/alchemy-effect/pull/113/commits/b91c9cf8):

- `read` now fails with `PlanetscaleConflict` directly when the observed `data.kind` doesn't match the resource type, with the same actionable message reconcile uses.
- Dropped the `as "mysql"` / `as "postgresql"` cast and replaced `Effect.map` with `Effect.flatMap` so the failure path is typed.
- Kept reconcile's check in place as defense-in-depth (the cloud can change between read and reconcile — TOCTOU).
- Applied symmetrically to `PostgresDatabase.ts`.

Re Mkassabov's "undefined behavior" point: yeah, the underlying scenario (manual replace with a different engine) is undefined. But surfacing it from read gives the user the right message at the earliest point, instead of relying on a downstream stage to catch a lie we told them.

-claude

packages/alchemy/src/Planetscale/MySQL/MySQLPassword.ts
Mkassabov: You were right on both counts. The bidirectional union was the root cause: `SensitiveString` was typed `string | Redacted<string>` for inputs (so callers can pass either), and the same schema was reused on responses, leaving consumers to coerce a runtime-impossible union at every read site.

Fixed upstream in [distilled#300](https://github.com/alchemy-run/distilled/pull/300):

- Split `Sensitive` into input-friendly (`SensitiveString`) and output-strict (`SensitiveOutputString`) variants. Output decodes to `Redacted<A>` only — no union.
- Generator now picks the strict variant for response schemas and the input-friendly one for request bodies via a `direction` flag.
- Released as `@distilled.cloud/*@0.21.3`.

Bumped distilled in alchemy and dropped the helpers in [`f86a2426`](https://github.com/alchemy-run/alchemy-effect/pull/113/commits/f86a2426):

- `MySQLPassword.toRedacted` — gone
- `MySQLMigrations.toRedacted` — gone
- `PostgresMigrations.toRedacted` — gone
- `PostgresRole.ts` `Redacted.make(Redacted.isRedacted(...) ? ... : ...)` dance — gone

Call sites now read `plaintext = created.plain_text` directly. `tsc -b` clean.

-claude

packages/alchemy/src/Planetscale/Postgres/PostgresDatabase.ts
Mkassabov: Resolved alongside the MySQL one in [`b91c9cf8`](https://github.com/alchemy-run/alchemy-effect/pull/113/commits/b91c9cf8) — `read` now fails with `PlanetscaleConflict` on kind mismatch instead of casting `data.kind` to a lie. See [the reply on the MySQLDatabase thread](https://github.com/alchemy-run/alchemy-effect/pull/113#discussion_r3257749887) for the rationale.

-claude

packages/alchemy/src/Planetscale/SqlFile.ts
Mkassabov: Renamed to `__alchemy_migrations` for the Planetscale resources (Mkassabov's suggestion). Heads up: Neon still uses `neon_migrations` and Cloudflare D1 still uses `d1_migrations` — those weren't touched in this PR since changing them would force a migration of every existing deployment's tracking table.

If you want to unify all three under `__alchemy_migrations`, happy to do that in a follow-up with a clear state-migration path.

-claude

Mkassabov (commented):
I think this is good. the last thing is that this flow isn't durable on create since we don't update state between the create call and the db being stable; so if a user interrupts that (kinda likely given these can take a while to provision and it can look like something going wrong) they can get a hung db.

@sam-goodwin is the right solution here to move database creation api call to pre-create so we save the name immediately and then reconcile handles waiting for it to be stable. not sure how that works when reconcile hits a state where the database doesn't exist though? is there some other way of having reconcile update state halfway through?

sam-goodwin:
Missing docs

packages/alchemy/src/Planetscale/Branch.ts
```diff
@@ -0,0 +1,509 @@
...
+                  organization,
+                  database: databaseName,
+                  branch: branchName,
+                })
+              : yield* demoteBranch({
+                  organization,
+                  database: databaseName,
+                  branch: branchName,
+                });
+          }
```
john-royal: isProduction/promote/demote is for Vitess/MySQL only: https://discord.com/channels/1134636291691126885/1392195845016912105/1448421285314035906

I wonder if we should omit this as an option for Postgres.

This message was from a while ago so my knowledge could be outdated.

packages/alchemy/src/Planetscale/Util.ts
```diff
@@ -1,8 +1,18 @@
...
 import * as Schedule from "effect/Schedule";
 
+/**
+ * Subset of {@link ScopedPlanStatusSession} used by the polling helpers.
+ * Typed structurally so callers can pass the full session without
+ * pulling the CLI types into this provider module.
+ */
+export interface Noter {
+  note: (message: string) => Effect.Effect<void>;
+}
```
sam-goodwin: get rid of this

packages/alchemy/src/Planetscale/Util.ts
```diff
@@ -63,23 +73,36 @@ export const pollUntil = <A, E, R>(
...
-            new NotReady({ description: `branch "${branch}" not found yet` }),
-          ),
+    Effect.gen(function* () {
+      if (session) {
+        const n = yield* Ref.updateAndGet(attempts, (i) => i + 1);
+        yield* session.note(
+          n === 1
+            ? `Waiting for branch "${branch}" to be ready...`
+            : `Waiting for branch "${branch}" to be ready... (${n})`,
+        );
```
sam-goodwin: better to use seconds instead of count and also provide expectations (e.g. "this can take a few minutes")

Mkassabov (approved):
I'm calling this approved. @sam-goodwin should approve the docs changes

@lucasthevenet Thank you so much for all the hard work and tolerating back and forth from us on this!

Mkassabov:
one quick note. the mysql example can't bundle because mysql2 has some cjs deps.
the actual resources work fine.

I think fixing our bundler is out of scope for this PR (and its not even this repo lol), and I don't want to delay planetscale (especially since the planetscale serverless driver still works for mysql, and postgres works fine).

lucasthevenet:
> I'm calling this approved. @sam-goodwin should approve the docs changes
> 
> @lucasthevenet Thank you so much for all the hard work and tolerating back and forth from us on this!

No problem! Happy to contribute :)

website/src/content/docs/tutorial/cloudflare/neon-hyperdrive.mdx
sam-goodwin, line 1: should we rename the file?

Mkassabov (approved):
finally! thank you!

PR #117, opened 2026-04-30
feat(cloudflare/Worker): Add option (`bundle:false`) to skip bundling by Rolldown

czxtm:
I discovered this when trying to deploy Fumadocs. This PR unblocks the ability to deploy Fumadocs and probably many other Next.js apps using `@opennextjs/cloudflare`:

When deploying a Worker whose `main` already points at a complete, runtime-ready ESM bundle produced by an external tool (OpenNext, wrangler, custom build pipelines, etc.), alchemy still runs that artifact through `cloudflareRolldown`.

`isExternal: true` already disables alchemy's Effect-wrapping virtual entry, but `prepareBundle` still calls `Bundle.build(...)` with `cloudflareRolldown` on the way through. There is currently no way to opt out of that re-bundle.

This adds a new `WorkerProps.bundle?: boolean` (default `true`). When set to `false`, `prepareBundle` reads `props.main` directly and returns a synthetic single-file `Bundle.BundleOutput` whose `content` is the file's bytes verbatim and whose `hash` is sha256(bytes). The downstream upload path is unchanged.

The accompanying integration test works as follows:

1. Deploy a hand-written ESM bundle that contains a SENTINEL comment
2. Verify the resulting `worker.hash?.bundle` equals `sha256(sourceBytes)` (which is impossible to satisfy if rolldown ran on the file)
3. Round-trip a request through the deployed Worker to confirm the literal sentinel string survives.

A reproduction of a failing deploy of Fumadocs (using `bunx create-fumadocs-app`) is available here: https://github.com/czxtm/repro-alchemy-bundle-false

The [main](https://github.com/czxtm/repro-alchemy-bundle-false) branch is a failing reproduction while the [fix-bundle-false](https://github.com/czxtm/repro-alchemy-bundle-false/tree/fix-bundle-false) branch contains a patched version of alchemy that includes the changes introduced in this PR. 

If `isExternal: true` is actually supposed to skip the re-bundling as well and just needs a bugfix, then this PR is not the right solution and can be closed.


sam-goodwin:
cc @john-royal

sam-goodwin:
> If isExternal: true is actually supposed to skip the re-bundling as well and just needs a bugfix, then this PR is not the right solution and can be closed.

isExternal is Effect vs Async style. It should not skip bundling

sam-goodwin:
Can we add an example that exercises this code path? Maybe examples/cloudflare-nextjs, is that what you're doing?

czxtm:
Yeah, I'll move the one in the repro to examples

czxtm:
@sam-goodwin is this still useful?

sam-goodwin:
> @sam-goodwin is this still useful?

@john-royal can you look at this and decide what we should do here?

czxtm:
@john-royal I know #142 is a no-go but can we push this one through, assuming this hasn't been implemented via some other means already? This would be enough to allow me (and others) to deploy nextjs without patching.

john-royal:
> @john-royal I know #142 is a no-go but can we push this one through, assuming this hasn't been implemented via some other means already? This would be enough to allow me (and others) to deploy nextjs without patching.

Yes, so sorry I forgot about this one.

Here's what I'd need in order to merge this:
- Remove the Next.js example (apologies for the flip-flop on that one, but it's best to keep Wrangler out of our examples).
- The regex parsing of code files to resolve dependencies is a little scary to me - even if it passes every test, using regex to parse code is inherently flaky and likely to surface edge cases in the future. I would be open to either of the following:
  1. Resolve files from patterns, similar to Alchemy V1 (implementation [here](https://github.com/alchemy-run/alchemy/blob/a56c9b008f2c21040bbec5eb6024fb586fee4991/alchemy/src/cloudflare/worker-bundle.ts#L183-L242)) and Wrangler.
  2. Use `oxc-parser` and `oxc-resolver` to properly parse files and resolve dependencies.

I lean towards the first option for simplicity, but I would be happy with either.

agcty:
We picked this up in #592 with the review feedback here applied — pattern-based module resolution per the v1 `WorkerBundleSource.FS` / Wrangler `findAdditionalModules` implementations linked above, and no example or wrangler dependency. Context: the React Router RSC investigation in #590 made a no-rebundle deploy path our critical need too. Thanks @czxtm for the groundwork — happy to adjust #592 in whatever direction works, or hand it back if you'd rather carry this one forward.

john-royal:
Replacing with #592

czxtm:
Yay! Thanks guys

PR #122, opened 2026-04-30
fix(core): treat raw Resource refs as upstream dependencies

m9tdev:
`upstreamAny`, `upstream` and `resolveUpstream` only walked Output expressions when collecting upstream resources, missing Resources passed directly as prop values (e.g. `prop: someResource`). Only Output-mediated refs (like `someResource.someAttribute` accessed inside `Output.interpolate`) were tracked.

The plan's per-resource `downstream` map is built from this traversal, and the destroy graph in `Apply.collectGarbage` uses `downstream` to order deletions in reverse. The gap meant a Resource and a downstream Resource that referenced it via a direct ref (rather than through an attribute access) could be deleted in parallel.

Detect raw Resource refs in all three traversers and key them under `{ [resource.FQN]: resource }`, matching the existing ResourceExpr path.

### DIFF
--- a/packages/alchemy/src/Output.ts
+++ b/packages/alchemy/src/Output.ts
@@ -499,7 +499,9 @@ export const upstreamAny = (
 ): {
   [ID in string]: Resource;
 } => {
-  if (isExpr(value)) {
+  if (isResource(value)) {
+    return { [value.FQN]: value as Resource };
+  } else if (isExpr(value)) {
     return upstream(value);
   } else if (Array.isArray(value)) {
     return Object.assign({}, ...value.map(resolveUpstream));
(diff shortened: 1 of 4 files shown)


sam-goodwin:
How did you run into this problem? Can you add a test to plan.test.ts and (if relevant) the apply.test.ts

m9tdev:
Hello @sam-goodwin! I'm trying to add Docker as provider and ran into this when it tried to delete the images used for the containers, before having finished deleting the containers.

I have a PoC running that uses the docker cli, but now adding the tcp api to `alchemy-run/distilled` and then I'll convert it to use that instead.

sam-goodwin:
Thanks for the patch. I added the tests. Looking forward to docker stuff!

PR #124, opened 2026-04-30
fix(cli): filter bun's "not in project directory" watcher warning

Mkassabov:
Use `@alchemy.run/node-utils`' vendored `foregroundChild` with its `stderrFilter` option to drop bun's hard-coded "not in project directory" watcher warnings — upstream `foreground-child` hardcodes `stdio = [0, 1, 2]` so there's no way to intercept stderr without owning the spawn call.

```diff
-import { foregroundChild } from "foreground-child";
+import { foregroundChild } from "@alchemy.run/node-utils";

-foregroundChild(runtime, args);
+foregroundChild(runtime, args, {
+  stderrFilter: (line) =>
+    !line.includes("is not in the project directory and will not be watched"),
+});
```

- Fix Windows Ctrl-C: handle `STATUS_CONTROL_C_EXIT` (3221225786) in addition to `SIGINT`
- Fix subpath imports (`@alchemy.run/node-utils/exit-hook` → `@alchemy.run/node-utils`)
- Drop `foreground-child` dependency

### DIFF
--- a/packages/alchemy/src/Auth/Lock.ts
+++ b/packages/alchemy/src/Auth/Lock.ts
@@ -1,4 +1,4 @@
-import { lock } from "@alchemy.run/node-utils/lockfile";
+import { lock } from "@alchemy.run/node-utils";
 import * as Effect from "effect/Effect";
 import * as fs from "node:fs/promises";
 import * as path from "pathe";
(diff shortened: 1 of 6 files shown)


packages/alchemy/src/Auth/Lock.ts
```diff
@@ -1,4 +1,4 @@
-import { lock } from "@alchemy.run/node-utils/lockfile";
```
john-royal, line 1: Why was this changed? The subpath imports fix an import resolution bug in some Node setups.

PR #131, opened 2026-05-01
fix(cloudflare): bootstrap regressions in SecretsStore and StateStore

sam-goodwin:
## Summary

`bun alchemy bootstrap cloudflare` was failing on any Cloudflare account that didn't already have a Secrets Store, blocking new users from deploying anything that needs alchemy state.

### Bug 1: SecretsStore create — `invalid_json_body`

Cloudflare's REST API for `POST /accounts/{id}/secrets_store/stores` takes a single `{ \"name\": \"...\" }` JSON object. The auto-generated `@distilled.cloud/cloudflare` SDK models the request body as `[{name}]` (an array), so we were sending `[{\"name\": \"...\"}]` and Cloudflare rejected it with code `1001 invalid_json_body`.

**Why nobody on the team could repro:** `SecretsStore` always tries to adopt an existing store first — Cloudflare allows exactly one Secrets Store per account, and deleting it permanently destroys every secret inside, so the resource is treated as account-level infrastructure that's adopted rather than re-created. On any account that had ever provisioned the store before (i.e. every internal account), the adopt path short-circuited and the broken create call never ran. Fresh accounts hit the create path on the very first bootstrap and got stuck.

**Fix:** bypass the SDK for this one call and POST the object shape via `HttpClient` directly. `MaximumStoresExceeded` is still surfaced as the SDK's tagged error so the existing adoption-race recovery is unchanged.

### Bug 2: StateStore worker — `[MISSING_EXPORT] \"default\" is not exported by \"bin/alchemy.js\"`

`StateStore/Api.ts` used `main: import.meta.path` to point the worker bundler at its own source file. That works in dev, but the published CLI is bundled by tsdown into a single `bin/alchemy.js` — `Api.ts` gets inlined, `import.meta.path` resolves to the CLI bundle (which has no `default` export), and the worker bundler crashes.

**Fix:** resolve the source path defensively. Use `import.meta.url` when it still ends in `.ts` (dev / monorepo), otherwise resolve back to the published `src/Cloudflare/StateStore/Api.ts` next to the bundled CLI.

## Tests

Added unit tests that mock `HttpClient` and assert the wire-level body shape, accountId encoding, and `MaximumStoresExceeded` tagged-error contract for the create path. The body-shape assertion is the regression check — the bug was invisible to the type system because the SDK's incorrect schema made the array body type-correct.

## Test plan

- [x] `bun x vitest run packages/alchemy/test/Cloudflare/SecretsStore` (3/3 passing)
- [x] `tsc --noEmit` clean
- [ ] Manual: `bun alchemy bootstrap cloudflare --force` against a fresh Cloudflare account (no existing Secrets Store)

Credit: reported by @garrett.

Made with [Cursor](https://cursor.com)


packages/alchemy/src/Cloudflare/SecretsStore/SecretsStore.ts
```diff
@@ -139,3 +146,85 @@ export const SecretsStoreProvider = () =>
...
+ * (code `1001`). The auto-generated `@distilled.cloud/cloudflare`
+ * `createStore` operation incorrectly models the body as an array of
+ * objects, so any call through it fails on accounts that don't already
+ * have a Secrets Store. This helper bypasses the SDK and posts the
+ * correct shape directly. The returned response shape matches the
+ * actual API: `result` is a single store object (not an array).
+ *
+ * Surfaces `MaximumStoresExceeded` as a tagged error so the caller can
+ * `Effect.catchTag` the same as if it had used the SDK's typed error.
+ */
```
sam-goodwin: Oh this is a bug in distilled?

PR #139, opened 2026-05-01
feat(aws): add AWS API Gateway REST resources

CharLEEx:
## Summary

Adds AWS API Gateway (REST API v1) resource providers to `packages/alchemy`, plus lifecycle tests, an `examples/aws-rest-api` reference stack, and tutorial content (`website/src/content/docs/tutorial/aws/api-gateway.mdx`).

## Includes

- REST API primitives: `RestApi`, `Deployment`, `Stage`, `Method`, `GatewayResource`, `Authorizer`, `ApiKey`, `UsagePlan`, `UsagePlanKey`, `DomainName`, `BasePathMapping`, `GatewayResponse`, `Account`, `VpcLink`, shared helpers
- Provider registration and `normalizeTags` export used by ApiGateway tagging
- `scripts/audit-service.ts` mapping for `apigateway`
- Root `tsconfig.json` project reference for `examples/aws-rest-api`

## Notes

- Integration tests require AWS credentials (e.g. configured profile in CI).
- Optional `VpcLink` test can be enabled with `ALCHEMY_TEST_VPC_LINK_TARGET_ARN`.

Refs: none

### DIFF
--- a/packages/alchemy/src/AWS/ApiGateway/ApiKey.ts
+++ b/packages/alchemy/src/AWS/ApiGateway/ApiKey.ts
@@ -1,0 +1,193 @@
+import { Region } from "@distilled.cloud/aws/Region";
+import * as ag from "@distilled.cloud/aws/api-gateway";
+import * as Effect from "effect/Effect";
+import { deepEqual, isResolved } from "../../Diff.ts";
+import * as Provider from "../../Provider.ts";
+import { Resource } from "../../Resource.ts";
+import type { Providers } from "../Providers.ts";
+import { createInternalTags } from "../../Tags.ts";
+
+import { apiKeyArn, syncTags } from "./common.ts";
+
+export interface ApiKeyProps {
+  name?: string;
+  description?: string;
+  enabled?: boolean;
+  generateDistinctId?: boolean;
+  /**
+   * Write-only value when creating; never stored in resource state or outputs.
+   */
+  value?: string;
+  stageKeys?: ag.StageKey[];
+  customerId?: string;
+  tags?: Record<string, string>;
+}
+
+export interface ApiKey extends Resource<
+  "AWS.ApiGateway.ApiKey",
+  ApiKeyProps,
+  {
+    id: string;
+    name: string | undefined;
+    enabled: boolean | undefined;
+    tags: Record<string, string>;
+  },
+  never,
+  Providers
+> {}
+
+/**
+ * API Gateway API key for usage plans and `apiKeyRequired` methods.
+ *
+ * @section API keys
+ * @example Generated key
+ * ```typescript
+ * const key = yield* ApiGateway.ApiKey("PartnerKey", {
+ *   name: "partner",
+ *   generateDistinctId: true,
+ * });
+ * ```
+ */
+const ApiKeyResource = Resource<ApiKey>("AWS.ApiGateway.ApiKey");
+
+export { ApiKeyResource as ApiKey };
+
+const toTags = (tags: ag.ApiKey["tags"]) =>
+  Object.fromEntries(
+    Object.entries(tags ?? {}).filter(
+      (e): e is [string, string] => e[1] !== undefined,
+    ),
+  );
+
+export const ApiKeyProvider = () =>
+  Provider.effect(
+    ApiKeyResource,
+    Effect.gen(function* () {
+      const awsRegion = yield* Region;
+
+      return {
+        stables: ["id"] as const,
+        diff: Effect.fn(function* ({ news: newsIn, olds }) {
+          if (!isResolved(newsIn)) return;
+          const news = newsIn as ApiKeyProps;
+          if (
+            news.value !== olds.value ||
+            news.customerId !== olds.customerId
+          ) {
+            return { action: "replace" } as const;
+          }
+        }),
+        read: Effect.fn(function* ({ output }) {
+          if (!output?.id) return undefined;
+          const k = yield* ag
+            .getApiKey({ apiKey: output.id, includeValue: false })
+            .pipe(
+              Effect.catchTag("NotFoundException", () =>
+                Effect.succeed(undefined),
+              ),
+            );
+          if (!k?.id) return undefined;
+          return {
+            id: k.id,
+            name: k.name,
+            enabled: k.enabled,
+            tags: toTags(k.tags),
+          };
+        }),
+        create: Effect.fn(function* ({ id, news: newsIn, session }) {
+          if (!isResolved(newsIn)) {
+            return yield* Effect.die("ApiKey props were not resolved");
+          }
+          const news = newsIn as ApiKeyProps;
+          const internalTags = yield* createInternalTags(id);
+          const allTags = { ...news.tags, ...internalTags };
+
+          const k = yield* ag.createApiKey({
+            name: news.name,
+            description: news.description,
+            enabled: news.enabled,
+            generateDistinctId: news.generateDistinctId,
+            value: news.value,
+            stageKeys: news.stageKeys,
+            customerId: news.customerId,
+            tags: allTags,
+          });
+          if (!k.id) return yield* Effect.die("createApiKey missing id");
+          yield* session.note(`Created API key ${k.id}`);
+          const full = yield* ag.getApiKey({
+            apiKey: k.id,
+            includeValue: false,
+          });
+          if (!full.id) return yield* Effect.die("getApiKey missing id");
+          return {
+            id: full.id,
+            name: full.name,
+            enabled: full.enabled,
+            tags: toTags(full.tags),
+          };
+        }),
+        update: Effect.fn(function* ({ id, news: newsIn, output, session }) {
+          if (!isResolved(newsIn)) {
+            return yield* Effect.die("ApiKey props were not resolved");
+          }
+          const news = newsIn as ApiKeyProps;
+          const patches: ag.PatchOperation[] = [];
+          if (news.name !== output.name) {
+            patches.push({
+              op: "replace",
+              path: "/name",
+              value: news.name ?? "",
+            });
+          }
+          if (news.description !== undefined) {
+            patches.push({
+              op: "replace",
+              path: "/description",
+              value: news.description ?? "",
+            });
+          }
+          if (news.enabled !== undefined && news.enabled !== output.enabled) {
+            patches.push({
+              op: "replace",
+              path: "/enabled",
+              value: String(news.enabled),
+            });
+          }
+          if (patches.length > 0) {
+            yield* ag.updateApiKey({
+              apiKey: output.id,
+              patchOperations: patches,
+            });
+          }
+
+          const internalTags = yield* createInternalTags(id);
+          const newTags = { ...news.tags, ...internalTags };
+          if (!deepEqual(output.tags, newTags)) {
+            yield* syncTags({
+              resourceArn: apiKeyArn(awsRegion, output.id),
+              oldTags: output.tags,
+              newTags,
+            });
+          }
+
+          yield* session.note(`Updated API key ${output.id}`);
+          const full = yield* ag.getApiKey({
+            apiKey: output.id,
+            includeValue: false,
+          });
+          return {
+            id: output.id,
+            name: full.name,
+            enabled: full.enabled,
+            tags: toTags(full.tags),
+          };
+        }),
+        delete: Effect.fn(function* ({ output, session }) {
+          yield* ag
+            .deleteApiKey({ apiKey: output.id })
+            .pipe(Effect.catchTag("NotFoundException", () => Effect.void));
+          yield* session.note(`Deleted API key ${output.id}`);
+        }),
+      };
+    }),
+  );
--- a/packages/alchemy/src/AWS/ApiGateway/ApiKey.ts
+++ b/packages/alchemy/src/AWS/ApiGateway/ApiKey.ts
@@ -1,25 +1,55 @@
 import { Region } from "@distilled.cloud/aws/Region";
 import * as ag from "@distilled.cloud/aws/api-gateway";
 import * as Effect from "effect/Effect";
+import * as Redacted from "effect/Redacted";
+import * as Stream from "effect/Stream";
 import { deepEqual, isResolved } from "../../Diff.ts";
+import { createPhysicalName } from "../../PhysicalName.ts";
 import * as Provider from "../../Provider.ts";
 import { Resource } from "../../Resource.ts";
 import type { Providers } from "../Providers.ts";
-import { createInternalTags } from "../../Tags.ts";
+import { createInternalTags, hasAlchemyTags, tagRecord } from "../../Tags.ts";
 
 import { apiKeyArn, syncTags } from "./common.ts";
 
 export interface ApiKeyProps {
+  /**
+   * Friendly name for the API key.
+   *
+   * If omitted, Alchemy generates a deterministic physical name from the
+   * stack, stage, logical ID, and instance ID.
+   */
   name?: string;
+  /**
+   * Human-readable description shown in API Gateway.
+   */
   description?: string;
+  /**
+   * Whether clients can use the key.
+   *
+   * @default true
+   */
   enabled?: boolean;
+  /**
+   * Appends a distinct suffix to the generated key value when AWS generates it.
+   */
   generateDistinctId?: boolean;
   /**
    * Write-only value when creating; never stored in resource state or outputs.
+   * Wrap with `Redacted.make` so state encoding preserves redaction.
    */
-  value?: string;
+  value?: Redacted.Redacted<string>;
+  /**
+   * Stage associations to attach directly to this API key.
+   */
   stageKeys?: ag.StageKey[];
+  /**
+   * External customer identifier associated with the key.
+   */
   customerId?: string;
+  /**
+   * User-defined tags. Alchemy internal tags are merged automatically.
+   */
   tags?: Record<string, string>;
 }
 
@@ -43,7 +73,6 @@ export interface ApiKey extends Resource<
  * @example Generated key
  * ```typescript
  * const key = yield* ApiGateway.ApiKey("PartnerKey", {
- *   name: "partner",
  *   generateDistinctId: true,
  * });
  * ```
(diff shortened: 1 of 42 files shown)


packages/alchemy/src/AWS/ApiGateway/ApiKey.ts
```diff
@@ -0,0 +1,193 @@
...
+  enabled?: boolean;
+  generateDistinctId?: boolean;
+  /**
+   * Write-only value when creating; never stored in resource state or outputs.
+   */
+  value?: string;
+  stageKeys?: ag.StageKey[];
+  customerId?: string;
+  tags?: Record<string, string>;
+}
```
sam-goodwin, line 54: Add thorough docs on each property (but not the interface)

packages/alchemy/src/AWS/ApiGateway/ApiKey.ts
```diff
@@ -0,0 +1,193 @@
...
+const ApiKeyResource = Resource<ApiKey>("AWS.ApiGateway.ApiKey");
+
+export { ApiKeyResource as ApiKey };
+
+const toTags = (tags: ag.ApiKey["tags"]) =>
+  Object.fromEntries(
+    Object.entries(tags ?? {}).filter(
+      (e): e is [string, string] => e[1] !== undefined,
+    ),
+  );
```
sam-goodwin: Does this belong in Tags.ts or do we already have it?

packages/alchemy/src/AWS/ApiGateway/ApiKey.ts
```diff
@@ -0,0 +1,193 @@
...
+
+      return {
+        stables: ["id"] as const,
+        diff: Effect.fn(function* ({ news: newsIn, olds }) {
+          if (!isResolved(newsIn)) return;
+          const news = newsIn as ApiKeyProps;
+          if (
+            news.value !== olds.value ||
+            news.customerId !== olds.customerId
+          ) {
```
sam-goodwin, line 125: Add some // comments helping justify important decisions like this

packages/alchemy/src/AWS/ApiGateway/ApiKey.ts
```diff
@@ -0,0 +1,193 @@
...
+        create: Effect.fn(function* ({ id, news: newsIn, session }) {
+          if (!isResolved(newsIn)) {
+            return yield* Effect.die("ApiKey props were not resolved");
+          }
+          const news = newsIn as ApiKeyProps;
+          const internalTags = yield* createInternalTags(id);
+          const allTags = { ...news.tags, ...internalTags };
+
+          const k = yield* ag.createApiKey({
+            name: news.name,
```
sam-goodwin: Make the name optional and use createPhysicalName to generate a deterministic, unique name that rotates on each place. 

Plenty of patterns in this repo to draw from.

packages/alchemy/src/AWS/ApiGateway/ApiKey.ts
```diff
@@ -0,0 +1,193 @@
...
+            description: news.description,
+            enabled: news.enabled,
+            generateDistinctId: news.generateDistinctId,
+            value: news.value,
+            stageKeys: news.stageKeys,
+            customerId: news.customerId,
+            tags: allTags,
```
sam-goodwin: If tags are added as part of create, that's great! It means we can use tags to attach alchemy metadata and use it to recover from "partial creates". 

Take a look at other AWS resources   

They all use Tags.ts to catch and recover from conflict exceptions

packages/alchemy/src/AWS/ApiGateway/Authorizer.ts
```diff
@@ -0,0 +1,210 @@
...
+  restApiId: Input<string>;
+  name: string;
+  type: ag.AuthorizerType;
+  providerARNs?: string[];
+  authType?: string;
+  authorizerUri?: string;
+  authorizerCredentials?: string;
+  identitySource?: string;
+  identityValidationExpression?: string;
+  authorizerResultTtlInSeconds?: number;
```
sam-goodwin, line 54: Docs

packages/alchemy/src/AWS/ApiGateway/ApiKey.ts
```diff
@@ -0,0 +1,193 @@
...
+
+export interface ApiKeyProps {
+  name?: string;
+  description?: string;
+  enabled?: boolean;
+  generateDistinctId?: boolean;
+  /**
+   * Write-only value when creating; never stored in resource state or outputs.
+   */
+  value?: string;
```
sam-goodwin: Is this a sensitive value? If so, wrap in `Redacted<string>`

packages/alchemy/src/AWS/ApiGateway/Authorizer.ts
```diff
@@ -0,0 +1,210 @@
...
+            news.name !== olds.name ||
+            news.type !== olds.type
+          ) {
+            return { action: "replace" } as const;
+          }
+          if (!deepEqual(news.providerARNs, olds.providerARNs)) {
+            return { action: "replace" } as const;
+          }
+          if (news.authType !== olds.authType) {
+            return { action: "replace" } as const;
```
sam-goodwin, line 118: // comments to add some concise details explaining the decisions

packages/alchemy/src/AWS/ApiGateway/DomainName.ts
```diff
@@ -0,0 +1,260 @@
...
+const DomainNameResource = Resource<DomainName>("AWS.ApiGateway.DomainName");
+
+export { DomainNameResource as DomainName };
+
+const toTags = (tags: ag.DomainName["tags"]) =>
+  Object.fromEntries(
+    Object.entries(tags ?? {}).filter(
+      (e): e is [string, string] => e[1] !== undefined,
+    ),
+  );
```
sam-goodwin: Duplicate. Use Tags.ts. Add to it if it's missing something

packages/alchemy/src/AWS/ApiGateway/DomainName.ts
```diff
@@ -0,0 +1,260 @@
...
+              op: news.certificateArn === undefined ? "remove" : "replace",
+              path: "/certificateArn",
+              value: news.certificateArn,
+            });
+          }
+          if (patches.length > 0) {
+            yield* ag.updateDomainName({
+              domainName: output.domainName,
+              patchOperations: patches,
+            });
```
sam-goodwin: I notice a discrete lack of catching and retries. Are there edge cases? Errors we may want to retry

packages/alchemy/src/AWS/ApiGateway/Method.ts
```diff
@@ -0,0 +1,421 @@
...
+/**
+ * Integration configuration for an API Gateway method (passed to `putIntegration`).
+ */
+export interface MethodIntegrationProps {
+  type: ag.IntegrationType;
+  integrationHttpMethod?: string;
+  uri?: Input<string>;
+  connectionType?: ag.ConnectionType;
+  connectionId?: string;
+  credentials?: string;
```
sam-goodwin, line 23: Redacted?

packages/alchemy/src/AWS/ApiGateway/RestApi.ts
```diff
@@ -0,0 +1,304 @@
...
+ * @example Regional REST API
+ * ```typescript
+ * import * as ApiGateway from "alchemy/AWS/ApiGateway";
+ *
+ * const api = yield* ApiGateway.RestApi("PublicApi", {
+ *   name: "my-api",
+ *   endpointConfiguration: { types: ["REGIONAL"] },
+ * });
+ * ```
+ */
```
sam-goodwin, line 114: This could be more thorough. Please update docs with various sections and examples (these end up on our api reference docs, more illustrative examples the better).

packages/alchemy/src/AWS/ApiGateway/RestApi.ts
```diff
@@ -0,0 +1,304 @@
...
+import * as Effect from "effect/Effect";
+import { deepEqual, isResolved } from "../../Diff.ts";
+import * as Provider from "../../Provider.ts";
+import { Resource } from "../../Resource.ts";
+import type { Providers } from "../Providers.ts";
+import { createInternalTags } from "../../Tags.ts";
+
+import { restApiArn, syncTags } from "./common.ts";
+
+export interface RestApiProps {
```
sam-goodwin, line 13: Can we use an OpenAPI schema?

Any opportunity to integrate with effect http api's generated schema?

Seems a lot easier than using these low level resources.

packages/alchemy/src/AWS/ApiGateway/RestApi.ts
```diff
@@ -0,0 +1,304 @@
...
+      patchReplace("/securityPolicy", news.securityPolicy ?? "TLS_1_0"),
+    );
+  }
+  if (news.endpointAccessMode !== prev.endpointAccessMode) {
+    patches.push(
+      patchReplace("/endpointAccessMode", news.endpointAccessMode ?? ""),
+    );
+  }
+  return patches;
+};
```
sam-goodwin, line 231: Do we have thorough tests for this?

packages/alchemy/src/AWS/ApiGateway/UsagePlan.ts
```diff
@@ -0,0 +1,331 @@
...
+const UsagePlanResource = Resource<UsagePlan>("AWS.ApiGateway.UsagePlan");
+
+export { UsagePlanResource as UsagePlan };
+
+const toTags = (tags: ag.UsagePlan["tags"]) =>
+  Object.fromEntries(
+    Object.entries(tags ?? {}).filter(
+      (e): e is [string, string] => e[1] !== undefined,
+    ),
+  );
```
sam-goodwin: Duplicate again

packages/alchemy/test/AWS/ApiGateway/ApiKey.test.ts
```diff
@@ -0,0 +1,42 @@
...
+import * as Effect from "effect/Effect";
+
+const { test } = Test.make({ providers: AWS.providers() });
+
+test.provider("create and delete API key", (stack) =>
+  Effect.gen(function* () {
+    const key = yield* stack.deploy(
+      Effect.gen(function* () {
+        return yield* AWS.ApiGateway.ApiKey("AgApiKey", {
+          name: "alchemy-test-ag-apikey",
```
sam-goodwin: Don't hard code names in tests. Make sure all names are optional and the resource provider generates a unique one using PhysicalName.ts. Look up the docs for that resource to confirm min/max length, illegal characters, etc

Hard coded names lead to conflicts.

packages/alchemy/test/AWS/ApiGateway/Authorizer.test.ts
```diff
@@ -0,0 +1,7 @@
+import * as AWS from "@/AWS";
+import * as Test from "@/Test/Vitest";
+import { describe } from "@effect/vitest";
+
+Test.make({ providers: AWS.providers() });
+
+describe.skip("ApiGateway.Authorizer", () => {});
```
sam-goodwin: Any reason we are skipping it?

packages/alchemy/test/AWS/ApiGateway/GatewayResponse.test.ts
```diff
@@ -0,0 +1,36 @@
+import * as AWS from "@/AWS";
+import * as Test from "@/Test/Vitest";
+import * as ag from "@distilled.cloud/aws/api-gateway";
+import { expect } from "@effect/vitest";
+import * as Effect from "effect/Effect";
+
+const { test } = Test.make({ providers: AWS.providers() });
+
+test.provider("create and delete gateway response", (stack) =>
```
sam-goodwin, line 9: A lot of tests are not thorough enough. 

Test the update flow too. Test edge cases.

sam-goodwin (changes requested):
Amazing PR. I think we can use this foundation to build an effect native abstraction on top in a separate PR. 

HttpApi schema -> api gateway config

packages/alchemy/src/AWS/ApiGateway/ApiKey.ts
CharLEEx: Addressed in 65fb3dd: added field-level JSDoc for every ApiKey prop, including generated names, write-only redacted values, stage keys, customer ID, enabled state, and tags. I left the interface itself without extra docs per the request.

packages/alchemy/src/AWS/ApiGateway/ApiKey.ts
CharLEEx: Addressed in 65fb3dd: moved this shared conversion into Tags.ts as tagRecord(...) and updated ApiKey, DomainName, UsagePlan, and VpcLink to use it instead of duplicating local helpers.

packages/alchemy/src/AWS/ApiGateway/ApiKey.ts
CharLEEx: Addressed in 65fb3dd and 275af43: added concise comments explaining why generated names are only patched when explicitly provided and why API key value rotation is modeled as replacement since API Gateway does not return the key value after create.

packages/alchemy/src/AWS/ApiGateway/ApiKey.ts
CharLEEx: Addressed in 65fb3dd: ApiKey.name is now optional and uses createPhysicalName({ id, maxLength: 128 }) when omitted, giving deterministic stack/stage/logical-ID based names with the instance suffix.

packages/alchemy/src/AWS/ApiGateway/ApiKey.ts
CharLEEx: Addressed in 65fb3dd: create now attaches Alchemy internal tags and catches ConflictException. On conflict it searches by name and only adopts an existing key when hasAlchemyTags(...) confirms it belongs to this stack/stage/id; otherwise the conflict is preserved.

packages/alchemy/src/AWS/ApiGateway/Authorizer.ts
CharLEEx: Addressed in 65fb3dd: added field-level JSDoc for Authorizer props, including restApiId, name, type, providerARNs, authType, authorizerUri, authorizerCredentials, identity source, validation expression, and TTL.

packages/alchemy/src/AWS/ApiGateway/ApiKey.ts
CharLEEx: Addressed in 65fb3dd: ApiKey.value is now Redacted.Redacted<string>. The provider unwraps it only when calling createApiKey, and the live test now uses Redacted.make(...). The value still is not returned in attributes/state.

packages/alchemy/src/AWS/ApiGateway/Authorizer.ts
CharLEEx: Addressed in 65fb3dd: added comments around the replacement decision for authorizer identity/kind fields, noting that restApiId/name/type define the authorizer shape and are clearer as replacement than in-place mutation.

packages/alchemy/src/AWS/ApiGateway/DomainName.ts
CharLEEx: Addressed in 65fb3dd: removed the duplicate DomainName tag helper and switched it to the shared tagRecord(...) helper in Tags.ts.

packages/alchemy/src/AWS/ApiGateway/DomainName.ts
CharLEEx: Addressed in 65fb3dd: added retry handling for DomainName update/delete around ConflictException and TooManyRequestsException, which are the likely transient cases while certificate, policy, or routing changes propagate.

packages/alchemy/src/AWS/ApiGateway/Method.ts
CharLEEx: Addressed in 65fb3dd: added docs clarifying Method.integration.credentials is an IAM role ARN or passthrough marker stored by API Gateway, not secret key material, so I did not wrap it in Redacted.

packages/alchemy/src/AWS/ApiGateway/RestApi.ts
CharLEEx: Addressed in 65fb3dd: expanded RestApi JSDoc with multiple API reference sections and examples for generated names, private APIs, binary media types, and disabling the default execute-api endpoint.

packages/alchemy/src/AWS/ApiGateway/RestApi.ts
CharLEEx: Addressed in 65fb3dd: documented RestApi as the low-level primitive intended to be composed directly today, with OpenAPI / Effect HttpApi schema generation called out as a higher-level layer to build separately on top of these resources.

packages/alchemy/src/AWS/ApiGateway/RestApi.ts
CharLEEx: Addressed in 65fb3dd: added explicit binary media type removal coverage in RestApi.test.ts in addition to the existing add/update test, so both patch directions are covered.

packages/alchemy/src/AWS/ApiGateway/UsagePlan.ts
CharLEEx: Addressed in 65fb3dd: removed the duplicate UsagePlan tag helper and switched it to tagRecord(...) from Tags.ts.

packages/alchemy/test/AWS/ApiGateway/ApiKey.test.ts
CharLEEx: Addressed in 65fb3dd: removed hard-coded physical names from the ApiGateway tests where providers can now generate deterministic names. ApiKey, UsagePlan, VpcLink, Authorizer, and RestApi names are optional and generated via createPhysicalName where appropriate.

packages/alchemy/test/AWS/ApiGateway/Authorizer.test.ts
CharLEEx: Addressed in 65fb3dd: replaced the empty skipped Authorizer suite with an env-gated live test. It runs when ALCHEMY_TEST_AUTHORIZER_URI is set and covers create plus TTL update behavior.

packages/alchemy/test/AWS/ApiGateway/GatewayResponse.test.ts
CharLEEx: Addressed in 65fb3dd: added a GatewayResponse update-flow test that changes statusCode and responseTemplates, then verifies the remote GatewayResponse reflects the update.

sam-goodwin:
I took this over and got it merged with tests passing https://github.com/alchemy-run/alchemy-effect/pull/169

Couldn't edit your PR for some reason. Sorry, the tests weren't passing and API gateway is notorious with its throttling limits, i had to add some logic to handle it.

Also, the design needed to use Bindings so that the Outputs properly connected the DAG or else deployment failed. The tests weren't in good shape. Now they are.

PR #141, opened 2026-05-01
fix: re-enable CLI bundle by routing import.meta through package name

Mkassabov:
[#135](https://github.com/alchemy-run/alchemy-effect/pull/135) disabled the `bin/alchemy.ts` tsdown bundle because relative `import.meta.*` references inside the CLI's import graph misresolve once inlined into `bin/alchemy.js` — they point at the bundle, not the original source file.

This applies the [#128](https://github.com/alchemy-run/alchemy-effect/pull/128) pattern (`import.meta.resolve("alchemy/...")` + matching `exports` entries) to the two real offenders, then re-enables the tsdown config so react/ink/pathe stay out of consumer-visible runtime deps.

- `Cloudflare/Local/Sidecar.ts` — package-name resolve `SidecarServer.ts` instead of `./SidecarServer.ts`.
- `Cloudflare/StateStore/Api.ts` — package-name resolve `Api.ts` instead of `import.meta.filename`. The bundled value is `bin/alchemy.js`, which has no `default` export and breaks the worker bundler with `[MISSING_EXPORT] "default" is not exported by "bin/alchemy.js"`.
- `package.json` — export `./Cloudflare/Local/SidecarServer.ts` and `./Cloudflare/StateStore/Api.ts`.
- `tsdown.config.ts` — uncomment `bin/alchemy.ts` entry; document the `import.meta` constraint inline.

`Agent/util/parser.ts` also uses `import.meta.url` but isn't in the CLI bundle's import graph (only reached via `Bundle/Docker.ts`), so it's left alone.

### DIFF
--- a/packages/alchemy/package.json
+++ b/packages/alchemy/package.json
@@ -246,7 +246,9 @@
     },
     "./bin/alchemy.js": "./bin/alchemy.js",
     "./bin/exec.ts": "./bin/exec.ts",
-    "./bin/exec.js": "./bin/exec.js"
+    "./bin/exec.js": "./bin/exec.js",
+    "./Cloudflare/Local/SidecarServer.ts": "./src/Cloudflare/Local/SidecarServer.ts",
+    "./Cloudflare/StateStore/Api.ts": "./src/Cloudflare/StateStore/Api.ts"
   },
   "publishConfig": {
     "access": "public"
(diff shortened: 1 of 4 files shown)


packages/alchemy/package.json
```diff
@@ -246,7 +246,9 @@
     },
     "./bin/alchemy.js": "./bin/alchemy.js",
     "./bin/exec.ts": "./bin/exec.ts",
-    "./bin/exec.js": "./bin/exec.js"
```
sam-goodwin, line 249: why do we export these?

PR #142, opened 2026-05-01
feat(cloudflare): Support `@cloudflare/opennextjs` / wrangler with full parity

czxtm:
> **This is a follow-up to #117** - Both PR's should be merged but I separated them for clarity.
> 
> #117 adds an option to opt-out of re-bundling by rolldown - that is a dependency for supporting opennext. #117 on its own is not sufficient to support all `cloudflare/opennextjs` apps. This PR fixes that.

After more testing I discovered that wrangler includes not just the main file, but any other file in the same directory. It also has a few other special-cased behaviors that are required.

This PR should allow any wrangler-deployed app to be deployed by Alchemy, as it actually uses wrangler under-the-hood (or uses esbuild equivalently) and therefore produces the exact same bundle. There are 2 approaches that could be taken - I included both so you can simply delete one of them after deciding which path to take.

`OpenNextWranglerSubprocess` spawns a subprocess and runs `wrangler --dry-run` which is the officially recommended way to produce the exact bundle that wrangler would upload.

`OpenNext` does not use a subprocess but instead imports it from wrangler to do the same thing programatically by calling esbuild the same way wrangler does. The obvious risk here is that if that ever changes, then there would be drift. It's also not part of the public API so there's also maintenance risk. The other downside is it introduces a couple dependencies.

Both of them produce the exact same thing and successfully deploy the example.


john-royal:
I'm going to have to close this for now.

We're open to supporting Next.js in the future, but only if we can do it without needing to bring Wrangler into the picture. Nesting Wrangler inside Alchemy - in other words, one deployment tool inside another - created compatibility, support, and UX challenges in Alchemy V1, and we don't want to repeat that mistake here.

In the future, I'm considering forking `@opennextjs/cloudflare` to remove the Wrangler dependency. For now, one alternative is [`vinext`](https://github.com/cloudflare/vinext), which re-implements the Next.js API surface in Vite. If you run into issues using `vinext` with Alchemy V2, feel free to file an issue.

czxtm:
@john-royal didnt know about vinext - wish i knew about it earlier! It's been a pain getting NextJS to work with alchemy V2. It's the de facto react framework so I definitely feel like Alchemy should support it somehow. I'll publish something if I can get it working well

PR #144, opened 2026-05-01
fix(Cloudflare): don't run update out of order for strict DAGs

sam-goodwin:
`Cloudflare.StaticSite` is a composition of `Build.Command` -> `Cloudflare.Worker`. Our convergence algorithm has support for circularity, but it had a bug where it would always run update out of order even if there was a strict DAG. This doesn't work well for things like "Build assets, then deploy". This change only allows the out of order behavior for cycles, which are less common (e.g. worker to worker bindings). We need to be careful with circularity. use it sparingly and design purposefully for it.


packages/alchemy/src/Cloudflare/Workers/Worker.ts
```diff
@@ -2438,16 +2450,36 @@ let _viteModule: ViteModule | null = null;
...
+      try {
+        const require = createRequire(path.join(root, "package.json"));
+        resolved = require.resolve("vite");
+        break;
+      } catch {
+        // try next candidate
+      }
+    }
+    if (!resolved) throw new Error("vite not resolvable");
+    vitePath = resolved;
```
sam-goodwin: @john-royal AI made this change, looks suss. Thoughts

packages/alchemy/src/Cloudflare/Workers/Worker.ts
```diff
@@ -2432,31 +2444,23 @@ const contentTypeFromExtension = (extension: string) => {
...
+    // On Windows, absolute paths must be file:// URLs for ESM import().
+    const viteUrl = pathToFileURL(vitePath);
+    return await import(/* @vite-ignore */ viteUrl.href);
   } catch {
-    // Fallback: use the Vite that ships with vinext (works for non-linked installs)
-    vitePath = "vite";
+    // Fallback: try to import vite from the global node_modules (works for non-linked installs)
+    // The fallback is a bare specifier and works as-is.
+    return await import("vite");
   }
```
john-royal, line 2465: @sam-goodwin Resolving from the actual project root instead of from `process.cwd()` if available was correct; the fallback and searching logic was slop IMO.

I ran `Website` and `Worker` tests and they worked with this change on my end, but if you could double check on your end that would be great.

john-royal (approved):

PR #160, opened 2026-05-02
fix(cloudflare/state-store): apply Cloudflare Access headers to HTTP requests

jacobiajohnson:
## Summary

When the deployed state-store worker hostname is fronted by Cloudflare Access (e.g. an org-wide policy on `*.workers.dev`), every request to the worker is intercepted before it reaches the script and 302'd to the Access SSO login page. The state-store HTTP client only attaches a bearer token, so `bun alchemy deploy` hangs forever in the 404 retry loop with:

```
HttpClientError: Decode error (404 PUT https://<worker>/state/stacks/CloudflareStateStore/...)
```

`beta.29` wired the `Access` service into `Cloudflare/EdgeSession.ts` for the edge-preview path, but `State/HttpStateStore.ts` was missed — even though `Access.AccessLive` is provided in the surrounding state-store layer (`Cloudflare/StateStore/State.ts:286`), nothing in the state-store request pipeline ever calls `getAccessHeaders`.

This PR mirrors the `Access.use((a) => a.getAccessHeaders(host))` pattern from `EdgeSession.ts:140` and merges the resulting headers (service-token pair or `cloudflared`-issued `CF_Authorization` cookie) into every state-store request alongside the bearer token.

## Repro

1. Have a Cloudflare account where Zero Trust Access fronts `*.workers.dev` (the failure mode in our case is a WorkOS-internal Access policy).
2. `bun alchemy deploy` on a stack that uses `Cloudflare.state()`.
3. The bootstrap deploys fine, then the first PUT to the state-store worker 302s to `*.cloudflareaccess.com` and the deploy hangs in the 404 retry loop.

With this patch + `cloudflared access login <state-store-url>` (or `CLOUDFLARE_ACCESS_CLIENT_ID` / `CLOUDFLARE_ACCESS_CLIENT_SECRET` env vars), the deploy completes normally.

## Test plan

- [x] Reproduced the hang against a Cloudflare Access-fronted account on beta.29
- [x] Verified the patched build deploys end-to-end (state-store + user stack + GET/PUT round trip through the deployed worker)
- [ ] CI

### DIFF
--- a/packages/alchemy/src/State/HttpStateStore.ts
+++ b/packages/alchemy/src/State/HttpStateStore.ts
@@ -1,9 +1,9 @@
 import * as Effect from "effect/Effect";
+import { identity } from "effect/Function";
 import * as Schedule from "effect/Schedule";
 import * as HttpClient from "effect/unstable/http/HttpClient";
 import * as HttpClientRequest from "effect/unstable/http/HttpClientRequest";
 import * as HttpApiClient from "effect/unstable/httpapi/HttpApiClient";
-import { Access } from "../Cloudflare/Access.ts";
 import { StateApi } from "./HttpStateApi.ts";
 
 import type { ReplacedResourceState, ResourceState } from "./ResourceState.ts";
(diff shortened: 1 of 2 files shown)


Mkassabov (approved):
Looks good to me. should wait for @john-royal to double check tho.

john-royal:
LGTM, running CI to double check.

jacobiajohnson:
@john-royal LMK if there's something I should do to fix the CI errors.

packages/alchemy/src/State/HttpStateStore.ts
```diff
@@ -17,10 +18,15 @@ export interface HttpStateStoreProps {
 
 export const makeHttpStateStore = ({ url, authToken }: HttpStateStoreProps) =>
   Effect.gen(function* () {
+    const accessHeaders = yield* Access.use((access) =>
+      access.getAccessHeaders(new URL(url).host),
+    );
```
sam-goodwin: Cloudflare specific stuff does not belong in the generic http state store.

sam-goodwin: Pushed a commit moving it

PR #163, opened 2026-05-03
fix(Cloudflare/SecretsStore): adoption opt-in by default

mstibbard:
Documentation suggests this should be opt-in by default but it is currently opted-out by default.

The initial Cloudflare state setup creates a state store and 2 secrets (`AlchemyStateStoreEncryptionKey` and `AlchemyStateStoreToken`), meaning any subsequent attempts to work with secrets results in the `maximum stores exceeded` error.


Mkassabov:
relevant disc thread: https://discord.com/channels/1359694195782320389/1500441429854388386

PR #176, opened 2026-05-04
fix(cloudflare): disable builder.sharedConfigBuild for vite build

john-royal:
This was causing Vite builds to fail for static websites.

I'm a little wary of breaking something else by removing this. If we pin down why this was added, maybe we can figure out the root reason.


sam-goodwin:
Is there a git blame we can look at

john-royal:
> Is there a git blame we can look at

Yes, Pear added it because of solid. Currently looking at root causes.

PR #178, opened 2026-05-04
fix(cloudflare/workers): drain pre-response handlers in serveWebRequest

### LINKED ISSUE #175: HttpMiddleware.cors() doesn't tag non-preflight responses when used inside a Cloudflare.Worker
Wrapping `HttpMiddleware.cors()` around the api handler returned by a `Cloudflare.Worker` resource only attaches CORS headers to **preflight** (`OPTIONS`) responses. Real responses (`GET` / `POST` / etc.) come back without any CORS headers, so browsers block the response.

## Reproduction

https://github.com/Mkassabov/alchemy-effect-cors-repro

```sh
bun install
bun alchemy deploy --profile <yours> --yes
# echo "WORKER_URL=<deployed url>" > .env
bun test
# → 1 pass, 1 fail
```

The two tests in [`test/cors.test.ts`](https://github.com/Mkassabov/alchemy-effect-cors-repro/blob/main/test/cors.test.ts) are the executable spec:
- ✅ `OPTIONS preflight carries Access-Control-Allow-Origin`
- ❌ `GET response carries Access-Control-Allow-Origin (BUG: missing)`

## Root cause

`HttpMiddleware.cors()` handles the two paths very differently ([effect source](https://github.com/Effect-TS/effect/blob/main/packages/effect/src/unstable/http/HttpMiddleware.ts)):

```ts
return httpApp => Effect.withFiber(fiber => {
  const request = ...;
  if (request.method === "OPTIONS") {
    // builds and returns a 204 with all the CORS headers itself
    return Effect.succeed(Response.empty({ status: 204, headers: ... }));
  }
  // for GET/POST/etc — just registers a callback on the request that
  // a downstream pre-response stage is supposed to invoke later
  appendPreResponseHandlerUnsafe(request, preResponseHandler);
  return httpApp;
});
```

- **OPTIONS preflight** — middleware constructs the response itself, headers ride along. ✅
- **GET / POST / etc.** — middleware does NOT touch the response. It pushes a `preResponseHandler` onto a queue on the request. That queue is meant to be drained right before the response is shipped — that drain step is what stamps `Access-Control-Allow-Origin` onto real responses.

When running an effect HTTP server (e.g. via `BunHttpServer.layer`), the server loop walks the pre-response queue. The Cloudflare.Worker pipeline takes the `Effect<HttpServerResponse>` returned from \`fetch\` and converts it directly to a Cloudflare \`Response\` — the pre-response queue is never drained, so CORS headers never make it onto real responses. Only preflight works because it short-circuits before that step.

Mkassabov:
`HttpMiddleware.cors()` (and any other middleware that registers a `preResponseHandler` for non-OPTIONS requests) was a no-op on real responses inside a `Cloudflare.Worker` — only OPTIONS preflights short-circuited with CORS headers. `serveWebRequest` ran the handler and converted the result straight to a web `Response`, never draining the pre-response queue.

Run the handler through effect's public `HttpEffect.toHandled` so the registered pre-response handler chain fires before the response is converted. `scopeTransferToStream` keeps streaming bodies' scopes closing on stream end.

```diff
-    const response = yield* handler.pipe(
-      Effect.provideService(HttpServerRequest.HttpServerRequest, request),
-      Effect.provideService(Request, webRequest as any),
-      Effect.catchCause(...),
-    );
-    return HttpServerResponse.toWeb(response, { context: yield* Effect.context() });
+    const httpApp = HttpEffectModule.toHandled(handler, (_req, response) => {
+      const transferred = HttpEffectModule.scopeTransferToStream(response);
+      resume(Effect.succeed(HttpServerResponse.toWeb(transferred, { context: parentContext })));
+      return Effect.void;
+    });
+    httpApp.pipe(
+      Effect.provideService(HttpServerRequest.HttpServerRequest, request),
+      Effect.provideService(Request, webRequest as any),
+      Effect.provide(parentContext),
+      Effect.runFork,
+    );
```

Verified against the repro from #175: both `OPTIONS` preflight and `GET` now carry `Access-Control-Allow-Origin`.

Fixes #175

### DIFF
--- a/packages/alchemy/src/Cloudflare/Workers/HttpServer.ts
+++ b/packages/alchemy/src/Cloudflare/Workers/HttpServer.ts
@@ -1,10 +1,10 @@
 import type * as cf from "@cloudflare/workers-types";
-import * as Cause from "effect/Cause";
 import * as Effect from "effect/Effect";
 import * as Option from "effect/Option";
 import type { Scope } from "effect/Scope";
 import type { HttpBodyError } from "effect/unstable/http/HttpBody";
-import * as HttpServerError from "effect/unstable/http/HttpServerError";
+import * as HttpEffectModule from "effect/unstable/http/HttpEffect";
+import type * as HttpServerError from "effect/unstable/http/HttpServerError";
 import * as HttpServerRequest from "effect/unstable/http/HttpServerRequest";
 import * as HttpServerResponse from "effect/unstable/http/HttpServerResponse";
 import * as Http from "../../Http.ts";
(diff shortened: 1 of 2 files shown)


sam-goodwin:
Closing in favor of https://github.com/alchemy-run/alchemy-effect/pull/404

PR #184, opened 2026-05-05
fix(aws/sqs): harden Queue reconcile + add lifecycle convergence tests

sam-goodwin:
Stacked on top of [#179](https://github.com/alchemy-run/alchemy-effect/pull/179). Hardens the SQS Queue resource as the first pass of a per-resource hardening sweep on top of the reconcile migration.

### Reconciler changes

```diff
- Effect.catch(() => Effect.succeed({} as Record<string, string>)),
+ Effect.catchTag("QueueDoesNotExist", () =>
+   Effect.succeed({} as Record<string, string>),
+ ),
```

`listQueueTags` was wrapped in a blanket `Effect.catch` that swallowed every error — including `InvalidSecurity` (auth) and `RequestThrottled` (throttling). A queue we couldn't list tags on would silently slip through as Unowned. Scoped to `QueueDoesNotExist` only.

```diff
  Effect.retry({
    while: (e) => e._tag === "QueueDeletedRecently",
    schedule: Schedule.fixed(1000).pipe(
      Schedule.tapOutput(...),
+     Schedule.both(Schedule.recurs(75)),
    ),
  }),
```

The recreate-after-recent-delete retry was unbounded — if SQS kept the name reserved past the documented 60s, we'd loop forever. Capped at 75s.

### New lifecycle tests

- reconcile resets attributes mutated out-of-band via the raw SDK
- reconcile re-creates a queue that was deleted out-of-band (rides the `QueueDeletedRecently` window)
- changing `queueName` triggers replace; old queue is deleted
- flipping `fifo` triggers replace
- updates all mutable standard attributes (`delaySeconds`, `maximumMessageSize`, `messageRetentionPeriod`, `receiveMessageWaitTimeSeconds`, `visibilityTimeout`) in one go
- updates all mutable FIFO-specific attributes (`contentBasedDeduplication`, `deduplicationScope`, `fifoThroughputLimit`)
- destroying an already-deleted queue is a no-op
- `adopt(true)` re-tags a foreign queue with internal alchemy tags

### Distilled patch

Several SQS errors were miscategorized so the AWS retry layer wouldn't back off on them. Filed as [alchemy-run/distilled#238](https://github.com/alchemy-run/distilled/pull/238) — once published, `RequestThrottled`, `KmsThrottled`, `PurgeQueueInProgress`, and `QueueDeletedRecently` will participate in the default retry policy. The reconciler-level `QueueDeletedRecently` handling stays for now since the AWS doc'd 60s window exceeds the default retry budget.

### Process

This is the template for a per-resource hardening sweep. After this lands, sub-agents will run the same loop against the rest of the surface area, each in its own worktree under `.worktrees/` (now in `.gitignore`).


sam-goodwin:

PR #185, opened 2026-05-05
fix(aws/dynamodb): harden Table reconcile + add lifecycle convergence tests

sam-goodwin:
Stacked on #179 (`refactor(core): collapse provider create+update into reconcile`). Per-resource hardening sweep for `AWS.DynamoDB.Table`.

### Reconciler changes

Wait for `ACTIVE` before mutating an observed-but-transient table. Pre-existing tables in `CREATING`/`UPDATING` (e.g. adoption mid-stabilization) would fall through to bare `updateTable` and 400 with `ResourceInUseException`.

```ts
} else if (state.table.TableStatus !== "ACTIVE") {
  yield* waitForTableActive(session, tableName);
  state = (yield* readTableState(tableName))!;
}
```

Wrap each control-plane `updateTable` / `tagResource` / `untagResource` in retry-while-`isRetryableControlPlaneError`. Without it, racing against a still-stabilizing GSI or a transient `InternalServerError` surfaces as a hard failure.

```ts
const retryControlPlaneUpdate = <A, E extends { _tag?: string }, R>(
) =>
  effect.pipe(
    Effect.retry({
      while: isRetryableControlPlaneError,
      schedule: Schedule.exponential(250).pipe(
        Schedule.both(Schedule.recurs(30)),
      ),
    }),
  );
```

Replace ad-hoc `Effect.fail(new Error(...))` sites with a typed `TableUnreadable` so the IaC error channel stays typed.

### New lifecycle tests

- Redeploy with same props is a no-op (reconcile is idempotent)
- Reconcile resets PITR / TTL / tags drift mutated out-of-band, including re-stamping internal alchemy tags after they're untagged
- Reconcile re-creates a table that was deleted out-of-band (asserts `tableId` regenerated, `tableArn` stable)
- Changing `tableName` triggers replace, old table is deleted
- Destroying an already-deleted table is a no-op
- Adding a GSI in-place without replacement (table arn / id stable)

### Distilled patch

https://github.com/alchemy-run/distilled/pull/239 — adds error categories so `ResourceInUseException`, `InternalServerError`, and `ThrottlingException` participate in the AWS retry layer instead of surfacing as hard `BadRequestError`/`AuthError`.

### Baseline fix

`packages/alchemy/test/test.resources.ts` lines 521-522 had a pre-existing tsc error on this branch (`output` is possibly undefined post-reconcile refactor). Carried forward the same fix the SQS hardening PR applied so the baseline `bun tsc -b` is clean.

### DIFF
--- a/packages/alchemy/src/AWS/DynamoDB/Table.ts
+++ b/packages/alchemy/src/AWS/DynamoDB/Table.ts
@@ -667,7 +667,7 @@ export const TableProvider = () =>
           const table = response.Table;
           if (!table?.TableArn) {
             return yield* Effect.fail(
-              new Error(`Table ${tableName} not found`),
+              new TableUnreadable({ tableName, phase: "describe" }),
             );
           }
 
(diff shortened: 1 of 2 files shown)


sam-goodwin:

PR #187, opened 2026-05-05
fix(aws/s3): harden Bucket reconcile + add lifecycle convergence tests

sam-goodwin:
Audits the S3 Bucket reconciler for the same classes of issues SQS had, and adds lifecycle convergence tests on top of #179.

### Reconciler changes

Scope blanket catches to documented signals so real failures surface:

```diff
- const exists = yield* s3.headBucket({ Bucket: bucketName }).pipe(
-   Effect.map(() => true),
-   Effect.catchTag("NotFound", () => Effect.succeed(false)),
-   Effect.catch(() => Effect.succeed(false)),
- );
+ const exists = yield* s3.headBucket({ Bucket: bucketName }).pipe(
+   Effect.map(() => true),
+   Effect.catchTag("NotFound", () => Effect.succeed(false)),
+ );
```

```diff
- Effect.catchTag("NoSuchTagSet", () => Effect.succeed({} as Record<string, string>)),
- Effect.catch(() => Effect.succeed({} as Record<string, string>)),
+ Effect.catchTag("NoSuchTagSet", () => Effect.succeed({} as Record<string, string>)),
+ Effect.catchTag("NoSuchBucket", () => Effect.succeed({} as Record<string, string>)),
```

Bound the `OperationAborted` / `ServiceUnavailable` retry on `createBucket` so we don't spin forever when the service stays in that state:

```diff
  Effect.retry({
    while: (e) => e._tag === "OperationAborted" || e._tag === "ServiceUnavailable",
-   schedule: Schedule.exponential(100),
+   schedule: Schedule.exponential(100).pipe(Schedule.both(Schedule.recurs(8))),
  }),
```

`syncBucketTags` always diffs against observed cloud tags and skips the API call when the diff is empty — avoids a spurious `deleteBucketTagging` on every reconcile of a bucket that already has no tags.

### New lifecycle tests

Each test runs `destroy -> deploy -> ... -> destroy` on a `ScratchStack` and asserts at every step:

- reconcile resets bucket tags mutated out-of-band
- reconcile resets a bucket policy mutated out-of-band
- reconcile re-creates a bucket that was deleted out-of-band (rides the `OperationAborted` window)
- changing `bucketName` triggers replace; old bucket is deleted
- flipping `objectLockEnabled` triggers replace
- destroying an already-deleted bucket is a no-op
- `adopt(true)` re-syncs tags on a foreign-tagged bucket

`test.resources.ts` picks up the same `output?.stableId` fix from the SQS PR so this branch type-checks in isolation.

### Distilled patch

S3 error category fixes at https://github.com/alchemy-run/distilled/pull/241.

### DIFF
--- a/packages/alchemy/src/AWS/S3/Bucket.ts
+++ b/packages/alchemy/src/AWS/S3/Bucket.ts
@@ -265,12 +265,16 @@ export const BucketProvider = () =>
         );
 
         // For us-east-1, BucketAlreadyOwnedByYou is not thrown, so we need to
-        // pre-emptively check if the bucket exists for idempotency
+        // pre-emptively check if the bucket exists for idempotency.
+        //
+        // Only swallow `NotFound` (the documented "no such bucket" signal).
+        // Auth/throttling/parse errors must surface so we don't blindly try
+        // to recreate a bucket that we don't own — which would race with the
+        // real owner and bounce off `BucketAlreadyExists`.
         if (region === "us-east-1") {
           const exists = yield* s3.headBucket({ Bucket: bucketName }).pipe(
             Effect.map(() => true),
             Effect.catchTag("NotFound", () => Effect.succeed(false)),
-            Effect.catch(() => Effect.succeed(false)),
           );
 
           yield* Effect.logInfo(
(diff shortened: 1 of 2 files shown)


sam-goodwin:

PR #189, opened 2026-05-05
fix(aws/ec2): harden Vpc reconcile + add lifecycle convergence tests

sam-goodwin:
Harden the EC2 `Vpc` reconciler against eventual-consistency races and remove brittle delete-error matching, then add lifecycle convergence tests on top of `ScratchStack`.

### Reconciler changes

```diff
-          const dnsSupportResult = yield* ec2.describeVpcAttribute({
-            VpcId: vpcId,
-            Attribute: "enableDnsSupport",
-          });
+          // describeVpcAttribute / modifyVpcAttribute / createTags / deleteTags
+          // right after createVpc can race ahead of the VPC's visibility in
+          // the read path even after `waitForVpcAvailable` succeeded.
+          const dnsSupportResult = yield* ec2
+            .describeVpcAttribute({
+              VpcId: vpcId,
+              Attribute: "enableDnsSupport",
+            })
+            .pipe(retryEventuallyConsistent);
```

```diff
-              Effect.retry({
-                while: (e) => {
-                  return (
-                    e._tag === "DependencyViolation" ||
-                    (e._tag === "ValidationError" &&
-                      e.message?.includes("DependencyViolation"))
-                  );
-                },
+              Effect.retry({
+                while: (e) => e._tag === "DependencyViolation",
```

The `ValidationError` substring branch was dead — distilled tags `DependencyViolation` directly via `withDependencyViolationError`, so the substring match never fired and any real `ValidationError` was being swallowed silently.

```diff
   /**
    * Whether instances launched in the VPC get DNS hostnames.
-   * @default true
+   * Matches the AWS API default for new non-default VPCs.
+   * @default false
    */
   enableDnsHostnames?: boolean;
```

Doc was inverted vs. what the code actually does (`?? false`).

`waitForVpcAvailable` previously hard-failed on `Error("VPC ${vpcId} not found")` (untagged, so `Effect.retry` couldn't ride it out). It now folds that case into the existing `VpcPending` tag and tolerates `InvalidVpcID.NotFound` from the lookup.

### New lifecycle tests

- reconcile resets DNS attributes mutated out-of-band
- reconcile resets tags mutated out-of-band (foreign tags removed, drifted tags reset, internal alchemy tags retained)
- reconcile re-creates a VPC that was deleted out-of-band
- changing cidrBlock triggers replace; old VPC is deleted
- destroying an already-deleted VPC is a no-op

The original happy-path test was `.skip`-ped; un-skipped it.

### Distilled patch

EC2 `IncorrectState` was uncategorized in distilled, so the AWS retry layer didn't ride it out. https://github.com/alchemy-run/distilled/pull/242 tags it as `ConflictError + RetryableError`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,7 +7,8 @@ dist/
 .DS_Store
 .env
 *.tsbuildinfo
-.claude/
+.claude/worktrees
+.worktrees
 .alchemy/
 # !.alchemy/github:alchemy/
 # examples/*/.alchemy
@@ -23,6 +24,7 @@ wrangler.jsonc
 .smoke.logs
 smoke
 .svelte-kit
+.claude/*.local.json
 .astro
 .wrangler
 
--- a/packages/alchemy/src/AWS/EC2/Vpc.ts
+++ b/packages/alchemy/src/AWS/EC2/Vpc.ts
@@ -77,7 +77,8 @@ export interface VpcProps {
 
   /**
    * Whether instances launched in the VPC get DNS hostnames.
-   * @default true
+   * Matches the AWS API default for new non-default VPCs.
+   * @default false
    */
   enableDnsHostnames?: boolean;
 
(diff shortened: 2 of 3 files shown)


sam-goodwin:

PR #190, opened 2026-05-05
fix(cloudflare/workers): harden Worker reconcile + add lifecycle convergence tests

sam-goodwin:
Three blanket `Effect.catch(() => Effect.succeed(...))` sites in the Worker reconciler were swallowing every error class, including auth, throttling, and 5xx. Scope or remove them so transient failures surface for engine-level retry instead of being misread as "no live state".

### Reconciler changes

```diff
   const liveAll = yield* listDomains({
     accountId,
     service: scriptName,
   }).pipe(
     Effect.map(...),
-    Effect.catch(() => Effect.succeed([])),
   );
```

```diff
   const otherOwner = yield* listDomains({
     accountId,
     hostname,
   }).pipe(
     Effect.map(...),
-    Effect.catch(() => Effect.succeed(undefined)),
   );
```

```diff
   const oldSettings =
     existingSettings ??
     (yield* getScriptSettings({ accountId, scriptName: name }).pipe(
       Effect.map((s) => s as typeof s | undefined),
-      Effect.catch(() => Effect.succeed(undefined)),
+      Effect.catchTag("WorkerNotFound", () => Effect.succeed(undefined)),
     ));
```

A failed `listDomains` used to be misread as "no domains live", which then re-PUT every desired hostname and dropped us straight into 409 conflicts. A failed `getScriptSettings` used to be misread as "fresh worker, no migration history", which would silently regenerate migration tags from scratch.

### New lifecycle tests

- redeploy with same code is idempotent (script identity + bundle hash + tags preserved)
- reconcile resets bindings mutated out-of-band via `patchScriptScriptAndVersionSetting`
- reconcile re-creates a worker that was deleted out-of-band via `deleteScript`
- changing the explicit physical `name` triggers replace; old script is gone
- adding then removing a KV binding propagates to the live worker
- destroying an already-deleted worker is a no-op (delete is idempotent)

### Distilled patch

No patch needed. Cloudflare's `client/api.ts` already maps HTTP status to categorized errors (`TooManyRequests` → throttling+retryable, 5xx → server+retryable), and the worker-named errors (`WorkerNotFound`, `DomainNotFound`, `HostnameAlreadyInUse`, ...) already cover the codes the reconciler routes on.

### DIFF
--- a/packages/alchemy/src/Cloudflare/Workers/Worker.ts
+++ b/packages/alchemy/src/Cloudflare/Workers/Worker.ts
@@ -1122,6 +1122,12 @@ export const LiveWorkerProvider = () =>
           // simply be out of date. Without this we PUT domains that are
           // already registered to this same Worker and Cloudflare
           // returns a confusing "hostname already in use" error.
+          //
+          // No blanket `Effect.catch` here: a list failure (auth,
+          // throttling, 5xx) must surface so the engine can retry or
+          // report cleanly. Treating any failure as "no domains live"
+          // would silently re-PUT every desired domain and drop us
+          // straight into 409 conflicts.
           const liveAll = yield* listDomains({
             accountId,
             service: scriptName,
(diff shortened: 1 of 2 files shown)


sam-goodwin:

PR #192, opened 2026-05-05
fix(aws/sns): harden Topic reconcile + add lifecycle convergence tests

sam-goodwin:
Stacked on top of [#179](https://github.com/alchemy-run/alchemy-effect/pull/179). Hardens the SNS Topic resource as part of the per-resource hardening sweep on top of the reconcile migration.

### Reconciler changes

```diff
   ],
   { concurrency: "unbounded" },
 ).pipe(
   Effect.catchTag("NotFoundException", () => Effect.succeed(undefined)),
-  Effect.catchTag("InvalidParameterException", () =>
-    Effect.succeed(undefined),
-  ),
 );
```

The outer `InvalidParameterException` catch wrapped *all three* parallel calls in `readTopic`. A 400 returned by `getTopicAttributes` or `listTagsForResource` would make the entire read return `undefined` — the engine then treats the live topic as missing and tries to recreate. Moved the catch onto `getDataProtectionPolicy` only (which legitimately returns 400 for FIFO topics).

```diff
- .pipe(
-   Effect.catchTag("NotFoundException", () => Effect.void),
-   Effect.catchTag("InvalidParameterException", () => Effect.void),
- );
+ .pipe(Effect.catchTag("NotFoundException", () => Effect.void));
```

`delete` was absorbing `InvalidParameterException` too — that hides malformed-ARN bugs persisted in state. Scoped to `NotFoundException` only.

### New lifecycle tests

- reconcile resets `DisplayName` and `DeliveryPolicy` mutated out-of-band via the raw SDK
- reconcile re-creates a topic that was deleted out-of-band
- changing `topicName` triggers replace; old topic is deleted
- flipping `fifo` triggers replace
- destroying an already-deleted topic is a no-op
- `adopt(true)` re-tags a foreign topic with internal alchemy tags (`alchemy::id`, `alchemy::stack`, `alchemy::stage`)

### Distilled patch

Several SNS errors were miscategorized so the AWS retry layer wouldn't back off on them. Filed as [alchemy-run/distilled#243](https://github.com/alchemy-run/distilled/pull/243) — once published, `Throttled`, `KMSThrottling`, `ConcurrentAccess`, and `StaleTag` participate in the default retry policy, and quota / not-found errors carry their semantic categories.

### DIFF
--- a/packages/alchemy/src/AWS/SNS/Topic.ts
+++ b/packages/alchemy/src/AWS/SNS/Topic.ts
@@ -290,14 +290,15 @@ export const TopicProvider = () =>
       };
     }),
     delete: Effect.fn(function* ({ output }) {
+      // SNS's deleteTopic is documented as idempotent (a missing topic still
+      // returns 200), but we still scope the catch to NotFoundException so a
+      // genuine InvalidParameterException (e.g. a malformed ARN persisted in
+      // state) surfaces instead of being silently swallowed.
       yield* sns
         .deleteTopic({
           TopicArn: output.topicArn,
         })
-        .pipe(
-          Effect.catchTag("NotFoundException", () => Effect.void),
-          Effect.catchTag("InvalidParameterException", () => Effect.void),
-        );
+        .pipe(Effect.catchTag("NotFoundException", () => Effect.void));
     }),
   });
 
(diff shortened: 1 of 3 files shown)


sam-goodwin:

PR #193, opened 2026-05-05
fix(aws/iam): harden Role reconcile + add lifecycle convergence tests

sam-goodwin:
Stacked on #179. Per-resource hardening sweep for `AWS.IAM.Role`.

### Reconciler changes

Wrap mutating IAM calls in `retryIamMutation`. `ConcurrentModificationException` and `EntityTemporarilyUnmodifiableException` are documented as transient — IAM serializes mutations to a role and surfaces these when two writes overlap. Without the wrap they surface as hard `ConflictError`.

```ts
const isRetryableIamControlPlaneError = (error: { _tag?: string }) =>
  error._tag === "ConcurrentModificationException" ||
  error._tag === "EntityTemporarilyUnmodifiableException";

const retryIamMutation = <A, E extends { _tag?: string }, R>(
) =>
  effect.pipe(
    Effect.retry({
      while: isRetryableIamControlPlaneError,
      schedule: Schedule.exponential(250).pipe(
        Schedule.both(Schedule.recurs(20)),
      ),
    }),
  );
```

Wrap post-create attach/put calls in `retryNoSuchEntityAfterCreate`. IAM is eventually consistent globally — a brand-new role can briefly `NoSuchEntity` from a different IAM host on subsequent mutating calls.

```ts
const retryNoSuchEntityAfterCreate = <A, E extends { _tag?: string }, R>(
) =>
  effect.pipe(
    Effect.retry({
      while: (e) =>
        e._tag === "NoSuchEntityException" ||
        isRetryableIamControlPlaneError(e),
      schedule: Schedule.exponential(500).pipe(
        Schedule.both(Schedule.recurs(8)),
      ),
    }),
  );
```

Bound the delete-role retry on `DeleteConflictException`. A role mid-detach from an instance profile clears within seconds; we cap the retry so a genuine, persistent reference still surfaces.

```diff
  yield* iam
    .deleteRole({ RoleName: output.roleName })
-   .pipe(Effect.catchTag("NoSuchEntityException", () => Effect.void));
+   .pipe(
+     Effect.retry({
+       while: (e) =>
+         e._tag === "DeleteConflictException" ||
+         isRetryableIamControlPlaneError(e),
+       schedule: Schedule.exponential(500).pipe(
+         Schedule.both(Schedule.recurs(10)),
+       ),
+     }),
+     Effect.catchTag("NoSuchEntityException", () => Effect.void),
+   );
```

Scope the implicit catches around `listRolePolicies` / `listAttachedRolePolicies` in delete to `NoSuchEntityException` so other failures (auth, throttling) surface.

### New lifecycle tests

- reconcile resets assume-role policy and inline policies mutated out-of-band via the raw IAM SDK
- reconcile re-creates a role that was deleted out-of-band (rides the eventual-consistency window)
- changing `roleName` triggers replace; old role is deleted
- adding/removing managed-policy attachments diffs against cloud state (out-of-band detach + foreign attach is reverted on next reconcile)
- destroying an already-deleted role is a no-op

The pre-existing `adopt(true)` test already covers re-tagging a foreign role with internal alchemy tags.

### Distilled patch

IAM error category fixes at https://github.com/alchemy-run/distilled/pull/244 — once published, `ConcurrentModificationException`, `EntityTemporarilyUnmodifiableException`, `ServiceFailureException`, and `RequestLimitExceeded` participate in the AWS retry layer instead of surfacing as hard `ConflictError` / `ServerError`. The reconciler-level retry helpers stay because the AWS retry layer is bounded; per-call control-plane retries fit within that budget.

### Baseline fix

`packages/alchemy/test/test.resources.ts` lines 521-522 had a pre-existing tsc error on this branch (`output` is possibly undefined post-reconcile refactor). Carried forward the same fix the SQS/DynamoDB/S3 hardening PRs applied so the baseline `bun tsc -b` is clean.

### DIFF
--- a/packages/alchemy/src/AWS/IAM/Role.ts
+++ b/packages/alchemy/src/AWS/IAM/Role.ts
@@ -1,5 +1,6 @@
 import * as iam from "@distilled.cloud/aws/iam";
 import * as Effect from "effect/Effect";
+import * as Schedule from "effect/Schedule";
 import { Unowned } from "../../AdoptPolicy.ts";
 import { isResolved } from "../../Diff.ts";
 import { createPhysicalName } from "../../PhysicalName.ts";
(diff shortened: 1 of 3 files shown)


sam-goodwin:

PR #194, opened 2026-05-05
fix(cloudflare/r2): harden Bucket reconcile + add lifecycle convergence tests

sam-goodwin:
Audits the R2 Bucket reconciler for the same classes of issues SQS / S3 / Worker had, and adds lifecycle convergence tests on top of #179.

### Reconciler changes

The `BucketAlreadyExists` re-fetch path in `reconcile` was inconsistent with the outer observe path:

```diff
   if (!observed) {
     observed = yield* createBucket({
       accountId: acct,
       name,
       storageClass: news.storageClass,
       jurisdiction: news.jurisdiction,
       locationHint: news.locationHint,
     }).pipe(
       Effect.catchTag("BucketAlreadyExists", () =>
         getBucket({
           accountId: acct,
           bucketName: name,
-          jurisdiction: news.jurisdiction,
+          jurisdiction,
         }),
       ),
     );
   }
```

The outer `getBucket` resolves `jurisdiction` via `output?.jurisdiction ?? news.jurisdiction ?? "default"`. The inner re-fetch was passing `news.jurisdiction`, which can be `undefined` while the outer call uses `"default"`. Same namespace in practice for default-jurisdiction buckets, but the inconsistency bites the moment a non-default jurisdiction is configured: the create hits one namespace, the re-fetch hits another, and we error out instead of converging.

No blanket `Effect.catch` sites to scope — the existing reconciler already routes only on the documented signals (`NoSuchBucket`, `BucketAlreadyExists`).

### New lifecycle tests

- reconcile resets `storageClass` mutated out-of-band via `r2.patchBucket`
- reconcile re-creates a bucket that was deleted out-of-band
- changing `name` triggers replace; old bucket is deleted
- destroying an already-deleted bucket is a no-op
- `adopt(true)` re-claims a foreign bucket under a new logical id and applies the new desired `storageClass`

`test.resources.ts` picks up the SQS PR's `output?.stableId` / `output?.stableArn` fix so this branch type-checks in isolation.

### Distilled patch

No patch needed. Cloudflare's HTTP status mapping in `client/api.ts` already attributes categories to `TooManyRequests` / `InternalServerError`, and R2's typed errors (`NoSuchBucket`, `BucketAlreadyExists`, `InvalidBucketName`, `InvalidRoute`, `NoRoute`) already cover every code the reconciler routes on.

### DIFF
--- a/packages/alchemy/src/Cloudflare/R2/R2Bucket.ts
+++ b/packages/alchemy/src/Cloudflare/R2/R2Bucket.ts
@@ -172,7 +172,11 @@ export const R2BucketProvider = () =>
 
           // Ensure — create if missing. R2 reports a concurrent create
           // (or partial state-persistence failure) as
-          // `BucketAlreadyExists`; tolerate by re-fetching the bucket.
+          // `BucketAlreadyExists`; tolerate by re-fetching the bucket
+          // through the same resolved `jurisdiction` we used for the
+          // initial observe. Using `news.jurisdiction` here (which can
+          // be undefined while the resolved value is "default") would
+          // query a different namespace than the create just hit.
           if (!observed) {
             observed = yield* createBucket({
               accountId: acct,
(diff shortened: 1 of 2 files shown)


sam-goodwin:

PR #195, opened 2026-05-05
fix(aws/logs): harden LogGroup reconcile + add lifecycle convergence tests

sam-goodwin:
Harden the `AWS.Logs.LogGroup` reconciler so observed cloud state — not `olds` or `output` — drives every sync step, and add lifecycle convergence tests that exercise the post-#179 reconcile shape.

### Reconciler changes

- KMS-key drift was unobservable: `kmsKeyId` was only honored when passed to `createLogGroup`. Added an `associateKmsKey` / `disassociateKmsKey` sync step diffed against `describeLogGroups`-observed state so adoption and out-of-band rotation converge.
- `listTagsForResource` was wrapped in a blanket `Effect.catch(() => {})`. That swallowed auth, throttling, and parse errors — meaning the reconciler would silently start from `{}` tags and rewrite the world. Scope to `ResourceNotFoundException` only.
- `read()` now lists live tags and returns `Unowned(attrs)` when the alchemy ownership tags are missing, so foreign-tagged log groups gate on `adopt(true)` instead of being silently taken over.

```diff
+ // KMS sync (new)
+ if (news.kmsKeyId !== observed?.kmsKeyId) {
+   if (news.kmsKeyId === undefined) {
+     yield* logs.disassociateKmsKey({ logGroupName }).pipe(
+       Effect.catchTag("ResourceNotFoundException", () => Effect.void),
+     );
+   } else {
+     yield* logs.associateKmsKey({ logGroupName, kmsKeyId: news.kmsKeyId });
+   }
+ }
```

```diff
- Effect.catch(() => Effect.succeed({} as Record<string, string>))
+ Effect.catchTag("ResourceNotFoundException", () =>
+   Effect.succeed({} as Record<string, string>),
+ )
```

### New lifecycle tests

`packages/alchemy/test/AWS/Logs/LogGroup.test.ts`:

- create + destroy with default props
- reconcile resets retention mutated out-of-band, then drops it when the prop is removed
- reconcile re-creates a log group that was deleted out-of-band
- changing `logGroupName` triggers replace and the old group is deleted
- destroying an already-deleted log group is a no-op
- owned (matching alchemy tags) log group is silently adopted without `--adopt`
- foreign-tagged log group requires `adopt(true)` to take over and gets re-tagged with internal alchemy tags

### Distilled patch

CloudWatch Logs error categories at https://github.com/alchemy-run/distilled/pull/245 — needed so `ThrottlingException` and `OperationAbortedException` participate in the AWS retry layer instead of surfacing as hard failures.

Also picks up the `output?.stableId ?? id` fallback in `test.resources.ts` from the #179 reconciler regression series.

### DIFF
--- a/packages/alchemy/src/AWS/Logs/LogGroup.ts
+++ b/packages/alchemy/src/AWS/Logs/LogGroup.ts
@@ -1,14 +1,19 @@
 import * as logs from "@distilled.cloud/aws/cloudwatch-logs";
 import { Region } from "@distilled.cloud/aws/Region";
 import * as Effect from "effect/Effect";
+import { Unowned } from "../../AdoptPolicy.ts";
 import { isResolved } from "../../Diff.ts";
 import { createPhysicalName } from "../../PhysicalName.ts";
 import * as Provider from "../../Provider.ts";
 import { Resource } from "../../Resource.ts";
-import type { Providers } from "../Providers.ts";
-import { createInternalTags, diffTags } from "../../Tags.ts";
+import {
+  createInternalTags,
+  diffTags,
+  hasAlchemyTags,
+} from "../../Tags.ts";
 import type { AccountID } from "../Environment.ts";
 import { AWSEnvironment } from "../Environment.ts";
+import type { Providers } from "../Providers.ts";
 import type { RegionID } from "../Region.ts";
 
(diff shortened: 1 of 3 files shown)


sam-goodwin:

PR #196, opened 2026-05-05
fix(cloudflare/d1): harden Database reconcile + add lifecycle convergence tests

sam-goodwin:
Two reconcile sites in the D1 Database provider were either swallowing all errors or under-catching the recoverable race. Tighten both, and pin lifecycle convergence with a new test suite.

### Reconciler changes

```diff
   const result = yield* executeSQL(
     accountId,
     databaseId,
     `PRAGMA table_info(${migrationsTable});`,
-  ).pipe(Effect.option);
+  );
```

`PRAGMA table_info(<missing>)` returns 200 OK with an empty result set, so the `Effect.option` was both unnecessary and dangerous: a transient 503 / auth blip would surface as "table doesn't exist", silently dropping us into the legacy-schema migration path on a perfectly normal database.

```diff
-  Effect.catchTag("InvalidProperty", () =>
-    Effect.gen(function* () {
-      const dbs = yield* listDbs({ accountId: acct, name });
-      ...
-    }),
-  ),
+  Effect.catchTag("DatabaseAlreadyExists", () => onConflict),
+  Effect.catchTag("InvalidProperty", () => onConflict),
```

`createDatabase` can race with itself when a previous run created the database but failed to persist output. Cloudflare reports the conflict as `DatabaseAlreadyExists` (canonical 409) — but the previous code only caught `InvalidProperty`, so a real race would surface as a hard error instead of being recovered via re-list.

### New lifecycle tests

- redeploy with same props is a no-op (physical id preserved)
- reconcile resets `read_replication.mode` mutated out-of-band via the raw D1 SDK
- reconcile re-creates a database that was deleted out-of-band
- changing `databaseName` triggers replace; old database is deleted
- migrations resume cleanly after a partial-failure mid-batch
- destroying an already-deleted database is a no-op (delete is idempotent)
- `adopt(true)` re-claims a foreign database under a new logical id

### Distilled patch

No patch needed. Cloudflare's `client/api.ts` already maps HTTP status to categorized errors (`TooManyRequests` → throttling+retryable, 5xx → server+retryable), and the D1-named errors (`DatabaseNotFound`, `DatabaseAlreadyExists`, `InvalidProperty`, ...) already cover the codes the reconciler routes on.

### DIFF
--- a/packages/alchemy/src/Cloudflare/D1/D1Database.ts
+++ b/packages/alchemy/src/Cloudflare/D1/D1Database.ts
@@ -397,6 +397,23 @@ export const DatabaseProvider = () =>
           let databaseName: string;
           const isFirstCreation = !observed;
           if (!observed) {
+            // Race recovery: another concurrent reconcile (or a previous
+            // run that failed to persist its output) may have created the
+            // database between our observation and our `createDatabase`.
+            // Cloudflare reports this as `DatabaseAlreadyExists` (the
+            // canonical 409 mapping) or, on older paths, as a generic
+            // `InvalidProperty` complaining about the name. Catch both,
+            // re-list, and adopt the existing row instead of failing.
+            const onConflict = Effect.gen(function* () {
+              const dbs = yield* listDbs({ accountId: acct, name });
+              const match = dbs.result.find((db) => db.name === name);
+              if (match) {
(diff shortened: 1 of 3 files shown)


sam-goodwin:

PR #197, opened 2026-05-05
fix(axiom): harden Dataset reconcile + add lifecycle convergence tests

sam-goodwin:
Stacked on top of [#179](https://github.com/alchemy-run/alchemy-effect/pull/179). Hardens the Axiom Dataset reconcile path the same way [#184](https://github.com/alchemy-run/alchemy-effect/pull/184) (SQS) and [#190](https://github.com/alchemy-run/alchemy-effect/pull/190) (Cloudflare Worker) did for their resources.

### Reconciler changes

The collision-recovery path on `createDataset` was casting the error channel to `{ _tag: string }`, hiding the real types, and only catching `Conflict | UnprocessableEntity` — but Axiom's `createDataset` doesn't actually declare `Conflict`, and 409s come back as `UnknownAxiomError` via the SDK's `HTTP_STATUS_MAP` fallback.

```diff
- ).pipe(
-   Effect.catchIf(
-     (e): e is { readonly _tag: "Conflict" | "UnprocessableEntity" } =>
-       e._tag === "Conflict" || e._tag === "UnprocessableEntity",
-     () => update({ ... }),
-   ),
- );
+ ).pipe(
+   Effect.catch((e: { readonly _tag?: string }) => {
+     const tag = e._tag;
+     if (
+       tag === "UnprocessableEntity" ||
+       tag === "Conflict" ||
+       tag === "UnknownAxiomError"
+     ) {
+       return Effect.gen(function* () {
+         const existing = yield* get({ dataset_id: news.name }).pipe(
+           Effect.catchTag("NotFound", () => Effect.succeed(undefined)),
+         );
+         if (existing === undefined) return yield* Effect.fail(e);
+         const ownership = parseMarker(existing.description);
+         const isOurs =
+           ownership === undefined ||
+           (ownership.stack === stack.name &&
+             ownership.stage === stage &&
+             ownership.id === id);
+         if (!isOurs) return yield* Effect.fail(e);
+         return yield* update({ dataset_id: existing.id, ... });
+       });
+     }
+     return Effect.fail(e);
+   }),
+ );
```

`Conflict` was unreachable as written, and the cast widened the channel to "anything tagged" so genuine `BadRequest` / `Forbidden` failures could fall through silently. Re-observing before update also closes a hijack window: if state was lost and a foreign owner held the same dataset name, the previous code would silently overwrite their description with our marker.

The sync step now diffs against OBSERVED state instead of `olds`, and recovers from a mid-reconcile out-of-band delete by recreating:

```diff
- const needsSync =
-   current.description !== desiredDescription ||
-   current.retentionDays !== news.retentionDays ||
-   current.useRetentionPeriod !== news.useRetentionPeriod;
+ const needsSync =
+   observed.description !== desiredDescription ||
+   observed.retentionDays !== news.retentionDays ||
+   observed.useRetentionPeriod !== news.useRetentionPeriod;
```

```diff
  const updated = yield* update({
    dataset_id: observed.id,
    description: desiredDescription,
    retentionDays: news.retentionDays,
    useRetentionPeriod: news.useRetentionPeriod,
- });
+ }).pipe(
+   Effect.catchTag("NotFound", () =>
+     create({ name: news.name, description: desiredDescription, ... }),
+   ),
+ );
```

### New lifecycle tests

- redeploy with same props is a no-op (id + created stable)
- reconcile resets description / `retentionDays` mutated out-of-band via `AxiomSdk.updateDataset` (also exercises marker restoration)
- reconcile re-creates a dataset that was deleted out-of-band via `AxiomSdk.deleteDataset`
- changing `name` triggers replace; old dataset is gone
- destroying an already-deleted dataset is a no-op (delete catches `NotFound`)
- `adopt(true)` re-claims a foreign-marked dataset and re-points the marker at the new logical id

### Distilled patch

No patch needed. Axiom's `client/api.ts` already routes HTTP status to typed errors via `HTTP_STATUS_MAP` (with `UnknownAxiomError` as the fallback) and the `Retry` Context.Service already drives backoff for transient categories. The reconciler change is purely about catching the right shapes alchemy-side.

### DIFF
--- a/packages/alchemy/src/Axiom/Dataset.ts
+++ b/packages/alchemy/src/Axiom/Dataset.ts
@@ -210,67 +210,119 @@ export const DatasetProvider = () =>
           // also stable input. Probe for live state by name (preferring the
           // cached `output.id` when present). `read` upstream has already
           // surfaced foreign datasets as `Unowned`, so by the time we land
-          // here mutation is safe.
+          // here mutation is safe — but we still re-confirm ownership on
+          // the conflict path below to avoid hijacking on a state-loss
+          // race where `read` returned `undefined` (transient NotFound or
+          // an out-of-band recreate by a foreign owner).
           const datasetId = output?.id ?? news.name;
           const observed = yield* get({ dataset_id: datasetId }).pipe(
             Effect.catchTag("NotFound", () => Effect.succeed(undefined)),
           );
 
-          // Ensure — POST creates the dataset. Tolerate Conflict/
-          // UnprocessableEntity as a race with a peer reconciler (or with
-          // upstream read↔create), falling through to the sync path.
-          let current = observed;
-          if (current === undefined) {
-            current = yield* (
-              create({
-                name: news.name,
-                description: augmentDescription(news.description, marker),
-                kind: news.kind,
-                retentionDays: news.retentionDays,
-                useRetentionPeriod: news.useRetentionPeriod,
-              }) as Effect.Effect<
-                Axiom.CreateDatasetOutput,
-                { readonly _tag: string },
-                never
-              >
-            ).pipe(
-              Effect.catchIf(
-                (
-                  e,
-                ): e is { readonly _tag: "Conflict" | "UnprocessableEntity" } =>
-                  e._tag === "Conflict" || e._tag === "UnprocessableEntity",
-                () =>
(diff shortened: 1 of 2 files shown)


sam-goodwin:

PR #199, opened 2026-05-05
fix(aws/ec2): add read lifecycle to enable Vpc adoption

sam-goodwin:
Stacked on top of [#179](https://github.com/alchemy-run/alchemy-effect/pull/179). VPC was the only AWS canonical resource without a `read` handler — flagged in [#189](https://github.com/alchemy-run/alchemy-effect/pull/189) as the gap that forced the adopt-takeover test to be dropped from the VPC hardening pass.

### What changed

```ts
// packages/alchemy/src/AWS/EC2/Vpc.ts
read: Effect.fn(function* ({ id, output }) {
  let vpc: EC2.Vpc | undefined;
  if (output?.vpcId) {
    const lookup = yield* ec2.describeVpcs({ VpcIds: [output.vpcId] }).pipe(
      Effect.catchTag("InvalidVpcID.NotFound", () =>
        Effect.succeed({ Vpcs: [] }),
      ),
    );
    vpc = lookup.Vpcs?.[0];
  }
  if (!vpc) {
    const filters = yield* createAlchemyTagFilters(id);
    const lookup = yield* ec2.describeVpcs({ Filters: filters });
    vpc = lookup.Vpcs?.[0];
  }
  if (!vpc) return undefined;
  // build attrs from vpc + tags …
  return (yield* hasAlchemyTags(id, tags)) ? attrs : Unowned(attrs);
}),
```

VPC ids are auto-assigned, so observation falls back to filtering `describeVpcs` by `alchemy::stack`/`alchemy::stage`/`alchemy::id`. `InvalidVpcID.NotFound` on the id-based first pass is treated as a race; the tag-filter pass returns empty when there's no match and we surface `undefined`.

Also picks up the same `output?.stableId ?? …` / `output?.stableArn ?? …` fallback in `test.resources.ts` that prior PRs in the sweep applied.

### New tests

- `owned vpc (matching alchemy tags) is silently adopted without --adopt` — wipe state, redeploy, assert vpcId/Arn carry over
- `foreign-tagged vpc requires adopt(true) to take over` — both `.skip`'d to match the cost profile of the existing `.skip`'d create/update/delete VPC test

### DIFF
--- a/packages/alchemy/src/AWS/EC2/Vpc.ts
+++ b/packages/alchemy/src/AWS/EC2/Vpc.ts
@@ -9,7 +9,14 @@ import { isResolved, somePropsAreDifferent } from "../../Diff.ts";
 import * as Provider from "../../Provider.ts";
 import { Resource } from "../../Resource.ts";
 import type { Providers } from "../Providers.ts";
-import { createInternalTags, createTagsList, diffTags } from "../../Tags.ts";
+import { Unowned } from "../../AdoptPolicy.ts";
+import {
+  createAlchemyTagFilters,
+  createInternalTags,
+  createTagsList,
+  diffTags,
+  hasAlchemyTags,
+} from "../../Tags.ts";
 import type { AccountID } from "../Environment.ts";
 import { AWSEnvironment } from "../Environment.ts";
 import type { RegionID } from "../Region.ts";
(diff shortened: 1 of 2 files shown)


sam-goodwin:

PR #200, opened 2026-05-05
fix(aws/scheduler): harden Schedule reconcile + add lifecycle convergence tests

sam-goodwin:
Stacked on top of [#179](https://github.com/alchemy-run/alchemy-effect/pull/179) and following the SQS hardening template from [#184](https://github.com/alchemy-run/alchemy-effect/pull/184). Hardens the EventBridge Scheduler `Schedule` resource.

### Reconciler changes

```diff
- if (!observed?.Arn) {
-   yield* scheduler.createSchedule({ … })
-     .pipe(Effect.catchTag("ConflictException", () => …listTagsForResource…))
-   observed = yield* scheduler.getSchedule({ … })
- }
- yield* scheduler.updateSchedule({ … })          // always called, even right after create
- // tag sync via tagResource(scheduleArn, …)     // not supported by AWS API
+ const arn = observed?.Arn
+   ? yield* conflictRetry(scheduler.updateSchedule({ … })).pipe(Effect.map(r => r.ScheduleArn))
+   : yield* scheduler.createSchedule({ … }).pipe(conflictRetry, Effect.map(r => r.ScheduleArn));
```

- Drop the always-on post-create `updateSchedule` call — `createSchedule` is already a full PUT, so an immediate update is wasted work and a `ConflictException` source.
- Drop tag-list / tag / untag flow on the schedule itself. EventBridge Scheduler **does not support tagging individual schedules** (only schedule groups), so `tagResource(scheduleArn, …)` would 400. The `tags` prop is removed from `ScheduleProps`; ownership is identified via the deterministic `(GroupName, Name)` tuple instead.
- Bounded-retry `ConflictException` on `updateSchedule`/`createSchedule`/`deleteSchedule` (15 × 2s). Pairs with the distilled patch below so the default AWS retry layer can also help.
- `read` returns `Unowned(attrs)` when `output === undefined` (engine has never owned this schedule), forcing `--adopt`/`adopt(true)` for first-touch adoption.
- `delete` now also rides `ConflictException`.

### New lifecycle tests

`packages/alchemy/test/AWS/Scheduler/Schedule.test.ts` — each test runs `destroy → deploy → … → destroy` on a `ScratchStack` and asserts convergence at every step:

- create + delete with default props
- reconcile resets `ScheduleExpression` / `State` / `FlexibleTimeWindow` / `Description` mutated out-of-band via the raw SDK
- reconcile re-creates a schedule deleted out-of-band
- changing `name` triggers replace; old schedule is deleted
- destroying an already-deleted schedule is a no-op
- foreign schedule (no engine state) requires `adopt(true)` to take over

### Distilled patch

`ConflictException` was tagged `ConflictError` only. Filed [alchemy-run/distilled#248](https://github.com/alchemy-run/distilled/pull/248) to add `RetryableError` so transient mid-update conflicts ride the default AWS retry layer instead of surfacing as hard failures. The reconciler-level bounded retry stays as a safety net.

### Other

- `test.resources.ts` `output?.stableId ?? …` fallback applied to fix the `'output' possibly undefined` tsc error in the static-stables resource test fixture.

### DIFF
--- a/packages/alchemy/src/AWS/Scheduler/Schedule.ts
+++ b/packages/alchemy/src/AWS/Scheduler/Schedule.ts
@@ -1,17 +1,13 @@
 import * as scheduler from "@distilled.cloud/aws/scheduler";
 import * as Effect from "effect/Effect";
+import * as Schedule_ from "effect/Schedule";
+import { Unowned } from "../../AdoptPolicy.ts";
 import { isResolved } from "../../Diff.ts";
 import type { Input } from "../../Input.ts";
 import { createPhysicalName } from "../../PhysicalName.ts";
 import * as Provider from "../../Provider.ts";
 import { Resource } from "../../Resource.ts";
 import type { Providers } from "../Providers.ts";
-import {
-  createInternalTags,
-  createTagsList,
-  diffTags,
-  hasTags,
-} from "../../Tags.ts";
 
 export interface ScheduleProps {
   /**
@@ -62,10 +58,6 @@ export interface ScheduleProps {
    * Action after a one-time schedule completes.
    */
   actionAfterCompletion?: string;
-  /**
-   * User-defined tags.
-   */
-  tags?: Record<string, string>;
 }
 
 /**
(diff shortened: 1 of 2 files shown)


sam-goodwin:

PR #201, opened 2026-05-05
fix(aws/rds): harden DBInstance reconcile + add lifecycle convergence tests

sam-goodwin:
Stacked on #179 (`refactor(core): collapse provider create+update into reconcile`). Per-resource hardening sweep for `AWS.RDS.DBInstance`.

### Reconciler changes

`waitForInstance` previously returned as soon as `DBInstanceArn` was non-null and capped at a 60s budget. RDS create cycles are 5-15 minutes, so the wait was strictly less than the typical create time and would return an instance still in `creating` status. Replace with `waitForStableInstance` which only succeeds on a terminal-for-our-purposes status and polls for up to 25min.

```ts
const isStableInstanceStatus = (status: string | undefined) =>
  status === "available" ||
  status === "stopped" ||
  status === "incompatible-network" ||
  // ...
  status === "failed";
```

`modifyDBInstance` was called blindly on every reconcile after observation — even when desired matched observed — triggering a needless `modifying` transition. Replace with a diff-driven `computeModifyPayload` that returns `undefined` on no-op so the API call is skipped entirely.

```ts
const modifyPayload = computeModifyPayload(observed, news);
if (modifyPayload) {
  yield* retryControlPlane(rds.modifyDBInstance(modifyPayload));
  observed = yield* waitForStableInstance(identifier, session);
}
```

Wrap `modifyDBInstance` / `addTagsToResource` / `removeTagsFromResource` in `retryControlPlane` which rides out `InvalidDBInstanceStateFault` and `DBInstanceNotFoundFault` races (instance mid-transition or just-tagged).

```ts
const retryControlPlane = <A, E extends { _tag?: string }, R>(
) =>
  effect.pipe(
    Effect.retry({
      while: (e) =>
        e._tag === "InvalidDBInstanceStateFault" ||
        e._tag === "DBInstanceNotFoundFault",
      schedule: Schedule.exponential("1 second").pipe(
        Schedule.both(Schedule.recurs(20)),
      ),
    }),
  );
```

`delete` now retries `InvalidDBInstanceStateFault` (so a delete issued during a backup or modify converges) and waits for the instance to actually leave RDS via `waitForInstanceDeleted` so a subsequent reconcile or replace doesn't race against `deleting` state.

If observation finds an existing instance in a non-stable status (e.g. mid-creation from a previous reconcile that crashed before persisting state), wait for stable before issuing modify to avoid `InvalidDBInstanceStateFault`.

Replace ad-hoc `Effect.fail(new Error(...))` sites with typed `DBInstanceUnreadable` / `DBInstanceNotStable` / `DBInstanceStillDeleting` so the IaC error channel stays typed.

Add `allocatedStorage` / `backupRetentionPeriod` / `preferredBackupWindow` props (and corresponding attributes) so the resource is usable for non-Aurora standalone instances and so the storage-scaling drift test has something to mutate.

### New lifecycle tests

Skipped by default — RDS provisioning is too costly for CI. Unskip when verifying.

- Redeploy with same props is a no-op (reconcile is idempotent)
- Reconcile resets `backupRetentionPeriod` / `preferredBackupWindow` / `autoMinorVersionUpgrade` / `copyTagsToSnapshot` / tags after out-of-band SDK mutation, including re-stamping internal alchemy tags after they're untagged
- Changing `dbInstanceIdentifier` triggers replace, old instance is deleted
- In-place modification of `allocatedStorage` scales a standalone instance (Aurora cluster members ignore it)
- Destroying an already-deleted instance is a no-op
- `adopt(true)` re-tags a foreign instance

### Distilled patch

https://github.com/alchemy-run/distilled/pull/249 — adds the missing `errorCategories` block so `InvalidDBInstanceState` / `Throttling` / `RequestLimitExceeded` / capacity faults participate in the AWS retry layer instead of surfacing as hard `BadRequestError`.

### Baseline fix

`packages/alchemy/test/test.resources.ts` had a pre-existing tsc error on this branch (`output` is possibly undefined post-reconcile refactor). Carried forward the same fix the SQS / DynamoDB hardening PRs applied so the baseline `bun tsc -b` is clean.

### DIFF
--- a/packages/alchemy/src/AWS/RDS/DBInstance.ts
+++ b/packages/alchemy/src/AWS/RDS/DBInstance.ts
@@ -1,4 +1,5 @@
 import * as rds from "@distilled.cloud/aws/rds";
+import * as Data from "effect/Data";
 import * as Effect from "effect/Effect";
 import * as Schedule from "effect/Schedule";
 import { isResolved } from "../../Diff.ts";
@@ -58,6 +59,20 @@ export interface DBInstanceProps {
    */
   copyTagsToSnapshot?: boolean;
   /**
+   * Allocated storage in GiB. Only used for non-Aurora instances —
+   * Aurora cluster members ignore this and consume the cluster volume.
+   */
+  allocatedStorage?: number;
+  /**
+   * Daily backup retention window in days. Aurora cluster members
+   * inherit this from the cluster, so set it on the cluster instead.
+   */
+  backupRetentionPeriod?: number;
+  /**
+   * Preferred backup window in `hh24:mi-hh24:mi` UTC format.
+   */
+  preferredBackupWindow?: string;
+  /**
    * User-defined tags.
    */
   tags?: Record<string, string>;
@@ -78,6 +93,9 @@ export interface DBInstance extends Resource<
     status: string | undefined;
     promotionTier: number | undefined;
     publiclyAccessible: boolean | undefined;
+    allocatedStorage: number | undefined;
+    backupRetentionPeriod: number | undefined;
+    preferredBackupWindow: string | undefined;
     dbSubnetGroupName: string | undefined;
     dbParameterGroupNames: string[];
     tags: Record<string, string>;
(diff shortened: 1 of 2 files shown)


sam-goodwin:

PR #202, opened 2026-05-05
fix(cloudflare/queue): harden Queue reconcile + add lifecycle convergence tests

sam-goodwin:
Five blanket `Effect.catch(() => Effect.succeed(...))` sites in the `Queue` and `QueueConsumer` providers were swallowing every error class, including auth, throttling, and 5xx. Scope each catch to the specific tag the path is recovering so transient failures surface for engine-level retry instead of being misread as "the resource is missing" or "this is a duplicate".

### Reconciler changes

```diff
   if (output?.queueId) {
     observed = yield* getQueue({
       accountId: acct,
       queueId: output.queueId,
-    }).pipe(Effect.catch(() => Effect.succeed(undefined)));
+    }).pipe(
+      Effect.catchTag("QueueNotFound", () => Effect.succeed(undefined)),
+    );
   }
```

```diff
   if (!observed) {
     observed = yield* createQueue({ accountId: acct, queueName }).pipe(
-      Effect.catch(() =>
+      Effect.catchTag("QueueAlreadyExists", () =>
         Effect.gen(function* () {
           const match = yield* findQueueByName(queueName);
           ...
         }),
       ),
     );
   }
```

```diff
   yield* deleteQueue({
     accountId: output.accountId,
     queueId: output.queueId,
-  }).pipe(Effect.catch(() => Effect.void));
+  }).pipe(Effect.catchTag("QueueNotFound", () => Effect.void));
```

`QueueConsumer` gets the same treatment scoped to `ConsumerNotFound` / `QueueNotFound` on get/read/delete, and `ConsumerAlreadyExists` on the create-race recovery.

A failed `getQueue` used to be misread as "queue is missing", which then fell through to `findQueueByName` — but with auth/throttling, the list scan also fails, and the next ensure step would try to create a duplicate. A failed `createQueue` of any kind used to be silently treated as "name taken", masking quota / network / 5xx issues until the subsequent list-scan also surfaced nothing useful.

### New lifecycle tests

`packages/alchemy/test/Cloudflare/Queue/Queue.test.ts` (new file):

- redeploy with same props is a no-op (queueId preserved)
- reconcile re-creates a queue that was deleted out-of-band via `queues.deleteQueue`
- changing physical `name` triggers replace; old queue is gone
- destroying an already-deleted queue is a no-op
- existing queue (matching name) is silently adopted without `--adopt`
- `adopt(true)` re-claims a queue under a different logical id
- consumer add / remove churn against a stable queue + worker
- consumer settings update propagates without replacing the consumer
- DLQ chaining survives redeploy (main + DLQ + consumer all stable)
- destroying an already-deleted consumer is a no-op

### Distilled patch

No patch needed. The existing `packages/cloudflare/patches/queues/*.json` patches in distilled already map the response codes the reconciler now routes on: `QueueAlreadyExists` (11009), `QueueNotFound` (11000), `ConsumerAlreadyExists` (11004), `ConsumerNotFound` (11006).

The pre-existing `output.stableId` TS error from #167/#179 in `test.resources.ts` is fixed the same way the prior hardening PRs did (`output?.stableId ?? ...`, `output?.stableArn ?? ...`).

### DIFF
--- a/packages/alchemy/src/Cloudflare/Queue/Queue.ts
+++ b/packages/alchemy/src/Cloudflare/Queue/Queue.ts
@@ -142,7 +142,9 @@ export const QueueProvider = () =>
 
           // Observe — re-fetch the cached queue; fall back to a name scan
           // when the cached id is gone (out-of-band delete or partial
-          // state-persistence failure).
+          // state-persistence failure). Only swallow `QueueNotFound`
+          // here so auth/throttling/5xx failures surface to the engine
+          // for retry rather than being misread as "queue is missing".
           let observed:
             | { queueId?: string | null; queueName?: string | null }
             | undefined;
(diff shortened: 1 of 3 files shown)


sam-goodwin:

PR #203, opened 2026-05-05
fix(aws/apigateway): harden RestApi reconcile + add lifecycle convergence tests

sam-goodwin:
Stacked on top of [#179](https://github.com/alchemy-run/alchemy-effect/pull/179). Hardens the ApiGateway RestApi (and Stage create-race) as the next pass of the per-resource hardening sweep.

### Reconciler changes

```diff
- if (!output?.restApiId) {
-   return yield* Effect.die(
-     "RestApi reconcile reached without a precreate output",
-   );
- }
- const observed = yield* ag.getRestApi({ restApiId: output.restApiId }).pipe(
-   Effect.catchTag("NotFoundException", () => Effect.succeed(undefined)),
- );
- if (!observed?.id) {
-   return yield* Effect.die(`RestApi ${output.restApiId} disappeared ...`);
- }
+ let observed = output?.restApiId
+   ? yield* ag.getRestApi({ restApiId: output.restApiId }).pipe(
+       Effect.catchTag("NotFoundException", () => Effect.succeed(undefined)),
+     )
+   : undefined;
+ if (!observed?.id) {
+   const created = yield* retryOnApiStatusUpdating(
+     ag.createRestApi({ name, ...desired, tags: desiredTags }),
+   );
+   observed = yield* ag.getRestApi({ restApiId: created.id! });
+ }
```

`reconcile` previously died if `output` was missing or pointed at a deleted-out-of-band API. The reconciler doctrine says reconcile must converge for greenfield, routine update, and adoption alike — folded the precreate body into the missing-observation branch so out-of-band deletion now self-heals.

```diff
- Effect.catchTag("BadRequestException", () => Effect.void),
+ Effect.catchTag("BadRequestException", (e) =>
+   isTagKeyAbsentError(e) ? Effect.void : Effect.fail(e),
+ ),
```

`syncTags` wrapped `untagResource` in a blanket `BadRequestException` swallow — every malformed tag key, invalid arn, or auth failure was silently treated as "tags converged". Scoped to the "tag key does not exist" message only.

```diff
  ag.createStage({ ... })
+   .pipe(Effect.catchTag("ConflictException", () => Effect.void))
```

Concurrent reconciles (or a state-loss replay) racing `createStage` would surface `ConflictException` instead of falling through to the sync step. Tolerated as a race, observed state then drives the diff.

### New lifecycle tests

Each test runs `destroy → deploy → … → destroy` on a `ScratchStack` and asserts convergence at every step. They run live under the `apigateway` vitest project (single-fork, no parallelism, 600s timeout).

- reconcile resets `description` / `binaryMediaTypes` mutated out-of-band
- reconcile resets `policy` mutated out-of-band
- reconcile re-creates a RestApi that was deleted out-of-band (rides the 30s `DeleteRestApi` throttle)
- destroying an already-deleted RestApi is a no-op
- silent adoption when tags match alchemy ownership tags
- `adopt(true)` re-tags a foreign RestApi
- a Stage with `tracingEnabled` + `variables` re-converges after drift

### Distilled patch

No patch needed. The api-gateway service in `@distilled.cloud/aws@0.16.3` already categorizes `TooManyRequestsException` and `LimitExceededException` as `ThrottlingError`, `ServiceUnavailableException` as `ServerError`, `ConflictException` as `ConflictError`, and `UnauthorizedException` as `AuthError` — which feeds the AWS retry layer correctly. The reconciler-level `retryOnApiStatusUpdating` stays for the 30s `DeleteRestApi` throttle window which exceeds the SDK's default 5-attempt budget.

### DIFF
--- a/packages/alchemy/src/AWS/ApiGateway/RestApi.ts
+++ b/packages/alchemy/src/AWS/ApiGateway/RestApi.ts
@@ -387,34 +387,54 @@ export const RestApiProvider = () =>
             return yield* Effect.die("RestApi props were not resolved");
           }
           const news = newsIn as RestApiProps;
+          const internalTags = yield* createInternalTags(id);
+          const desiredTags = { ...(news.tags ?? {}), ...internalTags };
 
-          // RestApi has a `precreate` that always runs before `reconcile`
-          // for greenfield deployments — it creates the REST API in full
-          // so that `restApi.restApiId` is resolvable and child resources
-          // (Method, Resource, Authorizer) can register themselves back on
-          // the API via `restApi.bind`. By the time `reconcile` runs the id
-          // is populated; we never expect `output === undefined` here, but
-          // we still handle it defensively.
-          if (!output?.restApiId) {
-            return yield* Effect.die(
-              "RestApi reconcile reached without a precreate output",
-            );
-          }
+          // Observe — fetch live cloud state by the stable id we cached in
+          // `output`. `output` is treated as a cache for the stable id
+          // only; it may not exist (greenfield, or state was lost) and the
+          // resource it points at may have been deleted out-of-band.
+          let observed = output?.restApiId
+            ? yield* ag
+                .getRestApi({ restApiId: output.restApiId })
+                .pipe(
+                  Effect.catchTag("NotFoundException", () =>
+                    Effect.succeed(undefined),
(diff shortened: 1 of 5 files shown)


sam-goodwin:

PR #204, opened 2026-05-05
fix(cloudflare/kv): harden Namespace reconcile + add lifecycle convergence tests

sam-goodwin:
Tighten the rename hop in `KVNamespace.reconcile` so it tolerates two real-world races that previously surfaced as hard failures, and add lifecycle convergence tests covering the standard reconcile shape.

The rest of the reconciler was already well-scoped: every catch is a `catchTag` (no blanket `Effect.catch`), there are no unbounded retries, and `getNamespace` already races against out-of-band deletes on the observation hop. The gap was on the rename-via-`updateNamespace` hop.

### Reconciler changes

```diff
   if (observed.title !== title) {
     const renamed = yield* updateNamespace({
       accountId: acct,
       namespaceId: observed.id,
       title,
-    });
+    }).pipe(
+      Effect.catchTag("NamespaceNotFound", () =>
+        // namespace deleted between observe and update — recreate
+        createNamespace({ accountId: acct, title }).pipe(
+          Effect.catchTag("NamespaceTitleAlreadyExists", () =>
+            Effect.gen(function* () {
+              const match = yield* findNamespaceByTitle(title);
+              if (match) return match;
+              return yield* Effect.die(...);
+            }),
+          ),
+        ),
+      ),
+      Effect.catchTag("NamespaceTitleAlreadyExists", () =>
+        // parallel deploy / dashboard rename took the title — adopt
+        Effect.gen(function* () {
+          const match = yield* findNamespaceByTitle(title);
+          if (match) return match;
+          return yield* Effect.die(...);
+        }),
+      ),
+    );
```

Without these catches a same-title redeploy that races an OOB delete would crash and leave engine state pointing at a dead namespace; a parallel rename would crash with a confusing 409 instead of converging onto the existing namespace.

### New lifecycle tests

- redeploy with same props is a no-op (reconciler is idempotent)
- reconcile resets a title mutated out-of-band via `kv.updateNamespace`
- reconcile re-creates a namespace deleted out-of-band via `kv.deleteNamespace`
- changing `title` triggers in-place rename (same `namespaceId`, not replace)
- destroying an already-deleted namespace is a no-op (delete is idempotent)
- `adopt(true)` re-claims a foreign namespace by title

### Distilled patch

No patch needed. The named KV errors (`NamespaceNotFound`, `NamespaceTitleAlreadyExists`, `TitleRequired`, `InvalidObjectIdentifier`, `MethodNotAllowed`) already cover every code the reconciler routes on, and Cloudflare's `client/api.ts` already attributes throttling/5xx categories so they participate in engine-level retry instead of surfacing as `UnknownCloudflareError`.

The pre-existing `output?.stableId` / `output?.stableArn` TS error from #167/#179 in `test.resources.ts` is fixed the same way the prior hardening PRs did.

### DIFF
--- a/packages/alchemy/src/Cloudflare/KV/KVNamespace.ts
+++ b/packages/alchemy/src/Cloudflare/KV/KVNamespace.ts
@@ -151,6 +151,17 @@ export const KVNamespaceProvider = () =>
           // Sync — KV's only mutable property is the title. Rename only
           // when the observed title drifts from desired so we avoid
           // unnecessary API calls on every reconcile.
+          //
+          // Two races are tolerated on the rename hop:
+          //   * `NamespaceNotFound` — the namespace was deleted between
+          //     observe and update. Recurse via reconcile rather than
+          //     leaving the engine with stale state pointing at a dead
+          //     namespace; the next call observes "missing" and creates
+          //     fresh.
+          //   * `NamespaceTitleAlreadyExists` — a parallel deploy or
+          //     dashboard rename took the desired title first. Re-find
+          //     the namespace by title and adopt; if it's not actually
+          //     there, surface the original error so it isn't masked.
           let namespaceId = observed.id;
           let resolvedTitle = observed.title;
           let supportsUrlEncoding = observed.supportsUrlEncoding ?? undefined;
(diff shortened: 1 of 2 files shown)


sam-goodwin:

PR #205, opened 2026-05-05
fix(aws/route53): harden Record reconcile + add lifecycle convergence tests

sam-goodwin:
Replace `Effect.die` and blanket `Effect.catchTag("InvalidChangeBatch")` in the Route 53 Record provider with targeted handling, and add lifecycle tests against a real hosted zone.

### Reconciler changes

- `waitForChange` now fails with a typed `Route53ChangePending` instead of `Effect.die`, so a stuck change batch surfaces cleanly instead of crashing the engine.
- Every mutating Route 53 call (`changeResourceRecordSets`, `getChange`, `listResourceRecordSets`) is wrapped in a `PriorRequestNotComplete` retry. AWS classifies that error as `BadRequestError` so the AWS retry layer doesn't auto-retry — but it's the standard hosted-zone-serialization signal and is always transient.
- `reconcile` observes the record first and skips the change batch entirely when the cloud state already matches desired. Same-spec redeploys no longer generate change batches at all.
- Re-read after upsert is bounded with `Route53RecordNotVisible` retry instead of `Effect.die`, to ride out the brief list/change ordering window.

```ts
// before
Effect.flatMap((changeInfo) =>
  changeInfo.Status === "INSYNC"
    ? Effect.succeed(changeInfo)
    : Effect.die(new Error("Route53ChangePending")),
),

// after
Effect.flatMap((changeInfo) =>
  changeInfo.Status === "INSYNC"
    ? Effect.succeed(changeInfo)
    : Effect.fail(new Route53ChangePending({ changeId })),
),
```

`delete` reads the observed record set and echoes it back to `DELETE`, so out-of-band drift can't trigger `InvalidChangeBatch`. Only `InvalidChangeBatch` carrying a "not found" message is tolerated.

### New lifecycle tests

`packages/alchemy/test/AWS/Route53/Record.test.ts` — each test provisions its own randomized public hosted zone, runs the scenario, and tears the zone down:

- reconcile resets TTL / values mutated out-of-band
- reconcile re-creates a record that was deleted out-of-band
- changing record `type` triggers replace
- changing record `name` triggers replace
- multiple records sharing a hosted zone all converge (`PriorRequestNotComplete` handling)
- destroying an already-deleted record is a no-op

### Distilled patch

https://github.com/alchemy-run/distilled/pull/250 — categorizes `PriorRequestNotComplete` as `ConflictError + RetryableError` and `ThrottlingException` as `ThrottlingError + RetryableError` so the AWS retry layer rides them out at the SDK level. The provider's local retry is kept as a defense-in-depth.

### DIFF
--- a/packages/alchemy/src/AWS/Route53/Record.ts
+++ b/packages/alchemy/src/AWS/Route53/Record.ts
@@ -1,4 +1,5 @@
 import * as route53 from "@distilled.cloud/aws/route-53";
+import * as Data from "effect/Data";
 import * as Effect from "effect/Effect";
 import * as Schedule from "effect/Schedule";
 import { isResolved } from "../../Diff.ts";
@@ -162,7 +163,7 @@ const toRecordSet = (props: RecordProps): route53.ResourceRecordSet => ({
         HostedZoneId: normalizeHostedZoneId(
           props.aliasTarget.hostedZoneId as string,
         ),
-        DNSName: props.aliasTarget.dnsName as string,
+        DNSName: normalizeName(props.aliasTarget.dnsName as string),
         EvaluateTargetHealth: props.aliasTarget.evaluateTargetHealth ?? false,
       }
     : undefined,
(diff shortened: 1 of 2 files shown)


sam-goodwin:

PR #206, opened 2026-05-05
fix(aws/eventbridge): harden Rule reconcile + add lifecycle convergence tests

sam-goodwin:
Harden the EventBridge Rule reconcile path against transient EventBridge faults and adoption / drift edge cases, and add the lifecycle test suite that proves it.

### Reconciler changes

```ts
// retry the EventBridge transients that previously surfaced as terminal failures
const retryEventBridgeTransients = <A, E, R>(effect: Effect.Effect<A, E, R>) =>
  effect.pipe(
    Effect.retry({
      while: (error: any) =>
        error?._tag === "ConcurrentModificationException" ||
        error?._tag === "InternalException",
      schedule: Schedule.exponential(200).pipe(Schedule.both(Schedule.recurs(8))),
    }),
  );
```

- Wrap `putRule` / `putTargets` / `removeTargets` / `listTargetsByRule` / `describeRule` and every tag operation in `retryEventBridgeTransients` so peer reconcilers don't take the deploy down.
- Catch `ResourceNotFoundException` on `tagResource` / `untagResource` / `listTagsForResource` (Rule + EventBus) so a delete-race during the tag sync degrades gracefully.
- `Rule.delete` and `EventBus.delete` are now idempotent (catch `ResourceNotFoundException`) and wrapped in the same retry helper.

### New lifecycle tests

`packages/alchemy/test/AWS/EventBridge/Rule.test.ts`:

- reconcile resets `ScheduleExpression` / `State` / `Description` / `EventPattern` mutated out-of-band
- reconcile re-creates a rule deleted out-of-band
- changing `name` triggers replace, old rule is deleted
- targets diff: add / remove / update converges
- destroying an already-deleted rule is a no-op
- silent adoption when alchemy tags match
- `adopt(true)` re-tags a foreign rule with the internal alchemy tags

### Distilled patch

https://github.com/alchemy-run/distilled/pull/251 — adds error categories so `ConcurrentModificationException`, `InternalException`, `ThrottlingException` participate in the AWS retry layer, plus `ResourceNotFoundException` on `deleteEventBus` (documented but missing from Smithy) so `catchTag` works.

### DIFF
--- a/packages/alchemy/src/AWS/EventBridge/EventBus.ts
+++ b/packages/alchemy/src/AWS/EventBridge/EventBus.ts
@@ -1,6 +1,7 @@
 import { Region } from "@distilled.cloud/aws/Region";
 import * as eventbridge from "@distilled.cloud/aws/eventbridge";
 import * as Effect from "effect/Effect";
+import * as Schedule from "effect/Schedule";
 
 import { Unowned } from "../../AdoptPolicy.ts";
 import { isResolved } from "../../Diff.ts";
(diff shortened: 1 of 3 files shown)


sam-goodwin:

PR #208, opened 2026-05-05
fix(aws/cloudfront): harden Distribution reconcile + add lifecycle convergence tests

sam-goodwin:
CloudFront Distribution reconcile was swallowing every error in the create path via a blanket `Effect.catch` + string-matched `isAccessDenied`, and updates never re-fetched the ETag on stale-version races, so any concurrent edit (or our own waitForDeployment racing the next reconcile) turned `PreconditionFailed`/`InvalidIfMatchVersion` into hard failures.

### Reconciler changes

```diff
-              .pipe(
-                Effect.catch((error) =>
-                  isAccessDenied(error)
-                    ? Effect.gen(function* () {
-                        // ... fallback create-then-tag
-                      })
-                    : Effect.fail(error),
-                ),
+              .pipe(
+                Effect.catchTag("AccessDenied", () =>
+                  Effect.gen(function* () {
+                    // create-then-tag fallback for create-allowed + tag-denied
+                  }),
+                ),
```

Update flow now reads the ETag fresh inside a closure and retries on stale-ETag via a tagged `DistributionStaleEtag` error with bounded `recurs(24) x 5s`:

```ts
const doUpdate = Effect.gen(function* () {
  const fresh = yield* getCurrent(distributionId);
  if (!fresh) return yield* Effect.fail(new DistributionMissing({ ... }));
  return yield* cloudfront.updateDistribution({
    Id: distributionId,
    IfMatch: fresh.etag,
    DistributionConfig: toConfig(fresh.config.CallerReference, news),
  });
});
const updated = yield* doUpdate.pipe(
  Effect.catchTag("InvalidIfMatchVersion", (e) => Effect.fail(new DistributionStaleEtag(...))),
  Effect.catchTag("PreconditionFailed",   (e) => Effect.fail(new DistributionStaleEtag(...))),
  Effect.retry({ while: (e) => e instanceof DistributionStaleEtag, schedule: ... }),
);
```

`PreconditionFailed` and `InvalidIfMatchVersion` stay context-dependent in distilled — only retried inside this read-modify-write block, never blanket-retryable.

Delete flow gets the same treatment for both the disable-update and the final `deleteDistribution`, plus tolerates `NoSuchDistribution` as already-gone. `getCurrent` tolerates `NoSuchDistribution`/`NoSuchResource` mid-flight so observation falls through to create during eventual-consistency races.

### New lifecycle tests

- reconcile resets `comment` mutated out-of-band via raw CloudFront SDK (exercises read-then-update with fresh ETag)
- destroying an already-deleted distribution is a no-op (catches `NoSuchDistribution`)
- `adopt(true)` re-tags a foreign distribution

Each is gated on `ALCHEMY_RUN_LIVE_AWS_WEBSITE_TESTS` and capped at 30 min — CloudFront create/update/delete is 5–15 min per op.

`test.resources.ts` adopts the same `output?.stableId ?? ...` fallback as PR #184 so the file type-checks under the reconciler's adoption signature.

### Distilled patch

No patch needed. `InvalidIfMatchVersion` is `BadRequestError`, `PreconditionFailed` is uncategorised, `IllegalUpdate` is `BadRequestError`, `InvalidArgument` is `BadRequestError` — all correct. Retries are scoped to the reconciler, never injected at the AWS retry layer.

### DIFF
--- a/packages/alchemy/src/AWS/CloudFront/Distribution.ts
+++ b/packages/alchemy/src/AWS/CloudFront/Distribution.ts
@@ -22,6 +22,18 @@ class DistributionPendingDeployment extends Data.TaggedError(
   message: string;
 }> {}
 
+class DistributionStaleEtag extends Data.TaggedError(
+  "DistributionStaleEtag",
+)<{
+  message: string;
+}> {}
+
+class DistributionMissing extends Data.TaggedError(
+  "DistributionMissing",
+)<{
+  message: string;
+}> {}
+
 export interface DistributionOrigin {
   /**
    * Unique origin identifier inside the distribution.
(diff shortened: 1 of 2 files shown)


sam-goodwin:

PR #209, opened 2026-05-05
fix(cloudflare/aigateway): harden AiGateway reconcile + add lifecycle convergence tests

sam-goodwin:
Stacked on top of [#179](https://github.com/alchemy-run/alchemy-effect/pull/179). Hardens the Cloudflare AiGateway resource as part of the per-resource hardening sweep on top of the reconcile migration.

### Reconciler changes

```diff
+ const ensure = Effect.gen(function* () {
+   const observed = yield* getAiGateway({ accountId: acct, id: gatewayId }).pipe(
+     Effect.catchTag("GatewayNotFound", () => Effect.succeed(undefined)),
+   );
+   if (observed !== undefined) return;
+   const request = yield* createRequest(id, news);
+   yield* createAiGateway(request).pipe(
+     Effect.catchTag("GatewayAlreadyExists", () =>
+       getAiGateway({ accountId: acct, id: request.id }),
+     ),
+   );
+ });

  yield* ensure;
  const update = yield* updateRequest(id, news, acct);
- const gateway = yield* updateAiGateway(update);
+ const gateway = yield* updateAiGateway(update).pipe(
+   Effect.catchTag("GatewayNotFound", () =>
+     Effect.gen(function* () {
+       yield* ensure;
+       return yield* updateAiGateway(update);
+     }),
+   ),
+ );
```

The Cloudflare AI Gateway create API doesn't accept `dlp` / `otel` / `storeId` / `stripe`, so the reconciler must always run `updateAiGateway` after ensure. If the gateway was deleted out-of-band between ensure and update — likely on adoption + concurrent operator paths — the update used to surface `GatewayNotFound` as a hard failure. Now we re-ensure (recreate) and retry the update once.

No blanket `Effect.catch` in this reconciler — every catch was already scoped to a single tag.

### New lifecycle tests

Each runs `destroy → deploy → … → destroy` on a `ScratchStack` and asserts convergence at every step.

- redeploy with same props is a no-op (`internalId` preserved)
- reconcile resets cache TTL / rate-limit / authentication mutated out-of-band via raw `updateAiGateway`
- reconcile re-creates a gateway that was deleted out-of-band via raw `deleteAiGateway`
- changing `id` triggers replace; old gateway is deleted
- destroying an already-deleted gateway is a no-op
- `adopt(true)` re-claims a foreign gateway (created out-of-band) and reconciles to desired settings

`test.resources.ts` picks up the same `output?.stableId ?? …` fallback prior hardening PRs used.

### Distilled patch

No patch needed. `GatewayNotFound` and `GatewayAlreadyExists` already cover every codepath the reconciler routes on — no `UnknownCloudflareError` surfaces.

### DIFF
--- a/packages/alchemy/src/Cloudflare/AiGateway/AiGateway.ts
+++ b/packages/alchemy/src/Cloudflare/AiGateway/AiGateway.ts
@@ -485,33 +485,46 @@ export const AiGatewayProvider = () =>
           const gatewayId =
             output?.gatewayId ?? (yield* createGatewayId(id, news.id));
 
-          // Observe — fetch the gateway's current state. The Cloudflare API
-          // returns 404 when the gateway is missing, which we tolerate so the
-          // reconciler can fall through to create.
-          const observed = yield* getAiGateway({
-            accountId: acct,
-            id: gatewayId,
-          }).pipe(
-            Effect.catchTag("GatewayNotFound", () => Effect.succeed(undefined)),
-          );
-
-          // Ensure — create if missing. Tolerate `GatewayAlreadyExists` for
+          // Ensure the gateway exists. Tolerate `GatewayAlreadyExists` for
           // idempotency: a peer reconciler may have created it concurrently,
           // or state persistence may have failed after a previous create.
-          if (observed === undefined) {
+          const ensure = Effect.gen(function* () {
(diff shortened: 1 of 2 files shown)


sam-goodwin:

PR #210, opened 2026-05-05
fix(aws/ecs): harden Cluster reconcile + add lifecycle convergence tests

sam-goodwin:
Stacked on top of [#179](https://github.com/alchemy-run/alchemy-effect/pull/179). Hardens the ECS Cluster resource as part of the per-resource hardening sweep.

### Reconciler changes

```diff
- const cluster = described.clusters?.[0];
+ const cluster = described.clusters?.find(
+   (c) =>
+     c.clusterName === clusterName &&
+     c.status !== "INACTIVE" &&
+     c.status !== "DEPROVISIONING",
+ );
```

ECS keeps INACTIVE clusters around in `describeClusters` for ~1h after delete. `read` was returning these tombstones as if alive, so out-of-band-deleted clusters never converged on redeploy.

```diff
- return { ... tags: output?.tags ?? {} };
+ const attrs = { ... tags: observedTags };
+ return (yield* hasAlchemyTags(id, observedTags))
+   ? attrs
+   : Unowned(attrs);
```

`read` had no adoption gate, so the engine would silently take over any cluster whose name happened to match. Wrap with `Unowned` based on observed alchemy tags so foreign clusters require `--adopt` / `adopt(true)`.

```diff
- yield* ecs.updateCluster({ ... });
+ yield* retryOnUpdateInProgress(ecs.updateCluster({ ... }));
```

`UpdateInProgressException` is transient — ECS surfaces it when a prior cluster mutation hasn't propagated. Bounded retry (60s) on `updateCluster` and `putClusterCapacityProviders`.

```diff
  yield* ecs.deleteCluster({ ... }).pipe(
    Effect.catchTag("ClusterNotFoundException", () => Effect.void),
-   Effect.catchTag("ClusterContainsServicesException", () => Effect.void),
-   Effect.catchTag("ClusterContainsTasksException", () => Effect.void),
  );
```

Silently swallowing `ClusterContains{Services,Tasks,ContainerInstances,CapacityProvider}Exception` made delete falsely report success while leaving the cluster live. Surface them so the engine reports a real dependency violation.

```diff
- yield* elbv2.deleteTargetGroup({ ... }).pipe(Effect.catch(() => Effect.void));
+ yield* elbv2.deleteTargetGroup({ ... });
```

`Service.delete`'s blanket catch on `deleteTargetGroup` would swallow `ResourceInUseException` — a real ordering bug, since the listener is removed first.

### New lifecycle tests

- reconcile resets `containerInsights` mutated out-of-band via the raw ECS SDK
- reconcile re-creates a cluster that was deleted out-of-band (rides the INACTIVE-tombstone window)
- changing `clusterName` triggers replace; old cluster is deleted
- `adopt(true)` re-tags a foreign cluster and preserves user tags
- foreign-tagged cluster requires `adopt(true)` to take over

### Distilled patch

ECS errors had no category annotations in the SDK, so `UpdateInProgressException` failed loud and `ClusterContains*` errors weren't surfaced as dependency violations. Filed as [alchemy-run/distilled#253](https://github.com/alchemy-run/distilled/pull/253). Once published, the AWS retry layer will participate in the transient `UpdateInProgressException` window automatically — the reconciler-level retry stays for now since the default budget may be tighter than ECS's propagation window.

The `test.resources.ts` fallback (`output?.stableId ?? `stable-${id}``) is included since the `StaticStablesResource` update path otherwise refuses to type-check on adoption (where `output` may be undefined at the type level).

### DIFF
--- a/packages/alchemy/src/AWS/ECS/Cluster.ts
+++ b/packages/alchemy/src/AWS/ECS/Cluster.ts
@@ -1,12 +1,14 @@
 import * as ecs from "@distilled.cloud/aws/ecs";
 import { Region } from "@distilled.cloud/aws/Region";
 import * as Effect from "effect/Effect";
+import * as Schedule from "effect/Schedule";
+import { Unowned } from "../../AdoptPolicy.ts";
 import { isResolved } from "../../Diff.ts";
 import { createPhysicalName } from "../../PhysicalName.ts";
 import * as Provider from "../../Provider.ts";
 import { Resource } from "../../Resource.ts";
 import type { Providers } from "../Providers.ts";
-import { createInternalTags, diffTags } from "../../Tags.ts";
+import { createInternalTags, diffTags, hasAlchemyTags } from "../../Tags.ts";
 import { AWSEnvironment, type AccountID } from "../Environment.ts";
 import type { RegionID } from "../Region.ts";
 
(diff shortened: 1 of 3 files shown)


sam-goodwin:

PR #211, opened 2026-05-05
fix(aws/acm): harden Certificate reconcile + add lifecycle convergence tests

sam-goodwin:
Audits the ACM Certificate reconciler for the same classes of issues SQS / S3 / IAM had, and adds lifecycle convergence tests on top of #179.

### Reconciler changes

`waitForValidationRecords`, `waitForIssued`, and `waitForChange` were retrying on bare `new Error("…")` sentinels and matching by `error.message`. Any other unrelated `Error` with the same message would have been retried; any genuine failure with a different message would short-circuit. Replaced with tagged data errors so retry conditions match precisely:

```diff
- return Effect.fail(new Error("CertificateValidationRecordPending"));
+ return Effect.fail(new ValidationRecordsPending());
  …
- while: (error) =>
-   error instanceof Error &&
-   error.message === "CertificateValidationRecordPending",
+ while: (error) => error._tag === "ValidationRecordsPending",
```

`delete` was unconditionally calling `deleteCertificate` and only catching `ResourceNotFoundException`. ACM forbids deleting a cert that's still attached to CloudFront / ELB / API Gateway, and detachment is eventually consistent — even after the user removes the association, ACM may continue surfacing `ResourceInUseException` for tens of seconds. Now bounded-retry (5 minutes total) on `ResourceInUseException`:

```diff
  acm.deleteCertificate({ CertificateArn: output.certificateArn }).pipe(
    Effect.catchTag("ResourceNotFoundException", () => Effect.void),
+   Effect.catchTag("ResourceInUseException", () =>
+     Effect.fail(new CertificateInUse()),
+   ),
+   Effect.retry({
+     while: (e) => e._tag === "CertificateInUse",
+     schedule: Schedule.fixed("5 seconds").pipe(
+       Schedule.both(Schedule.recurs(60)),
+     ),
+   }),
  )
```

`read` now supports adoption. Previously `findManagedCertificate` filtered out every cert that didn't carry our alchemy ownership tags, so when state was lost the engine could never see a foreign cert and `adopt(true)` had no effect. The new helper returns the first cert matching domain + SAN regardless of tags and reports whether it's owned; `read` returns `Unowned(attrs)` for foreign matches so the engine gates on `--adopt`:

```diff
- const certificate = output?.certificateArn
-   ? yield* describeCertificate(output.certificateArn)
-   : yield* findManagedCertificate(id, olds!);
- return certificate ? toAttrs(olds!, certificate, tags) : undefined;
+ // Slow path: search by domain + SAN. If the cert lacks our ownership
+ // tags, return Unowned(attrs) so the engine gates on --adopt.
+ const found = yield* findCertificateByDomain(id, olds);
+ if (!found) return undefined;
+ const attrs = toAttrs(olds, found.detail, found.tags);
+ return found.owned ? attrs : Unowned(attrs);
```

`reconcile` now also accepts any matching cert (owned or foreign) — by the time `reconcile` runs, the engine has already passed the adopt gate, so reconciling is safe and the existing tag-sync block reasserts ownership.

### New lifecycle tests

Each test runs `destroy -> deploy -> ... -> destroy` on a `ScratchStack` and asserts at every step:

- create and delete a DNS-validated certificate
- reconcile resets tags mutated out-of-band via raw ACM SDK (and re-asserts internal `alchemy:fqn` / `alchemy:stage` tags)
- transparency-logging drift behavior (replace-only via diff; redeploy keeps the same ARN)
- changing `domainName` triggers replace; old cert is deleted
- destroying an already-deleted certificate is a no-op
- `adopt(true)` re-tags a foreign certificate (asserts internal alchemy tags appear after takeover)
- DNS-validated certificate enters `PENDING_VALIDATION` with populated `DomainValidationOptions` (uses `*.example.com` domains; never validates, just asserts shape)

`test.resources.ts` picks up the same `output?.stableId` fix from prior hardening PRs so this branch type-checks in isolation.

### Distilled patch

ACM error category fixes at https://github.com/alchemy-run/distilled/pull/252 — `ThrottlingException` is currently tagged `BadRequestError` (because AWS surfaces it as HTTP 400) so the AWS retry layer doesn't back off on it. `LimitExceededException` is uncategorized; semantically it's a quota error. `RequestInProgressException` becomes `ConflictError + RetryableError`. `ResourceInUseException` (semantically `DependencyViolationError`) and `ResourceNotFoundException` (`NotFoundError`) are not patched — those category tags are gated on alchemy-run/distilled#246. The reconciler-level `Effect.catchTag("ResourceInUseException", …)` retry handles the in-use case directly without depending on category-based behavior.

### DIFF
--- a/packages/alchemy/src/AWS/ACM/Certificate.ts
+++ b/packages/alchemy/src/AWS/ACM/Certificate.ts
@@ -1,8 +1,10 @@
 import { Region as AwsRegion } from "@distilled.cloud/aws/Region";
 import * as acm from "@distilled.cloud/aws/acm";
 import * as route53 from "@distilled.cloud/aws/route-53";
+import * as Data from "effect/Data";
 import * as Effect from "effect/Effect";
 import * as Schedule from "effect/Schedule";
+import { Unowned } from "../../AdoptPolicy.ts";
 import { deepEqual, isResolved } from "../../Diff.ts";
 import * as Provider from "../../Provider.ts";
 import { Resource } from "../../Resource.ts";
@@ -13,6 +15,15 @@ import {
   diffTags,
   hasAlchemyTags,
 } from "../../Tags.ts";
+
+class ValidationRecordsPending extends Data.TaggedError(
+  "ValidationRecordsPending",
+) {}
+class CertificatePendingValidation extends Data.TaggedError(
+  "CertificatePendingValidation",
+) {}
+class Route53ChangePending extends Data.TaggedError("Route53ChangePending") {}
+class CertificateInUse extends Data.TaggedError("CertificateInUse") {}
 
 export interface CertificateProps {
   /**
@@ -156,7 +167,12 @@ export const CertificateProvider = () =>
         );
       });
 
-      const findManagedCertificate = Effect.fn(function* (
+      // Find any certificate matching the desired domain + SAN combo.
+      // Returns the first match plus a flag indicating whether it carries
+      // our alchemy ownership tags. Callers branch on `owned` to decide
+      // between silent adoption (owned), `Unowned(attrs)` gating, or a
+      // greenfield request.
+      const findCertificateByDomain = Effect.fn(function* (
         id: string,
         props: CertificateProps,
       ) {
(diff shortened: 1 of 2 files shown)


sam-goodwin:

PR #213, opened 2026-05-05
fix(aws/ecs): harden Service reconcile + add lifecycle convergence tests

sam-goodwin:
Stacked on top of [#179](https://github.com/alchemy-run/alchemy-effect/pull/179). Hardens the ECS Service resource as part of the per-resource hardening sweep.

### Reconciler changes

```diff
- const service = described?.services?.[0];
+ const service = described?.services?.find(
+   (s) =>
+     s.serviceName === serviceName &&
+     s.status !== "INACTIVE" &&
+     s.status !== "DRAINING",
+ );
```

ECS keeps INACTIVE services in `describeServices` for ~1h after delete, and `DRAINING` services are mid-replace. `read` was returning these tombstones as if alive, so out-of-band-deleted services and in-flight replacements never converged on redeploy.

```diff
- return { ... };
+ const observedTags = tagsFromService(service);
+ const attrs = { ... tags: observedTags };
+ return (yield* hasAlchemyTags(id, observedTags))
+   ? attrs
+   : Unowned(attrs);
```

`read` had no adoption gate. The engine would silently take over any service whose name matched in the cluster. Wrap with `Unowned` based on observed alchemy tags so foreign services require `--adopt` / `adopt(true)`.

```diff
+ const observedTags = tagsFromService(observed);
+ const { removed, upsert } = diffTags(observedTags, desiredTags);
+ if (serviceArn && upsert.length > 0) {
+   yield* ecs.tagResource({ resourceArn: serviceArn, tags: ... });
+ }
+ if (serviceArn && removed.length > 0) {
+   yield* ecs.untagResource({ resourceArn: serviceArn, tagKeys: removed });
+ }
```

The update path never reconciled tags, so adoption (where observed tags don't match what we last persisted) and out-of-band tag drift were both invisible. Diff against observed cloud tags and apply the delta via `tagResource` / `untagResource`.

```diff
- yield* ecs.updateService({ ... });
+ yield* retryOnUpdateInProgress(ecs.updateService({ ... }));
```

`UpdateInProgressException` is transient — ECS surfaces it when a prior service mutation hasn't propagated. Bounded retry (60s) on `updateService` (both reconcile and the drain-to-zero in delete) and `deleteService`.

```diff
  if (output.targetGroupArn) {
-   yield* elbv2
-     .deleteTargetGroup({ TargetGroupArn: output.targetGroupArn })
-     .pipe(Effect.catch(() => Effect.void));
+   yield* elbv2.deleteTargetGroup({
+     TargetGroupArn: output.targetGroupArn,
+   });
  }
```

The blanket catch on `deleteTargetGroup` would swallow `ResourceInUseException` — ELBv2 raises that only if the listener is still attached, which would be a real ordering bug since the listener is removed first.

### New lifecycle tests

Each test runs `destroy → deploy → ... → destroy` on a `ScratchStack` and asserts convergence at every step. Tests register an inline task definition, deploy with `desiredCount: 0` (no actual containers run), and clean up the task def on completion.

- reconcile resets `desiredCount` mutated out-of-band via the raw ECS SDK
- rolling task-definition update converges to the new revision
- changing `serviceName` triggers replace; old service is deleted
- destroying an already-deleted service is a no-op
- `adopt(true)` re-tags a foreign service and preserves user tags
- foreign-tagged service requires `adopt(true)` to take over

### Distilled patch

ECS errors are already covered by [alchemy-run/distilled#253](https://github.com/alchemy-run/distilled/pull/253) (sibling cluster PR). `ServiceNotFoundException`, `ServiceNotActiveException`, `UpdateInProgressException`, `PlatformUnknownException`, `PlatformTaskDefinitionIncompatibilityException`, `UnsupportedFeatureException`, etc. all carry the right categories — no additional patch needed.

The `test.resources.ts` fallback (`output?.stableId ?? `stable-${id}``) is included since the `StaticStablesResource` update path otherwise refuses to type-check on adoption (where `output` may be undefined at the type level).

### DIFF
--- a/packages/alchemy/src/AWS/ECS/Service.ts
+++ b/packages/alchemy/src/AWS/ECS/Service.ts
@@ -1,13 +1,15 @@
 import * as ecs from "@distilled.cloud/aws/ecs";
 import * as elbv2 from "@distilled.cloud/aws/elastic-load-balancing-v2";
 import * as Effect from "effect/Effect";
+import * as Schedule from "effect/Schedule";
+import { Unowned } from "../../AdoptPolicy.ts";
 import { deepEqual, isResolved } from "../../Diff.ts";
 import type { Input } from "../../Input.ts";
 import { createPhysicalName } from "../../PhysicalName.ts";
 import * as Provider from "../../Provider.ts";
 import { Resource } from "../../Resource.ts";
 import type { Providers } from "../Providers.ts";
-import { createInternalTags } from "../../Tags.ts";
+import { createInternalTags, diffTags, hasAlchemyTags } from "../../Tags.ts";
 import type { AccountID } from "../Environment.ts";
 import type { RegionID } from "../Region.ts";
 import type { ClusterArn } from "./Cluster.ts";
@@ -173,6 +175,13 @@ export interface Service extends Resource<
      * ARN of the generated listener, when `public: true`.
      */
     listenerArn?: string;
+
+    /**
+     * Resolved tag map currently applied to the ECS service. Combines
+     * internal alchemy tags with user-supplied tags. Read directly from
+     * the cloud during reconcile so adoption converges correctly.
+     */
+    tags: Record<string, string>;
   },
   never,
   Providers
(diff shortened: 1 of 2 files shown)


sam-goodwin:

PR #214, opened 2026-05-05
fix(aws/secretsmanager): harden Secret reconcile + add lifecycle convergence tests

sam-goodwin:
Harden the SecretsManager `Secret` reconciler so out-of-band drift, soft-delete races, and idempotent redeploys converge cleanly. Adds lifecycle convergence tests mirroring SQS/PR #184.

### Reconciler changes

- Restore a scheduled-for-deletion secret before mutating it. `CreateSecret`/`UpdateSecret` both reject `DeletedDate != null` with `InvalidRequestException`, so observe and `RestoreSecret` first.
- Skip the unconditional `updateSecret` when nothing has drifted, so idempotent redeploys don't bump `versionId`.
- Diff `Description` / `KmsKeyId` against observed cloud state and only send what's drifted — converges out-of-band mutations without round-tripping `undefined` -> `""` for callers who never set a description.
- Scope `readSecret`'s swallowing to `ResourceNotFoundException`. Auth, throttling, and server errors now surface so the engine retries or fails loudly.
- Tolerate `InvalidRequestException` on delete (already-scheduled) so engine-level double-destroy is safe.
- Add `forceDelete` (default `true`) + `recoveryWindowInDays` props so production users can opt into the soft-delete window while tests/dev keep clean teardown.

```ts
export interface SecretProps {
  // ...
  /** @default true */
  forceDelete?: boolean;
  /** @default 30 */
  recoveryWindowInDays?: number;
}
```

### New lifecycle tests

`packages/alchemy/test/AWS/SecretsManager/Secret.test.ts`:

- idempotent redeploy is a no-op (no `versionId` bump)
- drift on `Description` reconciles back
- new `secretString` advances `versionId`
- recreate after raw-SDK `ForceDeleteWithoutRecovery` out-of-band
- restore after raw-SDK schedule-delete out-of-band
- replace on `name` change
- double-destroy idempotency
- silent adoption when tags match
- `adopt: true` re-tags a foreign secret

### Distilled patch

[`alchemy-run/distilled#254`](https://github.com/alchemy-run/distilled/pull/254) — categorize `InternalServiceError` as `ServerError` (so it joins the AWS retry layer) and `ResourceExistsException` as `ConflictError` (proper race semantics).

### DIFF
--- a/packages/alchemy/src/AWS/SecretsManager/Secret.ts
+++ b/packages/alchemy/src/AWS/SecretsManager/Secret.ts
@@ -51,6 +51,23 @@ export interface SecretProps {
    * User-defined tags for the secret.
    */
   tags?: Record<string, string>;
+  /**
+   * When the resource is destroyed, immediately delete the secret without
+   * the soft-delete recovery window. Defaults to `true` so engine-driven
+   * `destroy` calls leave nothing behind that would block re-creating a
+   * secret with the same name.
+   *
+   * Set to `false` if you want AWS to schedule deletion with the
+   * `recoveryWindowInDays` window so the secret can be restored.
+   * @default true
+   */
+  forceDelete?: boolean;
+  /**
+   * Recovery window (in days) used when `forceDelete` is `false`. AWS
+   * accepts 7–30. Ignored when `forceDelete` is `true`.
+   * @default 30
+   */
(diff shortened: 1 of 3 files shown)


sam-goodwin:

PR #215, opened 2026-05-05
fix(aws/ec2): harden Subnet reconcile + add lifecycle convergence tests

sam-goodwin:
Harden the EC2 Subnet reconciler so adoption, drift, and out-of-band deletes converge cleanly, and add lifecycle convergence tests covering each path. Mirrors the VPC #189/#199 and SQS #184 hardening pattern.

### Reconciler changes

- Add `read` lifecycle. Subnet has no physical name we can derive — `read` looks up by cached `output.subnetId` first, falling back to a tag-filter search via `createAlchemyTagFilters(id)` when the id is missing or stale. Foreign-tagged subnets surface as `Unowned(attrs)`; alchemy-owned ones return raw attrs.
- Reconcile observes the same way: cached id → tag filter → `createSubnet`. An interrupted prior run that lost state after `createSubnet` recovers without leaking a duplicate.
- Wrap `createSubnet` in a bounded retry on `InvalidVpcID.NotFound` (≤15s) so a fresh-VPC-in-same-plan race rides out propagation.
- Wrap every post-create `modifySubnetAttribute` / `createTags` / `deleteTags` / final `describeSubnets` call in a bounded `InvalidSubnetID.NotFound` retry (≤15s). Distilled doesn't tag this as retryable, but it's transient when we just created the subnet ourselves.
- Each attribute (`MapPublicIpOnLaunch`, `AssignIpv6AddressOnCreation`, `EnableDns64`, private DNS hostname options) and tags now diff against **observed cloud state**, not `olds`/`output.tags`, so adoption and out-of-band drift converge.
- `delete` retries `DependencyViolation` on a 5s/60-attempt schedule (~5min) and treats `InvalidSubnetID.NotFound` as success — already-deleted is idempotent.
- Add `tags` to the Subnet attributes shape so observed cloud tags round-trip through state.

### New lifecycle tests

```
- reconcile resets MapPublicIpOnLaunch + tags mutated out-of-band via raw EC2 SDK
- reconcile re-creates a subnet deleted out-of-band (state still references old subnetId)
- changing cidrBlock triggers replace; old subnet is torn down
- changing availabilityZone triggers replace
- destroying an already-deleted subnet is a no-op
- owned subnet (matching alchemy tags) is silently adopted after state wipe
```

The foreign-tagged-takeover test is skipped because Subnet has no physical name to look up by — pre-creating a subnet branded with the new resource's `alchemy::id` would short-circuit the takeover semantics. The owned-adoption case exercises the read + tag-sync path end-to-end. Same gap acknowledged in VPC adoption (#199).

### Distilled patch

No patch needed. EC2's `IncorrectState` is correctly classified as `ConflictError` (PR #242) and `DependencyViolation` is its own retryable category. `InvalidSubnetID.NotFound` / `InvalidVpcID.NotFound` are intentionally untagged and handled locally where they're known to be transient.

### DIFF
--- a/packages/alchemy/src/AWS/EC2/Subnet.ts
+++ b/packages/alchemy/src/AWS/EC2/Subnet.ts
@@ -3,12 +3,19 @@ import * as Data from "effect/Data";
 import * as Effect from "effect/Effect";
 import * as Schedule from "effect/Schedule";
 
+import { Unowned } from "../../AdoptPolicy.ts";
 import type { ScopedPlanStatusSession } from "../../Cli/Cli.ts";
 import { isResolved, somePropsAreDifferent } from "../../Diff.ts";
 import * as Provider from "../../Provider.ts";
 import { Resource } from "../../Resource.ts";
 import type { Providers } from "../Providers.ts";
-import { createInternalTags, createTagsList, diffTags } from "../../Tags.ts";
+import {
+  createAlchemyTagFilters,
+  createInternalTags,
+  createTagsList,
+  diffTags,
+  hasAlchemyTags,
+} from "../../Tags.ts";
 import type { AccountID } from "../Environment.ts";
 import type { RegionID } from "../Region.ts";
 import type { VpcId } from "./Vpc.ts";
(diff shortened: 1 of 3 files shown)


sam-goodwin:

PR #216, opened 2026-05-05
fix(cloudflare/tunnel): harden Tunnel reconcile + add lifecycle convergence tests

sam-goodwin:
Stacked on top of [#179](https://github.com/alchemy-run/alchemy-effect/pull/179). Hardens the Cloudflare Tunnel resource as part of the per-resource hardening sweep (mirrors [#184](https://github.com/alchemy-run/alchemy-effect/pull/184) and [#190](https://github.com/alchemy-run/alchemy-effect/pull/190)).

### Reconciler changes

```diff
- ).pipe(Effect.catch(() => Effect.succeed(undefined)))
+ ).pipe(
+   Effect.catchTag("TunnelNotFound", () => Effect.succeed(undefined)),
+ )
```

Every catch in `reconcile`/`read`/`delete` was scoped to the specific error tag. A blanket `Effect.catch` on `getTunnel`/`getToken` was silently treating auth/throttling/5xx as "tunnel missing" and dropping reconcile straight into a re-create — masking real failures and risking duplicate tunnels.

```diff
+ if (observed?.deletedAt) {
+   observed = undefined;
+ }
```

Cloudflare retains a soft-deleted record briefly after `deleteTunnelCloudflared`. `getTunnel` returns it with `deletedAt` set; both `reconcile` and `read` now treat that as "missing" so the engine re-creates instead of resurrecting a zombie.

```diff
- Effect.catchTag("DuplicateTunnelName", () => findTunnelByName(name)),
+ Effect.catchTag("DuplicateTunnelName", (err) =>
+   Effect.gen(function* () {
+     if (!news.adopt) return yield* Effect.fail(err);
+     const existing = yield* findTunnelByName(name);
+     if (!existing || !existing.id) return yield* Effect.fail(err);
+     return existing;
+   }),
+ ),
```

`DuplicateTunnelName` only triggers takeover when `adopt: true`. Without `adopt`, the error surfaces — refusing accidental takeover of a foreign tunnel.

```diff
- // skip PUT if ingress unchanged vs olds
+ // always PUT when configSrc === "cloudflare"
```

Configuration PUT is unconditional (when `configSrc` is `"cloudflare"`) so out-of-band ingress drift in the Cloudflare dashboard is overwritten by the next reconcile.

`delete` only swallows `UnknownCloudflareError` with `code === 1002` (the not-found race). Auth/throttling/5xx and the "tunnel has active connections" failure still bubble.

### New lifecycle tests

Each runs `destroy → deploy → … → destroy` on a `ScratchStack`.

- reconcile resets ingress mutated out-of-band via the raw Cloudflare API
- reconcile re-creates a tunnel deleted out-of-band
- changing `name` triggers replace; old tunnel is deleted
- destroying an already-deleted tunnel is a no-op
- `adopt(true)` takes over a foreign tunnel matching by name; without `adopt` the duplicate-name error surfaces

### Distilled patch

No patch needed — `UnknownCloudflareError` only appears in `delete` (the documented `code: 1002` not-found race), already handled by the resource. All other reconcile paths use typed errors (`TunnelNotFound`, `TunnelTokenNotFound`, `DuplicateTunnelName`) from the existing distilled spec.

### DIFF
--- a/packages/alchemy/src/Cloudflare/Tunnel/Tunnel.ts
+++ b/packages/alchemy/src/Cloudflare/Tunnel/Tunnel.ts
@@ -148,11 +148,19 @@ export const TunnelProvider = () =>
         originRequest: Tunnel.OriginRequestConfig | undefined,
       ) =>
         Effect.gen(function* () {
-          if (!ingress && !originRequest) return;
+          // Always PUT for `cloudflare`-managed tunnels: the API treats
+          // the configuration as the canonical state, so an unconditional
+          // PUT lets reconcile clear out-of-band drift even when the user
+          // removed every ingress rule. An empty `ingress`/`originRequest`
+          // gets normalised to `undefined` so the request body matches
+          // what `getTunnelCloudflaredConfiguration` will return.
           yield* putConfiguration({
             accountId,
             tunnelId,
-            config: { ingress, originRequest },
+            config: {
+              ingress: ingress && ingress.length > 0 ? ingress : undefined,
+              originRequest,
+            },
           });
         });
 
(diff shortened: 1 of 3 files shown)


sam-goodwin:

PR #217, opened 2026-05-05
fix(aws/eks): harden Cluster reconcile + add lifecycle convergence tests

sam-goodwin:
Stacked on top of [#179](https://github.com/alchemy-run/alchemy-effect/pull/179). Hardens the AWS EKS Cluster (and Addon) reconciler as part of the per-resource hardening sweep on top of the reconcile migration.

### Reconciler changes

```diff
- yield* eks.updateClusterConfig({ … });
+ yield* eks.updateClusterConfig({ … }).pipe(
+   Effect.retry({
+     while: (e) => e._tag === "ResourceInUseException",
+     schedule: concurrentUpdateRetrySchedule, // exp 2s, max 60 retries (~5min)
+   }),
+ );
```

`updateClusterConfig`, `updateClusterVersion`, and the disable-deletion-protection update now ride a bounded `ResourceInUseException` retry. EKS surfaces this when the control plane is mid-update from a prior reconcile or a peer; it's genuinely transient. `InvalidRequestException` is intentionally NOT retried — EKS reuses it for both transient and permanent (bad input) cases, and we don't want a typo to retry for 5 minutes.

```diff
  yield* eks.deleteCluster({ name: output.clusterName }).pipe(
    Effect.catchTag("ResourceNotFoundException", () => Effect.void),
+   Effect.catchTag("ResourceInUseException", () => Effect.void),
  );
  yield* waitForClusterDeleted(output.clusterName);
```

`delete` now tolerates `ResourceInUseException` (cluster already in `DELETING` from a prior partial run) and falls through to the wait-for-deleted polling.

```diff
- const disableDeletionProtection = yield* eks.updateClusterConfig({ … });
+ const disableDeletionProtection = yield* eks.updateClusterConfig({ … }).pipe(
+   Effect.retry({ while: (e) => e._tag === "ResourceInUseException", … }),
+   Effect.catchTag("ResourceNotFoundException", () =>
+     Effect.succeed({ update: undefined }),
+   ),
+ );
```

If the cluster vanished out-of-band between describe and disable-deletion-protection, treat it as already-deleted instead of failing the destroy.

`Addon.updateAddon` gets the same bounded `ResourceInUseException` retry.

### New lifecycle tests

Gated behind `RUN_LIVE_EKS_TESTS=true` because EKS clusters take ~15 min to provision. Each test runs `destroy → deploy → … → destroy`.

- reconcile resets `logging` mutated out-of-band via the raw EKS SDK
- reconcile resets `endpointPublicAccess` mutated out-of-band
- changing `clusterName` triggers replace; old cluster is deleted
- `adopt(true)` re-tags a foreign cluster

### Distilled patch

Not needed. `InvalidRequestException` is correctly tagged `BadRequestError` only (no `RetryableError`, per the rule that context-dependent errors must not be globally retryable). `ResourceInUseException` is `ConflictError` only — we handle the bounded retry at the reconciler level so the retry budget is per-call and observable.

### Outstanding

PR is **draft**: all integration tests are skipped by default and have not been run live in this session. Set `RUN_LIVE_EKS_TESTS=true` and budget ~30 min per test before marking ready for review.

### DIFF
--- a/packages/alchemy/src/AWS/EKS/Addon.ts
+++ b/packages/alchemy/src/AWS/EKS/Addon.ts
@@ -247,18 +247,32 @@ export const AddonProvider = () =>
             state.configurationValues !== news.configurationValues ||
             podIdentityChanged
           ) {
-            yield* eks.updateAddon({
-              clusterName,
-              addonName,
-              addonVersion: news.addonVersion,
-              serviceAccountRoleArn: news.serviceAccountRoleArn as
-                | string
-                | undefined,
-              resolveConflicts: news.resolveConflicts,
-              configurationValues: news.configurationValues,
-              podIdentityAssociations: news.podIdentityAssociations,
-              clientRequestToken: yield* toClientRequestToken(id, "update"),
-            });
+            // `ResourceInUseException` here means a prior update on this
+            // addon hasn't finished applying. Retry bounded — the addon
+            // controller usually settles in seconds. We do NOT retry
+            // `InvalidRequestException` because EKS reuses it for both
+            // transient and permanent failures.
+            yield* eks
+              .updateAddon({
+                clusterName,
+                addonName,
+                addonVersion: news.addonVersion,
+                serviceAccountRoleArn: news.serviceAccountRoleArn as
+                  | string
+                  | undefined,
+                resolveConflicts: news.resolveConflicts,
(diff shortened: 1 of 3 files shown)


sam-goodwin:

PR #218, opened 2026-05-05
fix(cloudflare/secretsstore): harden SecretsStore reconcile + add lifecycle convergence tests

sam-goodwin:
Harden the Cloudflare `SecretsStore` and child `Secret` providers and add live convergence tests.

`SecretsStore` is account-singleton infrastructure (`Props = {}`, `delete` is a no-op, `read` always adopts the account's existing store), so the mutable surface lives entirely on `Secret`. The reconciler is the single observe → ensure → sync flow required by the doctrine; this change fixes a few NotFound-shape gaps and removes a stale doc example for a non-existent `name` prop.

### Reconciler changes

```diff
// Secret.ts — observe + delete: tolerate Cloudflare's generic 404 shape
   .pipe(
     Effect.catchTag("SecretNotFound", () => Effect.succeed(undefined)),
     Effect.catchTag("StoreNotFound", () => Effect.succeed(undefined)),
+    // Cloudflare also surfaces a generic 404 as `NotFound` (e.g. when both
+    // store and secret are missing simultaneously) — treat as "observed nothing".
+    Effect.catchTag("NotFound", () => Effect.succeed(undefined)),
   );

// Secret.ts — sync PATCH: tolerate out-of-band delete races
   patchStoreSecret({ ... }).pipe(
     Effect.catchTag("SecretNotFound", () => Effect.succeed(undefined)),
+    Effect.catchTag("StoreNotFound", () => Effect.succeed(undefined)),
   );
```

```diff
// SecretsStore.ts — drop stale `@example Adopt a specific named store`
- * @example Adopt a specific named store
- * ```typescript
- * const store = yield* Cloudflare.SecretsStore("MyStore", { name: "production-secrets" });
- * ```
```

`SecretsStore.Props` is `{}`; the example referenced a prop that doesn't exist. The provider already always-adopts the account's single store via `read`/`reconcile`.

### New lifecycle tests

`packages/alchemy/test/Cloudflare/SecretsStore/SecretsStore.test.ts` — seven `test.provider` cases against the live API:

- create + redeploy with byte-identical props is a no-op (`secretId` stable)
- reconcile resets `comment`/`scopes` mutated out-of-band via raw API
- value rotation triggers replace; new `secretId`, old one gone
- reconcile re-creates a secret deleted out-of-band
- changing `name` triggers replace; old name is deleted
- destroying an already-deleted secret is a no-op
- `adopt(true)` re-claims a foreign secret created out-of-band; same `secretId`, scopes/comment converge

Random-suffix names per test (no `Date.now()`); `assertSecretDeleted` retries on eventual consistency and tolerates `SecretNotFound`/`StoreNotFound`/`NotFound`.

### Distilled patch

No patch needed — the existing distilled `secrets-store` error tags (`SecretNotFound`, `StoreNotFound`, `NotFound`, `SecretNameAlreadyExists`, `MaximumStoresExceeded`) are sufficient.

### DIFF
--- a/packages/alchemy/src/Cloudflare/SecretsStore/Secret.ts
+++ b/packages/alchemy/src/Cloudflare/SecretsStore/Secret.ts
@@ -148,6 +148,10 @@ export const StoreSecretProvider = () =>
                 Effect.succeed(undefined),
               ),
               Effect.catchTag("StoreNotFound", () => Effect.succeed(undefined)),
+              // Cloudflare also surfaces a generic 404 as `NotFound` (e.g.
+              // when both the store and secret are missing simultaneously),
+              // so treat that as "observed nothing" too.
+              Effect.catchTag("NotFound", () => Effect.succeed(undefined)),
             );
           }
           if (!observed) {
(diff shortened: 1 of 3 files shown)


sam-goodwin:

PR #219, opened 2026-05-05
fix(github): harden Secret reconcile + add lifecycle convergence tests

sam-goodwin:
Hardens the GitHub `Secret` and `Variable` reconcilers as the GitHub pass of the per-resource hardening sweep. Octokit errors (which arrive as `unknown`) are now mapped to typed `Data.TaggedError` shapes so individual cases can be caught precisely instead of via blanket `try/catch (error: any)` blocks.

### Reconciler changes

```diff
- } catch (error: any) {
-   if (error.status !== 404) throw error;
- }
+ Effect.tryPromise({
+   try: ...,
+   catch: (e) => mapOctokitError(e, ctx),
+ }).pipe(Effect.catchTag("GitHubSecretNotFound", () => Effect.void))
```

`mapOctokitError` translates Octokit `status` → `GitHubSecretNotFound` (404), `GitHubSecretValidationError` (422), or a typed `GitHubSecretError` fallback so 422s from libsodium/key-id mistakes never get retried as if they were transient, and arbitrary error shapes never escape as `unknown`.

```diff
+ yield* upsertSecret(octokit, news).pipe(
+   Effect.retry({
+     while: (e) => e._tag === "GitHubSecretNotFound",
+     schedule: Schedule.fixed(1000).pipe(
+       Schedule.both(Schedule.recurs(30)),
+     ),
+   }),
+ );
```

Repo creation in the same plan is eventually consistent — a 404 from `getRepoPublicKey`/`createOrUpdate*Secret` right after the repo lands is a race, not a permanent failure. Capped at 30s so a genuinely missing repo still surfaces.

`Variable` gets the same Octokit-error mapping plus a `GitHubVariableAlreadyExists` (422) catch around `createRepoVariable`, so a peer reconciler racing the create falls through to `PATCH` instead of failing the whole stack.

### New lifecycle tests

`packages/alchemy/test/GitHub/Secret.test.ts` runs `destroy → deploy → … → destroy` against a real test repo (gated on `GITHUB_TEST_OWNER` / `GITHUB_TEST_REPO` / `GITHUB_TOKEN`):

- create + delete a repository secret
- redeploy with same value is convergent (re-PUT, no error)
- rotating value updates in place — `created_at` preserved, `updated_at` advances
- reconcile re-creates a secret deleted out-of-band
- changing `name` triggers replace; old secret is deleted
- destroying an already-deleted secret is a no-op
- `adopt(true)` re-claims a foreign secret

### DIFF
--- a/packages/alchemy/src/GitHub/Secret.ts
+++ b/packages/alchemy/src/GitHub/Secret.ts
@@ -1,6 +1,8 @@
 import type { Octokit } from "@octokit/rest";
+import * as Data from "effect/Data";
 import * as Effect from "effect/Effect";
 import * as Redacted from "effect/Redacted";
+import * as Schedule from "effect/Schedule";
 import * as Provider from "../Provider.ts";
 import { Resource } from "../Resource.ts";
 import { GitHubCredentials } from "./Credentials.ts";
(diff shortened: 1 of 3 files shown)


sam-goodwin:

PR #221, opened 2026-05-05
fix(cloudflare/vpcservice): harden VpcService reconcile + add lifecycle convergence tests

sam-goodwin:
Hardens the Cloudflare VpcService resource as part of the per-resource hardening sweep (mirrors [#184](https://github.com/alchemy-run/alchemy-effect/pull/184), [#190](https://github.com/alchemy-run/alchemy-effect/pull/190), and [#216](https://github.com/alchemy-run/alchemy-effect/pull/216)).

### Reconciler changes

```diff
- }).pipe(Effect.catch(() => Effect.succeed(undefined)));
+ }).pipe(
+   Effect.catchTag("VpcServiceNotFound", () => Effect.succeed(undefined)),
+ );
```

Every blanket `Effect.catch` in `reconcile`/`read`/`delete` was scoped to its specific typed error. A blanket catch on `getService` was silently treating auth/throttling/5xx as "service missing" and dropping reconcile straight into a re-create — masking real failures and risking duplicate services.

```diff
- Effect.catch((err: unknown) =>
+ Effect.catchTag("VpcServiceNameAlreadyExists", (err) =>
    Effect.gen(function* () {
      if (!news.adopt) return yield* Effect.fail(err);
      const existing = yield* findServiceByName(name);
      ...
    }),
  ),
```

`VpcServiceNameAlreadyExists` only triggers takeover when `adopt: true`. Without `adopt`, the duplicate-name error surfaces — refusing accidental takeover of a foreign service. Other create errors (including `VpcTunnelNotFound`) now bubble instead of being miscategorised as a name conflict.

```diff
- }).pipe(Effect.catch(() => Effect.void));
+ }).pipe(Effect.catchTag("VpcServiceNotFound", () => Effect.void));
```

`delete` only swallows `VpcServiceNotFound`. Auth/throttling/5xx still bubble.

### New lifecycle tests

Each runs `destroy → deploy → … → destroy` on a `ScratchStack`.

- reconcile resets settings mutated out-of-band via the raw Cloudflare API
- reconcile re-creates a service deleted out-of-band
- changing `name` updates in place (same `serviceId`)
- destroying an already-deleted service is a no-op
- `adopt(true)` re-claims a foreign service matching by name; without `adopt` the duplicate-name error surfaces

### Distilled patch

No patch needed — `VpcServiceNotFound`, `VpcServiceNameAlreadyExists`, and `VpcTunnelNotFound` are already typed errors in `@distilled.cloud/cloudflare/connectivity`. `UnknownCloudflareError` does not appear on any of the connectivity ops the reconciler uses.

### DIFF
--- a/packages/alchemy/src/Cloudflare/VpcService/VpcService.ts
+++ b/packages/alchemy/src/Cloudflare/VpcService/VpcService.ts
@@ -170,12 +170,21 @@ export const VpcServiceProvider = () =>
           // Observe — re-fetch the cached service; fall back to a name
           // scan so we recover from out-of-band deletes or partial state
           // persistence failures.
+          //
+          // Scope the catch to `VpcServiceNotFound` only. A blanket
+          // catch would silently treat auth/throttling/5xx failures as
+          // "service missing" and drop us straight into a re-create —
+          // masking real failures and risking duplicate services.
           let observed: connectivity.GetDirectoryServiceResponse | undefined;
           if (output?.serviceId) {
             observed = yield* getService({
               accountId: acct,
               serviceId: output.serviceId,
-            }).pipe(Effect.catch(() => Effect.succeed(undefined)));
+            }).pipe(
+              Effect.catchTag("VpcServiceNotFound", () =>
+                Effect.succeed(undefined),
+              ),
+            );
           }
           if (!observed) {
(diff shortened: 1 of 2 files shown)


sam-goodwin:

PR #222, opened 2026-05-05
fix(aws/ec2): harden SecurityGroup reconcile + add lifecycle convergence tests

sam-goodwin:
Harden the EC2 `SecurityGroup` reconciler against eventual-consistency races and rule drift, then add lifecycle convergence tests on top of `ScratchStack`.

### Reconciler changes

```diff
-      const describeSecurityGroup = (groupId: string) =>
-        ec2.describeSecurityGroups({ GroupIds: [groupId] }).pipe(
-          Effect.map((r) => r.SecurityGroups?.[0]),
-          Effect.flatMap((sg) =>
-            sg
-              ? Effect.succeed(sg)
-              : Effect.fail(new Error(`Security Group ${groupId} not found`)),
-          ),
-        );
+      const retryEventuallyConsistent = <A, E, R>(
+        eff: Effect.Effect<A, E, R>,
+      ) =>
+        eff.pipe(
+          Effect.retry({
+            while: (e) =>
+              e?._tag === "InvalidGroup.NotFound" ||
+              e?._tag === "SecurityGroupNotVisible",
+            schedule: Schedule.exponential(100).pipe(
+              Schedule.both(Schedule.recurs(10)),
+            ),
+          }),
+        );
+      const describeSecurityGroup = (groupId: string) =>
+        retryEventuallyConsistent(
+          ec2.describeSecurityGroups({ GroupIds: [groupId] }).pipe(
+            Effect.flatMap((r) =>
+              r.SecurityGroups?.[0]
+                ? Effect.succeed(r.SecurityGroups[0])
+                : Effect.fail({
+                    _tag: "SecurityGroupNotVisible" as const,
+                    groupId,
+                  }),
+            ),
+          ),
+        );
```

Rule sync used to revoke every observed rule and re-authorize the desired set every reconcile. Replaced with a canonical-key diff (protocol + ports + source + description) so we only apply the delta:

```diff
-          if (currentIngress.length > 0) {
-            yield* ec2.revokeSecurityGroupIngress({ ... all observed ... });
-          }
-          if (news.ingress) {
-            yield* ec2.authorizeSecurityGroupIngress({ ... all desired ... });
-          }
+          const observedKeys = new Map(observed.map((r) => [canonicalRuleKey(...), r]));
+          const desiredKeys = new Map(desired.map((r) => [canonicalRuleKey(...), r]));
+          const toRevoke = [...observedKeys].filter(([k]) => !desiredKeys.has(k));
+          const toAuthorize = [...desiredKeys].filter(([k]) => !observedKeys.has(k));
```

`read` now searches by alchemy tag filters when state is missing, returns `Unowned(attrs)` for foreign-tagged SGs (gating adoption behind `adopt(true)`), and treats a vanished SG as `undefined` so the engine recreates it.

```diff
-        read: Effect.fn(function* ({ output }) {
-          if (!output) return undefined;
-          const sg = yield* describeSecurityGroup(output.groupId);
-          ...
+        read: Effect.fn(function* ({ id, output }) {
+          const sg = output?.groupId
+            ? yield* findSecurityGroup(output.groupId)
+            : (yield* ec2.describeSecurityGroups({
+                Filters: yield* createAlchemyTagFilters(id),
+              })).SecurityGroups?.[0];
+          if (!sg) return undefined;
+          ...
+          return (yield* hasAlchemyTags(id, tagRecord)) ? attrs : Unowned(attrs);
```

`delete` previously also matched `e._tag === "ValidationError" && e.message?.includes("DependencyViolation")`. That branch was dead — distilled tags `DependencyViolation` directly via `withDependencyViolationError`, so the substring match never fired and any real `ValidationError` was being swallowed silently. Drop it.

### New lifecycle tests

- reconcile resets ingress + egress rules mutated out-of-band
- reconcile re-creates an SG deleted out-of-band
- changing `groupName` triggers replace; old SG is deleted
- adding/removing/changing rules diffs against cloud state
- destroying an already-deleted SG is a no-op
- foreign-tagged SG requires `adopt(true)` to take over

### Distilled patch

No patch needed — error categorization for SG flows (`InvalidGroup.NotFound`, `InvalidPermission.NotFound`, `DependencyViolation`) is already correct in distilled.

`.worktrees/` joins `.gitignore` (matching the VPC hardening PR) so per-resource hardening worktrees stay untracked.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,6 +8,7 @@ dist/
 .env
 *.tsbuildinfo
 .claude/
+.worktrees/
 .alchemy/
 # !.alchemy/github:alchemy/
 # examples/*/.alchemy
--- a/packages/alchemy/src/AWS/EC2/SecurityGroup.ts
+++ b/packages/alchemy/src/AWS/EC2/SecurityGroup.ts
@@ -2,12 +2,19 @@ import * as ec2 from "@distilled.cloud/aws/ec2";
 import { Region } from "@distilled.cloud/aws/Region";
 import * as Effect from "effect/Effect";
 import * as Schedule from "effect/Schedule";
+import { Unowned } from "../../AdoptPolicy.ts";
 import { isResolved } from "../../Diff.ts";
 import { createPhysicalName } from "../../PhysicalName.ts";
 import * as Provider from "../../Provider.ts";
 import { Resource } from "../../Resource.ts";
 import type { Providers } from "../Providers.ts";
-import { createInternalTags, createTagsList, diffTags } from "../../Tags.ts";
+import {
+  createAlchemyTagFilters,
+  createInternalTags,
+  createTagsList,
+  diffTags,
+  hasAlchemyTags,
+} from "../../Tags.ts";
 import type { AccountID } from "../Environment.ts";
 import { AWSEnvironment } from "../Environment.ts";
 import type { RegionID } from "../Region.ts";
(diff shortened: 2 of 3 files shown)


sam-goodwin:

PR #224, opened 2026-05-05
fix(aws/rds): harden DBCluster reconcile + add lifecycle convergence tests

sam-goodwin:
Hardens the AWS RDS `DBCluster` reconciler in line with the per-resource hardening sweep started by #184 and tracked by sibling PR #201 for `DBInstance`. Cluster control-plane calls fail with `InvalidDBClusterStateFault` mid-transition; the previous reconciler issued `modifyDBCluster` blindly, didn't wait for stable status, and surfaced transient state-machine races as terminal errors.

### Reconciler changes

```diff
- // unconditional modify, no observed-vs-desired diff
- yield* rds.modifyDBCluster({ DBClusterIdentifier: identifier, EngineVersion: news.engineVersion, ... });
- observed = yield* waitForCluster(identifier);
+ // wait for stable status before mutating
+ if (!isStableClusterStatus(observed.Status)) {
+   observed = yield* waitForStableCluster(identifier, session);
+ }
+ // diff observed → desired, skip the API on no-op
+ const modifyPayload = computeModifyPayload(observed, news, credentials.MasterUserPassword);
+ if (modifyPayload) {
+   yield* retryControlPlane(rds.modifyDBCluster(modifyPayload));
+   observed = yield* waitForStableCluster(identifier, session);
+ }
```

```diff
  delete: Effect.fn(function* ({ output }) {
    yield* rds.deleteDBCluster({ ... }).pipe(
+     Effect.retry({
+       while: (e) => e._tag === "InvalidDBClusterStateFault",
+       schedule: controlPlaneRetryPolicy,
+     }),
      Effect.catchTag("DBClusterNotFoundFault", () => Effect.void),
    );
+   yield* waitForClusterDeleted(output.dbClusterIdentifier);
  }),
```

- `InvalidDBClusterStateFault` is treated as retryable **only** in scoped contexts where we know we're polling a transitioning resource — never globally tagged retryable, since it's context-dependent (writer being modified vs. genuine `ConflictError`).
- New props/attrs: `backupRetentionPeriod`, `preferredBackupWindow`, `preferredMaintenanceWindow`. Existing `deletionProtection` and `copyTagsToSnapshot` are now reflected in attrs.
- `computeModifyPayload` diffs each mutable field (engine version, parameter group, security groups, port, IAM/HTTP endpoint, serverless v2 scaling, backup window/retention, deletion protection, copy-tags-to-snapshot) and returns `undefined` on a clean no-op so the modify call is skipped entirely.
- Tags are diffed against observed cloud tags (not `output.tags`) so adoption converges.
- `waitForClusterDeleted` polls until RDS drops the cluster, so a subsequent reconcile or replace doesn't race against `deleting`.

### New lifecycle tests

`packages/alchemy/test/AWS/RDS/DBCluster.test.ts` runs `destroy → deploy → ... → destroy` on `ScratchStack` and asserts convergence at every step. All tests are `test.provider.skip` because Aurora cluster create is 5–15 minutes per test; they're intended to be unskipped by hand against an isolated test account.

- reconcile resets `backupRetentionPeriod` / `preferredBackupWindow` / `copyTagsToSnapshot` / internal alchemy tags mutated out-of-band
- changing `dbClusterIdentifier` triggers replace; old cluster is deleted
- in-place modification of `engineVersion` (minor)
- `adopt(true)` re-tags a foreign cluster

### Distilled patch

No patch needed — the existing distilled error tagging is sufficient; `InvalidDBClusterStateFault` is intentionally **not** marked globally retryable since it's context-dependent (only the reconciler knows whether the transition is one it can ride out), and `InsufficientDBClusterCapacityFault` recovery takes minutes-to-hours which exceeds the default retry budget.

### DIFF
--- a/packages/alchemy/src/AWS/RDS/DBCluster.ts
+++ b/packages/alchemy/src/AWS/RDS/DBCluster.ts
@@ -1,5 +1,6 @@
 import * as rds from "@distilled.cloud/aws/rds";
 import * as secretsmanager from "@distilled.cloud/aws/secrets-manager";
+import * as Data from "effect/Data";
 import * as Effect from "effect/Effect";
 import * as Redacted from "effect/Redacted";
 import * as Schedule from "effect/Schedule";
@@ -76,6 +77,19 @@ export interface DBClusterProps {
    */
   kmsKeyId?: string;
   /**
+   * Daily backup retention window in days.
+   */
+  backupRetentionPeriod?: number;
+  /**
+   * Preferred backup window in `hh24:mi-hh24:mi` UTC format.
+   */
+  preferredBackupWindow?: string;
+  /**
+   * Preferred weekly maintenance window in
+   * `ddd:hh24:mi-ddd:hh24:mi` UTC format.
+   */
+  preferredMaintenanceWindow?: string;
+  /**
    * Let RDS manage the master user password in Secrets Manager.
    */
   manageMasterUserPassword?: boolean;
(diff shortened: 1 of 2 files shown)


sam-goodwin:

PR #225, opened 2026-05-05
fix(axiom): harden Monitor + Notifier reconcile + add lifecycle convergence tests

sam-goodwin:
Hardens the Axiom Monitor and Notifier reconcile paths the same way [#184](https://github.com/alchemy-run/alchemy-effect/pull/184) (SQS) and [#197](https://github.com/alchemy-run/alchemy-effect/pull/197) (Axiom Dataset) did for their resources.

### Reconciler changes

`updateMonitor` and `updateNotifier` both declare `NotFound` as an error — if a peer reconciler / human deletes the resource between our `get` and our `update`, the engine would surface a hard failure instead of converging. Catch `NotFound` and re-create.

```diff
- return yield* update({ ...news, id: observed.id });
+ return yield* update({ ...news, id: observed.id }).pipe(
+   Effect.catchTag("NotFound", () => create(news)),
+ );
```

```diff
- const result = yield* update({ ...news, id: observed.id! });
+ const result = yield* update({ ...news, id: observed.id! }).pipe(
+   Effect.catchTag("NotFound", () =>
+     create(news).pipe(Effect.map((r) => ({ ...r, id: r.id ?? "" }))),
+   ),
+ );
```

### New lifecycle tests

For Monitor:
- redeploy with same props is a no-op (id stable)
- reconcile resets `threshold` / `description` mutated out-of-band via `AxiomSdk.updateMonitor`
- reconcile re-creates a monitor deleted out-of-band via `AxiomSdk.deleteMonitor`
- changing `name` updates in place (id stays stable)
- changing `type` triggers replace; old monitor id is gone
- destroying an already-deleted monitor is a no-op (delete catches `NotFound`)

For Notifier:
- redeploy with same props is a no-op (id stable)
- reconcile resets `properties.email.emails` mutated out-of-band via `AxiomSdk.updateNotifier`
- reconcile re-creates a notifier deleted out-of-band via `AxiomSdk.deleteNotifier`
- changing `name` updates in place
- destroying an already-deleted notifier is a no-op

`adopt(true)` re-claim is intentionally omitted: Monitor / Notifier identifiers are server-assigned with no name uniqueness or ownership marker, so after a state wipe the engine has no way to discover the prior cloud resource by logical id — adoption is N/A for these resources today.

### Distilled patch

No patch needed. The Axiom client already routes HTTP status to typed errors via `HTTP_STATUS_MAP` and `updateMonitor` / `updateNotifier` already declare `NotFound` in their error channel; the reconciler change is purely about catching the right shape alchemy-side.

### DIFF
--- a/packages/alchemy/src/Axiom/Monitor.ts
+++ b/packages/alchemy/src/Axiom/Monitor.ts
@@ -117,9 +117,14 @@ export const MonitorProvider = () =>
             return yield* create(news);
           }
 
-          // Sync — the monitor exists; PATCH against its id with the
+          // Sync — the monitor exists; PUT against its id with the
           // desired props. `type` is replacement-only (handled in diff).
-          return yield* update({ ...news, id: observed.id });
+          // If the monitor disappeared between observe and update
+          // (deleted out-of-band mid-reconcile), recover by recreating
+          // instead of surfacing NotFound.
+          return yield* update({ ...news, id: observed.id }).pipe(
+            Effect.catchTag("NotFound", () => create(news)),
+          );
         }),
         delete: Effect.fn(function* ({ output }) {
           yield* del({ id: output.id }).pipe(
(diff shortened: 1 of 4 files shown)


sam-goodwin:

PR #226, opened 2026-05-05
fix(github): harden Variable + Comment reconcile + add lifecycle convergence tests

sam-goodwin:
Hardens the GitHub `Variable` and `Comment` resources, mirroring the per-resource sweep started in #184. Single GitHub scope; `Secret` is being hardened separately.

### Reconciler changes

`Variable` — tolerate the create race and skip the PATCH on a true no-op so `updatedAt` doesn't churn:

```diff
- yield* Effect.tryPromise(() => octokit.rest.actions.createRepoVariable({ … }));
- return { updatedAt: new Date().toISOString() };
+ const created = yield* Effect.tryPromise({
+   try: async () => {
+     try { await octokit.rest.actions.createRepoVariable({ … }); return true; }
+     catch (e: any) { if (e.status === 422) return false; throw e; }
+   },
+   catch: (e) => e as Error,
+ });
+ if (created) return { updatedAt: new Date().toISOString() };
+ // … fall through to the PATCH path below
```

```diff
- if (observed.value !== news.value) { /* PATCH */ }
- return { updatedAt: new Date().toISOString() };
+ if (observed === undefined || observed.value !== news.value) {
+   /* PATCH */ ; return { updatedAt: new Date().toISOString() };
+ }
+ return { updatedAt: output?.updatedAt ?? new Date().toISOString() };
```

`Comment` — only PATCH when the observed body drifts from desired (instead of always issuing the call), so a redeploy with the same `body` is a true no-op:

```diff
- // always PATCH (idempotent on the server, but churns updatedAt)
- const { data } = yield* Effect.tryPromise(() => octokit.rest.issues.updateComment({ … }));
+ if (observed.body !== body) {
+   const { data } = yield* Effect.tryPromise(() => octokit.rest.issues.updateComment({ … }));
+   return { commentId: data.id, htmlUrl: data.html_url, updatedAt: data.updated_at };
+ }
+ return { commentId: observed.id, htmlUrl: observed.htmlUrl, updatedAt: observed.updatedAt };
```

Both resources gain a `diff` that triggers `replace` when the GitHub-side identity changes — `(owner, repository, name)` for `Variable`, `(owner, repository, issueNumber)` for `Comment`. Previously a name/target change would silently create a second resource and orphan the old one.

```ts
diff: Effect.fn(function* ({ news, olds }) {
  if (!isResolved(news)) return undefined;
  if (
    news.owner !== olds.owner ||
    news.repository !== olds.repository ||
    news.name !== olds.name // issueNumber for Comment
  ) {
    return { action: "replace" } as const;
  }
  return undefined;
}),
```

### New lifecycle tests

Each test runs `destroy → deploy → … → destroy` on a `ScratchStack` and asserts convergence at every step. Gated on `GITHUB_TEST_OWNER` / `GITHUB_TEST_REPO` / `GITHUB_TEST_ISSUE_NUMBER` (+ `GITHUB_TEST_ISSUE_NUMBER_2` for the Comment replace test); skipped otherwise.

`Variable`:

- redeploy with same props is a no-op (`updatedAt` preserved)
- reconcile resets a value mutated out-of-band via the raw Octokit SDK
- reconcile re-creates a variable deleted out-of-band
- changing `name` triggers replace; old variable is deleted
- destroying an already-deleted variable is a no-op
- `adopt(true)` re-claims a foreign variable

`Comment`:

- redeploy with same props is a no-op (`commentId` and `updatedAt` preserved)
- reconcile resets a body mutated out-of-band
- reconcile re-creates a comment deleted out-of-band (new `commentId`)
- changing `issueNumber` triggers replace
- destroying an already-deleted comment is a no-op

### DIFF
--- a/packages/alchemy/src/GitHub/Comment.ts
+++ b/packages/alchemy/src/GitHub/Comment.ts
@@ -1,5 +1,6 @@
 import { Octokit } from "@octokit/rest";
 import * as Effect from "effect/Effect";
+import { isResolved } from "../Diff.ts";
 import * as Provider from "../Provider.ts";
 import { Resource } from "../Resource.ts";
 import { dedent } from "../Util/dedent.ts";
@@ -163,6 +164,21 @@ export const CommentProvider = () =>
   Provider.succeed(Comment, {
     stables: ["commentId"],
 
+    // The `(owner, repository, issueNumber)` tuple identifies the
+    // host issue/PR. A comment can't be moved between issues — change
+    // any of them and we must replace.
+    diff: Effect.fn(function* ({ news, olds }) {
+      if (!isResolved(news)) return undefined;
+      if (
+        news.owner !== olds.owner ||
+        news.repository !== olds.repository ||
+        news.issueNumber !== olds.issueNumber
+      ) {
+        return { action: "replace" } as const;
+      }
+      return undefined;
+    }),
+
     reconcile: Effect.fn(function* ({ news, output }) {
       const octokit = createClient(news);
       const body = dedent(news.body);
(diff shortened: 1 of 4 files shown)


sam-goodwin:

PR #227, opened 2026-05-05
fix(aws/autoscaling): harden AutoScalingGroup reconcile + add lifecycle convergence tests

sam-goodwin:
Hardens the AutoScalingGroup resource as part of the per-resource hardening sweep on top of the reconcile migration.

### Reconciler changes

```diff
- Effect.catch((error: any) =>
-   error?._tag === "AlreadyExistsFault"
-     ? Effect.void
-     : Effect.fail(error),
- ),
+ Effect.catchTag("AlreadyExistsFault", () => Effect.void),
```

Blanket `Effect.catch` swallowed every error. Scoped to `AlreadyExistsFault` only — race / restart-after-crash recovery — so auth, throttling, and validation errors propagate.

```diff
  existing = yield* describeGroup(autoScalingGroupName).pipe(
-   Effect.filterOrFail(Boolean, () => new Error(...)),
+   Effect.flatMap((group) =>
+     group ? Effect.succeed(group)
+           : Effect.fail(new AutoScalingGroupNotReadyAfterCreate()),
+   ),
    Effect.retry({
-     while: () => true,
+     while: (e) => e._tag === "AutoScalingGroupNotReadyAfterCreate",
      schedule: Schedule.recurs(8).pipe(...),
    }),
  );
```

The post-create read previously retried on *any* error including auth/validation. Now retries only the actual "still missing in describe" case.

```diff
+ .pipe(
+   Effect.retry({
+     while: (e) =>
+       e._tag === "ScalingActivityInProgressFault" ||
+       e._tag === "ResourceContentionFault",
+     schedule: Schedule.fixed("2 seconds").pipe(
+       Schedule.both(Schedule.recurs(15)),
+     ),
+   }),
+ );
```

`updateAutoScalingGroup` and `deleteAutoScalingGroup` now retry on transient `ScalingActivityInProgressFault` and `ResourceContentionFault`. Delete also retries on `ResourceInUseFault`. Attach/Detach LB target groups retry on the errors their distilled unions actually carry (`ResourceContention`, `InstanceRefreshInProgress`).

```diff
+ const attrs = toAttributes(group);
+ return (yield* hasAlchemyTags(id, attrs.tags))
+   ? attrs
+   : Unowned(attrs);
```

`read` now marks foreign ASGs as `Unowned`, gating takeover behind `adopt(true)` instead of silently re-tagging.

### New lifecycle tests

`packages/alchemy/test/AWS/AutoScaling/AutoScalingGroup.test.ts`. Skipped unless `TEST_LAUNCH_TEMPLATE_ID` and `TEST_SUBNET_IDS` are set (real EC2 fleet required).

- reconcile resets `desiredCapacity` mutated out-of-band via `setDesiredCapacity`
- changing `autoScalingGroupName` triggers replace
- in-place modification of `minSize`/`maxSize`/`desiredCapacity`
- destroying an already-deleted ASG is a no-op
- `adopt(true)` re-tags a foreign ASG

### Distilled patch

No patch needed for this pass. `ScalingActivityInProgressFault` is genuinely transient and a candidate for `withRetryableError` in distilled, but the reconciler-level retry handles it for now without depending on the AWS retry layer. `InvalidInstanceLifecycleStateException` is context-dependent and stays `BadRequestError + ConflictError` only.

### DIFF
--- a/packages/alchemy/src/AWS/AutoScaling/AutoScalingGroup.ts
+++ b/packages/alchemy/src/AWS/AutoScaling/AutoScalingGroup.ts
@@ -1,13 +1,15 @@
 import * as autoscaling from "@distilled.cloud/aws/auto-scaling";
+import * as Data from "effect/Data";
 import * as Effect from "effect/Effect";
 import * as Schedule from "effect/Schedule";
+import { Unowned } from "../../AdoptPolicy.ts";
 import { deepEqual, isResolved } from "../../Diff.ts";
 import type { Input } from "../../Input.ts";
 import { createPhysicalName } from "../../PhysicalName.ts";
 import * as Provider from "../../Provider.ts";
 import { Resource } from "../../Resource.ts";
 import type { Providers } from "../Providers.ts";
-import { createInternalTags, diffTags } from "../../Tags.ts";
+import { createInternalTags, diffTags, hasAlchemyTags } from "../../Tags.ts";
 import type { SubnetId } from "../EC2/Subnet.ts";
 import type {
   LaunchTemplateId,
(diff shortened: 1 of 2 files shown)


sam-goodwin:

PR #228, opened 2026-05-05
fix(aws/ec2): harden InternetGateway/NatGateway/EIP/Route/RouteTable reconcile + add lifecycle convergence tests

sam-goodwin:
Per-resource hardening sweep for EC2 networking on top of the reconcile migration. Mirrors the loop from #184 (SQS Queue) and #189 (EC2 Vpc).

### Reconciler changes

InternetGateway — scope a blanket `Effect.catch` and drop the dead `ValidationError` substring branch (same as #189):

```diff
-).pipe(Effect.catch(() => Effect.succeed({ Attachments: [] })));
+.pipe(
+  Effect.map((r) => r.InternetGateways?.[0]),
+  Effect.catchTag("InvalidInternetGatewayID.NotFound", () =>
+    Effect.succeed(undefined),
+  ),
+);
```

```diff
-while: (e) => {
-  return (
-    e._tag === "DependencyViolation" ||
-    (e._tag === "ValidationError" &&
-      e.message?.includes("DependencyViolation"))
-  );
-},
+while: (e) => e._tag === "DependencyViolation",
```

EIP — `read` returns `undefined` instead of failing when the address was released out-of-band, so the engine recreates rather than crashing. Also drops a `Effect.catchTag("AuthFailure", () => Effect.void)` that turned credential errors into silent destroy successes.

NatGateway — `read` tolerates `NatGatewayNotFound` and treats `deleting`/`deleted` states as "missing" so reconcile recreates instead of returning a tombstone.

EgressOnlyInternetGateway — `read` returns `undefined` on `InvalidEgressOnlyInternetGatewayId.NotFound` instead of failing.

### New lifecycle tests

InternetGateway:
- reconcile re-attaches a detached IGW
- destroying an already-deleted IGW is a no-op

EIP:
- reconcile re-creates an EIP that was released out-of-band
- destroying an already-released EIP is a no-op

RouteTable + Route + RouteTableAssociation (shared fixtures):
- redeploy with same props is a no-op (and reconcile does NOT duplicate the route)
- destroying an already-deleted RouteTable + association is a no-op

NatGateway (single happy-path test, ~5min):
- create + delete a public NAT Gateway with an attached EIP

EgressOnlyInternetGateway:
- destroying an already-deleted EIGW is a no-op

### Distilled patch

No patch landed in this PR. The audit surfaced four operations whose error sets in `@distilled.cloud/aws` are too narrow:

- `createRoute` — missing `RouteAlreadyExists`
- `deleteNatGateway` — missing `DependencyViolation`
- `releaseAddress` — missing `DependencyViolation`
- `disassociateRouteTable` — missing `InvalidRouteTableID.NotFound`

Each is documented inline in the source where the corresponding retry/catch would otherwise sit. They need follow-up patches in distilled, after which we can wire the retries up.

### DIFF
--- a/packages/alchemy/src/AWS/EC2/EIP.ts
+++ b/packages/alchemy/src/AWS/EC2/EIP.ts
@@ -128,15 +128,20 @@ export const EIPProvider = () =>
 
         read: Effect.fn(function* ({ output }) {
           if (!output) return undefined;
-          const result = yield* ec2.describeAddresses({
-            AllocationIds: [output.allocationId],
-          });
+          const result = yield* ec2
+            .describeAddresses({
+              AllocationIds: [output.allocationId],
+            })
+            .pipe(
+              Effect.catchTag("InvalidAllocationID.NotFound", () =>
+                Effect.succeed({ Addresses: [] }),
+              ),
+            );
 
           const address = result.Addresses?.[0];
           if (!address) {
-            return yield* Effect.fail(
-              new Error(`EIP ${output.allocationId} not found`),
-            );
+            // Released out-of-band — let the engine recreate via reconcile.
+            return undefined;
           }
 
           return {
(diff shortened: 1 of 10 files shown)


sam-goodwin:

PR #229, opened 2026-05-05
fix(cloudflare/apitoken): harden Account/User API tokens reconcile + add lifecycle convergence tests

sam-goodwin:
Hardens the `Cloudflare.AccountApiToken` and `Cloudflare.UserApiToken` reconcilers as the next pass of the per-resource hardening sweep on top of the reconcile migration. Both tokens get observe-vs-desired drift detection, an explicit `rollKey` for secret rotation, name-as-replace, and a `read` that discovers by name and brands as `Unowned`.

### Reconciler changes

```diff
  reconcile: Effect.fn(function* ({ id, news, output }) {
    ...
    const observed = output?.tokenId
      ? yield* getToken({ ... }).pipe(
          Effect.catchTag("TokenNotFound", () => Effect.succeed(undefined)),
          Effect.catchTag("InvalidRoute", () => Effect.succeed(undefined)),
        )
      : undefined;

    if (observed === undefined) { /* create */ }

-   // Always call updateToken on reconcile.
-   const result = yield* updateToken({ ... });
+   // Diff observed cloud state against desired and update only on drift.
+   const policyDrift = observedPolicyFingerprint(observed.policies) !== desiredPolicyFp;
+   const condDrift   = conditionFingerprint(observedCondition(observed.condition)) !== desiredCondFp;
+   const nameDrift   = (observed.name ?? "") !== name;
+   const windowDrift = ...;
+   if (policyDrift || condDrift || nameDrift || windowDrift) {
+     current = yield* updateToken({ ... });
+   }
```

Reconcile previously trusted `diff` (which only compares `olds` vs `news`) and unconditionally called `updateToken` on every reconcile. An out-of-band permission mutation never showed up in `olds`, so subsequent deploys with unchanged props left the drift in place. The new flow makes `getToken` the source of truth.

```diff
+ if (news.rollKey !== undefined && news.rollKey !== output!.rolledFor) {
+   const newValue = yield* putTokenValue({ ... });
+   value = Redacted.make(newValue);
+   rolledFor = news.rollKey;
+ }
```

New `rollKey` prop drives `PUT /tokens/{id}/value` (account + user variants). The new bearer is captured into a fresh `Redacted` and persisted as `output.value`; `output.rolledFor` records the marker so subsequent reconciles with the same `rollKey` are no-ops.

```diff
  diff: Effect.fn(function* ({ id, olds, news, output }) {
-   if (oldName !== newName || ... ) { return { action: "update" }; }
+   if (oldName !== newName) { return { action: "replace" }; }
+   if (oldPolicyFp !== newPolicyFp || oldCondFp !== newCondFp || ...) {
+     return { action: "update" };
+   }
  }),
```

Name now forces replace — name is part of token identity for audit/CI consumers and a clean cycle (mint new + retire old) is preferable to renaming in place.

```diff
  read: Effect.fn(function* ({ id, olds, output }) {
    if (output?.tokenId) { /* fast path: getToken */ }
+   const name = yield* resolveName(id, olds?.name);
+   const match = yield* findTokenByName(name);   // paginated listTokens
+   if (!match) return undefined;
+   return Unowned(buildAttributes(match, ...));
  }),
```

Tokens have no tag-based ownership signal, so a discovered token is always `Unowned`. Engine routes silent adoption only when `output.tokenId` round-trips through `getToken`; foreign discovery requires `adopt(true)` / `--adopt`.

### New lifecycle tests

Each test runs `destroy → deploy → … → destroy` on a `ScratchStack` and asserts convergence at every step. Same set for both `AccountApiToken` and `UserApiToken`:

- redeploy with same props is a no-op (token id and secret value preserved)
- update changes policies in place (id and secret value preserved)
- reconcile resets permissions mutated out-of-band via the raw SDK
- `rollKey` rotation: bumping the marker advances the secret value but keeps the id
- reconcile re-creates a token deleted out-of-band
- changing `name` triggers replace; old token is gone
- destroying an already-deleted token is a no-op
- `adopt(true)` re-claims a token after a state-store wipe

### Distilled patch

No patch needed. `accounts.{getToken,createToken,updateToken,deleteToken,listTokens,putTokenValue}` and the matching `user.*` operations already surface `TokenNotFound` and `InvalidRoute` as tagged errors (verified against `/Users/samgoodwin/workspaces/distilled/packages/cloudflare/src/services/{accounts,user}.ts`); no `UnknownCloudflareError` paths surface in the reconciler.

### DIFF
--- a/packages/alchemy/src/Cloudflare/ApiToken/AccountApiToken.ts
+++ b/packages/alchemy/src/Cloudflare/ApiToken/AccountApiToken.ts
@@ -1,6 +1,7 @@
 import * as accounts from "@distilled.cloud/cloudflare/accounts";
 import * as Effect from "effect/Effect";
 import * as Redacted from "effect/Redacted";
+import { Unowned } from "../../AdoptPolicy.ts";
 import { isResolved } from "../../Diff.ts";
 import { createPhysicalName } from "../../PhysicalName.ts";
 import * as Provider from "../../Provider.ts";
@@ -10,6 +11,8 @@ import type { Providers } from "../Providers.ts";
 import {
   buildConditionPayload,
   conditionFingerprint,
+  observedCondition,
+  observedPolicyFingerprint,
   policyFingerprint,
   resolvePolicies,
   type ApiTokenProps,
@@ -29,6 +32,12 @@ export type AccountApiToken = Resource<
      */
     value: Redacted.Redacted<string>;
     accountId: string;
+    /**
+     * Opaque marker for the most recent successful `rollKey`. Persisting
+     * this lets the reconciler detect when the user has bumped `rollKey`
+     * and trigger a `putTokenValue` rotation.
+     */
+    rolledFor?: string;
   },
   never,
   Providers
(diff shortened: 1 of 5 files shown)


sam-goodwin:

PR #230, opened 2026-05-05
fix(aws/scheduler): harden ScheduleGroup reconcile + add lifecycle convergence tests

sam-goodwin:
Following the SQS hardening template from [#184](https://github.com/alchemy-run/alchemy-effect/pull/184) and the sibling Schedule PR [#200](https://github.com/alchemy-run/alchemy-effect/pull/200). Hardens the EventBridge Scheduler `ScheduleGroup` resource. Unlike `Schedule`, schedule groups DO support tagging — so ownership detection, adoption gating, and tag-drift convergence are all in scope here.

### Reconciler changes

```diff
- if (!observed?.Arn) {
-   yield* scheduler.createScheduleGroup({ Name, Tags })
-     .pipe(Effect.catchTag("ConflictException", () => …listTagsForResource…hasTags…throw…))
-   observed = yield* scheduler.getScheduleGroup({ Name })
- }
- // tag sync diffed against observedTagsResp — kept
+ const arn = observed?.Arn
+   ?? yield* scheduler.createScheduleGroup({ Name, Tags })
+        .pipe(conflictRetry, Effect.map(r => r.ScheduleGroupArn));
```

- `read` now returns `Unowned(attrs)` when the group lacks the alchemy internal tag set, so first-touch adoption requires `--adopt` / `adopt(true)` instead of being silently taken over. Subsequent reconciles trust the persisted output.
- Drop the bespoke `ConflictException` ownership check on `createScheduleGroup`. The previous code fetched tags, called `hasTags`, and threw a plain `Error` if they didn't match — duplicating ownership logic that now lives entirely in `read` + the engine's `--adopt` gate.
- Replace it with a bounded `ConflictException` retry (15 × 2s) on `createScheduleGroup`, `tagResource`, `untagResource`, and `deleteScheduleGroup`. EventBridge Scheduler returns `ConflictException` while a group is in the transient `DELETING`/`CREATING` state, so re-creating a name through a residual `DELETING` window now converges instead of failing.
- Re-read the final `State` from the cloud after reconcile rather than echoing `observed?.State` (which may be stale by one round-trip).

`ConflictException` is intentionally NOT tagged `RetryableError` in distilled — it's context-dependent (`DELETING` race vs. genuine name collision). The reconciler-level bounded retry stays scoped.

### New lifecycle tests

`packages/alchemy/test/AWS/Scheduler/ScheduleGroup.test.ts` — each test runs `destroy → deploy → … → destroy` on a `ScratchStack` and asserts convergence at every step:

- create + delete with default props
- reconcile resets tags mutated out-of-band via the raw Scheduler SDK
- reconcile re-creates a group deleted out-of-band (rides the `DELETING` window)
- changing `name` triggers replace; old group is deleted
- destroying an already-deleted group is a no-op
- foreign-tagged group requires `adopt(true)` to take over and rewrites internal alchemy tags

### Distilled patch

No patch needed — `packages/aws/patches/scheduler.json` (per the existing process) already covers `ConflictException`, and tagging it `RetryableError` would over-broaden coverage. The reconciler-level bounded retry handles the genuinely transient cases.

### DIFF
--- a/packages/alchemy/src/AWS/Scheduler/ScheduleGroup.ts
+++ b/packages/alchemy/src/AWS/Scheduler/ScheduleGroup.ts
@@ -1,16 +1,18 @@
 import * as scheduler from "@distilled.cloud/aws/scheduler";
 import * as Effect from "effect/Effect";
+import * as Schedule_ from "effect/Schedule";
+import { Unowned } from "../../AdoptPolicy.ts";
 import { isResolved } from "../../Diff.ts";
 import { createPhysicalName } from "../../PhysicalName.ts";
 import * as Provider from "../../Provider.ts";
 import { Resource } from "../../Resource.ts";
-import type { Providers } from "../Providers.ts";
 import {
   createInternalTags,
   createTagsList,
   diffTags,
-  hasTags,
+  hasAlchemyTags,
 } from "../../Tags.ts";
+import type { Providers } from "../Providers.ts";
 
 export interface ScheduleGroupProps {
   /**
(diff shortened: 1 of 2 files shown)


sam-goodwin:

PR #231, opened 2026-05-05
fix(aws/organizations): harden Organizations resources reconcile + add lifecycle convergence tests

sam-goodwin:
Per-resource hardening sweep for AWS Organizations. No live integration tests added — they require a real management account and consume a non-renewable account-creation quota; flagged as outstanding.

### Reconciler changes

Add `PolicyChangesInProgressException` to the transient set and bump the retry budget. AWS Organizations serializes policy changes per-org and surfaces this tag mid-apply; it's genuinely transient.

```diff
 const isRetryableOrganizationsError = (error: any) =>
   error?._tag === "ConcurrentModificationException" ||
   error?._tag === "TooManyRequestsException" ||
   error?._tag === "ServiceException" ||
-  error?._tag === "FinalizingOrganizationException";
+  error?._tag === "FinalizingOrganizationException" ||
+  error?._tag === "PolicyChangesInProgressException";
```

Bound `retryDeleteEmptiness` for `*NotEmptyException` / `PolicyInUseException` on delete. These are genuine dependency violations, but right after the engine deletes children the parent's listing is eventually consistent and can briefly still report non-empty. Caps at ~10 attempts of exponential backoff so a real persistent reference still surfaces.

```ts
export const retryDeleteEmptiness = <A, E, R>(effect: Effect.Effect<A, E, R>) =>
  effect.pipe(
    Effect.retry({
      while: (error: any) =>
        isRetryableOrganizationsError(error) ||
        error?._tag === "OrganizationNotEmptyException" ||
        error?._tag === "OrganizationalUnitNotEmptyException" ||
        error?._tag === "PolicyInUseException",
      schedule: Schedule.exponential(500).pipe(
        Schedule.both(Schedule.recurs(10)),
      ),
    }),
  );
```

Wrap `Organization.delete`, `OrganizationalUnit.delete`, `Policy.delete` with `retryDeleteEmptiness`.

Stable-stringify policy documents in `Policy` and `OrganizationResourcePolicy` reconcilers. AWS doesn't preserve object key order on round-trip, so the naive `JSON.stringify` comparison flapped and re-issued `updatePolicy` / `putResourcePolicy` on every reconcile.

```diff
- const observedContent = JSON.stringify(state.document);
+ const observedContent = stableStringify(state.document);
```

Bump `Account` create-status poll cap from 240s (120 × 2s) to 600s (300 × 2s). Account creation typically completes in ~90s but can extend past 5 minutes; the old cap surfaced spurious failures.

### New lifecycle tests

None. Organizations reconcile/delete tests need a real management account, can only run sequentially, and exhaust the non-renewable account-creation quota. Flagged as outstanding — running them requires an opted-in test account.

### Distilled patch

No patch needed. The relevant errors (`PolicyChangesInProgressException`, `*NotEmptyException`, `PolicyInUseException`, `ConcurrentModificationException`) already carry the right base categories in distilled (`ConflictError`); the per-call retry helpers handle the context-specific transience that the AWS retry layer can't infer.

### DIFF
--- a/packages/alchemy/src/AWS/Organizations/Account.ts
+++ b/packages/alchemy/src/AWS/Organizations/Account.ts
@@ -301,10 +301,14 @@ const waitForCreateAccount = (requestId: string) =>
 
     return status;
   }).pipe(
+    // Account creation is asynchronous on AWS's side. Typical completion is
+    // ~90s but it can extend past 5 minutes under load or when CloudTrail /
+    // billing setup is slow. Cap at ~10 minutes (300 polls × 2s) so a stuck
+    // request eventually surfaces.
     Effect.retry({
       while: (error: any) => error?._tag === "CreateAccountInProgress",
       schedule: Schedule.spaced("2 seconds").pipe(
-        Schedule.both(Schedule.recurs(120)),
+        Schedule.both(Schedule.recurs(300)),
       ),
     }),
   );
--- a/packages/alchemy/src/AWS/Organizations/Organization.ts
+++ b/packages/alchemy/src/AWS/Organizations/Organization.ts
@@ -3,7 +3,7 @@ import * as Effect from "effect/Effect";
 import * as Provider from "../../Provider.ts";
 import { Resource } from "../../Resource.ts";
 import type { Providers } from "../Providers.ts";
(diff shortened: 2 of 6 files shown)


sam-goodwin:

PR #232, opened 2026-05-05
fix(aws/identitycenter): harden PermissionSet + AccountAssignment reconcile + add lifecycle convergence tests

sam-goodwin:
Per-resource hardening sweep for `AWS.IdentityCenter.PermissionSet` and convergence test coverage that also exercises `AccountAssignment`-adjacent flows. Distilled `ConflictException` is already classified `ConflictError`; the AWS retry layer doesn't pick it up by default, so the reconciler-level retry (`retryIdentityCenter`) stays as the source of truth for IdentityCenter-specific concurrent-modify recovery.

### Reconciler changes

Add ownership tagging to PermissionSet. SSO Admin supports `tagResource` / `listTagsForResource` / `untagResource` on permission sets, so the resource now carries the standard `alchemy::stack` / `alchemy::stage` / `alchemy::id` tags and gates adoption via `Unowned()` in `read`.

```ts
read: Effect.fn(function* ({ id, olds, output }) {
  const observed = /* describe by ARN or by name */;
  if (!observed) return undefined;
  return (yield* hasAlchemyTags(id, observed.tags))
    ? observed
    : Unowned(observed);
}),
```

Tags are seeded on `createPermissionSet` and reconciled against observed cloud state on every reconcile via `syncPermissionSetTags` (uses `diffTags` from `Tags.ts`). Tag drift, foreign-ownership takeover, and `--adopt` re-tagging all converge in one pass.

```ts
const { added, updated, removed } = diffTags(observedTags, desiredTags);
const upsert = [...added, ...updated];
if (upsert.length > 0) yield* retryIdentityCenter(ssoAdmin.tagResource({ ... }));
if (removed.length > 0) yield* retryIdentityCenter(ssoAdmin.untagResource({ ... }));
```

Replace the two ad-hoc `new Error(...)` failures after create/update with tagged errors so callers can pattern-match.

```ts
class PermissionSetMissingAfterCreate extends Data.TaggedError(...) {}
class PermissionSetMissingAfterUpdate extends Data.TaggedError(...) {}
```

### New lifecycle tests

`packages/alchemy/test/AWS/IdentityCenter/PermissionSet.test.ts` runs `destroy → deploy → … → destroy` on a `ScratchStack`. Tests are gated on `ALCHEMY_RUN_LIVE_AWS_IDENTITY_CENTER_TESTS=true` because they need a live IAM Identity Center instance.

- reconcile resets `description` and `relayState` mutated out-of-band via the raw SDK
- changing `name` triggers replace; old permission set is deleted
- destroying an already-deleted permission set is a no-op
- `adopt(true)` re-tags a foreign permission set (created out-of-band without alchemy tags)

### Distilled patch

No patch needed. `ConflictException` is already `ConflictError`; we deliberately keep it on the IdentityCenter-specific retry rather than promoting it to `RetryableError` globally — concurrent-modify there is genuinely transient but the recovery semantics (poll the assignment-status request id, etc.) are context-dependent and not safe to bake into a generic `RetryableError` policy.

### DIFF
--- a/packages/alchemy/src/AWS/IdentityCenter/PermissionSet.ts
+++ b/packages/alchemy/src/AWS/IdentityCenter/PermissionSet.ts
@@ -1,9 +1,17 @@
 import * as ssoAdmin from "@distilled.cloud/aws/sso-admin";
+import * as Data from "effect/Data";
 import * as Effect from "effect/Effect";
 import * as Stream from "effect/Stream";
+import { Unowned } from "../../AdoptPolicy.ts";
 import { isResolved } from "../../Diff.ts";
 import * as Provider from "../../Provider.ts";
 import { Resource } from "../../Resource.ts";
+import {
+  createInternalTags,
+  createTagsList,
+  diffTags,
+  hasAlchemyTags,
+} from "../../Tags.ts";
 import type { Providers } from "../Providers.ts";
 import { resolveInstance, retryIdentityCenter } from "./common.ts";
 
@@ -14,7 +22,7 @@ export interface PermissionSetProps {
    */
   instanceArn?: string;
   /**
-   * Permission set name.
+   * Permission set name. Stable — changing this triggers a replace.
    */
   name: string;
   /**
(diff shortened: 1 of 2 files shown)


sam-goodwin:

PR #233, opened 2026-05-05
fix(axiom): harden Dashboard/View/VirtualField/Annotation/ApiToken reconcile + add lifecycle convergence tests

sam-goodwin:
Stacked on top of [#197](https://github.com/alchemy-run/alchemy-effect/pull/197). Hardens the remaining Axiom REST resources (Dashboard, View, VirtualField, Annotation, ApiToken) the same way #197 did for Dataset.

### Reconciler changes

- **View** — name is the path identifier and `createView` declares only `UnprocessableEntity`, but 409s fall through `HTTP_STATUS_MAP` to `Conflict` / `UnknownAxiomError`. Replace the previous unsafe collision recovery with the observe-then-update pattern: re-`get` on conflict, fail loudly if ownership has flipped, otherwise PUT.

```diff
- if (observed === undefined) {
-   const result = yield* create(news);
-   return { ...result, id: news.name };
- }
+ if (observed === undefined) {
+   const result = yield* create(desired).pipe(
+     Effect.catch((e: { readonly _tag?: string }) => {
+       const tag = e._tag;
+       if (tag === "UnprocessableEntity" || tag === "Conflict" || tag === "UnknownAxiomError") {
+         return Effect.gen(function* () {
+           const existing = yield* get({ id: news.name }).pipe(
+             Effect.catchTag("NotFound", () => Effect.succeed(undefined)),
+           );
+           if (existing === undefined) return yield* Effect.fail(e);
+           const ownership = parseMarker(existing.description);
+           const isOurs = ownership === undefined ||
+             (ownership.stack === stack.name && ownership.stage === stage && ownership.id === id);
+           if (!isOurs) return yield* Effect.fail(e);
+           return yield* update({ ...desired, id: news.name });
+         });
+       }
+       return Effect.fail(e);
+     }),
+   );
+   return { ...result, id: news.name };
+ }
```

- **Dashboard / View / VirtualField / Annotation** — sync-path `update` now catches `NotFound` and recreates if the resource was deleted out-of-band mid-reconcile. `updateDashboard` doesn't declare `NotFound` in its typed errors, but the SDK's `HTTP_STATUS_MAP` still maps 404 → `NotFound`, so we narrow on `_tag` instead of `Effect.catchTag`.

```diff
- return toAttrsFromCreate(
-   yield* update({ ...news, uid: observed.uid, overwrite: true }),
- );
+ return toAttrsFromCreate(
+   yield* update({ ...desired, uid: observed.uid, overwrite: true }).pipe(
+     Effect.catch((e: { readonly _tag?: string }) => {
+       if (e._tag === "NotFound") return create(desired);
+       return Effect.fail(e);
+     }),
+   ),
+ );
```

- **Dashboard / View / VirtualField / Annotation** — added the same description-marker ownership scheme Dataset uses, so `read` can return `Unowned(attrs)` when it finds a foreign-tagged resource. `reconcile` augments `description` with a deterministic `[alchemy:stack=…;stage=…;id=…]` marker on every write so subsequent reads can tell ours apart from someone else's.

- **ApiToken** — replaced `Effect.die` with `Effect.fail` on the (unreachable) "create returned no token" branch, so a misbehaving Axiom response surfaces as a typed error instead of crashing the engine.

### New lifecycle tests

One file per resource at `packages/alchemy/test/Axiom/{Dashboard,View,VirtualField,Annotation,ApiToken}.test.ts`. Each runs `destroy -> deploy -> ... -> destroy` on a `ScratchStack` and asserts convergence at every step.

- redeploy with same props is a no-op (id stable, no rotation for ApiToken)
- reconcile resets fields mutated out-of-band via the raw Axiom API
- reconcile re-creates a resource deleted out-of-band
- changing name (or `description` for ApiToken) — replace vs in-place update, depending on the resource
- destroying an already-deleted resource is a no-op
- (View) `adopt(true)` re-claims a foreign-marked view by name

### Distilled patch

No patch needed. Same reasoning as #197: Axiom's `client/api.ts` already routes HTTP status to typed errors via `HTTP_STATUS_MAP` (with `UnknownAxiomError` as the fallback) and the `Retry` Context.Service drives backoff. The reconciler change is purely about catching the right shapes alchemy-side.

### DIFF
--- a/packages/alchemy/src/Axiom/Annotation.ts
+++ b/packages/alchemy/src/Axiom/Annotation.ts
@@ -1,8 +1,36 @@
 import * as Axiom from "@distilled.cloud/axiom";
 import * as Effect from "effect/Effect";
+import { Unowned } from "../AdoptPolicy.ts";
 import * as Provider from "../Provider.ts";
 import { Resource } from "../Resource.ts";
+import { Stack } from "../Stack.ts";
+import { Stage } from "../Stage.ts";
 import type { Providers } from "./Providers.ts";
+
+const MARKER_RE = /\s*\[alchemy:stack=([^;]+);stage=([^;]+);id=([^\]]+)\]\s*$/;
+
+const buildMarker = (stack: string, stage: string, id: string) =>
+  `[alchemy:stack=${stack};stage=${stage};id=${id}]`;
+
+const stripMarker = (description: string | undefined) =>
+  (description ?? "").replace(MARKER_RE, "").trimEnd();
+
+const augmentDescription = (
+  description: string | undefined,
+  marker: string,
+) => {
+  const base = stripMarker(description);
+  return base.length > 0 ? `${base}\n${marker}` : marker;
+};
+
+const parseMarker = (
+  description: string | undefined,
+): { stack: string; stage: string; id: string } | undefined => {
+  if (!description) return undefined;
+  const m = description.match(MARKER_RE);
+  if (!m) return undefined;
+  return { stack: m[1], stage: m[2], id: m[3] };
+};
 
 export type AnnotationProps = Axiom.CreateAnnotationInput;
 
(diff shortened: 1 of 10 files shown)


sam-goodwin:

PR #234, opened 2026-05-05
fix(cloudflare/workflow): harden Workflow reconcile + add lifecycle convergence tests

sam-goodwin:
Hardens the Cloudflare Workflow reconciler. Previously every reconcile did a blind PUT regardless of live state, so every redeploy re-PUT and there was no signal when desired state already matched.

### Reconciler changes

```diff
+ const observed = yield* getWorkflow({
+   accountId: acct,
+   workflowName: news.workflowName,
+ }).pipe(
+   Effect.map((r) => r as workflows.GetWorkflowResponse | undefined),
+   Effect.catchTag("WorkflowNotFound", () => Effect.succeed(undefined)),
+ );
+
+ if (
+   observed !== undefined &&
+   observed.className === news.className &&
+   observed.scriptName === news.scriptName
+ ) {
+   return {
+     workflowId: observed.id,
+     workflowName: observed.name,
+     className: observed.className,
+     scriptName: observed.scriptName,
+     accountId: acct,
+   };
+ }

  const result = yield* putWorkflow({ ... });
```

Observe live state first via `getWorkflow`. Only `WorkflowNotFound` is swallowed — auth, throttling, and 5xx errors must surface so the engine can retry or report cleanly. Skip the PUT when observed state already matches desired (no API churn on no-op redeploys).

### New lifecycle tests

Each test runs `destroy → deploy → … → destroy` against a real Cloudflare account.

- redeploy with same props is a no-op; workflow id is stable
- reconcile resets `scriptName` re-targeted out-of-band via raw `putWorkflow`
- reconcile re-creates a workflow deleted out-of-band (rides `WorkflowNotFound` in observe)
- destroying an already-deleted workflow is a no-op (`deleteWorkflow` catches `WorkflowNotFound`)

Workflow has no `read` lifecycle so `adopt(true)` and "owned without `--adopt`" cases don't apply here. Physical-name (`workflowName`) replace is also outside scope: `workflowName` derives from the user-facing class name in `Cloudflare.Workflow("X", ...)` and the engine's Resource diff handles class-rename as a logical-id change.

### Distilled patch

No patch needed. Cloudflare's `client/api.ts` already maps HTTP status to categorized errors (`TooManyRequests` → throttling+retryable, 5xx → server+retryable), and `WorkflowNotFound` covers the only NotFound-like code the reconciler routes on.

### DIFF
--- a/packages/alchemy/src/Cloudflare/Workers/Workflow.ts
+++ b/packages/alchemy/src/Cloudflare/Workers/Workflow.ts
@@ -475,6 +475,7 @@ export const WorkflowProvider = () =>
     WorkflowResource,
     Effect.gen(function* () {
       const { accountId } = yield* CloudflareEnvironment;
+      const getWorkflow = yield* workflows.getWorkflow;
       const putWorkflow = yield* workflows.putWorkflow;
       const deleteWorkflow = yield* workflows.deleteWorkflow;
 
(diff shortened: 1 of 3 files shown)


sam-goodwin:

PR #235, opened 2026-05-05
fix(kubernetes): harden ServiceAccount/ConfigMap/Job/Object reconcile + add lifecycle convergence tests

sam-goodwin:
Tighten the Kubernetes REST client and surface a few real props on the resource wrappers. Out-of-band edits and transient API failures used to fall through `requestJson` as untyped errors; the reconciler now reacts to them deterministically.

### Reconciler changes

Typed errors per HTTP status, instead of the single `KubernetesApiError` catch-all:

```ts
404 -> KubernetesObjectNotFound   // idempotent delete swallows
409 -> KubernetesConflict         // retried on apply (SSA race), context-dependent on create
410 -> KubernetesGone             // non-retryable
401/403 -> KubernetesUnauthorized // non-retryable
5xx -> KubernetesServerError      // retried with exponential backoff
422/other -> KubernetesApiError   // generic, NOT auto-retried
```

Apply and delete now ride out transient races:

```ts
applyObject(...).pipe(
  Effect.retry({
    while: (e) => e instanceof KubernetesConflict || e instanceof KubernetesServerError,
    schedule: Schedule.exponential(Duration.millis(250)).pipe(
      Schedule.both(Schedule.recurs(5)),
    ),
  }),
);
```

`deleteObject` uses `propagationPolicy: Background` so dependents (Job-managed Pods) are reaped without blocking, and 404 is treated as the desired terminal state.

Apply also stops sending `application/apply-patch+yaml` for non-PATCH verbs:

```ts
"Content-Type": isApplyPatch(method, body)
  ? "application/apply-patch+yaml"
  : "application/json",
```

Resource wrappers grew the props that practical use needs:

- `ConfigMap` — `binaryData`, `immutable`
- `ServiceAccount` — `automountServiceAccountToken`, `imagePullSecrets`
- `Job` — `parallelism`, `completions`, `backoffLimit`, `activeDeadlineSeconds`, `ttlSecondsAfterFinished`, `annotations`

JSDoc on `Object` now spells out that custom resources work as long as their `apiVersion`/`kind` is registered in `supportedKinds`.

### New lifecycle tests

Pure unit tests in `packages/alchemy/test/Kubernetes/client.test.ts` (24 cases). Integration coverage is deferred — there is no kind/minikube fixture in this repo yet, so per-resource lifecycle convergence (no-op redeploy, drift reconcile, OOB delete recovery, name-change replace, double-destroy) cannot be exercised end-to-end without standing one up.

- `classifyKubernetesStatus` — every status maps to its narrowest typed error; 422 stays untyped (context-dependent, not auto-retried)
- `isKubernetesKindSupported` — accepts the six canonical kinds, rejects unknown CRDs without throwing
- `buildKubernetesObjectPath` — core vs grouped vs cluster-scoped; throws on missing namespace for namespaced kinds
- `chunkByApplyRank` / `sortRefsForDelete` — Namespace -> SA -> CM -> Job on apply, exact reverse on delete

### DIFF
--- a/packages/alchemy/src/Kubernetes/ConfigMap.ts
+++ b/packages/alchemy/src/Kubernetes/ConfigMap.ts
@@ -16,9 +16,25 @@ export interface ConfigMapProps {
    */
   name?: string;
   /**
-   * Config map string data.
+   * Config map UTF-8 string data, keyed by file name.
    */
   data?: Record<string, string>;
+  /**
+   * Config map binary data, keyed by file name. Values must be base64-encoded.
+   *
+   * Useful for embedding non-UTF-8 payloads (e.g. images, certificates) that
+   * `data` cannot represent without corruption.
+   */
+  binaryData?: Record<string, string>;
+  /**
+   * If `true`, the config map is marked immutable on the API server. Subsequent
+   * updates that change `data`/`binaryData` will be rejected by Kubernetes —
+   * the only way to change values is to delete and recreate. Use for
+   * high-fan-out config maps where reduced kubelet watch load matters.
+   *
+   * @default false
+   */
+  immutable?: boolean;
   /**
    * Config map labels.
    */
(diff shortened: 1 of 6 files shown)


sam-goodwin:

PR #237, opened 2026-05-05
feat(cloudflare/images): add Images binding

agcty:
Adds a Cloudflare Images binding with a full Effect-native client.

- `Cloudflare.Images()` — pure binding marker, attachable via `Worker({ bindings: { ... } })` or the init phase.
- `Cloudflare.Images.bind(images)` returns an `ImagesClient` whose methods take `Stream.Stream<Uint8Array>` and return `Effect`s tagged with `ImagesError`. The runtime ReadableStream conversion is internal.

```ts
const images = yield* Cloudflare.Images.bind(Pipeline);

// info ─ Stream<Uint8Array> in, Effect<ImageInfoResponse> out
const info = yield* images.info(request.stream);

// transform ─ chainable, single Effect at the end
const out = yield* (yield* images.input(request.stream))
  .transform({ width: 128 })
  .output({ format: "image/jpeg" });
```

- Wires Images through Worker binding normalization, `InferEnv`, and the deploy-time policy registered in `Cloudflare.providers()`.
- Tests: two metadata tests assert `{ type: "images", name }` lands in the script settings (declarative + init-phase). A third end-to-end test deploys an `ImagesWorker` fixture, POSTs a 1×1 PNG, and asserts Cloudflare Images parses it (`format: "image/png"`, `width: 1`, `height: 1`) — exercising `Images.bind`, the `Stream<Uint8Array>` → runtime `ReadableStream` adapter, and the live runtime call.


agcty:
Verification update:

- Focused binding tests passed: `bun vitest run packages/alchemy/test/Cloudflare/Images/Images.test.ts`
- This is a compile-time Worker binding metadata resource; no Cloudflare live resources are needed for the test.

packages/alchemy/test/Cloudflare/Images/Images.test.ts
```diff
@@ -0,0 +1,106 @@
...
+      Effect.provideService(Stage, "test"),
+    );
+
+const emptyProviders = Layer.empty as unknown as Layer.Layer<any, never, any>;
+
+const compileStack = compileStackWithProviders(emptyProviders);
+
+const compileStackWithImagesBinding = compileStackWithProviders(
+  Cloudflare.ImagesBindingPolicyLive,
+);
```
sam-goodwin: You don't need this, provide `Cloudflare.providers()` to your `Test.make` and use the provider level testing.

see: https://v2.alchemy.run/guides/custom-provider/#test-the-lifecycle

sam-goodwin: Example: https://github.com/alchemy-run/alchemy-effect/blob/main/packages/alchemy/test/Cloudflare/R2/Bucket.test.ts

sam-goodwin (changes requested):

packages/alchemy/test/Cloudflare/Images/Images.test.ts
agcty: updated this to use the provider test harness with `Cloudflare.providers()`, as requested. typecheck passes. i also tried a live run, but the token i have does not include workers script permissions, so the worker script settings endpoint is blocked by credentials.

baptisteArno:
Need this guys, rtr? :)

sam-goodwin:
It was missing the effect-native interface and behavior tests. Pushed those and docs

PR #238, opened 2026-05-05
feat(cloudflare/ratelimit): add RateLimit binding

agcty:
Adds a Cloudflare `RateLimit` binding for counting arbitrary keys inside Workers. `RateLimit({...})` returns a marker that is a plain data structure in the `env` position **and** yieldable inside an Effect-native Worker — both resolve to the native `cf.RateLimit` binding.

### Declare via `env` (async or effect Worker)

`InferEnv` maps the marker to the native `cf.RateLimit`, so a plain async handler calls `env.THROTTLE.limit(...)`:

```ts
export const Worker = Cloudflare.Worker("Worker", {
  main: "./src/worker.ts",
  env: {
    THROTTLE: Cloudflare.RateLimit({
      namespaceId: 1001,
      simple: { limit: 2, period: 10 },
    }),
  },
});

// worker.ts
export default {
  fetch: async (req: Request, env: Cloudflare.InferEnv<typeof Worker>) => {
    const { success } = await env.THROTTLE.limit({ key: "ip" });
    return new Response(success ? "ok" : "rate limited", {
      status: success ? 200 : 429,
    });
  },
};
```

### Bind inside an effect-native Worker

`yield* Cloudflare.RateLimit({...})` attaches the binding to the surrounding Worker **and** returns the runtime client in one step — no separate `.bind(...)`:

```ts
Cloudflare.Worker("Worker", { main: "./src/worker.ts" },
  Effect.gen(function* () {
    const throttle = yield* Cloudflare.RateLimit({
      namespaceId: 1001,
      simple: { limit: 2, period: 10 },
    });

    return {
        const { success } = yield* throttle.limit({ key: "ip" });
        return HttpServerResponse.text(success ? "ok" : "rate limited");
      }),
    };
  }).pipe(Effect.provide(Cloudflare.RateLimitBindingLive)),
);
```

The dual behavior is achieved by giving the marker a `[Symbol.iterator]` (wrapping the bind Effect) without branding it as an `Effect`: it stays a plain `RateLimit` value for `InferEnv` / `env`, but `yield*` runs the binding.

### Tests

End-to-end behavioral coverage over `HttpClient` against deployed Workers — one suite run against both the async (`env`) and effect (`yield* RateLimit`) styles. The Worker drives `limit({ key })` `n` times within a single isolate, so throttling is deterministic without relying on edge propagation between requests. Assertions check the observable contract (first call allowed, burst eventually throttled, per-key isolation) rather than an exact allowed count, since Cloudflare's limiter is best-effort/per-colo.


agcty:
Verification update:

- Focused binding tests passed: `bun vitest run packages/alchemy/test/Cloudflare/RateLimit/RateLimit.test.ts`
- This PR depends on the Distilled upload schema fix in alchemy-run/distilled#256 for Worker script uploads containing `ratelimit` bindings.

packages/alchemy/test/Cloudflare/RateLimit/RateLimit.types.ts
```diff
@@ -0,0 +1,19 @@
...
+  },
+});
+
+type WorkerEnv = Cloudflare.InferEnv<typeof Worker>;
+
+declare const env: WorkerEnv;
+
+const result = env.RATE_LIMIT.limit({ key: "user-1" });
+
+void result;
```
Mkassabov: why do we have a .types.ts file that does more than just export types (which we also generally should avoid if possible)

Mkassabov (commented):
can you rebase off latest main please

PR #239, opened 2026-05-05
feat(cloudflare/zone): add Zone resource

agcty:
Adds a first-class Cloudflare Zone resource for zone-scoped infrastructure.

- supports creating or adopting zones by name
- exposes zone attributes needed by follow-on resources
- supports mutable `type`, `paused`, and `vanityNameServers` patching
- supports `delete: false` for retained production zones
- registers Zone in the Cloudflare provider collection

Verification run:

- `bun vitest run packages/alchemy/test/Cloudflare/Zone/Zone.test.ts`
- `bun tsc -b --pretty false`
- `git diff --check`

### DIFF
--- a/packages/alchemy/src/Cloudflare/Providers.ts
+++ b/packages/alchemy/src/Cloudflare/Providers.ts
@@ -32,6 +32,7 @@ import * as Tunnel from "./Tunnel/index.ts";
 import * as VpcService from "./VpcService/index.ts";
 import * as Workers from "./Workers/index.ts";
 import * as Workflows from "./Workers/Workflow.ts";
+import * as Zone from "./Zone/index.ts";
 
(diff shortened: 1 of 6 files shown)


agcty:
Verification update:

- Focused test passed without live env: `bun vitest run packages/alchemy/test/Cloudflare/Zone/Zone.test.ts` (skipped as expected).
- Live test passed against the real Cloudflare zone `sneef.xyz` using env credentials and an isolated temporary Alchemy profile. The test adopted the existing zone with `delete: false`, preserved current mutable zone settings, destroyed local state, and verified the zone was retained.

sam-goodwin:
Closing in favor of https://github.com/alchemy-run/alchemy-effect/pull/493

PR #240, opened 2026-05-05
feat(cloudflare/ruleset): add Ruleset resource

agcty:
Adds a first-class Cloudflare Ruleset resource for zone phase entrypoints.

- supports zone-scoped phase entrypoint ownership via `rulesets.putPhas`
- accepts `{ zoneId }`, zone ID strings, and hostname strings for zone selection
- clears the owned phase entrypoint on destroy
- adds a focused live test gated by `CLOUDFLARE_TEST_RULESET_ZONE_ID`
- registers Ruleset in the Cloudflare provider collection

Depends on alchemy-run/distilled#257 for generated Rulesets phase path support.

Verification run:

- `bun vitest run packages/alchemy/test/Cloudflare/Ruleset/Ruleset.test.ts`
- `bun tsc -b --pretty false`
- `git diff --check`


agcty:
Verification update:

- Focused test passes without live env: `bun vitest run packages/alchemy/test/Cloudflare/Ruleset/Ruleset.test.ts` (skipped as expected).
- Live test is currently blocked by alchemy-run/distilled#257. With the currently installed `@distilled.cloud/cloudflare@0.16.7`, the generated rulesets client calls `/{accountOrZone}/{accountOrZoneId}/rulesets/phases/...`, and Cloudflare returns `No route for that URI`. This is the exact path-generation bug fixed in the Distilled PR, so I am leaving this PR draft until that dependency is available here or otherwise linked for an end-to-end live run.

agcty:
Marking this ready for review now. One dependency remains for full live-test verification: alchemy-run/distilled#257 needs to land or be otherwise consumed by this repo so the generated Rulesets client uses the correct phase-entrypoint path. The current implementation and focused tests are ready for review, and the live-test blocker is documented above.

sam-goodwin:
@Mkassabov we need to figure out what we want to do for https://github.com/alchemy-run/distilled/pull/257

SO we can get this merged. It's a high demand/high utility resource

agcty:
Update: Distilled `0.19.1` is published now, so this branch has been rebased and updated to consume the scoped Rulesets operations from the released package (`getPhasForZone` / `putPhasForZone`).

Verified:
- `bun install --frozen-lockfile`
- `bun tsc -b --pretty false`
- `bun test packages/alchemy/test/Cloudflare/Ruleset/Ruleset.test.ts` (skips without `CLOUDFLARE_TEST_EMPTY_RULESET_ZONE_ID`)

The previous live-test path issue is resolved by the new Distilled package. Full live create/update/delete still needs a Cloudflare token with Zone Rulesets read/write for the test zone; the earlier token reached the correct endpoint but Cloudflare returned `Forbidden`.

@sam-goodwin @Mkassabov

agcty:
pushing some fixes rn, testing with my own zone

sam-goodwin:
Thanks. Please make sure we have adequate regression tests for anything you find on your zone

agcty:
small dependency note from live testing: create/update/delete with a real Cloudflare token and `sneef.xyz` reached the scoped zone phase endpoint correctly with `@distilled.cloud/cloudflare@0.19.1`.

the remaining failure is after delete / emptying the entrypoint. Cloudflare returns 200 with a valid ruleset result but no `rules` key. Distilled currently decodes `getPhas` / `putPhas` as if `rules` is required, so the SDK turns that 200 into a schema parse error and Alchemy retries/times out.

opened https://github.com/alchemy-run/distilled/pull/289 to fix that at the generated SDK layer. once released, this PR can stay idiomatic and just treat missing `rules` as an empty array instead of doing raw fetch / body parsing in Alchemy.

agcty:
@sam-goodwin seems good to go now! one thing i noticed while working in both distilled and alchemy effect repos is that having two repos is slowing down agentic work quite a bit, might be worth exploring a single monorepo

sam-goodwin:
I am upstreaming the distilled change and then will merge into this PR

PR #241, opened 2026-05-05
feat(cloudflare/r2): add bucket custom domains

agcty:
Adds R2 bucket custom-domain support to the existing Cloudflare R2Bucket resource.

- accepts `domains` as a single custom domain or an array
- supports `{ zoneId }`, zone ID strings, and hostname strings for zone selection
- creates, updates, and removes only domains tracked by the bucket resource state
- recreates a custom domain when its zone changes because Cloudflare does not expose `zoneId` on the update endpoint
- adds a focused live test gated by `CLOUDFLARE_TEST_R2_DOMAIN_ZONE_ID` and `CLOUDFLARE_TEST_R2_DOMAIN`

Verification run:

- `bun vitest run packages/alchemy/test/Cloudflare/R2/CustomDomain.test.ts`
- `bun tsc -b --pretty false`
- `git diff --check`

### DIFF
--- a/packages/alchemy/src/Cloudflare/R2/R2Bucket.ts
+++ b/packages/alchemy/src/Cloudflare/R2/R2Bucket.ts
@@ -19,7 +19,7 @@ export type R2BucketCustomDomain = {
   /**
    * Custom domain name to attach to the bucket.
    */
-  domain: string;
+  name: string;
   /**
    * Zone that contains the custom domain. If omitted, the zone is inferred
    * from `domain`. Pass a zone ID string, a hostname in the zone, or any object
@@ -65,7 +65,7 @@ export type R2BucketProps = {
   /**
    * Custom domain or domains to attach to the bucket.
    */
-  domains?: R2BucketCustomDomain | R2BucketCustomDomain[];
+  domain?: R2BucketCustomDomain | R2BucketCustomDomain[];
 };
 
 export type R2Bucket = Resource<
@@ -228,9 +228,7 @@ export const R2BucketProvider = () =>
           const observedByDomain = new Map(
             observed.map((domain) => [domain.domain, domain]),
           );
-          const desiredDomains = new Set(
-            desired.map((domain) => domain.domain),
-          );
+          const desiredDomains = new Set(desired.map((domain) => domain.name));
 
           // Remove domains that are no longer desired. Domains that keep the
           // same hostname but move zones are intentionally skipped here and
@@ -261,9 +259,9 @@ export const R2BucketProvider = () =>
                 const zoneId = yield* Zone.resolveZoneId({
                   accountId,
                   zone: domain.zone,
-                  hostname: domain.domain,
+                  hostname: domain.name,
                 });
-                const observedDomain = observedByDomain.get(domain.domain);
+                const observedDomain = observedByDomain.get(domain.name);
 
                 if (
                   observedDomain &&
@@ -280,7 +278,7 @@ export const R2BucketProvider = () =>
                   yield* deleteBucketDomainCustom({
                     accountId,
                     bucketName,
-                    domain: domain.domain,
+                    domain: domain.name,
                     jurisdiction,
                   }).pipe(
                     Effect.catchIf(
@@ -295,7 +293,7 @@ export const R2BucketProvider = () =>
                     accountId,
                     bucketName,
                     jurisdiction,
-                    domain: domain.domain,
+                    domain: domain.name,
                     enabled: domain.enabled ?? true,
                     zoneId,
                     ciphers: domain.ciphers,
@@ -312,7 +310,7 @@ export const R2BucketProvider = () =>
                 const updated = yield* updateBucketDomainCustom({
                   accountId,
                   bucketName,
-                  domain: domain.domain,
+                  domain: domain.name,
                   jurisdiction,
                   enabled: domain.enabled ?? true,
                   ciphers: domain.ciphers,
@@ -361,7 +359,7 @@ export const R2BucketProvider = () =>
               stables: oldName === name ? ["bucketName"] : undefined,
             } as const;
           }
-          if (!deepEqual(olds.domains, news.domains)) {
+          if (!deepEqual(olds.domain, news.domain)) {
             return { action: "update" } as const;
           }
         }),
@@ -429,7 +427,7 @@ export const R2BucketProvider = () =>
           const domains = yield* reconcileCustomDomains(
             attrs.bucketName,
             attrs.jurisdiction,
-            normalizeDomains(news.domains),
+            normalizeDomains(news.domain),
             output?.domains ?? [],
           );
 
@@ -487,7 +485,7 @@ const r2CustomDomainConsistencySchedule = Schedule.exponential(100).pipe(
 );
 
 const normalizeDomains = (
-  domains: R2BucketProps["domains"],
+  domains: R2BucketProps["domain"],
 ): R2BucketCustomDomain[] =>
   domains === undefined ? [] : Array.isArray(domains) ? domains : [domains];
 
--- a/packages/alchemy/src/Cloudflare/R2/R2Bucket.ts
+++ b/packages/alchemy/src/Cloudflare/R2/R2Bucket.ts
@@ -1,8 +1,6 @@
 import * as r2 from "@distilled.cloud/cloudflare/r2";
-import * as zones from "@distilled.cloud/cloudflare/zones";
 import * as Effect from "effect/Effect";
-import * as Option from "effect/Option";
-import * as Stream from "effect/Stream";
+import * as Schedule from "effect/Schedule";
 
 import { deepEqual, isResolved } from "../../Diff.ts";
 import { createPhysicalName } from "../../PhysicalName.ts";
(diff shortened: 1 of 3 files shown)


agcty:
Verification update:

- Focused test passed without live env: `bun vitest run packages/alchemy/test/Cloudflare/R2/CustomDomain.test.ts` (skipped as expected).
- Live test passed against real Cloudflare resources under `sneef.xyz`: created an R2 bucket, attached a generated disposable custom domain, updated the custom domain, destroyed the stack, and verified cleanup. Cloudflare returns `NoSuchBucket` after bucket deletion when checking the custom domain, which the test now accepts as successful cleanup.

packages/alchemy/src/Cloudflare/R2/R2Bucket.ts
```diff
@@ -230,13 +500,28 @@ export const R2BucketProvider = () =>
...
+                  storageClass: bucket.storageClass ?? "Standard",
+                  jurisdiction,
+                  location: normalizeLocation(bucket.location),
+                  accountId: acct,
+                  domains: existingDomains.filter((domain) =>
+                    ownedDomains.has(domain.domain),
+                  ),
+                };
+              }),
+            ),
```
john-royal: We can probably remove this bit entirely since we don't actually read this result.

packages/alchemy/src/Cloudflare/R2/R2Bucket.ts
```diff
@@ -126,6 +236,141 @@ export const R2BucketProvider = () =>
...
+                domain: previousDomain.domain,
+                jurisdiction,
+              }).pipe(
+                Effect.catchIf(
+                  isMissingCustomDomainOrBucket,
+                  () => Effect.void,
+                ),
+              );
+            }
+          }
```
john-royal: Can we do this in parallel instead of sequentially using `Effect.forEach` with `concurrency: "unbounded"`?

packages/alchemy/src/Cloudflare/R2/R2Bucket.ts
```diff
@@ -126,6 +236,141 @@ export const R2BucketProvider = () =>
...
+
+          const applied: R2Bucket.CustomDomain[] = [];
+          for (const domain of desired) {
+            const zoneId = yield* resolveZoneId(domain.zone);
+            const observedDomain = observedByDomain.get(domain.domain);
+            if (!sameCustomDomainConfig(observedDomain, domain, zoneId)) {
+              if (observedDomain) {
+                if (observedDomain.zoneId !== zoneId) {
```
john-royal: Same thing here. I'd change to `const applied = yield* Effect.forEach(..., { concurrency: "unbounded" })` so these run in parallel instead of sequentially.

packages/alchemy/src/Cloudflare/R2/R2Bucket.ts
```diff
@@ -126,6 +236,141 @@ export const R2BucketProvider = () =>
...
+
+            const refreshed = yield* r2
+              .getBucketDomainCustom({
+                accountId,
+                bucketName,
+                domain: domain.domain,
+                jurisdiction,
+              })
+              .pipe(Effect.map(toCustomDomainAttributes));
+            applied.push(refreshed);
```
john-royal: Do the `createBucketDomainCustom` and `updateBucketDomainCustom` functions return the attributes we're looking for? If yes, we can optimize by relying on those results instead of making another API call here.

packages/alchemy/src/Cloudflare/R2/R2Bucket.ts
```diff
@@ -98,8 +137,75 @@ export declare namespace R2Bucket {
...
+  observed.minTLS === desired.minTLS;
+
+const isMissingCustomDomainOrBucket = (error: unknown): boolean =>
+  typeof error === "object" &&
+  error !== null &&
+  (("status" in error && (error as { status: unknown }).status === 404) ||
+    ("_tag" in error &&
+      ((error as { _tag: unknown })._tag === "DomainNotFound" ||
+        (error as { _tag: unknown })._tag === "NoSuchBucket")));
+
```
john-royal: Nit: please move these to the bottom of the file (helps with being able to read the file from top to bottom)

john-royal (changes requested):
Just some minor changes here. Once you make those, should be good to go.

packages/alchemy/src/Cloudflare/R2/R2Bucket.ts
```diff
@@ -11,6 +14,37 @@ import { R2BucketBinding } from "./R2BucketBinding.ts";
...
+  /**
+   * Custom domain name to attach to the bucket.
+   */
+  domain: string;
+  /**
+   * Zone that contains the custom domain. Pass a zone ID string, a hostname in
+   * the zone, or any object with a `zoneId` attribute such as
+   * `Cloudflare.Zone`.
+   */
+  zone: R2BucketCustomDomainZone;
```
sam-goodwin: Is this required? If the user provides `domain`, can't we infer it?

packages/alchemy/src/Cloudflare/R2/R2Bucket.ts
```diff
@@ -126,6 +194,140 @@ export const R2BucketProvider = () =>
...
+            Stream.runHead,
+            Effect.map(Option.getOrUndefined),
+          );
+          if (!match) {
+            return yield* Effect.fail(
+              new Error(`Cloudflare zone not found for ${zone}`),
+            );
+          }
+          return match.id;
+        });
```
sam-goodwin: This should be a shared utility. Lots of places will likely need this. Put it in `Zone.ts` (create if it does not exist)

packages/alchemy/src/Cloudflare/R2/R2Bucket.ts
```diff
@@ -126,6 +194,140 @@ export const R2BucketProvider = () =>
...
+      ) =>
+        listBucketDomainCustoms({
+          accountId,
+          bucketName,
+          jurisdiction,
+        }).pipe(
+          Effect.map((response) =>
+            response.domains.map(toCustomDomainAttributes),
+          ),
+          Effect.catchTag("NoSuchBucket", () => Effect.succeed([])),
```
sam-goodwin: `undefined` might be a better return type as a way to indicate to the callers whether it existed or not? What should we do if the bucket does not exist, do we not want to retry or something?

packages/alchemy/src/Cloudflare/R2/R2Bucket.ts
```diff
@@ -126,6 +194,140 @@ export const R2BucketProvider = () =>
...
+                    domain: previousDomain.domain,
+                    jurisdiction,
+                  }).pipe(
+                    Effect.catchIf(
+                      isMissingCustomDomainOrBucket,
+                      () => Effect.void,
+                    ),
+                  ),
+            { concurrency: "unbounded" },
+          );
```
sam-goodwin, line 305: Why are we deleting twice? We delete here and again in the `applied` function below

packages/alchemy/src/Cloudflare/R2/R2Bucket.ts
```diff
@@ -126,6 +194,140 @@ export const R2BucketProvider = () =>
...
+                    accountId,
+                    bucketName,
+                    domain: domain.domain,
+                    jurisdiction,
+                  }).pipe(
+                    Effect.catchIf(
+                      isMissingCustomDomainOrBucket,
+                      () => Effect.void,
+                    ),
+                  );
```
sam-goodwin, line 340: More deleting? Have the coding agent justify these decisions with a `//` comment to help explain the flow.

I also don't see much retry logic going on. Are there any confict/not found eventual consistency edge cases to consider?

agcty: @sam-goodwin will do in a sec!

agcty: pushed `e13bb37d` for this. the flow now has `//` comments explaining the stale-domain prune vs the zone-change delete/recreate path, and custom-domain list/create/update retries the narrow `NoSuchBucket` consistency lag after bucket creation. i also made `zone` optional by inferring it from `domain`, moved zone resolution into shared `Cloudflare/Zone.ts`, and changed the list helper to return `undefined` for a missing bucket instead of silently treating it as an empty domain list. checks passed: `bun tsc -b --pretty false`; focused R2 custom-domain test also passed live against the real zone.

packages/alchemy/src/Cloudflare/Zone.ts
```diff
@@ -0,0 +1,44 @@
...
+  accountId: string;
+  zone: ZoneReference | undefined;
+  hostname: string;
+}) =>
+  Effect.gen(function* () {
+    if (typeof zone === "object") return zone.zoneId;
+    if (typeof zone === "string" && isZoneId(zone)) return zone;
+
+    const lookup = zone ?? hostname;
+    const match = yield* zones.listZones.items({}).pipe(
```
sam-goodwin: anyway to filter?

agcty: yes, pushed `03cb8fd7` to use the Cloudflare `/zones` server-side filters instead of streaming every zone. `resolveZoneId` now tries exact suffix candidates (`foo.example.com`, then `example.com`) with `name`, `account.id`, and `per_page=1`. checks passed: `bun tsc -b --pretty false`; live R2 custom-domain test passed against the real zone.

packages/alchemy/src/Cloudflare/Zone.ts
```diff
@@ -26,19 +27,62 @@ export const resolveZoneId = ({
...
-      Effect.map(Option.getOrUndefined),
+    for (const candidate of zoneNameCandidates(lookup)) {
+      const match = yield* findZoneByName({ accountId, name: candidate });
+      if (match) return match.id;
+    }
+    return yield* Effect.fail(
+      new Error(`Cloudflare zone not found for ${lookup}`),
     );
-    if (!match) {
+  });
```
sam-goodwin: No, this was right before. Use the .items() stream stuff. i was just asking if `.items({ ?? })` accepts a query filter so we don't have to scan. Using Stream to scan was right before

sam-goodwin: Oh sorry, i see what you're doing. Hmm

sam-goodwin: You're right, we can just go straight to it. Much better

PR #242, opened 2026-05-05
fix(cloudflare): enable hyperdrive in dev mode

john-royal:


packages/alchemy/src/Cloudflare/Hyperdrive/Hyperdrive.ts
```diff
@@ -66,9 +65,16 @@ export type HyperdriveMtls = {
   caCertificateId?: string;
   mtlsCertificateId?: string;
   /**
-   * @default "verify-full"
+   * @default "require"
    */
-  sslmode?: string;
+  sslmode?: "require" | "verify-ca" | "verify-full";
```
john-royal, line 70: The default is `require` per CF docs: https://developers.cloudflare.com/hyperdrive/configuration/tls-ssl-certificates-for-hyperdrive/#server-certificates-tlsssl-modes

examples/cloudflare-dev/src/EffectWorker.ts
```diff
@@ -11,11 +36,43 @@ export default class EffectWorker extends Cloudflare.Worker<EffectWorker>()(
   },
   Effect.gen(function* () {
     const kv = yield* Cloudflare.KVNamespace.bind(KV);
+    const hyperdrive = yield* Cloudflare.Hyperdrive.bind(Hyperdrive);
+
+    const pgLayer = hyperdrive.connectionString.pipe(
+      Effect.map((connectionString) =>
+        PgClient.layer({
+          url: Redacted.make(connectionString),
```
john-royal: Should our connection string stay redacted so you don't have to call `Redacted.make`?

examples/cloudflare-dev/src/EffectWorker.ts
```diff
@@ -11,11 +36,43 @@ export default class EffectWorker extends Cloudflare.Worker<EffectWorker>()(
...
+            return HttpServerResponse.text("Not found", { status: 404 });
+        }
+      }).pipe(Effect.provide(pgLayer), Effect.orDie),
     };
-  }).pipe(Effect.provide(Cloudflare.KVNamespaceBindingLive)),
+  }).pipe(
+    Effect.provide(
+      Layer.merge(
+        Cloudflare.KVNamespaceBindingLive,
+        Cloudflare.HyperdriveConnectionLive, // should this be called HyperdriveBindingLive for consistency?
```
john-royal: Should we rename this to `HyperdriveBindingLive`? Found the inconsistency a little confusing.

examples/cloudflare-dev/src/EffectWorker.ts
```diff
@@ -1,8 +1,33 @@
...
+export const Hyperdrive = Cloudflare.Hyperdrive("Hyperdrive", {
+  // should this support a raw connection string?
+  origin: {
+    // TODO: provision with Neon
+    scheme: "postgresql",
+    host: process.env.PGHOST!,
+    database: process.env.PGDATABASE!,
+    user: process.env.PGUSER!,
+    password: Redacted.make(process.env.PGPASSWORD!),
+  },
```
john-royal: I kinda miss being able to pass in a raw connection string here, though I understand the idea is that you provision your database through Alchemy so you don't have to specify all these.

packages/alchemy/src/Cloudflare/Hyperdrive/HyperdriveBinding.ts
```diff
@@ -40,7 +25,20 @@ export const HyperdriveBinding = Effect.gen(function* () {
...
+                scheme: dev.scheme,
+                host: dev.host,
+                port: dev.port ?? defaultPort(dev.scheme),
+                user: dev.user,
+                database: dev.database,
+                password: dev.password,
+                sslmode: dev.sslmode ?? "prefer",
+              },
+            }))
+          : undefined,
```
john-royal, line 159: Should we just use the production origin if a dev one isn't specified?

packages/alchemy/src/Drizzle/Postgres.ts
```diff
@@ -53,6 +53,9 @@ export const postgres = <E, R>(connectionString: Effect.Effect<string, E, R>) =>
     const cached = yield* Effect.cached(
       Effect.gen(function* () {
         const url = yield* connectionString;
+        // TODO: after a few requests to a Cloudflare Worker, this causes requests to fail with:
+        // "The Workers runtime canceled this request because it detected that your Worker's code had hung and would never generate a response. Refer to: https://developers.cloudflare.com/workers/observability/errors/"
+        // We probably need a better solution for request-scoped stuff in runtimes that require it.
```
john-royal: This was a problem while testing dev mode because you can't share connections between requests in a Worker: https://developers.cloudflare.com/hyperdrive/concepts/connection-lifecycle/#cleaning-up-client-connections

@sam-goodwin Do you have any ideas for how we can scope this to the request in runtimes that require this? I love the DX of this function but the errors were a problem.

examples/cloudflare-dev/src/EffectWorker.ts
sam-goodwin: Yes, that makes sense

examples/cloudflare-dev/src/EffectWorker.ts
Mkassabov: Is there a reason we don't support both?

packages/alchemy/src/Cloudflare/Hyperdrive/HyperdriveBinding.ts
```diff
@@ -1,46 +1,163 @@
...
+        ],
+        hyperdrives: ctx.dev
+          ? (Output.map(
+              Output.all(hyperdrive.hyperdriveId, Output.asOutput(origin)),
+              ([id, origin]) => ({
+                [id]: origin,
+              }),
+            ) as unknown as Record<string, Required<HyperdriveDevOrigin>>)
+          : undefined,
+      });
```
john-royal, line 160: @sam-goodwin Not sure if these are intentional or if they're bugs in core, so want to call this out:
1. The `as unknown as` here fixes a type error - should the types for `.bind` be updated to accept outputs?
2. Here, `origin` fails to resolve unless I wrap it in `Output.asOutput` - that definitely seems like a bug.

PR #250, opened 2026-05-07
fix(cli): correct args for dev command in node

john-royal:


packages/alchemy/src/Cli/commands/dev.ts
```diff
@@ -32,14 +32,13 @@ export const devCommand = Command.make(
...
       bin,
-      [
-        "run",
-        "--watch",
-        "--no-clear-screen",
-        fileURLToPath(import.meta.resolve(execEntry)),
-      ],
+      {
+        bun: ["run", "--watch", "--no-clear-screen", main],
+        node: ["--watch", main],
```
sam-goodwin: Is there an alternative to --no-clear-screen on node?

sam-goodwin: Do we need to detect node version and provide proper args to erase types?

john-royal: For `--no-clear-screen`: Yes, just added it, it's `--watch-preserve-output`.

Type stripping is enabled in Node 22.18 and later. However this only handles non-erasable types, so I added `--experimental-transform-types` (available in Node 22.7 and later).

Native TypeScript handling in Node doesn't honor your `tsconfig.json`. It's kinda dumb. One of our users seems to be using `tsx` anyway.

packages/alchemy/src/Cli/commands/dev.ts
```diff
@@ -32,14 +32,19 @@ export const devCommand = Command.make(
...
-        "run",
-        "--watch",
-        "--no-clear-screen",
-        fileURLToPath(import.meta.resolve(execEntry)),
-      ],
+      {
+        bun: ["run", ...process.execArgv, "--watch", "--no-clear-screen", main],
+        node: [
+          ...process.execArgv,
+          "--experimental-transform-types",
```
sam-goodwin, line 42: I recall not wanting to provide this for node 24 or even node 22 after some version? Alchemy v1 has something for this. People might get warnings if we always provide it?

There's also some gotcha with enums? There's an option for how transform enums, do we need to consider that?

PR #254, opened 2026-05-07
test(harden): stabilize live suite — bundle resolve, S3 SlowDown retry, SQS post-create race

sam-goodwin:
Stabilizes the live test suite — 3 consecutive clean runs (673 passed / 42 skipped) after these fixes. All but one of these failure modes show up only under parallel-suite load.

### Bundle: resolve `alchemy/Stack` (and friends) without requiring `lib/`

Rolldown was picking the `import` condition in `alchemy`'s `package.json#exports`, which points at `./lib/*.js`. On a fresh checkout `lib/` doesn't exist; rolldown emits `[UNRESOLVED_IMPORT]`, treats the path as external, and the Lambda crashes on init. Surfaces as opaque "Function URL returned 502" timeouts across SQS/SNS/DynamoDB/Kinesis Bindings tests and the Lambda Function test.

```ts
const SOURCE_CONDITIONS = ["bun", "workerd", "import", "default"];
```

Resolves straight to source `.ts`. Rolldown handles TS natively and every package exposing a `bun` source condition also publishes `src/`.

### S3 SlowDown — patched in distilled

S3 503 `SlowDown` was previously surfaced as `UnknownAwsError`, which the blanket `awsRetryFactory` in `Providers.ts` couldn't recognize. Patched in distilled to classify it as `ThrottlingError + RetryableError`: alchemy-run/distilled#270. Once that lands, the existing 10-attempt jittered-exponential retry handles it transparently — no alchemy-side change needed.

### SQS Queue reconcile: tolerate post-create eventual consistency

```ts
sqs.getQueueAttributes({ QueueUrl: queueUrl, AttributeNames: ["All"] }).pipe(
  Effect.retry({
    while: (e) => e._tag === "QueueDoesNotExist",
    schedule: Schedule.fixed(500).pipe(Schedule.both(Schedule.recurs(60))),
  }),
)
```

Same-tick `GetQueueAttributes` after `CreateQueue` can return `QueueDoesNotExist` for a few hundred ms.

### DynamoDB Bindings test: bump readiness budget

75 -> 100 retries (150s -> 200s). The Bindings handler's `/scan` probe was returning 500 the entire 150s window under parallel-suite load while IAM scan policy propagated. Still inside the `beforeAll` 240s timeout.

### Hyperdrive test: skip pending real Postgres origin

Cloudflare validates not just DNS but TCP-level connectivity at create time, so any placeholder host (`db.example.com`, `example.com:5432`, etc.) is rejected. Skipped both cases with a `TODO(hyperdrive)` to provision a Neon project as the fixture.


packages/alchemy/src/AWS/Assets.ts
```diff
@@ -5,10 +5,30 @@ import * as Context from "effect/Context";
...
+ * they're not part of the official PutObject error model. They are always
+ * safe to retry with backoff — the request never reached S3's storage
+ * tier — and the same alchemy stack uploading lots of Lambda code in a
+ * burst will trip this on a busy account.
+ */
+const isS3SlowDown = (error: unknown): boolean => {
+  if (typeof error !== "object" || error === null) return false;
+  const e = error as { _tag?: string; errorTag?: string };
+  return e._tag === "UnknownAwsError" && e.errorTag === "SlowDown";
+};
```
sam-goodwin: this looks like something we need to patch in distilled for s3

sam-goodwin: Patched in distilled — alchemy-run/distilled#270. Reverted the alchemy-side workaround in 90cf3d8a; once distilled releases, the existing blanket awsRetryFactory will retry SlowDown via the new ThrottlingError + RetryableError categories.

PR #269, opened 2026-05-07
feat(cloudflare): worker support non-string env bindings

Mkassabov:
Cloudflare Workers support [non-string env bindings](https://developers.cloudflare.com/workers/configuration/environment-variables/) (number, boolean, array, plain object) via the `json` binding kind. Widen `Worker.env` to accept those values and emit them as `type: "json"` bindings.

```ts
yield* Worker("api", {
  main,
  env: {
    NAME: "alchemy",        // plain_text
    SECRET: Redacted.make("..."),  // secret_text
    MAX_RETRIES: 3,         // json
    ENABLE_CACHE: true,     // json
    FEATURES: ["a", "b"],   // json
    CONFIG: { foo: "bar" }, // json
  },
});
```

Applies to both the remote put-script path and the local dev sidecar. `StaticSite` JSON-stringifies non-string values when forwarding `env` to its build `Command` (child-process env vars must be strings).

### DIFF
--- a/packages/alchemy/src/Cloudflare/Workers/Worker.ts
+++ b/packages/alchemy/src/Cloudflare/Workers/Worker.ts
@@ -1657,7 +1657,7 @@ export const LiveWorkerProvider = () =>
               metadataBindings.push({
                 type: "json",
                 name: key,
-                json: JSON.stringify(value),
+                json: value,
               });
             }
           }
--- a/packages/alchemy/src/Cloudflare/Workers/Worker.ts
+++ b/packages/alchemy/src/Cloudflare/Workers/Worker.ts
@@ -260,16 +260,6 @@ export interface WorkerProps<
   };
   limits?: WorkerLimits;
   placement?: WorkerPlacement;
-  /**
-   * Environment variables exposed on the Worker's `env` object.
-   *
-   * - `string` values become `plain_text` bindings.
-   * - `Redacted<string>` values become `secret_text` bindings.
-   * - Any other JSON value (`number`, `boolean`, plain object, array, `null`)
-   *   becomes a `json` binding and is delivered to the Worker as the parsed
-   *   JS value rather than a string. See
-   *   https://developers.cloudflare.com/workers/configuration/environment-variables/
-   */
   env?: Record<
     string,
     | string
--- a/packages/alchemy/src/Cloudflare/Workers/Worker.ts
+++ b/packages/alchemy/src/Cloudflare/Workers/Worker.ts
@@ -260,7 +260,26 @@ export interface WorkerProps<
   };
   limits?: WorkerLimits;
   placement?: WorkerPlacement;
-  env?: Record<string, string | Redacted.Redacted<string>>;
+  /**
+   * Environment variables exposed on the Worker's `env` object.
+   *
+   * - `string` values become `plain_text` bindings.
+   * - `Redacted<string>` values become `secret_text` bindings.
+   * - Any other JSON value (`number`, `boolean`, plain object, array, `null`)
+   *   becomes a `json` binding and is delivered to the Worker as the parsed
+   *   JS value rather than a string. See
+   *   https://developers.cloudflare.com/workers/configuration/environment-variables/
+   */
+  env?: Record<
+    string,
+    | string
+    | number
+    | boolean
+    | null
+    | readonly unknown[]
+    | { readonly [key: string]: unknown }
+    | Redacted.Redacted<string>
+  >;
   exports?: string[];
   bindings?: Bindings;
   /**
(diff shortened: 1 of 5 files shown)


Mkassabov:
does `Redacted.make` work for non-string values?
What if I do
```
CONFIG: { foo: Redacted.make("bar") },
```

Mkassabov (commented):
can you add tests?

Mkassabov:
Nested `Redacted` inside a JSON value isn't supported — only top-level `Redacted` becomes a `secret_text` binding. Anything else gets `JSON.stringify`'d, and `Redacted` deliberately serializes to `"<redacted>"` (its `toJSON` hides the value), so a nested secret would arrive in the worker as the literal string `"<redacted>"` rather than the underlying value.

If you need a secret, lift it to a top-level key. Could add a runtime warning if we want to be louder about it, but I'd rather not silently support a pattern that ends up shipping non-secret values via `secret_text` once you start mixing.

\- claude

Mkassabov:
Added in 6876465 — deploys a worker with `string`, `number`, `boolean`, `object`, `array`, and `Redacted<string>` env values, then asserts the runtime echoes them as typed JS values (e.g. `"NUM":42` not `"NUM":"42"`, `typeof env.OBJ === "object"`, etc.).

\- claude

packages/alchemy/test/Cloudflare/Workers/Worker.test.ts
```diff
@@ -712,4 +713,68 @@ describe.concurrent("Cloudflare.Worker", () => {
...
+      );
+      yield* expectUrlContains(worker.url!, '"STR":"hello"', {
+        timeout: "60 seconds",
+        label: "string env binding",
+      });
+      yield* expectUrlContains(
+        worker.url!,
+        '"NUM":"number","BOOL":"boolean","OBJ":"object","ARR":"array","STR":"string"',
+        { timeout: "60 seconds", label: "runtime typeof matches" },
+      );
```
sam-goodwin: nit: thisis complicated. why not just fetch and expect.toMatchObject or something?

Mkassabov: Yeah fair, fixed in 08f601a — single `expectUrlContains` to wait for propagation (returns the body), then `JSON.parse` + `toMatchObject` for the assertion.

\- claude

packages/alchemy/src/Cloudflare/Workers/Worker.ts
```diff
@@ -264,7 +264,26 @@ export interface WorkerProps<
...
+  /**
+   * Environment variables exposed on the Worker's `env` object.
+   *
+   * - `string` values become `plain_text` bindings.
+   * - `Redacted<string>` values become `secret_text` bindings.
+   * - Any other JSON value (`number`, `boolean`, plain object, array, `null`)
+   *   becomes a `json` binding and is delivered to the Worker as the parsed
+   *   JS value rather than a string. See
+   *   https://developers.cloudflare.com/workers/configuration/environment-variables/
+   */
```
Mkassabov: remove this comment please

Mkassabov: Done in 6ab6c79.

\- claude

Mkassabov:
moving to draft cause this test doesn'tpass. claude is fixing it

Mkassabov:
Dug into this. The failure isn't a flake — Cloudflare's `json` binding wants the *raw JSON value* embedded in the metadata struct (wrangler does `json: { foo: "bar" }`, not `json: "{\"foo\":\"bar\"}"`), and Cloudflare doesn't auto-parse a string `json` field on the way out.

The `@distilled.cloud/cloudflare` SDK enforces `json: Schema.String` though, so `Schema.encodeSync` rejects non-string values before the request leaves the client. Bypassing that means rebuilding the multipart `putScript` call from scratch — out of scope for this PR.

So the honest behaviour: alchemy widens the `env` type, JSON-stringifies non-string values, and ships them as `type: "json"` bindings. They arrive in `env` as JSON-encoded strings; consumers `JSON.parse(env.X)`. Updated the worker fixture and assertion in b48ab35 to match. The "transparent typed values" feature can land once the SDK schema is fixed upstream.

\- claude

Mkassabov:
Fixed properly in 6c31f0a — bumped `@distilled.cloud/cloudflare` to the pkg.ing build of [alchemy-run/distilled#277](https://github.com/alchemy-run/distilled/pull/277) which widens the `json` binding value to `unknown`. Now alchemy sends the raw value, Cloudflare parses it on the runtime side, and `env.X` arrives as the typed JS value (number/boolean/object/array) — no `JSON.parse` needed.

Live test passes:
```
Test Files  1 passed (1)
     Tests  1 passed | 11 skipped (12)
```

The catalog dep is pinned to the commit URL; needs to be reverted to a real semver once the distilled PR lands and a release ships.

\- claude

Mkassabov:
[alchemy-run/distilled#277](https://github.com/alchemy-run/distilled/pull/277) is in the merge queue. Once it lands and the next `@distilled.cloud/cloudflare` release ships, I'll flip the catalog dep here from the pkg.ing commit URL back to a real semver and this is ready to merge.

\- claude

package.json
```diff
@@ -69,7 +69,7 @@
       "@distilled.cloud/cloudflare-rolldown-plugin": "0.3.0",
       "@distilled.cloud/cloudflare-runtime": "0.2.1",
       "@distilled.cloud/cloudflare-vite-plugin": "0.2.0",
-      "@distilled.cloud/cloudflare": "^0.17.0",
+      "@distilled.cloud/cloudflare": "https://pkg.distilled.cloud/cloudflare/4878164",
```
sam-goodwin: revert

PR #272, opened 2026-05-07
fix(cloudflare): cloudflare vite dev bindings

nr1brolyfan:
- Simplifies Cloudflare.Vite usage with TanStack Start
- Auto-injects the Cloudflare Vite plugin in dev mode
- Makes Alchemy bindings work without a manual wrangler.jsonc
- Fixes local dev rootDir resolution
- Supports main: import.meta.url
- Scopes ALCHEMY_BUILD to builds started by Alchemy

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -298,7 +298,7 @@
     },
     "packages/alchemy": {
       "name": "alchemy",
-      "version": "2.0.0-beta.33",
+      "version": "2.0.0-beta.34",
       "bin": {
         "alchemy": "./bin/cli.js",
       },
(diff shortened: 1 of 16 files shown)


Mkassabov:
@nr1brolyfan We've made a lot of changes to the way dev vite works so I think we're going to close this.

Can you clarify more about what your problems were (and if they still exist on latest) and what your expected solution was?

nr1brolyfan:
> @nr1brolyfan We've made a lot of changes to the way dev vite works so I think we're going to close this.
> 
> Can you clarify more about what your problems were (and if they still exist on latest) and what your expected solution was?

I'm closing this now as it's outdated and/or being worked on. This was only a temporary fix.

PR #273, opened 2026-05-07
fix: dev mode node compatibility and profiles

john-royal:
Apologies for the sprawling change - I kept running the `cloudflare-dev` example and kept changing things until it worked in both Node and Bun.

## Changes
- Add missing `.ts` extensions for imports in the CLI.
- **Avoid forcing Bun/TypeScript when running locally because this prevents us from testing with Node.** Bun/TypeScript is still used when Bun is the invoker.
- Use the `.js` extension to invoke the sidecar when the invoker is using the `.js` file.
- Update the bundler to handle `import.meta.url` as the file extension because `import.meta.url` is undefined in Node.
- Add a Node-compatible sidecar RPC implementation using the `ws` package. The implementation is swapped automatically with the Bun one using a layer.
- Avoid setting `detached: true` for the sidecar because this messes things up in Windows.
- Corrected the `ConfigProvider` implementation for the sidecar:
  - Pass through the selected env file because Node doesn't read `.env` files automatically like Bun.
  - Pass through the selected profile.

## Notes
`LocalWorkerProvider` and `LiveWorkerProvider` were using different, copy/pasted implementations for Rolldown bundling, compatibility dates, and name generation. This came up when updating the bundler to support `import.meta.url`. To fix it, I:
- Updated `WorkerBundle.ts` with the missing features from `LiveWorkerProvider`.
- Updated `LiveWorkerProvider` to use the shared implementations of `createWorkerName`, `getCompatibility`, and `WorkerBundle`.


examples/cloudflare-dev/src/EffectWorker.ts
```diff
@@ -7,7 +7,7 @@ export const KV = Cloudflare.KVNamespace("KV");
 export default class EffectWorker extends Cloudflare.Worker<EffectWorker>()(
   "EffectWorker",
   {
-    main: import.meta.path,
+    main: import.meta.url,
```
sam-goodwin: Is this because it's not supported in node?

examples/cloudflare-dev/alchemy.run.ts
```diff
@@ -24,7 +24,7 @@ export default Alchemy.Stack(
   "CloudflareDev",
   {
     providers: Cloudflare.providers(),
-    state: Cloudflare.state(),
+    state: Alchemy.localState(),
```
sam-goodwin: Is this intentional?

packages/alchemy/src/Cli/exec.ts
```diff
@@ -4,17 +4,13 @@ import * as Layer from "effect/Layer";
...
-import { ProfileLive } from "../Auth/Profile.ts";
+import { AlchemyContextLive } from "../AlchemyContext.ts";
 import { PlatformServices } from "../Util/PlatformServices.ts";
 import { execStack, ExecStackOptions } from "./commands/deploy.ts";
 import { selectCli } from "./selectCli.ts";
 
 const services = Layer.mergeAll(
   Layer.provideMerge(AlchemyContextLive, PlatformServices),
-  Layer.provide(ProfileLive, PlatformServices),
-  Layer.provide(CredentialsStoreLive, PlatformServices),
```
sam-goodwin, line 17: regression? these were added by a recent PR i did to put the auth provider stuff in layers

packages/alchemy/src/Cloudflare/Workers/WorkerBundle.ts
```diff
@@ -85,22 +82,51 @@ export const WorkerBundle = Effect.gen(function* () {
...
+    }).pipe(
+      Effect.flatMap((path) => fs.realPath(path)),
+      Effect.mapError(
+        (cause) =>
+          new Bundle.BundleError({
+            message: `Failed to find real path for bundle: ${main}`,
+            cause,
+          }),
+      ),
+    );
```
sam-goodwin, line 107: Lambda and every other platform resource will need this

packages/alchemy/src/Sidecar/RpcPaths.ts
```diff
@@ -4,6 +4,7 @@ import * as FileSystem from "effect/FileSystem";
 import * as Hash from "effect/Hash";
 import * as Layer from "effect/Layer";
 import * as Path from "effect/Path";
+import { fileURLToPath } from "node:url";
```
sam-goodwin, line 7: node effect platform have this? Not that it matters really

packages/alchemy/src/Sidecar/RpcPaths.ts
```diff
@@ -27,3 +28,11 @@ export const layer = (main: string) =>
       });
     }),
   );
+
+export const sanitizeMain = (url: string) => {
+  let main = fileURLToPath(url);
+  if (import.meta.url.endsWith(".js")) {
+    main = main.replace(".ts", ".js");
+  }
+  return main;
```
sam-goodwin: nit: use pure syntax
```ts
  const main = fileURLToPath(url);
  if (import.meta.url.endsWith(".js")) {
    return main.replace(".ts", ".js");
  }
  return main;
```

packages/alchemy/package.json
```diff
@@ -345,7 +347,8 @@
     "drizzle-kit": ">=1.0.0-rc.1",
     "drizzle-orm": ">=1.0.0-rc.1",
     "effect": "catalog:",
-    "vite": "catalog:"
+    "vite": "catalog:",
+    "ws": "^8.20.0"
```
sam-goodwin: put in the catalog

examples/cloudflare-dev/src/EffectWorker.ts
john-royal: Yes

examples/cloudflare-dev/alchemy.run.ts
sam-goodwin: why change?

packages/alchemy/src/Sidecar/RpcPaths.ts
john-royal: Nope

packages/alchemy/src/Sidecar/RpcPaths.ts
john-royal: Done

examples/cloudflare-dev/src/EffectWorker.ts
john-royal: Switched to `.filename`

sam-goodwin (approved):

Issue #274, opened 2026-05-07
Better-auth + Drizzle.postgres integration


heyanay2105:
Trying to use better-auth's drizzle adapter with `Drizzle.postgres(hd.connectionString)` from a CF Worker, mirroring the existing [CloudflareD1 better-auth layer](https://github.com/alchemy-run/alchemy-effect/blob/38319fa09e00297bc1921a4434f3efa6003fbe19/packages/better-auth/src/CloudflareD1.ts).

[The adapter](https://github.com/better-auth/better-auth/blob/main/packages/drizzle-adapter/src/drizzle-adapter.ts) expects a plain promise-based drizzle client — it calls `await db.select().from(table)` and expects rows. `Drizzle.postgres` returns an Effect-style proxy that resolves on `yield*`, not `await`, so better-auth gets the proxy object back instead of rows and crashes with `b.map is not a function`.

Would be great to have a way to get a promise-based drizzle client backed by the same underlying connection.

PR #280, opened 2026-05-08
feat(drizzle): expose vanilla node-postgres client on Drizzle.postgres

heyanay2105:
Refs #274 — building block for a `BetterAuth.DrizzlePostgres` layer (separate PR).

Adds `Drizzle.postgresRaw(...)` — a vanilla `drizzle-orm/node-postgres` client for libraries that need a regular promise-shaped drizzle (e.g. better-auth's `drizzleAdapter`).

```ts
const raw = yield* Drizzle.postgresRaw(hd.connectionString);

const auth = betterAuth({
  database: drizzleAdapter(raw, { provider: "pg", schema }),
});
```

Lives in its own module so apps using only the Effect-flavored `Drizzle.postgres` don't pull `drizzle-orm/node-postgres` into the bundle. The drizzle client is cached per-`ExecutionContext` — the pool is built at most once per request.

### DIFF
--- a/packages/alchemy/src/Drizzle/Postgres.ts
+++ b/packages/alchemy/src/Drizzle/Postgres.ts
@@ -2,6 +2,10 @@ import * as PgClient from "@effect/sql-pg/PgClient";
 import type { AnyRelations, EmptyRelations } from "drizzle-orm";
 import type { EffectPgDatabase } from "drizzle-orm/effect-postgres";
 import * as PgDrizzle from "drizzle-orm/effect-postgres";
+import {
+  drizzle as nodePgDrizzle,
+  type NodePgDatabase,
+} from "drizzle-orm/node-postgres";
 import * as Effect from "effect/Effect";
 import * as Layer from "effect/Layer";
 import * as Redacted from "effect/Redacted";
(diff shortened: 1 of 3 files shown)


packages/alchemy/src/Drizzle/Postgres.ts
```diff
@@ -51,28 +68,49 @@ export const postgres = <
...
+
+    return new Proxy(effectDb as any, {
+      get(target, prop, receiver) {
+        if (prop === "raw") return raw;
+        return Reflect.get(target, prop, receiver);
+      },
+    }) as EffectPgDatabase<TRelations> & {
+      $client: PgClient.PgClient;
+      raw: Effect.Effect<NodePgDatabase<TRelations>, E, R | ExecutionContext>;
+    };
```
sam-goodwin: Why is it not using `proxyChain` helper anymore?

heyanay: Restructured — `.raw` moved out of `Drizzle.postgres` entirely and into a new `Drizzle.postgresRaw` export, per the tree-shaking feedback on https://github.com/alchemy-run/alchemy-effect/pull/281#discussion_r3221226852.

sam-goodwin:
Please open an issue if the  current sql support is not enough. This PR got stale (sorry about that). Would like to start fresh.

PR #281, opened 2026-05-08
feat(better-auth): add DrizzlePostgres layer

### LINKED ISSUE #274: Better-auth + Drizzle.postgres integration

heyanay2105:
Closes #274. Stacked on #280.

Adds `drizzlePostgresFromHyperdrive(...)` and `drizzlePostgresFromUrl(...)` — Layer factories that wire `better-auth` to a Postgres database via `drizzle-orm`. Mirrors `CloudflareD1`.

```ts
import { BetterAuth, drizzlePostgresFromHyperdrive } from "@alchemy.run/better-auth";

.pipe(
  Effect.provide(
    drizzlePostgresFromHyperdrive({
      hyperdrive: MyHyperdrive,
      secret: process.env.BETTER_AUTH_SECRET!,
      baseURL: "https://app.example.com",
      basePath: "/api/auth",
      schema: { user, session, account, verification },
      plugins: [emailOTP({ ... })],
    }),
  ),
)
```

Then in routes:

```ts
const auth = yield* BetterAuth;
if (url.pathname.startsWith("/api/auth/")) return yield* auth.fetch;
```

For non-Hyperdrive Postgres (Neon direct, Supabase direct, env URL):

```ts
drizzlePostgresFromUrl({
  url: process.env.DATABASE_URL!,
  secret: process.env.BETTER_AUTH_SECRET!,
  schema: { user, session, account, verification },
})
```

### Options

Accepts the full `BetterAuthOptions` (minus `database`, supplied internally via `Drizzle.postgresRaw(...)` from #280) plus a flat pass-through of `drizzleAdapter` config — `schema`, `usePlural`, `debugLogs`, `camelCase`, `transaction`. All plugins, hooks, advanced cookie config, rate limit, etc. work as you'd expect.

### Per-request init (no `Effect.cached` on auth)
Unlike `CloudflareD1`, auth is rebuilt per request — Cloudflare Worker TCP sockets can't be reused across requests. The `pg` pool is cached per-`ExecutionContext` via `Drizzle.postgresRaw` and the auth wrapper matches that lifecycle.


packages/better-auth/src/DrizzlePostgres.ts
```diff
@@ -0,0 +1,170 @@
...
+ *     DrizzlePostgres.fromUrl({
+ *       url: process.env.DATABASE_URL!,
+ *       secret: process.env.BETTER_AUTH_SECRET!,
+ *       schema: { user, session, account, verification },
+ *     }),
+ *   ),
+ * )
+ * ```
+ */
+export const DrizzlePostgres = {
```
sam-goodwin: Don't export objects like this. Just `èxport const fromHyperdrive =` from the module. Better for tree-shaking

packages/alchemy/src/Drizzle/Postgres.ts
```diff
@@ -51,28 +68,49 @@ export const postgres = <
...
       >,
     );
+
+    const raw = Effect.gen(function* () {
+      const ctx = yield* ExecutionContext;
+      return yield* (ctx.cache[rawSymbol] ??= yield* Effect.gen(function* () {
+        const url = Redacted.value(yield* connectionString);
+        return nodePgDrizzle<TRelations>(url);
+      }).pipe(Effect.cached));
+    }) as Effect.Effect<NodePgDatabase<TRelations>, E, R | ExecutionContext>;
```
sam-goodwin: I think for tree-shaking reasons, we should probably make this a different binding.

`Drizzle.postgresRaw`.

Effect's PgClient.layer doesn't go through this import, so we don't want to bloat the bundle by having them all in the same binding

heyanay: done — `Drizzle.postgresRaw` lives in its own module in #280

packages/better-auth/src/DrizzlePostgres.ts
heyanay: flattened into top-level `drizzlePostgresFromHyperdrive` and `drizzlePostgresFromUrl`

packages/alchemy/src/Drizzle/PostgresRaw.ts
```diff
@@ -0,0 +1,48 @@
...
+ * Effect runtime), use `Drizzle.postgres` instead. The two live in
+ * separate modules so apps that only need one don't pay the bundle
+ * cost of the other.
+ *
+ * The drizzle client is cached per-`ExecutionContext` — the pool is
+ * built at most once per request.
+ *
+ * @binding
+ */
+export const postgresRaw = <
```
sam-goodwin, line 61: Add example usage to the docs

packages/better-auth/src/DrizzlePostgres.ts
```diff
@@ -0,0 +1,168 @@
...
+ *
+ * @example
+ * ```typescript
+ * import { BetterAuth, drizzlePostgresFromHyperdrive } from "@alchemy.run/better-auth";
+ *
+ * .pipe(
+ *   Effect.provide(
+ *     drizzlePostgresFromHyperdrive({
+ *       hyperdrive: MyHyperdrive,
+ *       secret: process.env.BETTER_AUTH_SECRET!,
```
sam-goodwin: We should not be using environment variables here. Use Random resource to generate this instead of expecting the user to manually juggle.

packages/better-auth/src/DrizzlePostgres.ts
```diff
@@ -0,0 +1,168 @@
...
+) => {
+  const { url, ...rest } = options;
+  const connectionString: Effect.Effect<Redacted.Redacted<string>> =
+    typeof url === "string"
+      ? Effect.succeed(Redacted.make(url))
+      : Redacted.isRedacted(url)
+        ? Effect.succeed(url)
+        : url;
+  return buildLayer<TRelations>(connectionString, rest);
+};
```
sam-goodwin: Can these be combined into one function with a more idiomatic name like postgresDrizzle (consistent prefix)

```ts 
import * as BetterAuth from "@alchemy.run/better-auth"

BetterAuth.postgresDrizzle
```

sam-goodwin (changes requested):

packages/better-auth/src/DrizzlePostgres.ts
sam-goodwin: Done in 120a61c1 — collapsed into a single `postgresDrizzle({ db })` that accepts a Hyperdrive resource, URL string, `Redacted<string>`, or `Effect<Redacted<string>>`. Old file renamed to `PostgresDrizzle.ts` for consistency.

packages/better-auth/src/DrizzlePostgres.ts
sam-goodwin: Done in 120a61c1 — removed the `secret` option, the layer now provisions an `Alchemy.Random("BetterAuthSecret")` resource internally (matching the `CloudflareD1` layer).

packages/alchemy/src/Drizzle/PostgresRaw.ts
sam-goodwin: Done in 120a61c1 — added `@section`/`@example` JSDoc on `postgresRaw` (Hyperdrive binding, plain URL, plugging into better-auth) so they land in the generated API reference.

heyanay2105:
thanks, also noticed we should update CloudflareD1 to accept better-auth's options but that can be done separately

sam-goodwin:
Please open an issue if the  current sql support is not enough. This PR got stale (sorry about that). Would like to start fresh.

PR #283, opened 2026-05-08
feat(cloudflare): add SendEmail Worker binding

### LINKED ISSUE #243: lack of EmailSender functions binding in alchemy-effect but exists in alchemy v1

baptisteArno:
Adds a `SendEmail` capability for Cloudflare Workers, wrapping the native `send_email` binding.

Closes #243.

```ts
const email = yield* Cloudflare.SendEmail.bind({
  name: "EMAIL",
  destinationAddress: "ops@example.com",
});

yield* email.send({
  from: "noreply@example.com",
  to: "ops@example.com",
  subject: "Hello",
  text: "Hi",
});
```

Provide `Cloudflare.SendEmailLive` on the Worker effect (same pattern as `QueueBindingLive`, `HyperdriveBindingLive`). `sendRaw` is also exposed for the `cloudflare:email` `EmailMessage` form.

Naming note: alchemy v1 called this `EmailSender`. Renamed to `SendEmail` to match the Cloudflare runtime binding type (`cf.SendEmail`) and the action-based naming used by other capabilities (`SendMessage`, `PutObject`, `Fetch`).

No dedicated test. Pure binding capabilities like `Workers.Fetch` don't have one either, and an integration test couldn't exercise actual delivery without a manually-verified destination address on the test account. Manually checked that `bun tsc -b` and `bun generate:api-reference` pass.

### DIFF
--- a/packages/alchemy/src/Cloudflare/Email/SendEmail.ts
+++ b/packages/alchemy/src/Cloudflare/Email/SendEmail.ts
@@ -1,0 +1,211 @@
+import type * as runtime from "@cloudflare/workers-types";
+import * as Data from "effect/Data";
+import * as Effect from "effect/Effect";
+import * as Layer from "effect/Layer";
+import * as Binding from "../../Binding.ts";
+import type { ResourceLike } from "../../Resource.ts";
+import { isWorker, WorkerEnvironment } from "../Workers/Worker.ts";
+
+/**
+ * Configuration for a `send_email` binding on a Worker.
+ *
+ * `name` becomes the env key (e.g. `env.EMAIL`) and uniquely identifies
+ * the binding within the Worker. Use either `destinationAddress` or
+ * `allowedDestinationAddresses` to restrict where the Worker can send.
+ * Omit both to allow sending to any verified destination address on the
+ * account. Destination and sender addresses must be verified in the
+ * Cloudflare Email Routing dashboard before they can be used.
+ */
+export interface SendEmailConfig {
+  /**
+   * Binding name. This is the key used to access the binding from the
+   * Worker's `env` (e.g. `env.EMAIL`). Must match
+   * `[A-Za-z][A-Za-z0-9_]*`.
+   */
+  name: string;
+  /**
+   * Restrict the Worker to send to a single verified destination
+   * address.
+   */
+  destinationAddress?: string;
+  /**
+   * Restrict the Worker to send to one of these verified destination
+   * addresses.
+   */
+  allowedDestinationAddresses?: string[];
+  /**
+   * Restrict the Worker to send from one of these sender addresses. The
+   * sender domain must be configured for Email Routing on the account.
+   */
+  allowedSenderAddresses?: string[];
+}
+
+/**
+ * Email body shape for the builder-form `send` call. Either `text`,
+ * `html`, or both must be provided.
+ */
+export interface SendEmailMessage {
+  from: string | runtime.EmailAddress;
+  to: string | string[];
+  subject: string;
+  replyTo?: string | runtime.EmailAddress;
+  cc?: string | string[];
+  bcc?: string | string[];
+  headers?: Record<string, string>;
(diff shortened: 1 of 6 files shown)


sam-goodwin:
What tests can we add?

sam-goodwin:
cc @john-royal who had opinions on send email's design

baptisteArno:
Added a deploy-time test that binds 4 `send_email` variants on a single Worker (no restriction, `destinationAddress`, `allowedDestinationAddresses`, `allowedSenderAddresses`), deploys, then asserts each binding round-trips via `getScriptScriptAndVersionSetting`. See 62b85437.

A real send-and-receive test would need a verified destination + sender domain pre-configured on the test account, which seemed out of scope for this PR. Happy to add one if you can point me at an account setup.

packages/alchemy/test/Cloudflare/Email/send-email-worker.ts
```diff
@@ -0,0 +1,42 @@
...
+      destinationAddress: "ops@example.com",
+    });
+    yield* Cloudflare.SendEmail.bind({
+      name: "EMAIL_ALLOWED_DESTS",
+      allowedDestinationAddresses: ["ops@example.com", "alerts@example.com"],
+    });
+    yield* Cloudflare.SendEmail.bind({
+      name: "EMAIL_ALLOWED_SENDERS",
+      allowedSenderAddresses: ["noreply@example.com"],
+    });
```
sam-goodwin, line 34: These should return a handle with methods that are called below. Why is the value ignored?

packages/alchemy/src/Cloudflare/Email/SendEmail.ts
```diff
@@ -0,0 +1,211 @@
...
+  /**
+   * Restrict the Worker to send to one of these verified destination
+   * addresses.
+   */
+  allowedDestinationAddresses?: string[];
+  /**
+   * Restrict the Worker to send from one of these sender addresses. The
+   * sender domain must be configured for Email Routing on the account.
+   */
+  allowedSenderAddresses?: string[];
```
sam-goodwin, line 40: What's the default? 

Also, i think we need an `EmailAddress` or `EmailSender` resources to configure email addresses that are allowed to be sent?

packages/alchemy/test/Cloudflare/Email/send-email-worker.ts
```diff
@@ -0,0 +1,42 @@
...
+export default class SendEmailWorker extends Cloudflare.Worker<SendEmailWorker>()(
+  "SendEmailTestWorker",
+  {
+    main: import.meta.filename,
+    compatibility: { date: "2024-09-23" },
+  },
+  Effect.gen(function* () {
+    yield* Cloudflare.SendEmail.bind({ name: "EMAIL_UNRESTRICTED" });
+    yield* Cloudflare.SendEmail.bind({
+      name: "EMAIL_DESTINATION",
```
sam-goodwin, line 24: We shouldn't need a name for the binding.

The `.bind` call will return the binding and automatically create the name.

packages/alchemy/test/Cloudflare/Email/SendEmail.test.ts
```diff
@@ -0,0 +1,81 @@
...
+
+        expect(sendEmailBindings).toHaveLength(4);
+
+        expect(sendEmailBindings).toContainEqual(
+          expect.objectContaining({
+            type: "send_email",
+            name: "EMAIL_UNRESTRICTED",
+          }),
+        );
+        expect(sendEmailBindings).toContainEqual(
```
sam-goodwin, line 51: I think we should take this PR over and set it up in our account so we can test e2e behavior instead of checking config values.

sam-goodwin:
Taking this over here https://github.com/alchemy-run/alchemy-effect/pull/314

Need it in a PR i own so we can run things in CI.

PR #285, opened 2026-05-08
fix(cloudflare): reconcile Worker preview subdomain state

Dawsson:
## Issue

Cloudflare Worker reconciliation enabled Preview URLs when it called the subdomain API, but it only diffed the observed `enabled` flag. If a Worker had `enabled: true` and `previewsEnabled: false`, Alchemy skipped reconciliation.

This matters for the Cloudflare state store bootstrap because it reads the auth secret through an edge-preview request. Cloudflare documents Preview URLs as a separate Worker setting, and the script subdomain API exposes `previews_enabled` separately from `enabled`.

Docs:
- https://developers.cloudflare.com/workers/configuration/previews/
- https://developers.cloudflare.com/api/resources/workers/subresources/scripts/

## Solution

- Compare `previewsEnabled` when the desired workers.dev subdomain is enabled.
- Keep the disabled case scoped to `enabled`, matching the existing API call that sends only `enabled: false`.
- Retry the state-store secret probe for Cloudflare preview configuration errors (`1031` / `Invalid Workers Preview configuration`) while preview routing propagates.

## Validation

- `node_modules/.bin/oxfmt packages/alchemy/src/Cloudflare/Workers/Worker.ts packages/alchemy/src/Cloudflare/StateStore/State.ts`
- `git diff --check`

I also tried a live Worker regression test that manually disabled Preview URLs and redeployed, but it timed out waiting for initial workers.dev subdomain propagation before reaching the preview assertion. I left that test out because it would be flaky.


sam-goodwin:
Please add regression tests

PR #286, opened 2026-05-08
feat(cloudflare): add Analytics Engine binding

Dawsson:
## Issue

Alchemy v2 could not model a Cloudflare Workers Analytics Engine binding. Cloudflare configures this through Worker metadata as an `analytics_engine` binding, and the runtime binding exposes `writeDataPoint()`.

Docs:
- https://developers.cloudflare.com/workers/runtime-apis/bindings/analytics-engine/

## Solution

- Add `Cloudflare.AnalyticsEngineDataset` as a pure Worker binding marker.
- Emit `{ type: "analytics_engine", name, dataset }` in Worker metadata.
- Include Analytics Engine in `Cloudflare.InferEnv` as `AnalyticsEngineDataset`.

## Validation

- `node_modules/.bin/oxfmt packages/alchemy/src/Cloudflare/AnalyticsEngine/AnalyticsEngineDataset.ts packages/alchemy/src/Cloudflare/AnalyticsEngine/index.ts packages/alchemy/src/Cloudflare/index.ts packages/alchemy/src/Cloudflare/Workers/Worker.ts packages/alchemy/src/Cloudflare/Workers/InferEnv.ts`
- `git diff --check`


Dawsson:
Updated the implementation to avoid using a Resource-like `Type` marker for Analytics Engine. It now uses a plain `kind` marker, matching the Rate Limit binding pattern, and includes a Worker binding metadata test.\n\nAdditional validation:\n- `node_modules/.bin/tsc -b packages/alchemy/tsconfig.json --pretty false`\n- `bunx vitest run packages/alchemy/test/Cloudflare/AnalyticsEngine/AnalyticsEngineDataset.test.ts`

packages/alchemy/test/Cloudflare/AnalyticsEngine/AnalyticsEngineDataset.test.ts
```diff
@@ -0,0 +1,54 @@
...
+): Effect.Effect<Stack.CompiledStack<A>, Err> =>
+  // @ts-expect-error - Stack.make's typing erases R unsoundly here
+  effect.pipe(
+    Stack.make({
+      name: "test",
+      providers: Layer.empty as unknown as Layer.Layer<any, never, any>,
+      state: inMemoryState(),
+    }),
+    Effect.provideService(Stage, "test"),
+  );
```
sam-goodwin: What is going on here. This should not be needed. Follow the same pattern as other test

packages/alchemy/test/Cloudflare/AnalyticsEngine/AnalyticsEngineDataset.test.ts
```diff
@@ -0,0 +1,54 @@
...
+            {
+              type: "analytics_engine",
+              name: "EVENTS",
+              dataset: "app-events",
+            },
+          ],
+        },
+      },
+    ]);
+  }),
```
sam-goodwin: This test is insufficient. We need ot make a call. We need to stand up a fixture and exercise the Effect native interface. Looks like we are missing that. This is just the async interface

sam-goodwin:
Look at the other bindings like Artifacts store and followtheir lead. You need to build an Effect native interface

AnalyticsEngineDataset.ts
AnalyticsEngineDatasetBinding.ts

PR #288, opened 2026-05-08
feat(cloudflare): add Worker cron triggers

Dawsson:
## Issue

Alchemy v2 `Cloudflare.Worker` could not declare Cloudflare Worker Cron Triggers. Cloudflare supports Worker cron triggers through Wrangler `triggers.crons`, and the API exposes `PUT /accounts/{account_id}/workers/scripts/{scriptName}/schedules` to replace a Worker's schedule list.

Docs:
- https://developers.cloudflare.com/workers/configuration/cron-triggers/
- https://developers.cloudflare.com/api/resources/workers/subresources/scripts/subresources/schedules/methods/update/

## Solution

- Add `crons?: string[]` to `Cloudflare.Worker` props.
- Read live Worker schedules during provider reads.
- Diff schedules from resource state and reconcile with the Cloudflare schedules API after script upload.
- Preserve `crons` in the local Worker provider output for consistent state shape.

## Validation

- `bunx oxfmt packages/alchemy/src/Cloudflare/Workers/Worker.ts packages/alchemy/src/Cloudflare/Workers/LocalWorkerProvider.ts`
- `git diff --check`


packages/alchemy/src/Cloudflare/Workers/Worker.ts
```diff
@@ -272,6 +272,12 @@ export interface WorkerProps<
   >;
   exports?: string[];
   bindings?: Bindings;
+  /**
+   * Cron expressions that trigger the Worker's scheduled handler.
+   *
+   * Pass an empty array to remove all Cron Triggers.
+   */
+  crons?: string[];
```
sam-goodwin, line 288: Doesn't this invoke a specific `cron` handler? This will need to follow our event source model instead of being a property on the Worker

We can leave the cron property for the worker for async, but effect-native needs its own event source. Look at SQSQueueEventSource and look at Worker.ts and WorkerBundle.ts to see how we hook into the worker's entrypoint and route events

packages/alchemy/src/Cloudflare/Workers/CronEventSource.ts
```diff
@@ -0,0 +1,89 @@
...
+      yield* policy(expression);
+
+      const ctx = (yield* RuntimeContext) as unknown as FunctionContext;
+      yield* ctx.listen<void, Req>((event) => {
+        if (!isWorkerEvent(event) || event.type !== "scheduled") return;
+
+        const controller = event.input as cf.ScheduledController;
+        if (controller.cron !== expression) return;
+
+        return process(controller).pipe(Effect.catchCause(() => Effect.void));
```
sam-goodwin, line 85: We should route based on the cron expression.

sam-goodwin:
Take a look at how i wrote the tests in my commits. Merging. I've improved the AGENTS.md, AI should do the right thing next time

PR #295, opened 2026-05-08
feat(stage): typesafe Stage via generated declaration merging

baptisteArno:
Currently `Stage` resolves to `string`, so `yield* Stage` returns `string` everywhere and typos in `if (stage === "prd")` slip through.

This makes Stage opt-in typesafe by declaring the allowed list once on the stack:

```ts
Alchemy.Stack("Typebot", {
  stages: ["dev", "staging", "prod"] as const,
  providers: ...,
  state: ...,
}, body);
```

A new `alchemy types` command parses `alchemy.run.ts` and writes `.alchemy/types.d.ts`:

```ts
declare module "alchemy" {
  interface Stages {
    "dev": true;
    "staging": true;
    "prod": true;
  }
}
```

`Stage` is wired so that `keyof Stages` flows through. After running `alchemy types` once, `yield* Stage` narrows to the literal union, indexing on the stage value works, and typos error at compile time:

https://github.com/user-attachments/assets/f30c3aec-dfd5-435f-93f1-216b954e1a7e

When no `stages` array is declared, `Stage` falls back to `string` — fully back-compat.

### Runtime validation

The `--stage` flag in `deploy`, `destroy`, `plan`, `dev`, `login`, `logs`, `tail`, and `state` validates against the declared whitelist:

```
$ alchemy plan --stage preprod
Invalid stage "preprod" for stack "Typebot". Allowed stages: "dev", "staging", "prod".
```

### Notes

- `Stage`'s declaration uses an `interface Stage` + `const Stage` shape so the emitted `.d.ts` keeps `StageName` symbolic. That's what lets a consumer-side augmentation actually narrow at compile time. Inline class form would freeze the shape to `string` at lib build.
- The `STAGES_META` symbol is attached to the effect returned by `Alchemy.Stack(...)` so the CLI can read the array via `import()` without compiling the stack.
- Tested manually with a stack carrying `stages: ["dev", "staging", "prod"] as const`: `alchemy types` emits the augmentation, `tsc` narrows `yield* Stage`, `--stage preprod` is rejected before reaching the cloud.


sam-goodwin:
We have a strict no-codegen policy for users. We don't want to generate code in the user's repository. It leads to problems like we see in wrangler and sst.

We can export `Stages` as an interface so that users can augment it if they want on their side, but we should not add a CLI command to generate types.

baptisteArno:
Understood!

PR #298, opened 2026-05-09
feat(cloudflare): add Alerting Webhook resource

Gerbuuun:
- Add webhook resource/provider wiring for Cloudflare alerting
- Export alerting APIs through Cloudflare providers and index
- Cover create, update, adoption, and delete behavior in tests

### DIFF
--- a/packages/alchemy/src/Cloudflare/Alerting/Webhook.ts
+++ b/packages/alchemy/src/Cloudflare/Alerting/Webhook.ts
@@ -1,0 +1,247 @@
+import * as alerting from "@distilled.cloud/cloudflare/alerting";
+import * as Effect from "effect/Effect";
+import * as Option from "effect/Option";
+import * as Redacted from "effect/Redacted";
+import * as Stream from "effect/Stream";
+import { isResolved } from "../../Diff.ts";
+import * as Provider from "../../Provider.ts";
+import { Resource } from "../../Resource.ts";
+import { CloudflareEnvironment } from "../CloudflareEnvironment.ts";
+import type { Providers } from "../Providers.ts";
+
+export type WebhookType =
+  | "datadog"
+  | "discord"
+  | "feishu"
+  | "gchat"
+  | "generic"
+  | "opsgenie"
+  | "slack"
+  | "splunk";
+
+export type WebhookProps = {
+  /**
+   * Name of the webhook destination. Cloudflare includes this name in
+   * notification payloads sent to the webhook.
+   */
+  name: string;
+  /**
+   * POST endpoint Cloudflare calls when dispatching a notification.
+   */
+  url: string;
+  /**
+   * Optional secret sent in the `cf-webhook-auth` header for generic
+   * webhook notifications or formatted for supported destination types.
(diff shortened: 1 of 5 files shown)


Mkassabov:
@Gerbuuun whats the status on this? do you plan to complete or do you want us to take it over?

Gerbuuun:
Yes, you can take it over. I was trying it out and prematurely created a PR. Can't use it myself because I don't yet have a pro domain

packages/alchemy/src/Cloudflare/Alerting/Webhook.ts
```diff
@@ -0,0 +1,247 @@
...
+ * ```
+ *
+ * @example Webhook with secret
+ * ```typescript
+ * import * as Redacted from "effect/Redacted";
+ *
+ * const webhook = yield* Cloudflare.Webhook("GenericAlerts", {
+ *   name: "Generic Alerts",
+ *   url: "https://example.com/cloudflare/alerts",
+ *   secret: Redacted.make(process.env.CLOUDFLARE_WEBHOOK_SECRET!),
```
sam-goodwin, line 80: Never use process.env, use effect/Config instead.

packages/alchemy/src/Cloudflare/Alerting/Webhook.ts
```diff
@@ -0,0 +1,247 @@
...
+ * notification policies.
+ *
+ * @section Creating a Webhook
+ * @example Basic webhook
+ * ```typescript
+ * const webhook = yield* Cloudflare.Webhook("Alerts", {
+ *   name: "Slack Alerts",
+ *   url: "https://hooks.slack.com/services/...",
+ * });
+ * ```
```
sam-goodwin, line 71: Add an effect native event source binding.

packages/alchemy/src/Cloudflare/Alerting/Webhook.ts
```diff
@@ -0,0 +1,247 @@
...
+const unwrapSecret = (
+  secret: Redacted.Redacted<string> | undefined,
+): string | undefined => (secret ? Redacted.value(secret) : undefined);
+
+const isNotFoundError = (error: unknown): boolean => {
+  if (!error || typeof error !== "object") return false;
+  const tag = (error as { _tag?: unknown })._tag;
+  const status = (error as { status?: unknown }).status;
+  return tag === "NotFound" || tag === "WebhookNotFound" || status === 404;
+};
```
sam-goodwin, line 99: Looks like there's a missing errr tag in distilled/cloudflare. We need to open a pr in distilled patching so that we don't need casting here.

packages/alchemy/src/Cloudflare/Alerting/Webhook.ts
sam-goodwin: See https://github.com/alchemy-run/alchemy-effect/pull/401

sam-goodwin (changes requested):

sam-goodwin:
This exists now, closing.

PR #305, opened 2026-05-11
fix(cloudflare): include wasm modules in local sidecar bundle

baptisteArno:
The local sidecar bundles `.wasm` chunks (emitted by `cloudflare-rolldown-plugin` as `CompiledWasm`) but `bundleOutputToWorkerModules` filters out every binary file before handing them to workerd. Imports like `loro-crdt` then fail at runtime with "No such module" while the same worker deploys fine to Cloudflare via `putScript`.

```diff
-    if (file.path.endsWith(".map") || file.content instanceof Uint8Array) {
+    if (file.path.endsWith(".map")) {
+      continue;
+    }
+    if (file.content instanceof Uint8Array) {
+      modules.push({
+        name: file.path,
+        type: file.path.endsWith(".wasm") ? "Wasm" : "Data",
+        content: file.content,
+      });
       continue;
     }
```

`.map` files stay filtered. Tested locally: a Worker importing `loro-crdt` now boots in `alchemy dev`.


sam-goodwin:
Pushed some improvements to this PR to test it and also get local dev tests working (you can run `bun test` and test workerd instead of live)

PR #328, opened 2026-05-13
fix(Cloudflare): Cloudflare Worker HTTP effect lifecycle

### LINKED ISSUE #327: Cloudflare Worker Effect HTTP bridge hangs with Effect RpcServer.toHttpEffect
## Summary

Alchemy's Effect-native Cloudflare Worker HTTP bridge appears to hang when serving Effect RPC with `RpcServer.toHttpEffect(...)`.

The same minimal RPC server works when served directly with `wrangler dev` and Effect's own `HttpEffect.toWebHandler(...)`, but fails through Alchemy's Cloudflare Worker dev harness with the Workers runtime cancellation error:

```text
The Workers runtime canceled this request because it detected that your Worker's code had hung and would never generate a response.
```

## Standalone repro

Cloneable repo:

https://github.com/wking-io/effect-rpc-workerd-repro

Install:

```sh
pnpm install
```

Broken Alchemy path:

```sh
pnpm dev:broken
pnpm request
```

If Alchemy chooses another port, use the printed URL:

```sh
RPC_URL=http://reproapi.localhost:1338/rpc pnpm request
```

Observed result:

```text
HTTP/1.1 500 Internal Server Error
Internal Server Error
```

Alchemy/workerd logs include:

```text
The Workers runtime canceled this request because it detected that your Worker's code had hung and would never generate a response.
```

Workaround path:

```sh
pnpm dev:workaround
pnpm request
```

Observed result:

```text
HTTP/1.1 200 OK
{"jsonrpc":"2.0","id":0,"result":{"ok":true,"message":"hello workerd"}}
```

Direct Wrangler control:

```sh
pnpm dev:wrangler:broken
RPC_URL=http://localhost:8787/rpc pnpm request
```

Observed result in local testing: direct Wrangler returns successfully. That narrows the issue to the Alchemy Worker / Effect request harness rather than generic `workerd` + Effect RPC.

## Relevant files in the repro

- `src/broken-alchemy.ts`: Alchemy Worker using built-in `RpcServer.toHttpEffect(ReproRpcs)`.
- `src/workaround-alchemy.ts`: same RPC group and handlers, but with a request-local `RpcServer.Protocol` bridge.
- `src/broken.ts`: direct Wrangler control using `HttpEffect.toWebHandler(...)` and `RpcServer.toHttpEffect(...)`.
- `scripts/request.mjs`: curl-backed JSON-RPC request helper.

## Why this looks Alchemy-specific

Effect RPC's HTTP server path does this internally:

```ts
const { httpEffect, protocol } = yield* makeProtocolWithHttpEffect
yield* make(group, options).pipe(
  Effect.provideService(Protocol, protocol),
  Effect.forkScoped
)
return httpEffect
```

Direct Wrangler control succeeds when the Worker delegates through Effect's own HTTP web adapter:

```ts
const handler = HttpEffect.toWebHandler(app)
export default { fetch: (request) => handler(request) }
```

Alchemy's Cloudflare Worker bridge appears to run a different lifetime/scheduling shape:

- `Cloudflare/Workers/HttpServer.ts` creates/provides `HttpServerRequest` manually in `serveWebRequest(...)`, awaits the full `HttpServerResponse`, then converts with `HttpServerResponse.toWeb(...)`.
- `Cloudflare/Workers/Worker.ts` runs user request effects with `Effect.runPromiseExit`, manually provides a request `Scope`, and closes that scope in `ctx.waitUntil(...)` after the promise settles.

That differs from Effect's `HttpEffect.toWebHandler(...)`, which forks the HTTP app and resolves the web `Response` via Effect's `toHandled(...)` response callback path.

The failure mode suggests Alchemy's bridge can deadlock/starve Effect HTTP apps that rely on scoped forked fibers and response queues, such as `RpcServer.toHttpEffect(...)`.

## Expected behavior

An Effect `fetch` handler returned from `Cloudflare.Worker(..., Effect.gen(...))` should be able to serve an Effect HTTP app like `RpcServer.toHttpEffect(...)` without hanging.

## Actual behavior

Through Alchemy's Worker dev harness, the request never completes and workerd cancels it as hung. The same minimal RPC group/handler works through direct Wrangler and also works through the request-local protocol workaround.

## Dependency / security note

The repro intentionally has a small dependency set:

- `effect@4.0.0-beta.66`
- `alchemy@2.0.0-beta.38`
- `@effect/platform-node@4.0.0-beta.66`
- `wrangler@4.90.0`
- `typescript@6.0.3`
- `@cloudflare/workers-types@4.20260511.1`

Before publishing the repro, I checked the May 12, 2026 Mini Shai-Hulud supply-chain incident against this dependency set. This repro does not depend on the reported affected package families/namespaces. The lockfile was prepared with lifecycle scripts disabled, and:

```sh
pnpm audit --audit-level moderate
```

reported no known vulnerabilities.

wking-io:
## Summary

Fixes the Cloudflare Worker HTTP adapter so it runs handlers through Effect's standard HTTP lifecycle while preserving Alchemy's Worker-specific request context.

The previous adapter manually provided `HttpServerRequest` and converted `HttpServerResponse` to a web `Response`. That works for simple HTTP handlers, but it diverges from Effect's `HttpEffect.toHandled` / `toWebHandler` lifecycle in ways that can break scoped HTTP apps such as `RpcServer.toHttpEffect` under workerd. In the repro from #327, the same Effect RPC app succeeds under raw Wrangler but hangs when served through Alchemy.

This keeps Alchemy's existing behavior for:

- `cf-connecting-ip` -> `remoteAddress`
- the original Cloudflare `Request` service
- patched `request.raw` access
- existing 500 response mapping from `serveWebRequest`

while adding Effect's handled HTTP lifecycle, stream scope transfer, HEAD body suppression, and request abort interruption.

## Tests

- `bun tsc -b packages/alchemy`
- `bunx vitest run packages/alchemy/test/Cloudflare/Workers/HttpServer.test.ts --project alchemy`
- External repro: `effect-rpc-workerd-repro` with local Alchemy checkout, `alchemy dev alchemy.broken.ts`, then `RPC_URL=http://reproapi.localhost:1337/rpc pnpm request` returns `HTTP/1.1 200 OK` with the expected JSON-RPC result.

Fixes #327.

### DIFF
--- a/packages/alchemy/src/Cloudflare/Workers/HttpServer.ts
+++ b/packages/alchemy/src/Cloudflare/Workers/HttpServer.ts
@@ -1,9 +1,12 @@
 import type * as cf from "@cloudflare/workers-types";
 import * as Cause from "effect/Cause";
+import * as Context from "effect/Context";
 import * as Effect from "effect/Effect";
 import * as Option from "effect/Option";
 import type { Scope } from "effect/Scope";
 import type { HttpBodyError } from "effect/unstable/http/HttpBody";
+import * as EffectHttp from "effect/unstable/http/HttpEffect";
+import { ClientAbort } from "effect/unstable/http/HttpServerError";
 import * as HttpServerError from "effect/unstable/http/HttpServerError";
 import * as HttpServerRequest from "effect/unstable/http/HttpServerRequest";
 import * as HttpServerResponse from "effect/unstable/http/HttpServerResponse";
@@ -48,6 +51,10 @@ export const serveWebRequest = <Req = never>(
   Exclude<Req, HttpServerRequest.HttpServerRequest | Scope>
 > =>
   Effect.gen(function* () {
+    const context =
+      yield* Effect.context<
+        Exclude<Req, HttpServerRequest.HttpServerRequest | Scope>
+      >();
     const request = HttpServerRequest.fromWeb(
       webRequest as any as globalThis.Request,
     ).modify({
@@ -61,9 +68,7 @@ export const serveWebRequest = <Req = never>(
         }),
     });
 
-    const response = yield* handler.pipe(
-      Effect.provideService(HttpServerRequest.HttpServerRequest, request),
-      Effect.provideService(Request, webRequest as any),
+    const safeHandler = handler.pipe(
       Effect.catchCause((cause) => {
         const message = Option.match(Cause.findErrorOption(cause), {
           onNone: () => "Internal Server Error",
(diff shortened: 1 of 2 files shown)


sam-goodwin:
I think we'll need to update this for src/AWS/Lambda/HttpServer.ts too? Can we share code so we don't have to do it every time? 

Containers and EC2/ECS are fine because they use the NodeServices/BunServices implementation provided by effect/platform.

sam-goodwin:
Nevermind, i'll merge this and follow up in a separate PR

PR #331, opened 2026-05-13
feat(cloudflare): vite dev

john-royal:


examples/cloudflare-tanstack/src/routes/api.hello.ts
```diff
@@ -63,6 +62,8 @@ export const Route = createFileRoute("/api/hello")({
           // option 3 — bind to your effect worker and call rpc method
           case "rpc":
             return trace("GET option 3 (backend.hello rpc)", async () => {
+              // Error: Calling `require` for "path" in an environment that doesn't expose the `require` function. See https://rolldown.rs/in-depth/bundling-cjs#require-external-modules for more details.
+              const Cloudflare = await import("alchemy/Cloudflare");
```
sam-goodwin: Hmm, this doesn't look good

examples/cloudflare-tanstack/src/env.ts
```diff
@@ -1,4 +1,10 @@
...
 import type { WebsiteEnv } from "../alchemy.run.ts";
 
-export const env = cf.env as WebsiteEnv;
+// In development mode with TanStack Start, `import { env } from "cloudflare:workers"` does not work at the top level.
+// As a workaround, we use a proxy to access the env object.
+export const env = new Proxy({} as WebsiteEnv, {
+  get(_, prop) {
+    return cf.env[prop as keyof typeof cf.env];
+  },
+});
```
sam-goodwin, line 10: That's surprising, why is that?

examples/cloudflare-tanstack/src/routes/api.hello.ts
john-royal: It's very not good. Fails with @cloudflare/vite-plugin too, and the errors are different but equally useless:

```
Unexpected error: no match for module: ../rolldown-binding.linux-x64-gnu.node.
```

```
Error: Cannot find native binding. npm has a bug related to optional dependencies (https://github.com/npm/cli/issues/4828). Please try `npm i` again after removing both package-lock.json and node_modules directory.
```

Not sure what to do about this. The dynamic import doesn't work either - it just prevents the whole thing from crashing because of this.

PR #339, opened 2026-05-14
feat(axiom): smartfilter chart subtype

Mkassabov:


Mkassabov:
I don't like that smartfilter is a `chart` but I also don't want to rework all of our type names here because I'm lazy and 99% of people would get confused if we called it `panel` or whatever since it will almost always be a chart

sam-goodwin (approved):

PR #345, opened 2026-05-15
fix(dev): use @alchemy.run/node-utils lockfile

sam-goodwin:
Replace the sidecar's custom PID-file + touch-loop lock with `@alchemy.run/node-utils`'s vendored `proper-lockfile`. Same library [Auth/Lock.ts](packages/alchemy/src/Auth/Lock.ts) already uses, so `proper-lockfile` / `@types/proper-lockfile` are dropped from `packages/alchemy/package.json`.

```ts
// Sidecar/Lock.ts
const acquire = Effect.gen(function* () {
  yield* ensureLockFile;
  yield* Effect.acquireRelease(
    Effect.tryPromise({
      try: () =>
        lockfile.lock(paths.lock, {
          stale: STALE_MS,
          update: UPDATE_MS,
          retries: 0,
        }),
      catch: (cause) =>
        new LockError({ reason: "Conflict", message: "...", cause }),
    }),
    (release) => Effect.promise(() => release().catch(() => {})),
  );
  return yield* Effect.never;
}).pipe(Effect.forkScoped);
```

The `update`/`stale` options replace the manual heartbeat fiber; release runs on scope close.


packages/alchemy/src/Sidecar/Lock.ts
```diff
@@ -30,161 +22,54 @@ export class Lock extends Context.Service<
...
-              reason: "Invalid",
-              message: "Lock file is invalid",
-            }),
-          )
-        : Effect.succeed(pid);
-    }),
-  );
+  const ensureLockFile = fs
+    .writeFileString(paths.lock, "", { flag: "a" })
+    .pipe(Effect.ignore);
```
john-royal: Shouldn't this be handled by the new package instead?

sam-goodwin: You're right — `realpath: false` skips the `fs.realpath` stat (uses `path.resolve` instead), so the anchor file never needs to exist. The lock dir is the only thing on disk. Dropped `ensureLockFile` in both [Sidecar/Lock.ts](packages/alchemy/src/Sidecar/Lock.ts) and [Auth/Lock.ts](packages/alchemy/src/Auth/Lock.ts) in a82b041.

PR #353, opened 2026-05-16
fix(aws,cloudflare): sanitize entrypoint urls

Mkassabov:


Mkassabov:
claude said json stringify over pathe because it handlesmore edgecases like `"` in path names which is valid on linux/macos (psycho shit)

PR #370, opened 2026-05-18
feat(cloudflare): add Turnstile widget resource

agcty:
### Summary

Adds `Cloudflare.TurnstileWidget` backed by distilled's Turnstile API, including create, update, read, delete, and explicit-sitekey adoption.

The provider is registered in `Cloudflare.providers()` and exported from `alchemy/Cloudflare`. The resource treats account changes, explicit sitekey changes/removal, and region changes as replacements, while mutable widget settings are updated in place. Domain comparison is order-insensitive because Cloudflare canonicalizes domain order in responses.

### Verification

- `bun tsc -p packages/alchemy/tsconfig.json --pretty false`
- `bun tsc -p packages/alchemy/tsconfig.test.json --pretty false`
- `bun vitest run packages/alchemy/test/Cloudflare/Turnstile/TurnstileWidget.test.ts`

The focused Turnstile test was run with a real Cloudflare API token/account. It covered create/update/delete, managing an existing widget by sitekey, replacing when `sitekey` is removed, the domain-order canonicalization regression, and verified no `alchemy-test-turnstile*` widgets were left behind.

### DIFF
--- a/packages/alchemy/test/Cloudflare/Turnstile/TurnstileWidget.test.ts
+++ b/packages/alchemy/test/Cloudflare/Turnstile/TurnstileWidget.test.ts
@@ -1,0 +1,192 @@
+import * as Cloudflare from "@/Cloudflare";
+import { CloudflareEnvironment } from "@/Cloudflare/CloudflareEnvironment";
+import * as Test from "@/Test/Vitest";
+import * as turnstile from "@distilled.cloud/cloudflare/turnstile";
+import { expect } from "@effect/vitest";
+import * as Data from "effect/Data";
+import * as Effect from "effect/Effect";
+import * as Redacted from "effect/Redacted";
+import { MinimumLogLevel } from "effect/References";
+import * as Schedule from "effect/Schedule";
+
+const { test } = Test.make({ providers: Cloudflare.providers() });
+
+const logLevel = Effect.provideService(
+  MinimumLogLevel,
+  process.env.DEBUG ? "Debug" : "Info",
+);
+
+test.provider(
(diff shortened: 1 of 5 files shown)


packages/alchemy/test/Cloudflare/Turnstile/TurnstileWidget.test.ts
```diff
@@ -0,0 +1,192 @@
...
+class WidgetStillExists extends Data.TaggedError("WidgetStillExists") {}
+
+const isMissingTurnstileWidget = (error: unknown): boolean =>
+  typeof error === "object" &&
+  error !== null &&
+  (("_tag" in error && (error as { _tag: unknown })._tag === "NotFound") ||
+    ("_tag" in error &&
+      (error as { _tag: unknown })._tag === "CloudflareHttpError" &&
+      "status" in error &&
+      (error as { status: unknown }).status === 404));
```
sam-goodwin, line 188: This looks like a missing error tag in distilled? Why do we need a complicated casting to catch an error tag?

the CloudflareHttpError+404 means we need to patch distilled.

agcty: Agreed, this belongs in Distilled. I opened alchemy-run/distilled#303 to map the raw Turnstile 404 to `NotFound`; once that lands/releases, this branch can drop the temporary `CloudflareHttpError` fallback and use `catchTag("NotFound")`.

sam-goodwin:
Thsi exists in Turnstile/Widget

PR #371, opened 2026-05-18
feat(cloudflare/zaraz): add config resource

agcty:
Add a Cloudflare Zaraz config resource and a type-only Zaraz event contract surface.

```ts
const zarazConfig = yield* Cloudflare.ZarazConfig("Analytics", {
  zone: "example.com",
  dataLayer: true,
  settings: {
    autoInjectScript: true,
    ecommerce: true,
    hideIPAddress: true,
  },
});
```

```ts
const zarazEvents = Cloudflare.defineZarazEvents<{
  Login: { method: "google" | "email" | "email-link" };
  "Button Clicked": { button_label: string; context?: string };
  __zarazSPA: undefined;
}>({ ecommerce: true });

export type AppZarazEvents = Cloudflare.InferZarazEvents<typeof zarazEvents>;
```

The resource retains omitted config fields, keeps the live Zaraz config on destroy by default, and supports `delete: true` to restore Cloudflare's default config. It also observes the current workflow, but does not reconcile workflow yet because the lower-level Distilled workflow PUT encoding needs https://github.com/alchemy-run/distilled/pull/301 first.

The event contract helper is intentionally type-only. Browser code should keep calling Cloudflare's injected `window.zaraz` API and import derived application event types, avoiding runtime imports from the main Alchemy package in client bundles.

`defineZarazEvents` only enables `zaraz.ecommerce()` typing when called with `{ ecommerce: true }`, matching Cloudflare's `settings.ecommerce` flag. When enabled, ecommerce typing uses Cloudflare's standard ecommerce events, including `Product Viewed`, `Checkout Started`, `Order Completed`, `Order Refunded`, and the rest of the supported event list.


packages/alchemy/src/Cloudflare/Zaraz/ZarazConfig.ts
```diff
@@ -0,0 +1,385 @@
...
+/**
+ * A Cloudflare Zaraz zone configuration.
+ *
+ * Zaraz is a zone-level singleton. This resource reconciles the current zone
+ * config to the desired values while retaining existing settings for fields
+ * omitted from props.
+ *
+ * Destroy keeps the current Zaraz config by default to avoid wiping unrelated
+ * zone-level analytics setup. Set `delete: true` to restore Cloudflare's
+ * default Zaraz config on destroy.
```
sam-goodwin, line 154: We should probably open the docs with a one-liner of what Zaraz is and link to relevant resources before going into mechanical details of how it works

packages/alchemy/src/Cloudflare/Zaraz/ZarazConfig.ts
```diff
@@ -0,0 +1,385 @@
...
+  if (Array.isArray(value)) {
+    return value.map((item) => stripFields(item, empty));
+  }
+  if (!value || typeof value !== "object") return value;
+  return Object.fromEntries(
+    Object.entries(value)
+      .filter(([, entry]) => entry !== empty)
+      .map(([key, entry]) => [key, stripFields(entry, empty)]),
+  );
+};
```
sam-goodwin: There are a lot of adhoc utilities here:
1. do we already have these somewhere?
2. can we share code so we don't need to keep re-writing this?

packages/alchemy/src/Cloudflare/Zaraz/ZarazEventTypes.ts
```diff
@@ -0,0 +1,246 @@
...
+  };
+  page?: {
+    title?: string;
+    url?: string;
+    referrer?: string;
+    encoding?: string;
+  };
+};
+
+type Simplify<T> = { readonly [K in keyof T]: T[K] } & {};
```
sam-goodwin: I suspect this already exists in src/Utils/ 

If not, add it.

packages/alchemy/src/Cloudflare/Zaraz/ZarazEventTypes.ts
sam-goodwin, line 1: Is this meant to be an Event Source?

packages/alchemy/src/Cloudflare/Zaraz/ZarazEvents.ts
```diff
@@ -0,0 +1,246 @@
...
+ * @example
+ * ```typescript
+ * const zaraz = Cloudflare.defineZarazEvents<{
+ *   Login: { method: "google" | "email" | "email-link" };
+ *   "Button Clicked": { button_label: string; context?: string };
+ * }>();
+ *
+ * const zarazWithEcommerce = Cloudflare.defineZarazEvents<{
+ *   Login: { method: "google" | "email" | "email-link" };
+ * }>({ ecommerce: true });
```
sam-goodwin: Why are these function calls instead of properties on the Zaraz resource?

packages/alchemy/test/Cloudflare/Zaraz/ZarazConfig.test.ts
```diff
@@ -0,0 +1,139 @@
...
+  if (Array.isArray(value)) {
+    return value.map((item) => stripFields(item, empty));
+  }
+  if (!value || typeof value !== "object") return value;
+  return Object.fromEntries(
+    Object.entries(value)
+      .filter(([, entry]) => entry !== empty)
+      .map(([key, entry]) => [key, stripFields(entry, empty)]),
+  );
+};
```
sam-goodwin: More duplicated utils. We can't keep duplicating this

packages/alchemy/test/Cloudflare/Zaraz/ZarazEvents.types.ts
sam-goodwin, line 1: I don't think we should have a .types file. Always a code smell, especially in tests. Please make sure all types are defined in the src/Zaraz folder.

sam-goodwin (changes requested):

Mkassabov:
@agcty released new version of distilled with this change

agcty:
updated this branch to `@distilled.cloud/cloudflare@0.21.4`, which includes the Zaraz workflow raw-body fix from distilled.\n\nverified locally:\n- `bun tsc -b packages/alchemy/tsconfig.json --pretty false --force`\n- `bun tsc -b packages/alchemy/tsconfig.test.json --pretty false --force`\n- `bun vitest run packages/alchemy/test/Cloudflare/Zaraz/ZarazConfig.test.ts packages/alchemy/test/Cloudflare/Zaraz/ZarazEventTypes.test.ts`\n\nthe live Zaraz config tests still skip locally without the Cloudflare Zaraz test env, but the dependency blocker is gone now.

PR #373, opened 2026-05-19
feat(cloudflare): add binding tag helpers

wking-io:
## Summary
- add a generic Cloudflare `BindingTag` helper for app-named runtime services
- expose `.Tag` helpers on Cloudflare object bindings: R2, KV, D1, Queue, AI Gateway, Analytics Engine, Artifacts, Images, Hyperdrive, SendEmail, and Secrets Store secrets
- preserve binding/resource discovery by delegating generated tags to the existing `XBinding.bind(resource)` path
- add source JSDoc examples for a basic D1 binding tag and multiple R2 binding tags

### Basic D1 binding tag

```typescript
const MyDB = Cloudflare.D1Database("MyDB");

class DB extends Cloudflare.D1Database.Tag<DB>()("DB", {
  resource: MyDB,
}) {}

  "Worker",
  Effect.gen(function* () {
    const DBLive = yield* DB.layer;

    return {
        const db = yield* DB;
        const user = yield* db
          .prepare("SELECT * FROM users WHERE id = ?")
          .bind("user-123")
          .first();

        return Response.json(user);
      }).pipe(Effect.provide(DBLive)),
    };
  }).pipe(Effect.provide(Cloudflare.D1ConnectionLive)),
);
```

### Multiple R2 binding tags

```typescript
const PublicBucket = Cloudflare.R2Bucket("PublicAssets");
const PrivateBucket = Cloudflare.R2Bucket("PrivateAssets");

class PublicAssets extends Cloudflare.R2Bucket.Tag<PublicAssets>()(
  "PublicAssets",
  { resource: PublicBucket },
) {}

class PrivateAssets extends Cloudflare.R2Bucket.Tag<PrivateAssets>()(
  "PrivateAssets",
  { resource: PrivateBucket },
) {}

  "Worker",
  Effect.gen(function* () {
    const PublicAssetsLive = yield* PublicAssets.layer;
    const PrivateAssetsLive = yield* PrivateAssets.layer;
    const AppBindings = Layer.mergeAll(PublicAssetsLive, PrivateAssetsLive);

    return {
        const publicAssets = yield* PublicAssets;
        const privateAssets = yield* PrivateAssets;

        yield* publicAssets.put("index.html", "<h1>Hello</h1>");
        yield* privateAssets.put("audit.json", "{}");

        return new Response("ok");
      }).pipe(Effect.provide(AppBindings)),
    };
  }).pipe(Effect.provide(Cloudflare.R2BucketBindingLive)),
);
```

### DIFF
--- a/packages/alchemy/src/Cloudflare/AiGateway/AiGateway.ts
+++ b/packages/alchemy/src/Cloudflare/AiGateway/AiGateway.ts
@@ -6,7 +6,7 @@ import * as Provider from "../../Provider.ts";
 import { Resource } from "../../Resource.ts";
 import { CloudflareEnvironment } from "../CloudflareEnvironment.ts";
 import type { Providers } from "../Providers.ts";
-import { AiGatewayBinding } from "./AiGatewayBinding.ts";
+import { AiGatewayBinding, AiGatewayTag } from "./AiGatewayBinding.ts";
 
 export type AiGatewayRateLimitingTechnique = "fixed" | "sliding";
 
@@ -269,6 +269,7 @@ export const AiGateway = Resource<AiGateway>("Cloudflare.AiGateway")({
    * client for the runtime AI Gateway binding.
    */
   bind: AiGatewayBinding.bind,
+  Tag: AiGatewayTag,
 });
 
 export const AiGatewayProvider = () =>
--- a/packages/alchemy/src/Cloudflare/AiGateway/AiGateway.ts
+++ b/packages/alchemy/src/Cloudflare/AiGateway/AiGateway.ts
@@ -262,6 +262,36 @@ export const isAiGateway = (value: unknown): value is AiGateway =>
  *   logManagementStrategy: "STOP_INSERTING",
  * });
  * ```
+ *
+ * @section Binding to a Worker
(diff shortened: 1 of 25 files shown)


Mkassabov:
In alchemy `.bind(resource)` is already the typed runtime handle. The pattern for using it is just be
```ts
Effect.gen(function* () {
  const database = yield* MyDB;
  const db = yield* Cloudflare.D1Connection.bind(database);
  return { fetch: Effect.gen(function* () { yield* db.exec(...) }) };
})
```

look at the examplesin the PR the tag + layer indirection happens inside the same `Effect.gen` scope; so I'm pretty sure this isn't actually decoupling anything and the dependency graph ends up exactly the same (except you're adding an extra context tag in there).

---

**Can you clarify what you are trying to achieve here?** 

As far as I can see this change has the downsides that we now have another api to explain in the docs, and it seems a footgun is introduced since yielding the `.layer` from inside the runtime phase would break discovery.

I'm open to there being a hole in our docs or a DX problem in the current way we do things, I'm just trying to get to the root of that.

wking-io:
I talked my way through it in discord 🤘

PR #375, opened 2026-05-19
chore(ci): pin GitHub Actions to latest releases

aeterno-caspian:
Pin GitHub Actions to immutable commit SHAs so CI runs the exact action code reviewed in this repo, instead of whatever a mutable version tag points to later.

```yaml
uses: actions/setup-node@48b55a011bda9f5d6aeb4c2d9c7362e8dae4041e # v6.4.0
```

The trailing release comment keeps upgrades readable, while the SHA makes workflow runs reproducible and reduces supply-chain risk from moved or compromised tags.

Applies the same pinned-action shape to the deploy, test, PR package, and release workflows.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -39,7 +39,7 @@ jobs:
       contents: read
       pull-requests: write
     steps:
-      - uses: actions/checkout@v4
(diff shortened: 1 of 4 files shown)


Mkassabov (approved):
thanks!
manually checked all the pinned commits and they do in fact line up with latest releases

PR #381, opened 2026-05-20
refactor(AWS/Lambda): split Function into bridge + runtime context

sam-goodwin:
Mirrors the convention established for Cloudflare Workers in #374. The
per-invocation runtime context moves out of Function.ts into its own
file, and a new FunctionBridge owns the per-invocation scope so each
Lambda invocation runs with a fresh Scope, request-scoped HandlerContext
and ExecutionContext, and the deploy-init service context re-provided
on top of the globalContext Layer.

The generated entrypoint collapses from a ~70-line inline layer setup
to a four-line default-export of makeFunctionBridge, matching the
Cloudflare bundle template shape.

### DIFF
--- a/packages/alchemy/src/AWS/Lambda/FunctionBridge.ts
+++ b/packages/alchemy/src/AWS/Lambda/FunctionBridge.ts
@@ -1,0 +1,138 @@
+import * as Credentials from "@distilled.cloud/aws/Credentials";
+import * as Region from "@distilled.cloud/aws/Region";
+import { NodeServices } from "@effect/platform-node";
+import type * as lambda from "aws-lambda";
+import * as Cause from "effect/Cause";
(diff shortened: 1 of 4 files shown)


packages/alchemy/src/AWS/Lambda/FunctionBridge.ts
```diff
@@ -0,0 +1,138 @@
...
+  >;
+
+  const exported = tag.pipe(
+    Effect.flatMap((func) => func.RuntimeContext.exports!),
+    Effect.flatMap((exports) =>
+      Effect.isEffect(exports.handler)
+        ? exports.handler
+        : Effect.succeed(exports.handler),
+    ),
+  ) as Effect.Effect<
```
sam-goodwin, line 50: Is .handler right? What if it's parameterized? Shouldn't we use default by default?

PR #382, opened 2026-05-20
fix(dev): dangling processes after dev server exits

john-royal:
Local providers used to run inside an in-process "sidecar" tied to the dev server. When the dev process exited (crash, Ctrl-C, hot reload restart), the child miniflare/runtime processes it had spawned could survive and leak — leftover workerd processes, ports stuck open.

This PR refactors the sidecar into a generic `Local/Rpc*` provider abstraction and moves local providers into a dedicated child process managed by an `RpcSpawner`. The spawner:

- Spawns provider servers via `bun run` or `node --experimental-transform-types` (whichever runtime the dev command was invoked with) so `.ts` entry points work in both.
- Registers an `exitHook` that `SIGKILL`s the entire process group on parent exit, plus a scoped finalizer for clean shutdown.
- Caches sessions per `(serverEntryUrl, alchemyContext, stack)` and re-spawns if the child has died.

```ts
export const LocalWorkerProvider = () =>
  RpcProvider.effect(
    Worker,
    import.meta.resolve(/* ./Local.ts | ./Local.js */),
    Effect.gen(function* () { /* … */ }),
  );
```

`RpcProvider.providerServicesEffect` lets heavy provider-only layers (e.g. the Cloudflare local runtime + proxy) build lazily inside the child instead of the parent.

### Notable

- `Sidecar/` and `Cloudflare/Local/Sidecar*` are removed; consumers move to `Local/Rpc*`.
- `dev` command now passes `ALCHEMY_RPC_SPAWNER_URL` to the inner exec process; `RpcProviderProxy.fromEnv()` picks it up.
- Tests added under `packages/alchemy/test/Local/` for serialization, server, session, spawner, and spawner cleanup (verifies orphaned-child cleanup on parent crash).


packages/alchemy/src/Cli/commands/dev.ts
```diff
@@ -15,41 +17,53 @@ export const devCommand = Command.make(
...
+        detached: false,
+      });
+      yield* child.exitCode;
+    },
+    (effect, args) =>
+      Effect.provide(
+        RpcSpawner.layerServer({
+          profile: args.profile,
+          envFile:
+            args.envFile._tag === "Some" ? args.envFile.value : undefined,
```
sam-goodwin: nit: Option.getOrUndefined

packages/alchemy/src/Cloudflare/Workers/LocalWorkerProvider.ts
```diff
@@ -1,37 +1,555 @@
...
+  RpcProvider.effect(
     Worker,
+    import.meta.resolve(
+      // `import.meta.resolve(<string>)` is a runtime API — TypeScript's
+      // `rewriteRelativeImportExtensions` does NOT touch the string literal, so
+      // we have to pick the right extension ourselves. `import.meta.url` reflects
+      // the actual on-disk extension of *this* file (`.ts` when loaded from
+      // `src/` under Bun or vitest, `.js` when loaded from the compiled `lib/`
+      // under Node), which is exactly the signal we need.
+      import.meta.url.endsWith(".ts") ? "../Local.ts" : "../Local.js",
```
sam-goodwin, line 101: is import.meta.url better to use than import.meta.filename?

packages/alchemy/src/Cloudflare/Workers/LocalWorkerProvider.ts
```diff
@@ -1,37 +1,555 @@
...
+          }
+        }
+        return modules;
+      });
+
+      const serveScoped = Effect.fnUntraced(function* (
+        worker: WorkerConfig,
+        bundle: Bundle.BundleOutput,
+      ) {
+        const scope = yield* Effect.flatMap(Effect.scope, Scope.fork);
```
sam-goodwin: nit: prefer left-to-right `Effect.scope.pipe(Scope.fork)`

sam-goodwin: Why fork a scope instead of make new?

packages/alchemy/src/Cloudflare/Local.ts
```diff
@@ -0,0 +1,24 @@
...
+  Layer.merge(
+    Credentials.fromAuthProvider(),
+    CloudflareEnvironment.fromProfile(),
+  ),
+  CloudflareAuth,
+);
+
+RpcServer.launch(
+  LocalWorkerProvider().pipe(
+    Layer.provide(localRuntimeServices({ port: 1337 })),
```
sam-goodwin: hard-coded? Will this work for monorepos?

packages/alchemy/src/Cloudflare/Local.ts
```diff
@@ -0,0 +1,24 @@
...
+    Layer.provide(cloudflareServices),
+  ),
+);
```
sam-goodwin: nit: left-to-right

`.pipe(.., RpcServer.launc)`?

packages/alchemy/src/Local/RpcProviderProxy.ts
```diff
@@ -0,0 +1,93 @@
...
+    Effect.suspend(() => {
+      const url = process.env[SPAWNER_URL_ENV_KEY];
+      if (!url) {
+        return Effect.die(
+          new Error(`${SPAWNER_URL_ENV_KEY} environment variable is not set`),
+        );
+      }
+      return make(url);
+    }),
+  );
```
sam-goodwin, line 86: Don't use `process.env`, use `Config`

packages/alchemy/src/Local/RpcSerialization.ts
```diff
@@ -0,0 +1,223 @@
...
+    );
+  }
+  return value;
+};
+
+const deserializeRpcArgs = (value: unknown): unknown => {
+  if (Array.isArray(value)) {
+    return value.map(deserializeRpcArgs);
+  } else if (typeof value === "object" && value !== null) {
+    if ("_tag" in value && value._tag === "Redacted" && "value" in value) {
```
sam-goodwin, line 204: `Redacted.isRedacted`

packages/alchemy/src/Local/RpcServerEnvironment.ts
```diff
@@ -0,0 +1,63 @@
...
+export const layer = (environment: RpcServerEnvironment) =>
+  Layer.mergeAll(
+    ProfileLive,
+    CredentialsStoreLive,
+    Layer.succeed(AuthProviders, {}),
+    ConfigProvider.layer(
+      loadConfigProvider(Option.fromUndefinedOr(environment.envFile)).pipe(
+        Effect.map((base) => withProfileOverride(base, environment.profile)),
+      ),
+    ),
```
sam-goodwin, line 35: Are we copy-pasting this everywhere?

packages/alchemy/src/Local/RpcServerEnvironment.ts
```diff
@@ -0,0 +1,63 @@
...
+    Layer.succeed(Stage, environment.stack.stage),
+  );
+
+export const RPC_SERVER_ENVIRONMENT_KEY =
+  "ALCHEMY_RPC_SERVER_ENVIRONMENT" as const;
+
+export const fromEnv = () => {
+  try {
+    const environment = JSON.parse(
+      process.env[RPC_SERVER_ENVIRONMENT_KEY]!,
```
sam-goodwin: no process.env, use Config

packages/alchemy/src/Local/RpcSpawner.ts
```diff
@@ -0,0 +1,196 @@
...
+        detached: false,
+        env: {
+          [RPC_SERVER_ENVIRONMENT_KEY]: JSON.stringify(environment),
+        },
+        extendEnv: true,
+      },
+    );
+    const handle = yield* spawner.spawn(command);
+    const unregister = exitHook(() => {
+      killProcessGroup(handle.pid, "SIGKILL");
```
sam-goodwin, line 91: Should we try SIGINT before SIGKILL?

packages/alchemy/test/Local/fixtures/rpc-spawner-parent.ts
```diff
@@ -0,0 +1,61 @@
...
+          dotAlchemy: "/tmp/.alchemy",
+          updateStateStore: false,
+          dev: true,
+          adopt: false,
+        },
+        stack: { name: "test", stage: "dev" },
+      }),
+      headers: { "content-type": "application/json" },
+    }).then((r) => r.text()),
+  );
```
sam-goodwin: Use HttpClient in tests, not fetch

packages/alchemy/test/Local/fixtures/rpc-spawner-parent.ts
```diff
@@ -0,0 +1,61 @@
...
+
+  // The child's pid is whatever owns the listening port returned in res.
+  // We surface it for the test harness via stdout.
+  process.stdout.write(`PARENT_PID=${process.pid}\n`);
+  process.stdout.write(`CHILD_URL=${res}\n`);
```
sam-goodwin: why not console.log?

packages/alchemy/test/Local/RpcServer.test.ts
```diff
@@ -0,0 +1,183 @@
...
+    stdio: ["ignore", "pipe", "pipe"],
+  });
+  let stdout = "";
+  let stderr = "";
+  child.stdout!.on("data", (d) => {
+    stdout += d.toString();
+  });
+  child.stderr!.on("data", (d) => {
+    stderr += d.toString();
+  });
```
sam-goodwin: Do we need this low level spawn code in tests vs the existing exec util?

packages/alchemy/test/Local/RpcServer.test.ts
```diff
@@ -0,0 +1,183 @@
...
+    kill: async () => {
+      if (child.exitCode === null && !child.killed) {
+        child.kill("SIGTERM");
+        setTimeout(() => {
+          if (child.exitCode === null) child.kill("SIGKILL");
+        }, 1_000);
+      }
+      return await exited;
+    },
+  };
```
sam-goodwin: re-write tests in effect, not async code

packages/alchemy/src/Local/RpcSerialization.ts
john-royal: These values are serialized as `{_tag: "Redacted", value: ...}` and `{_tag: "Output", description: ...}`, so we need to detect them manually - Redacted.isRedacted and Output.isOutput do not work.

Added a comment to clarify.

packages/alchemy/src/Local/RpcSpawner.ts
john-royal: The `exitHook` is a last-ditch, synchronous attempt to clean up in case the scope finalizer doesn't run.

The scope finalizer runs `SIGINT`, followed by `SIGKILL` if that doesn't success, and finally unregisters the exit hook. So we do try both when we're able to.

packages/alchemy/src/Cloudflare/Workers/LocalWorkerProvider.ts
john-royal: Generally I prefer `import.meta.filename`, but `import.meta.resolve` takes and returns a URL IIRC

packages/alchemy/src/Cloudflare/Local.ts
john-royal: Yes, if 1337 is taken, cf-runtime retries on next available port.

It was hard-coded before. Perhaps we should make it configurable separately?

PR #401, opened 2026-05-21
feat(cloudflare/workers): add Email event source for inbound mail

sam-goodwin:
Adds `Cloudflare.email({ zone }).subscribe(handler)` — one call wires both halves of an Email Worker consumer:

- **Runtime**: registers the `email` event listener on the Worker.
- **Deploy-time**: yields a sibling `Email.Routing` toggle on the zone plus the routing resource whose `actions: [{ type: "worker", … }]` targets this Worker — `Email.CatchAll` for a catch-all subscription, `Email.Rule` for anything more specific.

```typescript
  "Inbox",
  { main: import.meta.url },
  Effect.gen(function* () {
    yield* Cloudflare.email({ zone: "example.com" }).subscribe((message) =>
      message.forward("ops@example.com"),
    );
    return {};
  }).pipe(Effect.provide(Cloudflare.EmailEventSourceLive)),
);
```

`alchemy.run.ts` no longer needs hand-rolled `Email.Routing` + `Email.Rule` calls.

### Message wrapper

`message.raw` is the underlying `cf.ForwardableEmailMessage` (escape hatch); ergonomic fields and Effect-wrapped actions sit on top. The cf names `raw` / `rawSize` are exposed as `body` / `bodySize` so `raw` can be the escape hatch.

```ts
interface ForwardableEmailMessage {
  readonly raw: cf.ForwardableEmailMessage;
  readonly from: string;
  readonly to: string;
  readonly headers: cf.Headers;
  readonly body: cf.ReadableStream<Uint8Array>;
  readonly bodySize: number;
  setReject(reason: string): Effect.Effect<void>;
  forward(rcptTo: string, headers?: cf.Headers): Effect.Effect<void, EmailError>;
  reply(message: cf.EmailMessage): Effect.Effect<void, EmailError>;
}
```

### Since the original revision

Rebased onto `main` and fixed three defects found while validating it against the real API.

**The documented default could never deploy.** `email({ zone })` with no matchers built the catch-all as an `Email.Rule` with `matchers: [{ type: "all" }]`. Cloudflare rejects that outright:

```
createRule(matchers: [{ type: "all" }])  ->  Conflict: Invalid rule operation
```

The zone catch-all is a per-zone singleton behind `/rules/catch_all` — which is exactly why `Rule`'s own `list()` filters those rows out. An omitted-or-`all` matcher now provisions `Email.CatchAll`.

**A failing handler silently destroyed mail.** The listener contract is `E = never`, so the event source logged the failure and then discarded it — accepting the envelope and dropping it, with nothing left to retry from. It now logs and re-raises as a defect, so Cloudflare turns it into a temporary SMTP failure and the sending server retries. Bounce deliberately with `message.setReject(...)`.

```diff
-          Effect.catchCause(() => Effect.void),
+          Effect.tapCause(logIt),
+          Effect.orDie,
```

**It did not compile.** `subdomain` is now `workersDev`, and `Email.test.ts` was the last file in the repo on the deleted `@/Test/Vitest` + `@effect/vitest` harness, so it could not run under the `alchemy-test` runner at all.

### Testing

`test/Cloudflare/Workers/EmailEventSource.test.ts`, run live against a scratch zone. The deploy-time half needs no mail to flow, so it is asserted out-of-band via distilled on **every** run rather than being skipped whenever the email env vars are absent; only the send→receive round trip stays gated on `CLOUDFLARE_TEST_EMAIL_FROM`.

Also validated in a real application: a production Cloudflare Email Routing relay migrated from a raw `email:` handler to `subscribe(...)`, then torn down and rebuilt from scratch on a different machine against a preview build of this branch.

Confirmed against live mail, not fixtures:

- inbound traverses MX → Email Routing → catch-all → the `subscribe` handler → R2 + D1
- outbound forwards back through the `send_email` binding
- the relay then ran unattended and handled real third-party mail on both paths — auto-forwarding for known senders, holding unknown ones for review
- `alchemy destroy` restores the zone's catch-all to its pre-management state

The teardown is what surfaced the `Email.CatchAll` destroy bug fixed in the last commit here: the captured pre-management action named a Worker that had since been deleted, so restoring it failed and stranded the Worker, bucket and database behind it.

### Depends on

#1349 — the auto-created rule names a Worker created in the same apply, which is the race that PR types and retries.


sam-goodwin:
@Mkassabov i don't have an account with email testing capability. Can you set up our test account and drive this home?

Issue #402, opened 2026-05-21
Worker doesn't run HTTP pre-response handlers before returning responses


lucasthevenet:
## Summary

Effect HTTP middleware can register pre-response handlers that run after the route handler returns but before the response is sent.
CORS middleware uses this for normal requests: it lets the handler produce the response, then adds headers like access-control-allow-origin in a pre-response handler.

The current Worker path converts the handler response directly into a Web Response, so those pre-response handlers never run. As a result, normal GET / POST requests can miss CORS headers.

OPTIONS preflight still works because CORS short-circuits preflight with a fully decorated response instead of relying on the pre-response handler path.

### Define a Worker with a cors middleware:

```ts
export default class TestWorker extends Cloudflare.Worker<TestWorker>()(
  "TestWorker",
  {
    main: import.meta.filename
  },
  Effect.gen(function* () {
    return {
      fetch: HttpRouter.empty.pipe(
        HttpRouter.post(
          "/",
          HttpServerResponse.json({ ok: true }),
        ),
        Layer.provide(HttpRouter.cors({
          allowedOrigins: ["*"],
          allowedMethods: ["GET", "POST", "OPTIONS"],
          allowedHeaders: ["Content-Type"],
        })),
        HttpRouter.toHttpEffect,
      ),
    };
  }),
) {}
```

### Send an actual request with an Origin header:
`curl -i \
  -X POST "$WORKER_URL/" \
  -H "Origin: https://example.com" \
  -H "Content-Type: application/json" \
  --data '{"ok":true}'`
#### Expected:
`access-control-allow-origin: *`
#### Actual:
no access-control-allow-origin header present

Preflight requests do include the header, which is why a preflight-only test misses the bug:
`curl -i \
  -X OPTIONS "$WORKER_URL/" \
  -H "Origin: https://example.com" \
  -H "Access-Control-Request-Method: POST" \
  -H "Access-Control-Request-Headers: content-type"`

sam-goodwin:
Is this on latest? We have this test, is it not testing the right thing? because it passes

https://github.com/alchemy-run/alchemy-effect/blob/7a2329904ad8f5530d42f58e71b25570fc034213/packages/alchemy/test/Cloudflare/Workers/HttpApi.test.ts#L79-L100

lucasthevenet:
Preflight request works fine. The issue is cors headers are not being returned with other methods.

[...]

PR #404, opened 2026-05-21
fix(cloudflare): run HttpServer pre-response handlers correctly

lucasthevenet:
Solves #402 

### Cause
Effect HTTP middleware can register pre-response handlers that run after the route handler returns but before the response is sent. CORS uses this for normal requests: the handler produces the response, then a pre-response handler attaches headers like `access-control-allow-origin`.
The Cloudflare Worker HTTP adapter converted the handler's response directly into a Web `Response` without going through Effect's handled-response lifecycle, so those pre-response handlers never ran.
`OPTIONS` preflight kept working because CORS short-circuits preflight with a fully decorated response and does not rely on the pre-response handler path. `GET` / `POST` responses silently went out without CORS headers.

### Solution
Route the handler through `HttpEffect.toHandled` and convert the final, post-pre-response-handler `HttpServerResponse` to a Web `Response` from inside the `handleResponse` callback. Use `scopeTransferToStream` so streaming bodies outlive the request scope, and pass `withoutBody: request.method === "HEAD"` to match the standard Effect HTTP adapter.
The previous `catchCause` fallback around the handler was removed: `Http.safeHttpEffect` already converts handler causes into safe `500` / `499` responses, and `toHandled` has error type `never` after that, so the fallback was unreachable and would have exposed raw error messages if it ever ran.

### Added
- Regression test that sends an actual `POST` with an `Origin` header and asserts `access-control-allow-origin` is present on the response. The existing preflight-only test did not catch this bug.


lucasthevenet:
Note: I'm not fully convinced it's the correct approach,`toHandled` was used in the adapter before and was intentionally removed in a later refactor, so it's worth confirming that bringing it back is the right call here rather than handling pre-response handlers some other way.

packages/alchemy/test/Cloudflare/Workers/HttpApi.test.ts
sam-goodwin, line 1: all these tests are now broken. Going to dig in and see if i can fix it

lucasthevenet:
toHandled seems to ignore completely what is being returned inside the callback and always returns void. I guess that is the reason for the separate variable assignment.
image

packages/alchemy/test/Cloudflare/Workers/HttpApi.test.ts
lucasthevenet: After some changes it looks like the HttpApi tests now pass correctly

sam-goodwin: It breaks `RpcHttp.test.ts`

lucasthevenet: If possible, could you share the test failure outputs? They aren't failing for me, even on a clean run.

sam-goodwin: That is interesting. I'll take a second look

packages/alchemy/test/Cloudflare/Workers/RpcHttp.test.ts
john-royal, line 1: @sam-goodwin I tracked the test issue you had down to an eventual consistency problem - some of the test cases in here retried on failure, others did not.

I updated this so it pings the worker until it's ready in `beforeAll`, and removed the retries elsewhere since they're no longer needed. Passes consistently.

Is this ok with you? If yes I'll merge.

PR #407, opened 2026-05-21
feat(cloudflare/vectorize): add Vectorize index as a bindable resource

DavidJFelix:
Add a Cloudflare Vectorize index resource modeled on D1, with a deploy-time Worker binding and a runtime Effect-native client.

- VectorizeIndex resource (create/read/delete; immutable so any config change replaces) backed by the distilled vectorize client
- IndexBinding attaches a `{ type: "vectorize" }` binding to a Worker
- VectorizeConnection Binding.Service + Policy wrap the runtime client (describe/query/queryById/insert/upsert/deleteByIds/getByIds)
- register provider + policy, add to WorkerBinding union and exports
- resource lifecycle tests + end-to-end Worker binding test

https://claude.ai/code/session_01FdfqB2qdRy9g4pGuzVqrTp


packages/alchemy/test/Cloudflare/Vectorize/fixtures/vectorize-worker.ts
sam-goodwin, line 1: move to a fixtures/ folder (like we do in test/Cloudflare/Workers/)

packages/alchemy/test/Cloudflare/Vectorize/fixtures/vectorize-worker.ts
```diff
@@ -0,0 +1,78 @@
...
+
+        if (request.method === "POST" && url.pathname === "/upsert") {
+          const mutation = yield* vec.upsert([
+            { id: "a", values: [0.1, 0.2, 0.3], metadata: { kind: "first" } },
+            { id: "b", values: [0.9, 0.8, 0.7], metadata: { kind: "second" } },
+            { id: "c", values: [0.4, 0.5, 0.6], metadata: { kind: "third" } },
+          ]);
+          return yield* HttpServerResponse.json({
+            mutationId: mutation.mutationId,
+          });
```
sam-goodwin, line 77: Nice, this is awesome

sam-goodwin (commented):
You need to update `InferEnv.ts` and `WorkerAsyncBindings.ts`

DavidJFelix:
Think I handled the feedback LMK. I'm a little new to effect and mostly just working with claude on this.

packages/alchemy/src/Cloudflare/Vectorize/VectorizeIndex.ts
sam-goodwin, line 1: Are we also missing VectorizeMetadataIndex?

Take a look at alchemy v1's:
https://github.com/alchemy-run/alchemy/blob/main/alchemy/src/cloudflare/vectorize-metadata-index.ts

packages/alchemy/src/Cloudflare/Vectorize/VectorizeIndex.ts
```diff
@@ -0,0 +1,249 @@
...
+import type { Providers } from "../Providers.ts";
+
+export type DistanceMetric = "cosine" | "euclidean" | "dot-product";
+
+export type VectorizePreset =
+  | "@cf/baai/bge-small-en-v1.5"
+  | "@cf/baai/bge-base-en-v1.5"
+  | "@cf/baai/bge-large-en-v1.5"
+  | "openai/text-embedding-ada-002"
+  | "cohere/embed-multilingual-v2.0";
```
sam-goodwin: always add `| (string & {})` so that we don't block users when cloudflare adds more

packages/alchemy/src/Cloudflare/Vectorize/VectorizeIndex.ts
DavidJFelix: Yeah that's missing from v2 as well. I overlooked it because I got snagged trying to bind vectorize and the metadata index doesn't require an additional binding. I haven't really used vectorize yet because of the prior snag and this sidetrack adventure to try to upstream some changes ;) Thanks for the catch, adding it now.

DavidJFelix: Added the resource -- working on some tests for it. Looking at a normal unit test and maybe spot checking a few things since it's a dependent resource. I'm thinking:

* what happens to metadata indexes if we delete the vectorize database they're attached to?
* what happens to metadata indexes if we update the vectorize database or replace it?

sam-goodwin: The metadata index will always be deleted first because it is tied to the index as a downstream dependency. 

You can't write code that drops the index without also removing the metadata index.

PR #408, opened 2026-05-21
fix(cloudflare): harden local durable object handling

john-royal:
The `LocalWorkerProvider` had a few issues regarding durable objects:
1. In the `durableObjectNamespaces` output, the `className` and `namespaceId` were reversed (see [Discord](https://discord.com/channels/1359694195782320389/1505638957168459866/1506693333379059843)).
2. We were using the binding name as a `namespaceId`, which can cause a mismatch at runtime if the binding name and the `className` are different. I updated the `cloudflare-dev` example to test this, and set the `namespaceId` to a combination of `${id}-${className}` so it's unique and deterministic.
3. DO bindings from workers other than the current one were being registered as part of the current one. This wouldn't cause an error, but it would mean that data would be out of sync between workers. I added a check of `binding.scriptName` to catch that.


examples/cloudflare-dev/src/AsyncWorker.ts
```diff
@@ -15,7 +15,7 @@ export default {
       const instance = (await WebAssembly.instantiate(wasm)) as AddInstance;
       return Response.json({ result: instance.exports.add(3, 4) });
     }
-    const counter = env.Counter.getByName("my-counter");
+    const counter = env.COUNTER.getByName("my-counter");
```
sam-goodwin, line 18: Is uppercase a requirement?

packages/alchemy/src/Cloudflare/Workers/LocalWorkerProvider.ts
```diff
@@ -196,8 +196,23 @@ export const LocalWorkerProvider = () =>
...
+              binding.type === "durable_object_namespace" &&
+              // The `durableObjectNamespaces` property is only used to declare DOs in this worker.
+              // Otherwise, it's a cross-worker durable object binding, which cloudflare-runtime handles automatically.
+              (!binding.scriptName || binding.scriptName === name)
+            ) {
+              if (!binding.className) {
+                return yield* new WorkerValidationError({
+                  message: `Durable object namespace ${binding.name} must have a className`,
+                  value: binding,
+                });
```
sam-goodwin: Why do we even allow it to be constructed incorrectly? Is it meant to default to id?

examples/cloudflare-dev/src/AsyncWorker.ts
john-royal: No, this is just to test one of the edge cases that this fixes.

The problem that's fixed in `LocalWorkerProvider` is that if the class name and the binding name don't match, this would cause an error in the local runtime. So I added this to test that this edge case is handled.

sam-goodwin: That sounds like confusing behavior. Why does the class name and binding name need to match ?

packages/alchemy/src/Cloudflare/Workers/LocalWorkerProvider.ts
john-royal: The type comes from the API as modeled by `@distilled.cloud/cloudflare`. Probably one of those cases where Cloudflare marks everything as optional for no reason, and we just inherited it downstream.

packages/alchemy/src/Cloudflare/Workers/LocalWorkerProvider.ts
```diff
@@ -196,8 +196,23 @@ export const LocalWorkerProvider = () =>
...
-            if (binding.type === "durable_object_namespace") {
-              durableObjectNamespaces[binding.name] = binding.className!;
+            if (
+              binding.type === "durable_object_namespace" &&
+              // The `durableObjectNamespaces` property is only used to declare DOs in this worker.
+              // Otherwise, it's a cross-worker durable object binding, which cloudflare-runtime handles automatically.
+              (!binding.scriptName || binding.scriptName === name)
+            ) {
+              if (!binding.className) {
+                return yield* new WorkerValidationError({
```
sam-goodwin: Does this die or fail? It's not an error we plan to handle, right? Just a named error to help understanding?

packages/alchemy/src/Cloudflare/Workers/LocalWorkerProvider.ts
sam-goodwin: Can we fix it to default to the logical id like the binding name? Or make it required

packages/alchemy/src/Cloudflare/Workers/LocalWorkerProvider.ts
john-royal: Correct, should I wrap in `Effect.die`?

sam-goodwin: I think so, yeah?

examples/cloudflare-dev/src/AsyncWorker.ts
john-royal: It doesn't need to match, but there was a bug where we were setting:

```
durableObjectNamespaces[binding.name] = binding.className
```

instead of

```
durableObjectNamespaces[binding.className] = <namespace id>
```

Which would cause an error at runtime, but only if `binding.name` didn't match `binding.className`, because we would use the object key to retrieve the class name:

```ts
const toRuntimeDurableObjectNamespaces = (
  namespaces: Record<string, string>,
): RuntimeDurableObjectNamespace[] => {
  return Object.entries(namespaces).map(([className, namespaceId]) => ({
    className,
    uniqueKey: namespaceId,
    sql: true,
  }));
};
```

The namespace ID can be whatever you want. The bug is we were putting the `className` in the wrong place.

packages/alchemy/src/Cloudflare/Workers/LocalWorkerProvider.ts
john-royal: Fix the upstream `className` type in distilled, or fix here?

sam-goodwin: Is distilled wrong?

examples/cloudflare-dev/src/AsyncWorker.ts
sam-goodwin: I still don't understand why we are validating that they are the same if we want them to be allowed to be different?

sam-goodwin: I'm probably being dumb. Bit sick today

packages/alchemy/src/Cloudflare/Workers/LocalWorkerProvider.ts
john-royal: The type of `binding` is `WorkerBinding`, which is an alias in Alchemy for `workers.PutScriptRequest["metadata"]["bindings"]`.

The `durable_object_namespace` type from there is:

```ts
{
  name: string;
  type: "durable_object_namespace";
  className?: string;
  environment?: string;
  namespaceId?: string;
  scriptName?: string;
}
```

I can't tell if this is a mistake. I don't think you can construct a DO binding without a class name. But I could be wrong, and `class_name` is also shown as optional in the [API docs](https://developers.cloudflare.com/api/resources/workers/subresources/scripts/methods/update#(resource)%20workers.scripts%20%3E%20(method)%20update%20%3E%20(params)%200%20%3E%20(param)%20metadata%20%3E%20(schema)%20%3E%20(property)%20bindings%20%3E%20(items)%20%3E%20(variant)%209):
image

sam-goodwin: Let's start by fixing it here at least? Maybe cloudflare spec is wrong (would not be surprised if they were just lazy and didn't use a union type or something )

john-royal: Right now `WorkerBinding` is a direct alias for the type from distilled:

```ts
// packages/alchemy/src/Cloudflare/Workers/WorkerBinding.ts
export type WorkerBinding = Exclude<
  workers.PutScriptRequest["metadata"]["bindings"],
  undefined
>[number];
```

That type is used by all worker bindings, not just DO.

What do you want me to do? We can:
a) Update distilled
b) Vendor in the whole union type, for every binding, and then have to update in multiple places going forward
c) Keep the type alias, but override it for DOs with a stricter one (e.g. `Omit<workers.PutScriptRequest["metadata"]["bindings"][number], { type: "durable_object_namespace" }> & { /* add our own DO type here */ }`

Any other ideas?

john-royal: I *think* the Cloudflare spec is wrong, but I don't know for sure that it is, and I don't know what it should be instead. That makes me a bit reluctant to change the type in distilled, even though I doubt that it would be a problem.

examples/cloudflare-dev/src/AsyncWorker.ts
john-royal: We are not validating that they're the same.

I changed the example to verify that it works when they're different. This is to verify the bug fix I made in `LocalWorkerProvider`.

Does that make sense?

packages/alchemy/src/Cloudflare/Workers/LocalWorkerProvider.ts
sam-goodwin: Can't you just intersect it to override without omit? Don't want to bike shed. Simplest solution for now

john-royal: That doesn't seem to work:

image


```ts
export type WorkerBinding =
  | Exclude<workers.PutScriptRequest["metadata"]["bindings"], undefined>[number]
  | {
      name: string;
      type: "durable_object_namespace";
      className: string;
      environment?: string;
      namespaceId?: string;
      scriptName?: string;
    };
```

Should I just use `Exclude` instead?

john-royal: Updated to use `Exclude` since the intersection type didn't work. Let me know if this is ok.

sam-goodwin: Hmm maybe just vendor it in (lol sorry for run around) because then it can be documented?

john-royal: I included a comment explaining what this is, and I feel like it's better than having to either:
- update the types every time
- get a weird type error if the API type changes and ours doesn't match, which Claude will probably `as any` away

```ts

export type WorkerBinding =
  // This is a workaround so that `className` is required in the type of `durable_object_namespace` bindings.
  | Exclude<
      NonNullable<workers.PutScriptRequest["metadata"]["bindings"]>[number],
      { type: "durable_object_namespace" }
    >
  | {
      name: string;
      type: "durable_object_namespace";
      className: string;
      environment?: string;
      namespaceId?: string;
      scriptName?: string;
    };
```

But I'm happy to just vendor it if that's what you'd prefer?

john-royal: Update: there's a `WorkerMetadataBinding` type from workers-sdk where `className` is required. So that might be our answer on whether distilled is correct ([ref](https://github.com/cloudflare/workers-sdk/blob/07336888e0bc82925e4023f5b72a0062f10d77b8/packages/workers-utils/src/types.ts#L56)).

john-royal: Relevant patch to distilled: https://github.com/alchemy-run/distilled/pull/305 (apologies - added mention to wrong PR!)

PR #411, opened 2026-05-22
fix(bundle): support Vite-style ?raw imports

john-royal:
Inline file contents as a string when imported with the Vite-style `?raw` suffix.

```ts
import schema from "./schema.sql?raw";
//        ^ string — file contents read as UTF-8
```

Implemented as a new built-in rolldown plugin (`alchemy:raw`) enabled by default in `builtInPlugins` — no option, nothing to configure.

`?url` and `?inline` are intentionally omitted: both rely on Vite's browser asset pipeline (URL serving, `data:` fallback) that has no analogue in the server-side bundles produced here (Workers, Lambda).

### DIFF
--- a/packages/alchemy/src/Bundle/Bundle.ts
+++ b/packages/alchemy/src/Bundle/Bundle.ts
@@ -11,6 +11,7 @@ import {
   type BundleAnalyzerPluginOptions,
 } from "./BundleAnalyzerPlugin.ts";
 import { purePlugin, type PurePluginOptions } from "./PurePlugin.ts";
+import { rawPlugin } from "./RawPlugin.ts";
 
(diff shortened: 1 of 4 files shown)


sam-goodwin (commented):
Is there a way to test that is an e2e fixture instead of spot checking these internals ?

john-royal:
> Is there a way to test that is an e2e fixture instead of spot checking these internals ?

The test actually does run `Bundle.build` a few times, so it's not just unit-testing internals. I'll reorder the tests to make this clearer.

Or are you asking for it to be tested in a worker or something like that? Feel like that's not necessary since it's just a bundling change, but happy to do it if you want.

PR #415, opened 2026-05-22
fix(cloudflare): register non-fetch ExportedHandler methods from Worker shape

### LINKED ISSUE #414: Worker shape methods for ExportedHandler (email, queue, scheduled) not registered as event listeners
## Bug

When returning non-`fetch` handler methods on the Worker init shape (e.g. `email`, `queue`), they are stored in `userShape` for RPC but never registered as event listeners. The `dispatch` function then fails with `No event handler found for event type 'email'`.

## Reproduction

```typescript
  "Worker",
  { main: import.meta.path },
  Effect.gen(function* () {
    return {
      fetch: Effect.gen(function* () { /* ... */ }),
      email: (message: ForwardableEmailMessage) =>
        Effect.gen(function* () {
          // handle inbound email
        }),
    };
  }),
);
```

Sending an email to the Worker produces:

```
Error: No event handler found for event type 'email'
  at WorkerRuntimeContext.ts:141
```

## Root cause

In `WorkerRuntimeContext.ts`, `serve()` only registers the `fetch` handler as a listener:

```typescript
serve: (handler, options) => {
  if (options?.shape) userShape = options.shape;
  return ctx.listen(makeRequestHandler(handler)); // only fetch
}
```

And in `Platform.ts`, only `fetch` is extracted from the shape:

```typescript
impl?.fetch
  ? runtimeContext.serve?.(impl.fetch, { shape: impl })
  : Effect.void
```

Other `ExportedHandlerMethods` (`email`, `queue`, `scheduled`, etc.) on the shape are ignored. The bridge creates dispatch entries for all of them, but no listeners are registered for non-fetch methods.

## Suggested fix

`serve()` (or the Platform init) should iterate `ExportedHandlerMethods` on the shape and register a listener for each one present, similar to how `makeRequestHandler` wraps `fetch`. Each listener would check `isWorkerEvent(event) && event.type === methodName` and call the corresponding shape method with `event.input`.

## Workaround

Use `RuntimeContext` + `ctx.listen()` manually:

```typescript
const ctx = yield* RuntimeContext;
yield* (ctx as any).listen((event: any) => {
  if (!isWorkerEvent(event) || event.type !== "email") return;
  return handleEmail(event.input as ForwardableEmailMessage);
});
```

Mkassabov:
## Summary

- When a Worker returns non-fetch `ExportedHandler` methods (`email`, `queue`, `scheduled`, etc.) on its shape, `serve()` now registers event listeners for them alongside the fetch handler.
- Previously only the fetch listener was registered via `makeRequestHandler`, so email and other event types caused `"No event handler found for event type 'email'"` errors at runtime.
- The fix filters `ExportedHandlerMethods` for any non-fetch methods present on the user's shape and registers a listener for each that dispatches `event.input` to the corresponding handler function.

Closes #414

## Test plan

- [ ] Deploy a Worker with an `email` handler on the shape
- [ ] Send an inbound email and verify it's dispatched to the handler instead of erroring
- [ ] Verify fetch requests still work as before
- [ ] Verify Workers without non-fetch handlers are unaffected (`nonFetchMethods.length === 0` early return)

### DIFF
--- a/packages/alchemy/src/Cloudflare/Workers/RpcAsync.ts
+++ b/packages/alchemy/src/Cloudflare/Workers/RpcAsync.ts
@@ -2,9 +2,17 @@ import type * as Effect from "effect/Effect";
 import type * as Stream from "effect/Stream";
 import type { Rpc } from "../../Rpc.ts";
 import { isRpcErrorEnvelope, isRpcStreamEnvelope } from "../Bridge.ts";
+import {
+  ExportedHandlerMethods,
+  type ExportedHandlerMethod,
+} from "./Worker.ts";
+
+const handlerMethodSet = new Set<string>(ExportedHandlerMethods);
 
 export type RpcAsync<Shape> = {
-  [K in keyof Shape as K extends "fetch" ? never : K]: Shape[K] extends (
+  [K in keyof Shape as K extends ExportedHandlerMethod
+    ? never
+    : K]: Shape[K] extends (
     ...args: infer A
   ) => Effect.Effect<infer T, any, any>
     ? (...args: A) => Promise<T>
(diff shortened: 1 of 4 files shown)


sam-goodwin:
We also need to exclude them from the RPC interface for the worker

sam-goodwin (commented):
Tests?

PR #417, opened 2026-05-22
docs(guides/secrets): clarify Effect Worker vs async Worker / Vite

Mkassabov:
Rewrites the secrets guide around the real axis of confusion: which env-var helper to use depends on **how the Worker is written**, not what kind of value it is.

### What changed

- Decision table up top: Effect Worker → `Alchemy.Secret/Variable`; async Worker / Vite → `yield* Config.string/redacted` inside `env: { … }`.
- New "Async Worker" section showing the Stack-level shape:

  ```ts
  env: {
    OPENAI_API_KEY: yield* Config.redacted("OPENAI_API_KEY"),
    VITE_APP_ENV:   yield* Config.string("VITE_APP_ENV"),
  },
  ```

- Caution block calling out the exact footgun (a `Config` value in `env` ships as JSON), with the diff:

  ```diff
    env: {
  -   VITE_APP_ENV: Config.string("VITE_APP_ENV"),
  +   VITE_APP_ENV: yield* Config.string("VITE_APP_ENV"),
    },
  ```

  Plus: if one entry needs `yield*`, every entry does.

- "Vite / static-asset apps" section noting `Alchemy.Secret/Variable` aren't an option — no edge init phase to attach an accessor to.

### Why

Users hitting `Alchemy.Variable` in an async-Worker `env` block were getting silently-wrong bindings (the `Config` object shipped as JSON) with no typed autocomplete on the env shape. The old guide only documented the Effect-Worker path, so this wasn't discoverable.


website/src/content/docs/guides/secrets.mdx
```diff
@@ -84,78 +89,202 @@ declares the binding as `type: "secret_text"`.
...
 
-The 2-arg form replaces `Config.redacted(name)` with whatever you
-want. Useful when the env var on disk has a different name, or when
-the value comes from a different system entirely:
+Read them inside the async handler the normal Cloudflare way:
 
 ```typescript
-import * as Config from "effect/Config";
+// src/worker.ts
+import type { WorkerEnv } from "../alchemy.run.ts";
```
sam-goodwin, line 146: Did we show this

website/src/content/docs/guides/secrets.mdx
```diff
@@ -84,78 +89,202 @@ declares the binding as `type: "secret_text"`.
...
+ import * as Effect from "effect/Effect";
+
+ export default Alchemy.Stack(
+   "App",
+   { providers: Cloudflare.providers(), state: Cloudflare.state() },
+   Effect.gen(function* () {
+     const worker = yield* Cloudflare.Worker("Worker", {
+       main: "./src/worker.ts",
++      env: {
++        OPENAI_API_KEY: yield* Config.redacted("OPENAI_API_KEY"),
```
sam-goodwin, line 132: Make sure to emphasize why the redacted is important

website/src/content/docs/guides/secrets.mdx
sam-goodwin: You'll need to pull the worker out of the Stack closure and export the type from it

sam-goodwin (approved):

sam-goodwin:
Superceded by https://github.com/alchemy-run/alchemy-effect/pull/445 ?

PR #422, opened 2026-05-23
fix(alchemy): move CLI, test, and DB driver deps out of dependencies

danieljvdm:
Reclassify dependencies that every consumer was installing but only a subset actually needs. A production install of `alchemy` no longer pulls React 19 + ink, a test framework, or three database drivers unless the relevant provider/CLI is actually used.

### `react`, `ink` → `devDependencies`

Only imported under `src/Cli/tui/**` and bundled into `bin/exec.js` by `tsdown` (matching its own stated intent), so they don't need to be runtime deps. `pathe` stays a dependency since `bin/cli.js` imports it unbundled.

### `@effect/vitest` → optional peer + devDep

Only used by `src/Test/Vitest.ts`, a helper consumers use in *their own* tests. No reason to ship a test framework into every production install.

### `pg`, `mysql2`, `@libsql/client` → optional peers + devDeps

Each is reachable only through a specific provider (`Neon` / `Planetscale` / `SQLite`) and absent from `State`/`Cli`/`Stack` core, so a Cloudflare- or AWS-only consumer no longer installs all three.

```diff
 "dependencies": {
-  "@effect/vitest": "catalog:",
-  "@libsql/client": "catalog:",
-  "ink": "^6.3.1",
-  "mysql2": "^3.15.3",
-  "pg": "^8.13.0",
-  "react": "^19.2.0",
 }
 "peerDependenciesMeta": {
+  "@effect/vitest": { "optional": true },
+  "@libsql/client": { "optional": true },
+  "mysql2": { "optional": true },
+  "pg": { "optional": true }
 }
```

`bun tsc -b` passes.


sam-goodwin:
We've discussed this before: https://github.com/alchemy-run/alchemy-effect/pull/312

Our plan is actually to split things into packages to solve this problem. Moving everything to peers makes for a high friction onboarding DX.

PR #428, opened 2026-05-24
feat(cloudflare/Worker): support for custom tld in `Vite` and `LocalWorkerProvider`

utopyin:
this is just a proof of concept, far from production grade

todo
- [ ] handle missing permissions (modifying /etc/hosts probably requires sudo)
- [ ] support windows hosts path

### DIFF
--- a/packages/alchemy/src/Cloudflare/Workers/LocalWorkerProvider.ts
+++ b/packages/alchemy/src/Cloudflare/Workers/LocalWorkerProvider.ts
@@ -1,8 +1,8 @@
 import {
-  layerLocalProxy,
-  layerRuntime,
   Runtime,
   RuntimeError,
+  layerLocalProxy,
+  layerRuntime,
   type BindingHook,
   type BindingServices,
   type HyperdriveOrigin,
@@ -48,8 +48,9 @@ import type * as Bundle from "../../Bundle/Bundle.ts";
 import { InstanceId } from "../../InstanceId.ts";
 import * as RpcProvider from "../../Local/RpcProvider.ts";
 import type { ResourceBinding } from "../../Resource.ts";
-import { Stack } from "../../Stack.ts";
+import { Stack, type StackSpec } from "../../Stack.ts";
 import { Stage } from "../../Stage.ts";
+import { syncLocalHosts } from "../../Util/LocalHosts.ts";
 import { CloudflareEnvironment } from "../CloudflareEnvironment.ts";
 import type { WorkerAssetsConfig, WorkerProps } from "../Workers/Worker.ts";
 import { getCompatibility } from "./Compatibility.ts";
@@ -309,6 +310,18 @@ export const LocalWorkerProvider = () =>
 
       const rootScope = yield* Effect.scope;
       const workerdScopes = new Map<string, Scope.Closeable>();
+      const localHostnames = new Map<string, string>();
+      const localHostsMarker = `cloudflare:${stack.stage}:${stack.name}`;
+      const syncStackHosts = () =>
+        stack.localDomain === undefined
+          ? Effect.void
+          : syncLocalHosts(localHostsMarker, localHostnames.values());
+      if (stack.localDomain !== undefined) {
+        yield* Scope.addFinalizer(
+          rootScope,
+          syncLocalHosts(localHostsMarker, []),
+        );
+      }
 
       const context = yield* Effect.context<RuntimeServices>();
       const instances = new Map<
@@ -331,7 +344,13 @@ export const LocalWorkerProvider = () =>
       }) {
         const { id, props, bindings } = options;
         const config = yield* buildConfig(options);
-        const url = yield* localProxy.registerWorker(id);
+        const proxyUrl = yield* localProxy.registerWorker(id);
+        const url = toLocalWorkerUrl(id, proxyUrl, stack.localDomain);
+        const localHostname = new URL(url).hostname;
+        if (stack.localDomain !== undefined) {
+          localHostnames.set(id, localHostname);
+          yield* syncStackHosts();
+        }
         if (props.vite) {
           const devServer = yield* Vite.viteDev(
             props.vite.rootDir,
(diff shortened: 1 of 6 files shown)


Mkassabov:
After talking about this internally I don't think we should be setting the hosts file since it requires sudo/admin permissions.

We should read the hosts file and if we dont find the TLD in the hosts then we show you an error with what to add to your hosts file to make it work

CC @john-royal

sam-goodwin:
A manual step is usually terrible for IaC. Maybe walk them through interactively?

utopyin:
yeah i agree requiring sudo is not great, maybe we could have a dedicated Alchemy command that sets up the hosts file for us based on the Alchemy "plan"

it would be a command you run once with sudo and the error thrown when the TLD is not found in the hosts could reference this command directly

still a manual step though, @sam-goodwin are you saying we could be prompted into setting up the hosts interactively when running alchemy dev?

Mkassabov:
> A manual step is usually terrible for IaC. Maybe walk them through interactively?

This isn't for the IaC part of alchemy it's for dev mode. The problem isn't being scared that we touch the hosts file, it's that it requires sudo and having the user restart dev mode with sudo, or always running with sudo is bad UX, interactivity doesn't fix that.

utopyin:
@Mkassabov i dont see why the interactive step couldn't spawn a child process with sudo and prompt the user to enable privileged access for this child process using `stdio: "inherit"`

```ts
import { spawn } from "node:child_process";

spawn("echo", ["This is not priviledged"], { stdio: "inherit" });
spawn("sudo", ["echo", "Hello sudo World!"], { stdio: "inherit" }); // root privilege scoped to this child process
```

Mkassabov:
> i dont see why the interactive step couldn't spawn a child process with sudo and prompt the user to enable privileged access for this child process using `stdio: "inherit"`

IIRC windows doesn't have an equivalent to this we can actually rely on since sudo.exe isn't always enabled; so we're making platform specific behavior for unix only and wed need a windows fallback anyway.

This would introduce a blocking interactive step when it's not actually blocking; IMO it should be a warning.

Also the user doesn't know what we actually ran to warrent sudo, so they are blindly giving us sudo access; which seems like poor security practice.

utopyin:
totally get the windows stuff, but i would not mind having a best effort solution for unix platform and a manual fallback for windows. This is purely a DX issue imo, so i feel like platform specific behavior is fine. Would be interesting to know how `portless` does it, since they do sync the `/etc/hosts` file as well

> This would introduce a blocking interactive step when it's not actually blocking; IMO it should be a warning.

it is kinda blocking IMO since without it the custom TLD property could not be respected

> Also the user doesn't know what we actually ran to warrent sudo, so they are blindly giving us sudo access; which seems like poor security practice.

i dont understand what you mean by "blindly"? We could explicitly warn why we need sudo permission, since the target audience is technical. They could always refuse, and add the hosts themselves before running `alchemy dev` again.

the warn could look something like:
```
We need sudo access to modify /etc/hosts in order to add the custom TLD ".test".
If you prefer not giving Alchemy sudo access temporarily, you can those 2 lines to `/etc/hosts`:

127.0.0.1 backend.web.test
127.0.0.1 website.web.test 

Grant sudo access? y/N
```

the wording could be better, but english is hard

Mkassabov:
> Would be interesting to know how portless does it, since they do sync the /etc/hosts file as well

I think john said portless requires you run sudo at the root.

> it is kinda blocking IMO since without it the custom TLD property could not be respected

It's not, alchemy is able to start the underlying processes just fine. Vite also doesn't update hosts for you and it's just doesn't work even if it's in `allowedHosts`. That seems to be standard behavior.

If you want a concrete example of how it's blocking: a minor backend update to just a worker that can be checked against localhost, is now blocked by a hosts update that isn't needed for the change (super minor, but it is friction there)

> i dont understand what you mean by "blindly"? We could explicitly warn why we need sudo permission, since the target audience is technical. 

Just because our comment says something doesn't garuntee the command we run does that thing. I'm not a macos/linux user but from my understanding it's bad practice to elevate mid process like that since you're running elevated commands the user doesn't know about. Given we have people complaining about the security implications of peer dependencies and actions being pinned to specific commits I feel like doing the more secure thing here is better (unless I don't understand the Linux security model, which is very possible)

> They could always refuse, and add the hosts themselves before running alchemy dev again.

User doesn't need to restart alchemy dev, just update hosts file (e.g. by running a sudo command we give them in a new terminal). I think copying a command from a warning and pasting it into a new terminal is low enough friction here.

utopyin:
fair points!

utopyin:
if we can have at least an alchemy command to update the hosts file matching the stack’s config that would be nice

Mkassabov:
> fair points

I am open to being wrong here;

> if we can have at least an alchemy command to update the hosts file matching the stack’s config that would be nice

we can have a `sudo alchemy update-hosts` I guess. is there something wrong with the warning printing `echo "127.0.0.1   things.alchemy.local things2.alchemy.local" | sudo tee -a /etc/hosts` (on win: `Add-Content -Path "$env:SystemRoot\System32\drivers\etc\hosts" -Value "127.0.0.1`tthings.alchemy.local things2.alchemy.local"; ipconfig /flushdns`); just concern that with a lot of resource the command can end up too long or is it something else?

utopyin:
> I am open to being wrong here;

your pushbacks are thoughtful, and i 100% trust your judgment on this kind of stuff.
I'm probably a bit too DX focused for my own use cases, and my understanding is that what i see as great DX could be a deal breaker for others more security-focused, so i totally get that

i think that command could use some formatting with `\` to break it into blocks, but since the only job of that command it to sync hosts i think we're fine with a long-ish print 

i can imagine adding a `--yes` option so that people can script it and run it on file changes, so we need to ensure it's idempotent which may add a bit of complexity. Then maybe it would make sense for the warning to simply explain what it's doing more than printing the script.

we could also check if the hosts are already configured, so that it succeeds fast

sam-goodwin:
I think this is no longer relevant because we don't use localhost subdomains anymore. Please re-open if there's stil la problem here.

PR #432, opened 2026-05-25
feat(cloudflare): worker.url now tries to infer canonical url & domain order is preserved

Mkassabov:
`domains` property on worker comes from a cloudflare api call, prior to this pr that means the order wasn't deterministic.

people do things like this:
```ts
  const worker = yield* SolZeroWorker
  const url = Output.map(Output.all(worker.domains, worker.url), ([domains, workerUrl]) => {
    const primaryDomain = domains[0]
    return primaryDomain ? `https://${primaryDomain.hostname}` : workerUrl
  })
```

which can produce different results based on cf outputs.

now we sort domains to mirror the input order and include other domain (just as the workers.dev url).

Worker.url now tries to infer the first value in domains as the canonical url

### DIFF
--- a/packages/alchemy/src/Cloudflare/Workers/LocalWorkerProvider.ts
+++ b/packages/alchemy/src/Cloudflare/Workers/LocalWorkerProvider.ts
@@ -378,7 +378,7 @@ export const LocalWorkerProvider = () =>
           url,
           tags: [],
           durableObjectNamespaces: config.durableObjectNamespaces,
-          urls: [url],
+          domains: [url],
           crons: Array.from(
             new Set([...getCronBindings(bindings), ...(props.crons ?? [])]),
           ),
--- a/packages/alchemy/src/Cloudflare/Workers/LocalWorkerProvider.ts
+++ b/packages/alchemy/src/Cloudflare/Workers/LocalWorkerProvider.ts
@@ -378,7 +378,7 @@ export const LocalWorkerProvider = () =>
           url,
           tags: [],
           durableObjectNamespaces: config.durableObjectNamespaces,
-          domains: [],
+          urls: [url],
           crons: Array.from(
             new Set([...getCronBindings(bindings), ...(props.crons ?? [])]),
           ),
--- a/packages/alchemy/src/Cloudflare/Workers/Worker.ts
+++ b/packages/alchemy/src/Cloudflare/Workers/Worker.ts
@@ -276,7 +276,7 @@ export type Worker<Bindings extends WorkerBindings = any> = Resource<
     tags: string[] | undefined;
     durableObjectNamespaces: Record<string, string>;
     accountId: string;
-    urls: string[];
+    domains: string[];
     crons: string[];
     hash?: {
       assets: string | undefined;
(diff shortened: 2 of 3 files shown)


packages/alchemy/src/Cloudflare/Workers/Worker.ts
```diff
@@ -1903,7 +1903,14 @@ export const LiveWorkerProvider = () =>
...
                 service: workerName,
-              }).pipe(Effect.map((r) => r.result ?? [])),
+              }).pipe(
+                Effect.map(
+                  (r) =>
+                    r.result.sort((a, b) =>
+                      (a.hostname ?? "").localeCompare(b.hostname ?? ""),
+                    ) ?? [],
+                ),
+              ),
```
sam-goodwin: we should match the order the user provides, not sort them

packages/alchemy/src/Cloudflare/Workers/Worker.ts
```diff
@@ -1905,6 +1905,20 @@ export const LiveWorkerProvider = () =>
...
+            // is appended after the user-ordered ones.
+            const userOrder = normalizeDomains(olds?.domain);
+            const orderedDomains = [
+              ...userOrder.flatMap(
+                (h) => domainsList.find((d) => d.hostname === h) ?? [],
+              ),
+              ...domainsList.filter(
+                (d) => !d.hostname || !userOrder.includes(d.hostname),
+              ),
+            ];
```
sam-goodwin, line 1985: Should we use the list they give us?

Mkassabov: not sure I follow? `const userOrder = normalizeDomains(olds?.domain);` is the users order.
do you mean omit the ones the user didn't tell us about?

sam-goodwin: I'm surprised there's so much code. Why is it not just this?

```
domain: news.domain
```

Mkassabov: They'd never get the workers.dev domain then. Fetching the domains from cloudflare and sorting them based on the users input feels more accurate than just returning the users inputs back to them

packages/alchemy/src/Cloudflare/Workers/Worker.ts
```diff
@@ -1967,6 +1905,20 @@ export const LiveWorkerProvider = () =>
...
+            // flip between deploys. Drift (domains we don't know about)
+            // is appended after the user-ordered ones.
+            const userOrder = normalizeDomains(olds?.domain);
+            const orderedDomains = [
+              ...userOrder.flatMap(
+                (h) => domainsList.find((d) => d.hostname === h) ?? [],
+              ),
+              ...domainsList.filter(
+                (d) => !d.hostname || !userOrder.includes(d.hostname),
+              ),
```
sam-goodwin: put the subdomain in here if listDomains does not return it

packages/alchemy/src/Cloudflare/Workers/LocalWorkerProvider.ts
```diff
@@ -378,7 +378,7 @@ export const LocalWorkerProvider = () =>
           url,
           tags: [],
           durableObjectNamespaces: config.durableObjectNamespaces,
-          domains: [],
+          urls: [url],
```
sam-goodwin: I think id prefer to keep it as domains?

sam-goodwin: Reason is because `worker.` will be confusing when there are both url and urls. 

Seems a bit easier to reason about `url` and `domains`?

PR #434, opened 2026-05-25
fix(better-auth): better auth as peer dep

Mkassabov:


packages/better-auth/package.json
just-be-dev, line 1: @Mkassabov you should probably add it as a dev dependency too so it's still present for tests, etc.

sam-goodwin (approved):

packages/better-auth/package.json
Mkassabov: bun peer deps are installed by default: https://bun.com/docs/guides/install/add-peer

double checked and it does install it correctly

PR #445, opened 2026-05-27
feat(core): replace Alchemy.Secret and Alchemy.Variable with effect/Config

sam-goodwin:
This PR removes the complicated `Alchemy.Variable` and `Alchemy.Secret` helpers in favor of basing everything on `effect/Config`.

We are designing for Alchemy's init vs runtime phase approach so that a `Config` `yield*`ed in `init` will bind that Config to the Platform (Worker, Lambda, etc.)

```ts
  "Worker", 
  Effect.gen(function*() {
    const apiKey = yield* Config.redacted("API_KEY");
    
    // now you can use the API_KEY immediately (it is available both in the runtime and init phases, it's just a Config)
    // const client = createClient(apiKey)
    
    return {
      fetch: Effect.gen(function*() {
        // or use apiKey here in the runtime phase
      })
    } 
  })
)
```

Before, with `Alchemy.Secret`, we forced you to use the Accessor pattern which is cumbersome:
```ts
  "Worker", 
  Effect.gen(function*() {
    const API_KEY = yield* Alchemy.Secret("API_KEY");
   
    return {
      fetch: Effect.gen(function*() {
        const apiKey = yield* API_KEY // <- you had to yield it at runtime to get it
      })
    } 
  })
)
```

If you wanted to do any transformations, you had to do this redundant step:
```ts
yield* Alchemy.Secret("API_KEY", Config.string("API_KEY").pipe(
  // transformations
))
```

And not being able to use the value in the init phase means initializing things like clients required dealing with Outputs and Accessors. Which is pointless since the Config values are expected to be there both during init and runtime. 

So, we now just lean in to `yield* Config.string(..)` or `Config.redacted` and hook it with our own ConfigProvider that intercepts when a Config is evaluated in the context of a Platform and binds it to the runtime automatically.

Caution: this design requires the user understand init vs runtime phase (a mandatory requirement for using alchemy v2 well). Beware of the following footgun:

```ts
  "Worker", 
  Effect.gen(function*() {
    // oops - didn't yield it in the init phase
    return {
      fetch: Effect.gen(function*() {
        // if you use config only at runtime phase, the value will not have been bound to your Worker
        const apiKey = yield* Config.redacted("API_KEY");
      })
    } 
  })
)
```

You can however still do this and, for example, set them globally:
```ts
  "Worker", 
  { 
    main: import.meta.filename,
    env: {
      API_KEY: <some value>
    }
  },
  Effect.gen(function*() {
    // oops - didn't yield it in the init phase
    return {
      fetch: Effect.gen(function*() {
        // if you use config only at runtime phase, the value will not have been bound to your Worker
        const apiKey = yield* Config.redacted("API_KEY");
      })
    } 
  })
)
```

But this is not recommended. Instead, users using Effect workers should use the constructor pattern and avoid evaluating `Config` in their runtime parts of the code. That way, Config bindings are always captured by init and automatically flows from Layers to compose and create Workers.


agcty:
I like the ergonomics of it, plus it feels a bit more effect native. About the phases I don't think it's much of an issue as long as agents understand it well!

filipfalcon:
Nice. The more native it gets the better.

PR #455, opened 2026-05-27
fix(cloudflare): reduce unnecessary updates in alchemy dev

john-royal:
Workers and bindings that depend on the `workerName`, such as `Workflow`, would update unnecessarily in `alchemy dev`. This is for two reasons:
1. The `workerName` wasn't marked as stable, triggering an update in downstream resources.
2. Some bindings would be duplicated in the `diff` function, causing the hashing to return an inconsistent result, triggering an update in the Worker itself.


packages/alchemy/src/Cloudflare/Workers/LocalWorkerProvider.ts
```diff
@@ -378,16 +379,26 @@ export const LocalWorkerProvider = () =>
...
           const options = {
             id,
             props: news,
-            bindings: newBindings,
+            // Filter out duplicate bindings by sid so the hash is stable.
+            // TODO(sam): Should we do this in core?
+            bindings: newBindings.filter(
+              (binding, index) =>
+                newBindings.findIndex((b) => b.sid === binding.sid) === index,
+            ),
```
john-royal: @sam-goodwin Wondering if we need a change in core.

Explanation: in the `cloudflare-dev` example, `KV` is bound to the `Worker` and to `NotifyWorkflow`. This caused the `KV` binding to appear twice in the `bindings` array, but only in the `diff` handler (it's deduplicated in `reconcile`).

Deduplicating by `sid` here works, but I'm wondering if that should be happening in core instead.

sam-goodwin: That does sound like a thing to handle in core.

PR #465, opened 2026-05-29
feat(cloudflare): add DnsRecord resource

sam-goodwin:
Adds a `Cloudflare.DnsRecord` resource for create/update/delete of DNS records of any type (A, AAAA, CNAME, MX, TXT, SRV, …).

```ts
const record = yield* Cloudflare.DnsRecord("www", {
  zone: { zoneId: zone.zoneId },
  name: "www.example.com",
  type: "A",
  content: "192.0.2.1",
  proxied: true,
});
```

### Lifecycle

Follows the reconciler doctrine — observe (cached id → name+type scan), ensure (POST, tolerating Cloudflare's `81057 "already exists"` race by adopting the match), sync (PUT only on drift).

- `diff` replaces only when the zone changes; everything else is an in-place PUT.
- `delete` is idempotent (404 → no-op).
- `read` returns `Unowned(...)` on a name+type match so adoption is gated behind `--adopt`.

### Why raw HttpClient instead of the distilled SDK

The installed `@distilled.cloud/cloudflare@0.21.5` collapses its create/update request schema to `type: Schema.Literal("A")`, so the typed methods can only create A records. To support every record type, the provider calls the DNS REST API directly through the Effect `HttpClient` (auth via the `Credentials` service), parsing the `{ success, result, errors }` envelope into a tagged `DnsApiError`.


sam-goodwin:
Got bundled in with https://github.com/alchemy-run/alchemy-effect/pull/570

PR #470, opened 2026-05-30
fix(aws): deliver S3 bucket notifications to Lambda event sources

konopkov:
Two fixes that together make `S3.notifications(bucket).subscribe(...)` actually deliver object events to a Lambda. Independently, each leaves the trigger broken; both are required for an end-to-end S3 → Lambda event source.

### `fix(aws/lambda)`: malformed invoke action

The S3 → Lambda invoke permission used `lambda.InvokeFunction` (dot) instead of `lambda:InvokeFunction` (colon). Lambda `AddPermission` requires the `service:Action` form, so AWS rejects the call with `ValidationException`.

```diff
  yield* Permission(`AWS.Lambda.InvokeFunction(${bucket.LogicalId})`, {
-   action: "lambda.InvokeFunction",
+   action: "lambda:InvokeFunction",
    functionName: host.functionName,
    principal: "s3.amazonaws.com",
    sourceArn: bucket.bucketArn,
  });
```

Every sibling call site already uses the colon form (`Function.ts`, `EventBridge/ToLambda.ts`, `Lambda/InvokeFunction.ts`, `Lambda/TopicEventSource.ts`).

### `fix(aws/s3)`: notification binding never applied

`subscribe(...)` records a bucket binding carrying `notificationConfiguration`, but the Bucket provider's `reconcile` only ran `ensureBucketExists → syncBucketTags → syncBucketPolicy`. Nothing ever read `binding.data.notificationConfiguration` or called `putBucketNotificationConfiguration`, so the permission was created, state said "configured", but the bucket's notification config stayed empty and no trigger ever existed.

A new `syncBucketNotifications` helper (mirroring `syncBucketPolicy`) is called from `reconcile` right after `syncBucketPolicy`. It:

- collects declared Lambda targets from bindings;
- canonicalizes them (ignoring S3-assigned `Id`s and event order) to avoid deploy churn;
- merges with any existing Topic/Queue/EventBridge config, managing only its own Lambda targets;
- PUTs with `SkipDestinationValidation: true`, since the invoke-permission resource may apply after the bucket reconcile — skipping S3's synchronous test-invoke avoids an ordering failure, and the permission exists by the time events are delivered.

### DIFF
--- a/packages/alchemy/src/AWS/Lambda/BucketEventSource.ts
+++ b/packages/alchemy/src/AWS/Lambda/BucketEventSource.ts
@@ -119,7 +119,7 @@ export const BucketEventSourcePolicyLive = BucketEventSourcePolicy.layer.effect(
     ) {
       if (Lambda.isFunction(host)) {
         yield* Permission(`AWS.Lambda.InvokeFunction(${bucket.LogicalId})`, {
-          action: "lambda.InvokeFunction",
+          action: "lambda:InvokeFunction",
           functionName: host.functionName,
           principal: "s3.amazonaws.com",
           sourceArn: bucket.bucketArn,
--- a/packages/alchemy/src/AWS/Lambda/BucketEventSource.ts
+++ b/packages/alchemy/src/AWS/Lambda/BucketEventSource.ts
@@ -113,11 +113,23 @@ export const BucketEventSourcePolicyLive = BucketEventSourcePolicy.layer.effect(
       bucket: Bucket,
       {
         events: Events = ["s3:ObjectCreated:*"],
+        prefix,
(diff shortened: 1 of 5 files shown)


sam-goodwin (changes requested):
Thanks for the patch. Can you add a fixture test that exercises this end to end. Bind to a function, put object to the bucket, process the notification and store a completion message in s3. Have the test deploy and assert everyhting flows end to end by making http requests against the function.

konopkov:
Hi @sam-goodwin Thanks for taking care of this and adding the e2e fixture test  appreciate you closing the loop and merging it 🙏Alchemy is awesome!

PR #472, opened 2026-05-30
fix(cloudflare): two vite fixes (draft)

john-royal:
Two bugs in this PR. I'll split these out into a separate PR, but I'm grouping them here to test against a user repro: https://github.com/lucasthevenet/repro-alchemy

1. `vite dev` listens on port 1337 instead of 5173, which is non-default behavior for Vite, and can cause conflicts with our server (need to figure out why the retry logic doesn't apply here, but this should work as a patch)
2. When binding an `Output` value to an external Worker, we `yield*` the output in `bindWorkerAsyncBindings`, which calls `.bind()` on it prematurely, causing a crash because `bind()` requires `RuntimeContext` which is not available at deploy-time. To fix this, I added a check to see if it's an `Output` before yielding.


john-royal:
Replacing with #513 and #514

PR #474, opened 2026-05-31
enable direct secret usage

harrysolovay:
Adds direct Cloudflare Secrets Store secret support as Worker env bindings.


sam-goodwin:
Closed in favor of https://github.com/alchemy-run/alchemy-effect/pull/475

Thanks for getting it started, I didn't see it until too late.

PR #493, opened 2026-06-01
feat(cloudflare): Zone

sam-goodwin:
Adds `Cloudflare.Zone` — a managed DNS zone resource.

```ts

const zone = yield* Cloudflare.Zone("MyZone", {
  name: "example.com",
});
```

Attributes mirror the Cloudflare zone object (`null`s normalized to `undefined`), so the full zone is typed:

```ts
zone.zoneId;
zone.nameServers;        // string[]
zone.status;             // "initializing" | "pending" | "active" | "moved"
zone.accountId;
zone.createdOn;
zone.developmentMode;
zone.meta;               // { foundationDns, phishingDetected, ... }
zone.owner;              // { id, name, type }
// ...activatedOn, modifiedOn, originalNameServers, vanityNameServers, etc.
```

A zone wraps a real registered domain, so deleting one is destructive and rarely what you want on a routine teardown. The resource defaults to **retain** — `stack.destroy()` leaves the zone untouched. Opt into deletion explicitly:

```ts
import { destroy } from "alchemy/RemovalPolicy";

yield* Cloudflare.Zone("MyZone", { name: "example.com" }).pipe(destroy());
```

Adopt a zone that already exists in Cloudflare. Zones carry no ownership markers, so the engine refuses to take one over unless you opt in with `adopt(true)` — otherwise it fails with `OwnedBySomeoneElse`:

```ts
import { adopt } from "alchemy/AdoptPolicy";

const zone = yield* Cloudflare.Zone("MyZone", { name: "example.com" }).pipe(adopt(true));
```

### Supporting changes

- `Resource(type, { defaultRemovalPolicy })` lets a resource type flip the engine's destroy-by-default; `Zone` uses `"retain"`.
- `destroy()` / `retain()` now default to `true` when called with no argument.
- `cloudflare bootstrap-token` gains zone read/settings scopes so a deploy token can manage zones.


sam-goodwin:
We should wait for distilled patch so that we can use error tags instead of regex https://github.com/alchemy-run/distilled/pull/315

PR #504, opened 2026-06-01
feat(apply): add --concurrency limit for deploy/destroy

sam-goodwin:
Cap how many resources apply at once. Defaults to unbounded.

```sh
# no more than 64 resources mid-lifecycle at any time
alchemy deploy --concurrency 64
alchemy destroy --concurrency 64

# omit the flag → unbounded (every ready resource applies at once)
alchemy deploy
```

Programmatic:

```ts
yield* apply(plan, { concurrency: 64 });   // bounded
yield* apply(plan);                         // { concurrency: "unbounded" } (default)
```

The limit is a shared semaphore taken only around the provider call, never around the dependency wait — so a resource blocked on its upstreams holds no permit and a tight limit (even `--concurrency 1`) can't deadlock the DAG.

### DIFF
--- a/packages/alchemy/src/Apply.ts
+++ b/packages/alchemy/src/Apply.ts
@@ -2,6 +2,7 @@ import * as Cause from "effect/Cause";
 import * as Deferred from "effect/Deferred";
 import * as Effect from "effect/Effect";
 import * as Option from "effect/Option";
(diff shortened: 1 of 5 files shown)


sam-goodwin:
@claude please merge latest main into this PR, fix any merge conflicts and push to this pr

PR #515, opened 2026-06-01
feat(cloudflare): provide ExecutionContext to Workflow runs

aryasaatvik:
Workflow runs weren't a first-class execution context. `WorkerBridge.processEvent` mints a fresh `ExecutionContext` (`{ scope, cache }`) per `fetch`/`queue`/`scheduled` event, but `WorkflowBridge.run` only provided `WorkflowEvent` + `WorkflowStep`. So any `@binding` helper that resolves per-execution resources through `ExecutionContext` — e.g. `Drizzle.postgres`, which builds its pool against `ctx.scope` and memoizes it in `ctx.cache` — died inside workflow steps with a missing-`ExecutionContext` defect. The requirement is type-erased at the `Drizzle.postgres` boundary, so `tsc` couldn't catch it.

This provides a fresh `ExecutionContext` (and the matching ambient `Scope`) per run, mirroring `processEvent`. `task` already threads the surrounding body context into every step via `Effect.context`, so the run-level context reaches each `task` automatically: one pool per run-invocation, reused across steps, scope closed on settle unless a binding ejected it.

```ts
// WorkflowBridge.run
const scope = Scope.makeUnsafe();
const exit = await Effect.runPromiseExit(
  fn(event.payload).pipe(
    Effect.provideService(WorkflowEventService, wrapWorkflowEvent(event)),
    Effect.provideService(WorkflowStep, wrapWorkflowStep(step)),
    Effect.provideService(ExecutionContext, { scope, cache: {} }),
    Effect.provideService(Scope.Scope, scope),
  ),
);
// No `waitUntil` here — close inline with the run's real exit; a failing
// finalizer is logged and ignored so it can't mask the outcome.
if (!isScopeEjected(scope)) {
  await Scope.close(scope, exit).pipe(
    Effect.ignoreCause({ log: "Warn", message: "Workflow run scope close failed" }),
    Effect.runPromise,
  );
}
if (Exit.isSuccess(exit)) return exit.value;
throw Cause.squash(exit.cause);
```

`ExecutionContext` is added to `WorkflowRunServices` so the body type reflects what the bridge provides.

A per-run scope also matches Cloudflare's workflow-hibernation model: a resumed run re-executes in a fresh isolate, so the scope (and its pool) is rebuilt per resume rather than shared across hibernations.

### Why `Scope.Scope` too

`processEvent` and `DurableObjectBridge` both provide the run scope as the ambient `Scope` service. `WorkflowServices` advertises `Scope` (via `PlatformServices`), so a body doing `yield* Scope.Scope` type-checks today but would fail at runtime. Providing it closes that parity gap. The targeted `Drizzle.postgres` binding doesn't need it (it reads `ctx.scope` as a value), but the other two bridges set the precedent.

### Drizzle.postgres docs

Corrects stale JSDoc: it claimed the pool was built on "an isolated, never-closing `Scope`" and "at most once per JS realm." The code actually memoizes one pool per `ExecutionContext` (per request / per workflow run), built against `ctx.scope`, torn down when that scope closes.

### Validation

`tsc -b` and lint pass (rebased on `2.0.0-beta.51`). Exercised downstream in a real app (5 `Cloudflare.Workflow`s using `Database.layer` over `Drizzle.postgres`): typechecks and the full test suite passes with this change applied.

Adds a gated live regression test (`test/Planetscale/Postgres/DrizzleWorkflow.test.ts`, behind `PLANETSCALE_TEST`): deploys a Planetscale Postgres + Hyperdrive, hosts a Workflow that runs `Drizzle.postgres` queries inside `task` steps, and asserts the run completes with the row it wrote (a read-back across two steps also proves the per-run pool is reused). It compiles, lints, and skips cleanly without creds; it has **not** been executed live yet (needs Planetscale + Cloudflare credentials).


sam-goodwin:
Thanks for the PR, looking at this now.

I think we want to actually get rid of ExecutionContext now that we figured out scopes properly. I'll do that in a subsequent PR tho

packages/alchemy/test/Planetscale/Postgres/fixtures/drizzle-workflow-db.ts
sam-goodwin, line 1: I think we shoukld use Neon instead for tests because Planetscale is very slow

PR #536, opened 2026-06-03
chore: remove `asEffect` from core constructs

john-royal:
The `asEffect()` pattern has mostly been removed from Effect 4, but there are still many instances of it in Alchemy. This PR replaces them with `Effectable.Prototype`, which simplifies our implementation and lets you use Effect helpers like `Effect.all` and `Effect.forEach`.

Updated core constructs: `Binding`, `Resource`, `Platform`
Updated resources consuming `asEffect`: `ContainerBinding`, `DurableObjectNamespace`, `RpcWorker`, `RpcDurableObjectNamespace`

Intentionally NOT updated:
- `Output` - because outputs aren't quite the same as effects, and there's a lot of custom logic involved
- some Cloudflare bindings (`Images`, `Browser`, `RateLimit`, `DynamicWorkerLoader`) - turning those into real effects would break our async bindings setup


examples/cloudflare-dev/alchemy.run.ts
```diff
@@ -40,14 +40,11 @@ export default Alchemy.Stack(
...
     // Spawn several additional workers to test concurrency.
-    // TODO: Effect.all doesn't work here; Platform needs to be updated to use Effectable.
-    const additionalWorkers = [
-      yield* makeAsyncWorker("AdditionalWorker1"),
-      yield* makeAsyncWorker("AdditionalWorker2"),
-      yield* makeAsyncWorker("AdditionalWorker3"),
-      yield* makeAsyncWorker("AdditionalWorker4"),
-      yield* makeAsyncWorker("AdditionalWorker5"),
-    ];
+    const additionalWorkers = yield* Effect.all(
```
sam-goodwin: nit: Effect.forEach

packages/alchemy/src/Util/effect.ts
```diff
@@ -22,18 +21,12 @@ export const effectClass: {
...
-            return pipeArguments(this, fns);
-          }
-        },
-        impl,
+        class {},
+        // Spreading the Effect prototype onto the class (static side) makes the
+        // class itself a real Effect — `Effect.isEffect(X)` is true, so
+        // `Effect.all([X])` / `Effect.forEach` work — and subclasses
+        // (`class Y extends effectClass(impl)`) inherit the protocol statically.
+        Effectable.Prototype({ label: "EffectClass", evaluate: () => impl }),
```
sam-goodwin: We should start namespacing our special krys`"alchemy/EffectClass"`

PR #537, opened 2026-06-03
feat(cloudflare): Build.DevServer resource used in Cloudflare.StaticSite

Mkassabov:
Adds `Cloudflare.DevCommand` — a resource for declaring a long-lived shell process during `alchemy dev`. Wires it into `Cloudflare.StaticSite` so `dev.command` (e.g. `npm run dev`) is spawned by the dev sidecar and survives user-code HMR.

```ts
Cloudflare.StaticSite("App", {
  command: "npm run build",
  outdir: "dist",
  main: "./src/worker.ts",
  dev: { command: "npm run dev" },
});
```

### Worker opt-out

`WorkerProps.dev` accepts `false`:

```ts
dev?: false | { host?; port?; strictPort? }
```


packages/alchemy/src/Cloudflare/Workers/LocalWorkerProvider.ts
```diff
@@ -361,6 +403,40 @@ export const LocalWorkerProvider = () =>
...
+              Object.entries(worker.env ?? {}).map(([k, v]) => [
+                k,
+                Redacted.isRedacted(v)
+                  ? Redacted.value(v)
+                  : typeof v === "string"
+                    ? v
+                    : JSON.stringify(v),
+              ]),
+            ),
+            PORT: String(port),
```
john-royal: Can we guarantee that the child process will honor this port?

packages/alchemy/src/Cloudflare/Website/StaticSite.ts
```diff
@@ -23,8 +24,49 @@ export interface StaticSiteProps<Bindings extends WorkerBindingProps = {}>
...
+     * Port the local proxy listens on.
+     * @default 1337
+     */
+    port?: number;
+    /**
+     * When `true`, fail instead of falling back to another port if {@link port}
+     * is already in use.
+     * @default false
+     */
+    strictPort?: boolean;
```
john-royal: Not sure if we should be proxying an external dev server, since we can't be 100% sure what port it's listening on, and the purpose of the proxy is to keep ports consistent during HMR.

packages/alchemy/src/Cloudflare/Website/StaticSite.ts
```diff
@@ -167,47 +213,58 @@ const makeStaticSite = <
...
+              ? Object.fromEntries(
+                  Object.entries(resolved.env).flatMap(([k, v]) => {
+                    if (v === undefined) return [];
+                    if (typeof v === "string" || Redacted.isRedacted(v))
+                      return [[k, v]];
+                    return [[k, JSON.stringify(v)]];
+                  }),
+                )
+              : undefined,
+          });
```
john-royal, line 253: I think local dev commands should be handled here - either within the `Command` resource, or using a dedicated `DevCommand` resource (which should be implemented using `RpcProvider` and included in `Cloudflare/Local.ts`).

Plumbing everything through to the worker provider doesn't feel right, especially since the build command is already externalized.

packages/alchemy/src/Cloudflare/Workers/LocalWorkerProvider.ts
```diff
@@ -66,6 +69,45 @@ import type { WorkerBinding } from "./WorkerBinding.ts";
...
+      } else {
+        resume(Effect.succeed(port));
+      }
+    });
+  });
+});
+
+const tryConnect = (port: number) =>
+  Effect.callback<void, Error>((resume) => {
+    const socket = net.createConnection({ host: "127.0.0.1", port });
```
john-royal: Should the `host` be a parameter?

packages/alchemy/src/Cloudflare/Workers/LocalWorkerProvider.ts
Mkassabov: Not really? Since the command could be whatever and the user could force a --port flag, or the underlying server might not be able to support multiple ports.

Would it be better to just not have this at all?

john-royal: Good call.

packages/alchemy/src/Cloudflare/Website/StaticSite.ts
```diff
@@ -167,47 +198,75 @@ const makeStaticSite = <
...
+      // process scoped to the stack and skip the build entirely. The dev
+      // server prints its own URL — alchemy does not proxy it.
+      if (useDevServer) {
+        const devCwd =
+          resolved.dev!.cwd ??
+          (typeof resolved.cwd === "string" ? resolved.cwd : undefined);
+        yield* spawnDevServer(resolved.dev!.command, devCwd).pipe(
+          Effect.forkScoped,
+        );
+      }
```
john-royal: `forkScoped` doesn't work like this in Alchemy. My life would be much simpler if it did.

To avoid dangling processes, you'll need to:
- Create a `DevCommand` resource and implement the provider using`RpcProvider.effect`, instead of `Provider.effect`, so that way it can run from a sidecar process that outlives the user process during HMR.
- Then, on every `diff`/`reconcile` call, you determine whether to keep the old process or kill the old one and create a new one based on whether the command or env has changed.
- The scope used for spawning will need to be cached because the scope inside a `reconcile` hook is closed after reconcile is complete.
- Add the new `DevCommandProvider` to `Cloudflare/Local.ts`:
  ```diff
  +Layer.merge(LocalWorkerProvider(), DevCommandProvider()).pipe(
  -LocalWorkerProvider().pipe(
    Layer.provide(localRuntimeServices()),
    Layer.provide(cloudflareServices),
    RpcServer.launch,
  );
  ```

You can use `LocalWorkerProvider` as an example, but your implementation will be much simpler since it's just handling dev commands.

Wiring up a local worker to use assets from a command instead of a directory would be more involved, so we can skip that for this first pass. For now, I would simply update `WorkerProps` and `LocalWorkerProvider` to handle something like `dev: false` so that worker doesn't try to start and potentially get in the way.

packages/alchemy/src/Cloudflare/Workers/Worker.ts
```diff
@@ -287,26 +287,33 @@ export interface WorkerProps<
...
-     * When `true`, fail instead of falling back to another port if {@link port}
-     * is already in use.
-     * @default false
-     */
-    strictPort?: boolean;
-  };
+  dev?:
+    | false
+    | string
+    | {
```
john-royal, line 290: Wonder if an object union might be clearer? e.g. `{ server?: true, port: ... } | { server: false; url?: string }`

packages/alchemy/src/Build/Local.ts
sam-goodwin, line 1: LocalDevCommand.ts or DevCommand.main.ts?

Might need a better convention here

packages/alchemy/src/Build/DevCommand.ts
```diff
@@ -0,0 +1,279 @@
...
+  {
+    /**
+     * URL extracted from the first matching stdout/stderr line. Best-effort:
+     * `undefined` if no URL appeared within {@link URL_EXTRACT_TIMEOUT}.
+     */
+    url?: string;
+  }
+> {}
+
+export const DevCommand = Resource<DevCommand>("Cloudflare.DevCommand");
```
sam-goodwin: Build.DevCommand and add typedoc with usage or tell people not to use it?

packages/alchemy/src/Build/DevCommand.ts
```diff
@@ -0,0 +1,279 @@
...
+}
+
+/**
+ * A long-lived shell process scoped to a stack instance, started during
+ * `alchemy dev` and torn down on stack teardown or when its `command` / `cwd`
+ * / `env` changes.
+ *
+ * The provider runs inside the dev sidecar (see `Cloudflare/Local.ts`) so the
+ * child process survives user-code HMR — Alchemy's user process can restart
+ * without killing your `npm run dev` server.
```
sam-goodwin: This would run during deploy, so this isn't true?

PR #540, opened 2026-06-03
fix(cloudflare/dev): cross-script durable object bindings not working

john-royal:
This is a draft. Verifying against user repro first; will update here with tests and a better description after.

This is a stacked change; please merge #542 first.


packages/alchemy/tsdown.config.ts
```diff
@@ -22,7 +22,7 @@ export default [
   // Bundling inlines those so they stay devDependencies (same rationale as
   // the cli bundle below).
   defineConfig({
-    entry: ["bin/exec.ts"],
+    entry: ["bin/alchemy.ts", "bin/exec.ts"],
```
john-royal: Our CLI doesn't work outside Bun! Set it here to get a usable preview package; will spin off into separate PR.

.github/workflows/pr-package.yml
sam-goodwin, line 1: Why did this change

john-royal: Trying to get a usable preview package. Looks like we're currently only publishing TS in preview packages; tsc isn't being run because the workflow runs build from a nested directory instead of from the root. I'll spin this off into a separate PR once I get it working.

packages/alchemy/src/Auth/AuthProvider.ts
```diff
@@ -33,7 +33,7 @@ export class AuthError extends Schema.TaggedErrorClass<AuthError>()(
   "AuthError",
   {
     message: Schema.String,
-    cause: Schema.optional(Schema.Defect),
+    cause: Schema.optional(Schema.Defect()),
```
john-royal, line 36: Effect changed this in beta-78. Will spin off tomorrow.

PR #548, opened 2026-06-05
fix(engine): honor retain removal policy on replace

john-royal:
`RemovalPolicy.retain` was only honored when a resource was removed from the stack (orphan delete / destroy), not when it was replaced. Replacement is create-new-then-delete-old, and the old generation's cleanup in `collectGarbage` called `provider.delete` unconditionally — so the previous physical resource was destroyed even though the user asked to retain it.

The fix guards the old-generation delete with the replaced node's removal policy (`ReplacedResourceState` already carries `removalPolicy`, set at replace time and preserved through chain pops). State cleanup (chain pop / collapse to `created`) still runs, so the resource converges normally; only the physical `provider.delete` is skipped.

```diff
-if (attr !== undefined) {
+const retainOldGeneration =
+  !isDeleteNode(node) && node.removalPolicy === "retain";
+
+if (attr !== undefined && !retainOldGeneration) {
   yield* provider.delete({ /* … */ });
 }
```

Behavior:

- Replace of a retained resource keeps the old physical instance (no `provider.delete`), still mints a fresh instance and converges to `created`.
- Nested replacement chains skip the delete on every pass while draining to `created`.
- A changed policy (`retain` → `destroy`) at replace time deletes the old generation, consistent with CloudFormation `UpdateReplacePolicy` reading current desired state.

Adds a `retain removal policy on replace` suite to `apply.test.ts` covering replace-with-retain, the destroy contrast (delete fires exactly once), nested chains, and an orphan-delete regression guard.

### DIFF
--- a/packages/alchemy/src/Apply.ts
+++ b/packages/alchemy/src/Apply.ts
@@ -1586,7 +1586,9 @@ const collectGarbage = Effect.fnUntraced(function* (
               });
               yield* report("deleted");
             } else {
-              yield* scopedSession.note("Cleaning up replaced resource...");
+              if (!retainOldGeneration) {
+                yield* scopedSession.note("Cleaning up replaced resource...");
+              }
               if (
                 node.old.status === "replacing" ||
                 node.old.status === "replaced"
@@ -1626,7 +1628,11 @@ const collectGarbage = Effect.fnUntraced(function* (
                   removalPolicy: node.removalPolicy,
                 });
               }
-              yield* scopedSession.note("Replaced resource cleanup complete.");
+              yield* scopedSession.note(
+                retainOldGeneration
+                  ? "Replaced resource retained."
+                  : "Replaced resource cleanup complete.",
+              );
             }
           }),
         ));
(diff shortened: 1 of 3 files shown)


packages/alchemy/src/Apply.ts
```diff
@@ -1544,7 +1544,20 @@ const collectGarbage = Effect.fnUntraced(function* (
               });
             }
 
-            if (attr !== undefined) {
+            // Honor `retain` for the old generation of a replacement, mirroring
+            // the orphan-delete path above. Delete-node retain is already
+            // handled with an early return; this guards the replaced
+            // old-generation physical delete.
+            const retainOldGeneration =
+              !isDeleteNode(node) && node.removalPolicy === "retain";
```
sam-goodwin, line 1552: Node is the item in the plan, right? Why does node.retain represent the old policy

john-royal: The old policy would be `node.old.removalPolicy`. `node.removalPolicy` is the current one.

PR #551, opened 2026-06-05
fix(drizzle): expose Schema `out` as a cwd-relative path

john-royal:
`Drizzle.Schema` exposed its `out` attribute as an absolute path (`path.resolve(process.cwd(), …)`), which baked machine-specific paths into persisted state. Emit it relative to the current working directory instead; internal filesystem ops still use the absolute form.

```diff
-    out,
+    out: relativeOut(out),
```

Downstream `migrationsDir` consumers (PlanetScale Postgres/MySQL, Neon.Branch, Cloudflare.D1Database) all read the value through `listSqlFiles`/`hashMigrations`, which resolve a relative path against cwd, so no consumer changes are needed.


sam-goodwin:
add test please

PR #552, opened 2026-06-05
[codex] Fix Effect beta pipe compatibility

brendadeeznuts1111:
Updates Test/Core and Artifacts usage from effect.pipe to pipe() for Effect beta compatibility.\n\nThis PR contains the existing local commit from this checkout after rebasing onto current upstream main. The dirty working tree was not staged or included.\n\nValidation: not run in this cleanup pass.


john-royal:
What's the exact Effect version you're on (e.g. `4.0.0-beta.75`)?

We haven't encountered this problem from other users or in tests, so I want to double check that this isn't because of an outdated version on your end.

sam-goodwin:
This changes doesn't seem to make sense and haven't heard back. closing

PR #557, opened 2026-06-05
feat(planetscale): Postgres logical database resource

jonbeckman:


packages/alchemy/src/Planetscale/Postgres/PostgresLogicalDatabase.ts
```diff
@@ -0,0 +1,628 @@
...
+  ownershipTable: string;
+}
+
+/**
+ * A logical PostgreSQL database created inside a PlanetScale PostgreSQL
+ * database branch.
+ *
+ * Use this when one PlanetScale PostgreSQL database/branch should host
+ * multiple app databases that each need separate migrations, imports, and app
+ * role grants.
```
Mkassabov: I think this is a little unclear to users who might just want a database and wont know if they should use this orthe normal PostgresDatabase resource; could we make this more explicit, and maybe link to the main planetscaleDatabase as well?

packages/alchemy/src/Planetscale/Postgres/PostgresLogicalDatabase.ts
```diff
@@ -0,0 +1,628 @@
...
+ *   inheritedRoles: [],
+ * });
+ *
+ * const logicalDb = yield* Planetscale.PostgresLogicalDatabase("AppDb", {
+ *   name: "app",
+ *   adminOrigin: admin.origin,
+ *   appRoleName: Planetscale.postgresRoleNameFromUsername(app.username),
+ *   migrationsDir: "./migrations",
+ * });
+ * ```
```
Mkassabov, line 152: this example feels particularly confusing since it obscures the fact that you ALSO need an actual postgres database to get the admin role; we should show the `const database = yield* Planetscale.PostgresDatabase`

packages/alchemy/src/Planetscale/Postgres/PostgresLogicalDatabase.ts
```diff
@@ -0,0 +1,628 @@
...
+const isObject = (value: unknown): value is Record<PropertyKey, unknown> =>
+  typeof value === "object" && value !== null;
+
+const isReadablePostgresOrigin = (value: unknown): value is PostgresOrigin =>
+  isObject(value) &&
+  typeof value.scheme === "string" &&
+  typeof value.host === "string" &&
+  typeof value.port === "number" &&
+  typeof value.user === "string" &&
+  value.password !== undefined;
```
Mkassabov: can we use effect predicate for these?

packages/alchemy/src/Planetscale/Postgres/PostgresLogicalDatabase.ts
```diff
@@ -0,0 +1,628 @@
...
+  input.output.appRolePrivilegesHash !==
+    appRolePrivilegesHash(input.appRolePrivileges) ||
+  input.output.appRolePrivilegesVersion !== input.appRolePrivilegesVersion ||
+  input.output.migrationsTable !== input.migrationsTable ||
+  input.output.importsTable !== input.importsTable ||
+  input.output.ownershipTable !== input.ownershipTable ||
+  !ownersEqual(input.output.owner, input.owner) ||
+  input.appRolePrivileges?.ready === false ||
+  !recordsEqual(input.migrations, input.output.migrationsHashes) ||
+  !recordsEqual(input.imports, input.output.importHashes);
```
Mkassabov, line 289: feels verbose to extract this to a utility if its single use

jonbeckman:
Hey @Mkassabov I addressed your comments. The only one I didn't agree with was `logicalDatabaseNeedsUpdate`, I like it as a function as it self documents its purpose and keeps the parent function clean.

Note that my branch is stacked on @john-royal's changes in john/vite-incorrect-esmexternalrequire because I needed them for an unrelated fix. Will move this to draft while we wait on that to land.

Mkassabov:
> The only one I didn't agree with was `logicalDatabaseNeedsUpdate`

Fair, I don't feel too strongly about that, keeping it as is is okay

jonbeckman:
Ah John had already merge his changes in. I have updated this branch so there are no merge conflicts. It is ready to go.

sam-goodwin:
This PR is missing a description with code snippet showing usage now supported

packages/alchemy/src/Planetscale/Postgres/PostgresLogicalDatabaseClient.ts
sam-goodwin, line 1: All this code is async, it should be using effect

Mkassabov:
Upon discussion this looks like it's out of scope for alchemy. Unless we are misunderstanding there isn't an actual cloud resource being provisioned here; so this abstraction risks being leaky and compromises some of our security mentalities.

This might be better off as a standalone library, and if you choose to go in that direction we'd support it.

jonbeckman:
That's fair.

jonbeckman:
For anyone who gets here and is looking for this resource, I built it into a library here https://www.npmjs.com/package/alchemy-planetscale-logical-db

PR #566, opened 2026-06-07
fix(core): resolve Config values in input props

john-royal:
`Config` values are valid `Input`s, but generic input resolution never handled them. `resolveInput` walked the `Config` object with `Object.entries`, stripping its prototype down to its enumerable `parse` property — so providers received `{ parse }` instead of the configured value (and `Config.isConfig` returned `false` on it downstream).

This went unnoticed for a Worker's `env` because those are resolved out of band by the bindings mechanism (yieldable values are resolved before binding metadata is derived). It surfaced when a user set `dev.port` to a `Config` — the `Input` type says that's allowed, but nothing actually resolved it.

Resolve `Config` against the deploy environment during input resolution instead of walking it:

```ts
// Plan.ts resolveInput / Output.ts evaluate
} else if (Config.isConfig(input)) {
  return yield* resolveInput(yield* input);
} else if (Duration.isDuration(input) || Redacted.isRedacted(input)) {
  // opaque — Redacted stays wrapped for the secrecy boundary
  return input;
}
```

Resolving in plan (rather than per-provider) also means the concrete value participates in diffing/hashing — an opaque `Config` hashes the same regardless of the underlying value, so env changes never triggered an update. `Config.redacted` resolves to a `Redacted`, which stays wrapped end-to-end.

### Notes

- `Redacted` and `Duration` are unchanged (still opaque).
- Raw `Effect` inputs are intentionally left unresolved for now.
- Dropped the now-dead defensive `Config` handling in the Cloudflare Worker provider (`Worker.ts` vite `define`, `WorkerAsyncBindings.ts` `toBinding`).
- `Output.evaluate` / `resolveInput` now surface `Config.ConfigError`, which the deploy pipeline already threads.

### Tests

- `Output.test.ts`: new `Config` block — top-level / nested / array resolution, resolution against a `ConfigProvider`, and `Config`→`Redacted` staying wrapped.
- `plan.test.ts`: new block driving the `resolveInput` path via `plan.resources.*.props`.

### DIFF
--- a/packages/alchemy/src/Cloudflare/Workers/Worker.ts
+++ b/packages/alchemy/src/Cloudflare/Workers/Worker.ts
@@ -1,8 +1,8 @@
 import type * as cf from "@cloudflare/workers-types";
 import * as workers from "@distilled.cloud/cloudflare/workers";
 import * as zones from "@distilled.cloud/cloudflare/zones";
+import type * as Config from "effect/Config";
 import type { ConfigError } from "effect/Config";
-import * as Config from "effect/Config";
 import * as Context from "effect/Context";
 import * as Data from "effect/Data";
 import * as Effect from "effect/Effect";
@@ -1178,7 +1178,7 @@ export const LiveWorkerProvider = () =>
                       : Redacted.isRedacted(value) &&
                           typeof Redacted.value(value) === "string"
                         ? Redacted.value(value)
-                        : Config.isConfig(value) || Effect.isEffect(value)
+                        : Effect.isEffect(value)
                           ? yield* value as Effect.Effect<any>
                           : undefined,
                   ];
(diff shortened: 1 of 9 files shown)


packages/alchemy/src/Cloudflare/Workers/WorkerAsyncBindings.ts
```diff
@@ -75,20 +75,14 @@ type BindingSpec = InputProps<
...
+): BindingSpec => {
   if (typeof binding === "string") {
     return {
       type: "plain_text",
       name: bindingName,
       text: binding,
     };
-  } else if (isRedacted(binding)) {
+  } else if (Redacted.isRedacted(binding)) {
     const val = Redacted.value(binding);
```
john-royal, line 86: This now type checks without needing a separate helper 🎉

`val` evaluates to type `Json`.

packages/alchemy/src/Cloudflare/Workers/Worker.ts
```diff
@@ -1178,7 +1178,7 @@ export const LiveWorkerProvider = () =>
                       : Redacted.isRedacted(value) &&
                           typeof Redacted.value(value) === "string"
                         ? Redacted.value(value)
-                        : Config.isConfig(value) || Effect.isEffect(value)
+                        : Effect.isEffect(value)
```
sam-goodwin, line 1181: is a Config an Effect?

sam-goodwin (approved):

PR #568, opened 2026-06-08
fix(pr-package): use the resolved value for Authentication

boarush:
`pr-package` uses a `Cloudflare.Secret` to store the secret for `Authentication` header against which the incoming `PUT` and `POST` requests are checked. `Cloudflare.Secret` returns a `Redacted<string>` which needs to resolved before checking against the incoming header. Directly using the `expected` value matches the header against `<redacted>` which is not intended.


boarush:
Closed in favor of #598

PR #579, opened 2026-06-10
feat(cloudflare/container): support image-only containers without a main entrypoint

agcty:
## Motivation

`Cloudflare.ContainerApplication` currently fails with `Container requires a \`main\` entrypoint.` even when the `dockerfile` prop fully describes the image. That blocks containers backed by prebuilt service images — e.g. `cloudflare/sandbox` or `ghcr.io/drizzle-team/gateway` attached to a Durable Object namespace — where there is no JS program to bundle and the image's own `ENTRYPOINT`/`CMD` should run.

## Change

- `main` is now optional when `dockerfile` is provided; omitting both still fails with a clear error (`Container requires a \`main\` entrypoint or a \`dockerfile\`.`).
- Image-only path: the user Dockerfile is used verbatim — no `WORKDIR`/`COPY`/`ENTRYPOINT` statements are appended, so the image's own entrypoint runs. The build-and-push flow is unchanged otherwise (the context just contains only the Dockerfile).
- The image hash covers the final Dockerfile alone when there is no bundle (`sha256Object` drops the undefined `bundleHash` key).
- The existing `entrypoint`/`command` configuration overrides keep working for both paths.
- `Cloudflare.Container` (the platform/layer form) is unaffected — its props still require `main`.

Example:

```ts
const sandbox = Cloudflare.ContainerApplication("sandbox", {
  dockerfile: "FROM docker.io/cloudflare/sandbox:0.7.18-python",
  instanceType: "lite",
  maxInstances: 5,
})
```

## Validation

- `bun tsc -b` clean.
- No Container test harness exists under `packages/alchemy/test/Cloudflare` yet; happy to add one if you can point me at the preferred seam for docker-dependent paths.

### DIFF
--- a/packages/alchemy/src/Cloudflare/Container/ContainerApplication.ts
+++ b/packages/alchemy/src/Cloudflare/Container/ContainerApplication.ts
@@ -90,6 +90,11 @@ export interface ContainerApplicationProps extends PlatformProps {
   /**
    * Main entrypoint for the container program. This file is bundled and
    * added to the Docker image as the container's entrypoint.
+   *
+   * Optional when {@link dockerfile} is provided: image-only containers
+   * use the Dockerfile verbatim and run its own `ENTRYPOINT`/`CMD` —
+   * e.g. prebuilt service images like `cloudflare/sandbox`. The
+   * {@link entrypoint} and {@link command} overrides remain available.
    */
   main?: string;
   /**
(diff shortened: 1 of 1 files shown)


sam-goodwin:
Solving same problem as https://github.com/alchemy-run/alchemy-effect/pull/583 ?

Made a comment on how i think the DX should look here https://github.com/alchemy-run/alchemy-effect/pull/583#discussion_r3393754059

PR #581, opened 2026-06-10
fix(cloudflare): export fetcher from bridge

john-royal:
Vite's dependency optimization doesn't play well with a full import of `alchemy/Cloudflare` because of Node and other native imports. As a result, we offer an `alchemy/Cloudflare/Bridge` export with functions that are safe to import from Vite.

Our `Fetcher` utility, including functions like `fromCloudflareFetcher`, was not exported, meaning that you couldn't do something like this:

```ts
const client = Cloudflare.toHttpClient( // returns an effect HttpClient
  Cloudflare.fromCloudflareFetcher(env.BACKEND),
);
const res = await client
  .execute(HttpClientRequest.fromWeb(request))
  .pipe(Effect.runPromise);
```

This PR re-exports `./Fetcher` so this is fixed. Also updated the TanStack Start example and integration tests to exercise this new code path.


sam-goodwin:
can we add a test?

PR #583, opened 2026-06-11
feat(cloudflare/container): build images from a prebuilt context directory

Butch78:
Build a Cloudflare Container straight from a self-contained Docker build context, skipping JS bundling entirely. Cloudflare Containers run any `linux/amd64` image; only the resource assumed a JS entrypoint.

```ts
export class SqlCompute extends Cloudflare.Container<SqlCompute>()("sql-compute", {
  context: "./sql-container",      // dir with its own Dockerfile
  // dockerfile: "Dockerfile.prod" // path, or inline contents; defaults to "Dockerfile"
  instanceType: "standard-1",
}) {}
```

- `context` (a build-context dir) replaces the JS-bundling pipeline when no `main` is set. `main` and `context` are mutually exclusive; `main` stays optional.
- `dockerfile` accepts a **path** or **inline contents** (told apart by whether it resolves to a file), in both modes. Context mode passes `docker build -f`; `dockerBuild` gains a `dockerfile` option.
- One `resolveContextMode` helper decides mode + Dockerfile location for both the hash and the build, so they can't drift.
- The context image hash folds in each file's contents (streamed, never buffered whole) plus its unix mode, honoring `.dockerignore` and always skipping `.git`, so content changes trigger a rebuild and rollout.

Used to deploy a Rust (DataFusion) container attached to a Durable Object: builds, pushes to `registry.cloudflare.com` with the content-addressed tag, and serves via `ctx.container`.


packages/alchemy/src/Cloudflare/Container/ContainerApplication.ts
```diff
@@ -645,34 +683,44 @@ await Effect.runPromise(serverEffect).catch((err) => {
...
-          platform: "linux/amd64",
-        });
+        if (props.prebuiltContext) {
+          // Prebuilt-context mode: build the caller's directory verbatim,
+          // with its own Dockerfile, no bundle files, no appended ENTRYPOINT.
+          yield* dockerBuild({
+            tag: imageRef,
+            context: props.prebuiltContext,
+            platform: "linux/amd64",
+          });
```
sam-goodwin: I don't think `prebuiltContext` is an intuitive name for this property. I do agree it should be possible to skip bundling.

Why not just call it `context?: string` and make `main?: string` optional. Skip JS bundling if `main` and `dockerfile` are not provided?

I also wonder if `dockerfile` should be a path (or maybe even allow it to be both a path or dockerfile contents because we can detect that with `path.resolve`?

```ts
Cloudflare.Container("container", {
  context: "./sql-container",
  // dockerfile: "Dockerfile.custom" // <- defaults to "Dockerfile"
})

Comment (Butch78) [reply packages/alchemy/src/Cloudflare/Container/ContainerApplication.ts]: Good call, done in 9fbf01d:

- Renamed `prebuiltContext` → `context`. JS bundling is skipped whenever there's no `main` (the two are now mutually exclusive).
- `dockerfile` accepts a **path** or **inline contents**, resolved by whether it points at an existing file (`path.resolve` against the context, then cwd). In context mode it defaults to `<context>/Dockerfile` and is passed via `docker build -f`.

```ts
Cloudflare.Container("container", {
  context: "./sql-container",
  // dockerfile: "Dockerfile.custom" // path, or inline "FROM …" contents
})
```

Butch78:
Thanks @Copilot — addressed in 9fbf01d (alongside the `prebuiltContext` → `context` rename @sam-goodwin asked for):

- **Path building / Windows** — now uses `path.join` everywhere instead of string concatenation; directory-walk keys are normalized to POSIX.
- **Reads every file into memory** — context files are now hashed with a streaming `sha256File` (chunks through `node:crypto`, never buffered whole). Added `.dockerignore` support plus an always-on `.git` skip so large/volatile dirs aren't walked.
- **File mode / symlinks** — each file's unix `mode` is folded into the hash. `fs.stat` follows symlinks (effect exposes no `lstat`), so links are hashed by their dereferenced target; documented on the helper.
- **Duplicated branching** — extracted a single `resolveContextMode` helper that decides the mode and resolves the Dockerfile location; both the hash step and the build step consume it, so they can't drift.
- **`main` optional weakens typing** — kept `main` optional per the API direction in this thread (`main`/`context` mutually exclusive, validated at runtime) rather than a discriminated union, since the union would have to duplicate the full shared prop set across both members.

Added unit coverage for streaming `sha256File` and for the new `docker build -f` path.

packages/alchemy/src/Bundle/Docker.ts
```diff
@@ -47,6 +53,20 @@ export const materializeDockerfile = Effect.fn(function* (
...
+ * Returns the lowercase hex digest.
+ */
+export const sha256File = Effect.fn(function* (filePath: string) {
+  const fs = yield* FileSystem.FileSystem;
+  const hash = createHash("sha256");
+  yield* fs
+    .stream(filePath)
+    .pipe(Stream.runForEach((chunk) => Effect.sync(() => hash.update(chunk))));
+  return hash.digest("hex");
+});
```
sam-goodwin: Do we already have a util for this in sha256.ts? 

If not, add it to that file for re-usability.

Butch78: Good idea — moved it to `Util/sha256.ts` as `sha256File` (streams via the FileSystem service through `node:crypto`, alongside `sha256`/`sha256Object`) in 607f4ed. Tests moved to `test/Util/sha256.test.ts`.

Butch78:
Thanks for the suggestions, @sam-goodwin — both are in: `context` + path-or-inline `dockerfile`, and the streaming hash now lives in `Util/sha256.ts`. Let me know if you'd like anything else tweaked.

sam-goodwin:
This should be fixed now - we support arbitrary dockerfiles/context directory and remote image references

PR #584, opened 2026-06-11
feat(cloudflare): run DO-attached containers in alchemy dev

Butch78:
Run Durable-Object-attached containers under `alchemy dev`, using workerd's native container support against the local Docker daemon (what `wrangler dev` / miniflare do).

A worker binds a container class to a DO namespace with an image name; locally that image is started per DO instance and exposed as `ctx.container`:

```ts
yield* worker.bind`my-container`({
  containers: [{ className: "MyDo", imageName: "my-image:dev" }],
});
```

- `containers` binding data gains an optional `imageName`; entries without one are remote-only attachments and are ignored locally (`ctx.container` stays absent)
- `LocalWorkerProvider` maps `className -> imageName` from binding data, attaches `container: { imageName }` to the matching workerd DO namespace, and configures a `localDocker` container engine (socket `unix:///var/run/docker.sock`)
- The egress-interceptor sidecar image defaults to the same pinned `cloudflare/proxy-everything` miniflare uses; `ALCHEMY_CONTAINER_EGRESS_IMAGE` overrides it (e.g. to work around cloudflare/workerd#6792 on hosts where TPROXY intercepts bridge-local traffic and hangs the sidecar readiness probe)

Draft because it activates only once `@distilled.cloud/cloudflare-runtime` threads the container config through to workerd: alchemy-run/cloudflare-tools#46, then a catalog bump here. The new fields typecheck against 0.10.12 and are dropped by its config assembly, so behavior is unchanged until that lands.

Validated end-to-end on the cloudflare-tools#46 runtime: a Rust DataFusion container attached to a DO class under `alchemy dev` cold-starts on first request, serves proxied traffic, and is destroyed by the DO's idle alarm.

### DIFF
--- a/packages/alchemy/src/Cloudflare/Workers/LocalWorkerProvider.ts
+++ b/packages/alchemy/src/Cloudflare/Workers/LocalWorkerProvider.ts
@@ -214,7 +214,27 @@ export const LocalWorkerProvider = () =>
             hyperdrives: worker.hyperdrives,
             durableObjectNamespaces: toRuntimeDurableObjectNamespaces(
               worker.durableObjectNamespaces,
+              worker.containerImages,
             ),
+            // Local containers: hand workerd the docker socket whenever a
+            // namespace has an attached image, mirroring `wrangler dev`.
+            // The egress-interceptor default is the image miniflare pins;
+            // ALCHEMY_CONTAINER_EGRESS_IMAGE overrides it (e.g. to work
+            // around cloudflare/workerd#6792 on hosts where TPROXY
+            // intercepts bridge-local traffic and hangs the sidecar
+            // readiness probe).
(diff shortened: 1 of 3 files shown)


Butch78:
Added unit tests in 472532b (`test/Cloudflare/Workers/LocalWorkerProvider.test.ts`). Pulled the inline `className -> imageName` loop out into an exported, documented `collectContainerImages` helper and exported `toRuntimeDurableObjectNamespaces`, so both are covered directly: remote-only attachments (no `imageName`) are skipped, className collisions take the last write, and the `container` field is attached only to namespaces that have a mapped image. Still draft pending cloudflare-tools#46 + the catalog bump.

john-royal:
Replaced by #636.

PR #587, opened 2026-06-11
feat(cloudflare): forward dev.env to the StaticSite dev server

agcty:
## Problem

`Build.DevServer` accepts `env` (including `Redacted` values, kept out of logs and state), but `StaticSite`'s `dev` prop only exposes `command`/`cwd`/`url` — the DevServer is constructed with command+cwd only. A `StaticSite` whose dev is delegated to an external command (the documented `StaticSite` pattern — e.g. `zola serve` in the `cloudflare-static-site` example, or any custom pipeline) has no supported way to pass extra environment to that process.

Today the only workaround is interpolating values into `dev.command`, which (a) is awkward for anything structured and (b) persists secrets in plaintext in resource state if the value is sensitive. `DevServer` already solves both: it merges `env` over `process.env` and keeps `Redacted` values out of logs/state. `StaticSite.dev` just doesn't forward it.

## Change

Adds `env?: Record<string, string | Redacted.Redacted<string>>` to `StaticSiteProps["dev"]` and forwards it to the `DevServer` call. Pure pass-through; `DevServer` already implements the semantics — this only closes the gap where `StaticSite.dev` is the one `DevServer` caller that drops a supported field.

## Example

```ts
Cloudflare.StaticSite("Site", {
  command: "zola build",
  outdir: "public",
  dev: {
    command: "zola serve",
    env: {
      // extra env for the dev process: ports, feature flags, external
      // service tokens. Secrets can travel as Redacted so they never
      // land in logs or state.
      ZOLA_BASE_URL: "http://localhost:1111",
      SOME_TOKEN: Redacted.make(process.env.SOME_TOKEN!),
    },
  },
})
```

### DIFF
--- a/packages/alchemy/src/Cloudflare/Website/StaticSite.ts
+++ b/packages/alchemy/src/Cloudflare/Website/StaticSite.ts
@@ -4,9 +4,10 @@ import * as Redacted from "effect/Redacted";
 import { AlchemyContext } from "../../AlchemyContext.ts";
 import { Command, type CommandProps } from "../../Build/Command.ts";
 import { DevServer } from "../../Build/DevServer.ts";
-import type { InputProps } from "../../Input.ts";
+import type { Input, InputProps } from "../../Input.ts";
 import * as Namespace from "../../Namespace.ts";
 import { effectClass } from "../../Util/effect.ts";
+import { asEffect } from "../../Util/types.ts";
 import type { Providers } from "../Providers.ts";
 import type { AssetsConfig } from "../Workers/Assets.ts";
 import {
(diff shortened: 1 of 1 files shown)


agcty:
Additional motivation from #590: plugin-rsc apps (React Router RSC, Waku) cannot run under the native dev path today — the dev environment assumes a single `ssr` worker — so `dev: { command }` is their structural dev path rather than an escape hatch. That makes env forwarding to the spawned command the only way for those apps to reach stage-scoped resources (e.g. Hyperdrive local connection strings) during dev.

john-royal (approved):

john-royal:
I'm wondering if we should merge the top-level `env` property?

agcty:
Good call — done in 53be7d66. The dev process now receives the top-level `env` as its base with `dev.env` as overrides, mirroring how the build command already receives it.

One boundary kept: only process-env material crosses — strings and `Redacted` values. Binding objects aren't environment variables, and deferred values (`Config`, resource references) can't be safely forced at construction time, so those are skipped rather than stringified.

Issue #590, opened 2026-06-11
Cloudflare.Vite: @vitejs/plugin-rsc apps (React Router RSC, Waku) cannot deploy — server bundle is captured in-memory from the "ssr" environment only


agcty:
## Repro

- React Router 7.15 RSC (`unstable_reactRouterRSC` + `@vitejs/plugin-rsc@0.5.27`) — applies to any `@vitejs/plugin-rsc` framework (Waku/vocs hit the dev-side half of this in alchemy-run/cloudflare-tools#43)
- alchemy `2.0.0-beta.55`, `Cloudflare.Vite()("website", { rootDir: import.meta.dirname, ... })`
- `alchemy deploy`: the vite build succeeds (all five plugin-rsc `buildApp` passes), client assets upload, then the worker upload fails:

```
ERROR (#1): ScriptModuleNotFound: Uncaught Error: No such module "__vite_rsc_assets_manifest.js".
  imported from "index.js"
```

## Root cause

`Workers/Vite.ts` captures the server bundle from the in-memory `generateBundle` of the environment literally named `"ssr"` (lines 50–58), and `Worker.ts` uploads exactly that `OutputBundle`. Three properties of plugin-rsc output break this:

1. **The worker environment is `rsc`, not `ssr`.** plugin-rsc apps put the request handler in the `rsc` environment (react-server resolve conditions); `ssr` is the HTML-renderer child loaded from it. The captured bundle is the wrong program — even if the upload succeeded, it would serve the SSR renderer with no RSC runtime. (`@distilled.cloud/cloudflare-vite-plugin` already accommodates plugin-rsc with `rsc: { serverHandler: false }`; this issue is the deploy-side half of that work.)
2. **`__vite_rsc_assets_manifest.js` never exists in any `OutputBundle`.** plugin-rsc derives it from the *client* bundle and writes it with `fs.writeFileSync` into the server outDirs after all environment builds complete, inside its own `buildApp` orchestration (`@vitejs/plugin-rsc` `writeAssetsManifest`). Server chunks import it via a relative specifier. An in-memory `generateBundle` capture is structurally unable to include it, regardless of which environment is captured.
3. **Cross-environment imports are relative paths between outDirs on disk.** `import.meta.viteRsc.loadModule("ssr", ...)` compiles to e.g. `import("./__ssr_build/index.js")` — resolvable only when the directory tree is uploaded as the module set with relative names intact.

## How the reference deploy path handles this

[...]

Issue #589, opened 2026-06-11
Worker bundler: default-on pure annotations break pre-bundled effect inputs at runtime


agcty:
## Symptom (2.0.0-beta.55)

A `StaticSite` whose `main` is an already-complete vite server bundle (React Router RSC app, effect-based request handling) deploys fine, but **every request** on the deployed Worker dies with:

```
TypeError: _(...).then is not a function
    at MixedScheduler.<anonymous> (Schema-<hash>.js:14:4727)
```

The same code is healthy under `vite dev` and in the vite build itself — the failure is introduced by the Worker upload re-bundle. Setting `build: { pure: false }` fixes it deterministically.

## Cause (corrected)

My first read of this was that the pure plugin was matching the `effect`/`@effect/*` packages on the pre-bundled input. That is **not** what happens, and it's worth correcting because it points at the real fix.

`purePlugin` matches by **package name resolved from the module's path**, not by anything in the code: `resolvePackageInfo` walks up to the nearest `package.json` and `packageNameFromId` reads the `node_modules/<pkg>` segment (`PurePlugin.ts` ~L120-121, L270-282, L294-366). In a vite/rolldown output (`build/server/index.js` + `build/server/assets/*.js`) the effect package boundaries no longer exist — walking up from any chunk lands on the **app's own `package.json`**, not `effect`. So the `effect`/`@effect/*` patterns never fire on this input.

What actually fires is `autoDetectEntryPackage` (default `true`, `PurePlugin.ts` L72/L91/L107-116): it detects the entry's owning package and adds it to the match set. The app package declares `"sideEffects": false`, so:

- every chunk under `build/server/` resolves to that one owning package and matches, and
- because `sideEffects` is `false`, the plugin applies **both** `/*#__PURE__*/` annotations **and** `moduleSideEffects: false` (`PurePlugin.ts` L133-145) across the whole pre-bundled output.

Marking an already-bundled vite artifact side-effect-free / pure lets rolldown drop top-level initializers the effect scheduler dereferences at request time — hence the per-request `TypeError`. `build: { pure: false }` disables the whole plugin, which is why it's a clean fix.

So the trigger is **autoDetectEntryPackage + the app package's `sideEffects: false`**, applied to a pre-bundled multi-chunk input — not effect-package matching.

## Suggestions

[...]

PR #592, opened 2026-06-11
feat(cloudflare/worker): bundle: false — deploy prebuilt workers without re-bundling

agcty:
Picks up #117 (thanks @czxtm for the original implementation and repro) with the review feedback applied. Adds `bundle?: boolean` to `Cloudflare.Worker`: when `false`, `main` is uploaded byte-for-byte — no rolldown, no minification, no pure-annotation pass.

```ts
Cloudflare.Worker("site", {
  main: "./.open-next/worker.js",
  bundle: false,
  assets: "./.open-next/assets",
});
```

Changes relative to #117, per review:

- **Pattern-based module resolution instead of regex import-scanning.** Additional modules are resolved from glob patterns the way Alchemy v1 (`WorkerBundleSource.FS`) and Wrangler (`findAdditionalModules`) do: the entry's directory is walked recursively, matching files are uploaded as modules named by their POSIX path relative to that directory. The entry is always first and never duplicated; `.js.map` files are not uploaded.
- **No Next.js example, no wrangler dependency.** The code path is exercised end-to-end by an integration test deploying a committed prebuilt fixture (nested ES modules + a text module) and asserting the response renders through the nested module graph — the bundle hash is derived from the verbatim source-byte hashes, so it cannot match a rolldown-processed output.
- Diffing needs no special casing: `prepareBundle` branches internally, so `hasChanged` compares the on-disk module-set hash against the stored `hash.bundle`.

Conscious decisions to sign off on:

- **`rules?: { globs: string[] }[]` replaces the defaults when provided** (Alchemy v1 semantics), rather than Wrangler's append-with-fallthrough-warnings.
- **Default rules are a superset of v1's** (`**/*.js`, `**/*.mjs`, `**/*.wasm`, plus Wrangler's `no_bundle` Text/Data rules: `**/*.txt`, `**/*.html`, `**/*.sql`, `**/*.bin`).
- **Glob matching includes dotfiles** (`dot: true`) — prebuilt output dirs commonly nest under hidden directories (`.open-next/`), and Wrangler's matching doesn't special-case them either.
- **No sourcemap upload**: v1 uploaded `.js.map` as modules; Wrangler routes sourcemaps through a separate upload channel that doesn't exist here yet, and maps-as-modules is arguably wrong. Omitted for now.
- **Durable Object / Workflow classes must be exported by the prebuilt entry itself** — `exports` is not applied when `bundle: false` (documented on the prop).

Also related:

- Users hitting #589 (re-bundle mangling a pre-built `main`) can now opt out entirely — `bundle: false` skips the bundler, including its pure-annotation pass. The default path's behavior there is unchanged.
- #590 — until `Cloudflare.Vite` captures multi-environment RSC builds, `bundle: false` is the working path for `@vitejs/plugin-rsc` apps: build with vite, deploy the emitted server directory as-is.

Not covered (same scope as #117): `alchemy dev` still rolldown-watches `main`; a no-bundle dev watcher (à la wrangler's `findAdditionalModuleWatchDirs`) is a follow-up.

### DIFF
--- a/packages/alchemy/src/Bundle/Bundle.ts
+++ b/packages/alchemy/src/Bundle/Bundle.ts
@@ -311,7 +311,7 @@ function bundleErrorFromUnknown(error: unknown): BundleError {
   });
 }
 
-function bundleOutputFromFiles(
+export function bundleOutputFromFiles(
   files: [BundleFile, ...BundleFile[]],
 ): Effect.Effect<BundleOutput> {
   return Effect.map(
--- a/packages/alchemy/src/Cloudflare/Workers/Worker.ts
+++ b/packages/alchemy/src/Cloudflare/Workers/Worker.ts
@@ -225,39 +225,6 @@ export interface WorkerProps<
    */
   main?: string;
   /**
-   * Whether to bundle {@link main} with rolldown before upload.
-   *
-   * Set to `false` when `main` already points at a complete,
-   * runtime-ready ESM Worker produced by an external tool (OpenNext,
-   * a separate rolldown/esbuild pipeline, etc.). The entry and every
-   * file around it matching {@link rules} are uploaded byte-for-byte —
-   * no bundling, no minification, no transformation. Module names are
-   * the files' POSIX paths relative to the entry's directory, matching
-   * Wrangler's `no_bundle` contract.
-   *
-   * Re-bundling such artifacts is unsafe: dynamic `import()` calls the
-   * upstream tool relies on can be rewritten in ways that break runtime
-   * behavior.
-   *
-   * Durable Object and Workflow classes must be exported by the prebuilt
-   * entry itself — {@link exports} is not applied when `bundle` is
-   * `false`.
-   *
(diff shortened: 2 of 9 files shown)


agcty:
Verification beyond the included tests: we deployed a production React Router RSC app through this branch — a 22 MB worker of 506 module files collected by the glob walk, including `@vitejs/plugin-rsc`'s disk-written `__vite_rsc_assets_manifest.js` and the nested `__ssr_build/` environment output (the exact files an in-memory capture misses, per #590), plus Durable Object classes exported from the prebuilt entry. `GET /` renders 200 end-to-end (RSC render → Hyperdrive → PlanetScale) and a DB-probing health route returns OK. Same app previously required `pure: false` to survive the re-bundle (#589) — with `bundle: false` no workaround is needed.

john-royal:
Beautiful! Going to give this a once-over right now but this is exactly what I wanted - thank you for helping to push this over the line!

PR #596, opened 2026-06-11
fix(core): resolve per-field Effect inputs in resource props

### LINKED ISSUE #588: Worker: `domain` prop is read without resolving per-field Effect inputs (crashes domain reconciliation)
## Repro (2.0.0-beta.55)

The class-form Worker/StaticSite pattern relies on a plain props literal so `InferEnv<typeof Website>` can infer the binding map. Stage-dependent values are therefore expressed as **per-field Effect inputs** (the `InputProps` machinery), not by branching the whole props object:

```ts
export class Website extends Cloudflare.StaticSite<Website>()("website", {
  // ...
  url: Alchemy.Stack.useSync(stack => stack.stage !== "prod"),
  domain: Alchemy.Stack.useSync(stack => (stack.stage === "prod" ? "example.com" : undefined)),
  env: { /* bindings — per-field Effects here resolve fine */ },
})
```

Deploying any stage fails at the final step:

```
[Worker] Reconciling custom domains (1) ...
[Worker] fail
ERROR: SchemaError: Expected string | undefined, got {"~effect/Effect/args":(fiber) => ...}
  at ["hostname"]
    at GET /accounts/{account_id}/workers/domains
```

## Cause

`Worker.ts` feeds the raw prop into `normalizeDomains(news.domain)` without resolving Input effects in that path. Env bindings and other props resolve fine — this domain path reads `news` directly. Two consequences:

1. The unresolved Effect object is neither `undefined` nor an array, so it normalizes to a one-element domain list — reconciliation runs even on stages where the value would resolve to `undefined`.
2. The object reaches the API client as `hostname` and fails schema validation.

`news.url !== false` a few lines down looks like the same unresolved read — an Effect object is truthy, so a `url` that resolves to `false` on some stages would still enable the workers.dev subdomain there.

## Why per-field inputs (not in-generator construction)

The obvious alternative — construct the Worker inside the stack generator with different props per stage — **breaks `InferEnv`**: once the props object is produced by a generator rather than a literal, the binding-map type is lost, so `Cloudflare.InferEnv<typeof Website>` (used to type `env.d.ts`) no longer works. Per-field Effect inputs are the only shape that keeps the class-form inference *and* lets values vary by stage. That makes consistent input resolution across props a requirement of the class form you ship, not a request for a new pattern.

## Expected

`domain` and `url` resolve per-field Effect inputs on the provider path the same way `env` bindings already do.

Happy to PR the fix if you point me at the intended resolution point for that path (same fork as #586/#587).

agcty:
Fixes #588.

## Problem

The class-form Worker/StaticSite pattern requires a plain props literal so `InferEnv<typeof Website>` can infer the binding map — stage-dependent values must therefore be **per-field Effect inputs** (`domain: Stack.useSync(stack => stack.stage === "prod" ? "example.com" : undefined)`), not whole-props branching. `Input<T>` admits `Effect<T>` for any prop, and the `isResolved` doc in `Diff.ts` already describes props as potentially containing "an `Effect` that was not fully evaluated by `resolveInput` in Plan.ts" — but neither `resolveInput` nor `Output.evaluate` had a branch for raw Effects:

- **`resolveInput` (Plan.ts)** passed the Effect through to `provider.diff` and the plan node's props, tripping providers' `isResolved` guards into conservative fallbacks.
- **`Output.evaluate` (Output.ts)** shredded it via the generic object walk into a plain `{"~effect/Effect/args": ...}` object that reached provider APIs — for the Worker's domain reconciliation, a crash in the Cloudflare client's schema validation; and because the unresolved object is truthy, a `domain` that should have resolved to `undefined` entered reconciliation at all. `news.url !== false` had the same hazard.

## Fix

Resolve raw Effects in both walkers, recursing into the result — mirroring the Config branches added in #566 (same two functions, same recursion shape). Repeated evaluation of the same input Effect across plan and apply matches how `EffectExpr` outputs already behave (see the existing memoization TODO on `Output.evaluate`).

Two kinds of Effects are *references*, not per-field values, and are carved out:

- **Function-form Effects** (resource classes, `effectClass` constructors, `Binding.Policy` tags) — class references carried in props (e.g. Worker `exports`); guarded by `typeof !== "function"`.
- **Non-class resource references** (`const db = Hyperdrive("db", …)` placed in `env`) — bare constructor Effects. The constructor's FQN derives from the ambient `Namespace` at execution time, so executing one during input resolution (no namespace in scope) re-derives a different FQN and mints a phantom resource — `MissingSourceError: Source db not found`. The constructor now brands its Effect (`Symbol.for("alchemy/ResourceEffect")`, the same duplicate-module-safe pattern as `alchemy/Expr`), and the walkers leave branded references opaque. They also count as resolved in `Diff.isResolved` (so provider diffs keep running) and are skipped by the Worker vite env mapping.

Plain Effects that *wrap* a reference (`AI_GATEWAY_ID: Effect.map(aiGateway, g => g.gatewayId)`) can't be detected without running them — `Effect.map` strips the brand — and per-field resolution must run them. To make that safe, the constructor additionally memoizes its result per stack (`WeakMap` keyed on the `Stack` service object): re-execution returns the resource registered during construction, and the resulting attribute expression resolves through the normal Output machinery. The `// TODO(sam): check if props are different and die` on the existing-FQN registry path is untouched — on `main` the only dedup is the stack registry (`stack.resources[fqn]`), and the memo adds construction identity alongside it without changing that behavior.

Worth saying plainly: for *resolved* resources (class form, or after `yield*`), attribute expressions (`backend.url`) are the canonical way to reference another resource's output. But for an **unyielded non-class reference inside a props literal** there is no attribute surface — the reference is the bare constructor Effect, so `Effect.map(resource, …)` is the only way to project a field from it. The memo is what makes that form safe instead of silently minting a phantom resource. (We initially tried migrating our app to `aiGateway.gatewayId` and the type-checker correctly rejected it — the runtime value would be `undefined`.)

## Out of scope, noted while auditing

- `StaticSite` stringifies non-string `env` values for the **build command's** environment at construction (`JSON.stringify(v)`), before any engine resolution — a separate provider-local issue for Effect-valued env there.
- `stringifyResolvedString` in AWS `Website/StaticSite.ts`/`Router.ts` composes per-field Effects into inner-resource props; those were latently broken the same way as #588 and are healed by this change.

## Tests

- `Output.test.ts`: evaluate resolves raw Effects (top-level/nested/array/recursive), resolved-`undefined` stays `undefined`, Redacted results stay wrapped, effect classes **and non-class resource references** pass through untouched.
- `plan.test.ts`: Effect props arrive concrete in plan props; resolved-`undefined` lands as `undefined`; diff observes resolved values (noop on unchanged); a resource reference stays opaque with no phantom plan node.
- `apply.test.ts`: end-to-end `Stack.useSync` props (including the resolved-`undefined` skip-guard case); a namespaced resource reference embedded in another resource's props deploys with the reference arriving un-executed (reproduces the `MissingSourceError` on main); a plain Effect wrapping a reference resolves to the real resource's attribute.
- `Diff.test.ts`: per-field Effects count as unresolved; branded references count as resolved.

All new tests fail on `main` and pass with the fix; `bun tsc -b`, oxlint, and the engine/Local/Build suites are green.

Verified beyond the test suite by deploying a production app through this branch: a Worker with a stage-conditional `domain: Stack.useSync(…)` (the #588 repro — crashed every stage before), ~20 bindings including Hyperdrive/R2/KV resource references in `env`, Durable Objects, and a wrapped-reference prop. Deploys cleanly; the `undefined`-resolved domain skips reconciliation; the app serves 200 end-to-end.

### DIFF
--- a/packages/alchemy/src/Cloudflare/Workers/Worker.ts
+++ b/packages/alchemy/src/Cloudflare/Workers/Worker.ts
@@ -21,7 +21,11 @@ import type { InputProps } from "../../Input.ts";
 import * as ProviderLayer from "../../Local/ProviderLayer.ts";
 import { Platform, type Main, type PlatformProps } from "../../Platform.ts";
 import * as Provider from "../../Provider.ts";
-import { Resource, type ResourceBinding } from "../../Resource.ts";
+import {
+  isResourceEffect,
+  Resource,
+  type ResourceBinding,
+} from "../../Resource.ts";
 import { Stack } from "../../Stack.ts";
 import { CloudflareEnvironment } from "../CloudflareEnvironment.ts";
 import type { HyperdriveDevOrigin } from "../Hyperdrive/Hyperdrive.ts";
@@ -1182,7 +1186,10 @@ export const LiveWorkerProvider = () =>
                       : Redacted.isRedacted(value) &&
                           typeof Redacted.value(value) === "string"
                         ? Redacted.value(value)
-                        : Effect.isEffect(value)
+                        : // Resource references are opaque handles, not env
+                          // var values — running one here would register a
+                          // resource outside the construction phase.
+                          Effect.isEffect(value) && !isResourceEffect(value)
                           ? yield* value as Effect.Effect<any>
                           : undefined,
                   ];
--- a/packages/alchemy/src/Diff.ts
+++ b/packages/alchemy/src/Diff.ts
@@ -3,7 +3,7 @@ import * as Redacted from "effect/Redacted";
 import type { Input } from "./Input.ts";
 import * as Output from "./Output.ts";
 import type { BindingNode } from "./Plan.ts";
-import type { ResourceBinding } from "./Resource.ts";
+import { isResourceEffect, type ResourceBinding } from "./Resource.ts";
 import { isPrimitive } from "./Util/data.ts";
 
 export type Diff = NoopDiff | UpdateDiff | ReplaceDiff;
@@ -45,6 +45,10 @@ export const isResolved = <T>(value: Input<T>): value is T =>
 
 const _hasUnresolved = (value: unknown): boolean => {
   if (value == null || isPrimitive(value)) return false;
+  // Non-class resource references are permanently opaque — `resolveInput`
+  // never executes them (see Plan.ts) — so they don't make props
+  // "unresolved": a provider's diff can still run against the rest.
+  if (isResourceEffect(value)) return false;
   if (Output.isExpr(value) || Effect.isEffect(value)) return true;
   if (Array.isArray(value)) return value.some(_hasUnresolved);
   if (typeof value === "object") {
--- a/packages/alchemy/src/Output.ts
+++ b/packages/alchemy/src/Output.ts
@@ -7,7 +7,12 @@ import type { Pipeable } from "effect/Pipeable";
 import * as Redacted from "effect/Redacted";
 import { SingleShotGen } from "effect/Utils";
 import { getRefMetadata, isRef, type Ref } from "./Ref.ts";
-import { isResource, type Resource, type ResourceLike } from "./Resource.ts";
+import {
+  isResource,
+  isResourceEffect,
+  type Resource,
+  type ResourceLike,
+} from "./Resource.ts";
 import { RuntimeContext, sanitizeKey } from "./RuntimeContext.ts";
 import { Stack } from "./Stack.ts";
 import { Stage } from "./Stage.ts";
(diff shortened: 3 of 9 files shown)


sam-goodwin:
I don't think we should do this, and we should instead change `Input<T>` to now allow an Effect because we can't capture the requirements and error channels of the effect, which breaks the type safety of Effect

agcty:
That's fair — a raw `Effect<T, E, R>` in props erases both the error and requirements channels, and resolving it in the walkers codifies that erasure rather than fixing it.

The thing that needs a home before `Input<T>` drops Effects is stage-conditional per-field values. The class-form Worker pattern needs a plain props literal for `InferEnv`, so whole-props branching is out — and today the only per-field form is `domain: Stack.useSync(stack => stack.stage === "prod" ? "example.com" : undefined)`, which is exactly the raw-Effect shape you want gone.

Proposal: have `Stack.useSync` (or a sibling) return a branded, channel-free expression — the same expression surface the Output machinery already resolves — instead of a raw Effect. Then `Input<T>` can stop admitting `Effect` at the type level, `Config` stays the secrets path, and field projection off resources stays attribute expressions on yielded resources. The walkers need no new Effect branch at all; the half-support gets deleted instead of completed.

If that matches what you have in mind, I'll rework this PR into that shape — the repro and most of the tests carry over as-is.

sam-goodwin:
The whole `props` can be an Effect, so you should use `Stack.useSync` and then return the props. Use `Effect.all` or `Effect.flatMap` to compose with other effects and then finally return the props for that resource. Captures the requirements and allows arbitrary data to come from 1..* effects.
```ts
Worker("worker", Stack.useSync(stack => ({
  domain: stack.stage === "prod" ? .. : ..
})
```

What am I missing?

sam-goodwin:
Please re-open if necessary

PR #598, opened 2026-06-11
fix(pr-package): unwrap Redacted on both sides of bearer-token auth

Mkassabov:
Supersedes [#568](https://github.com/alchemy-run/alchemy-effect/pull/568). That PR fixes the worker side of the auth check, but the bug is symmetric — fixing only one side breaks every existing publisher.

### The symmetric bug

`Cloudflare.Secret.bind(...)` resolves to `Redacted<string>`. `Redacted.toString()` returns the literal string `"<redacted>"`, and `toJSON()` delegates to `toString()`.

Worker side (in `packages/pr-package/src/Worker.ts`):

```ts
const expected = yield* authToken;            // Redacted<string>
if (authHeader !== `Bearer ${expected}`) {... // compares against "Bearer <redacted>"
```

Stack-output side (in `stacks/pr-package.ts`):

```ts
return {
  authToken: authToken.text,  // Redacted<string> — JSON-serializes to "<redacted>"
};
```

When that stack output is persisted to state and pulled into Doppler / `PR_PACKAGE_TOKEN`, the env var ends up as the literal string `<redacted>`. The publisher then sends `Authorization: Bearer <redacted>`, and the worker's broken check also produces `"Bearer <redacted>"`. They match. Two bugs cancelling — which is why this "worked" in prod.

### Fix

Unwrap on both sides.

```diff
// packages/pr-package/src/Worker.ts
-  if (!authHeader || authHeader !== `Bearer ${expected}`) {
+  if (!authHeader || authHeader !== `Bearer ${Redacted.value(expected)}`) {
```

```diff
// stacks/pr-package.ts
-    authToken: authToken.text,
+    authToken: Redacted.value(authToken.text),
```

README example updated to match.

### Rollout note

The currently-distributed `PR_PACKAGE_TOKEN` is the literal string `<redacted>`. After this lands, re-deploy the stack and re-publish the new `authToken` output to Doppler / GH Actions secrets so publishers send the real token.

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -702,6 +702,17 @@
         "effect": "catalog:",
       },
     },
+    "examples/cloudflare-worker-auth": {
+      "name": "cloudflare-worker-auth",
+      "version": "0.0.0",
+      "dependencies": {
+        "@cloudflare/workers-types": "catalog:",
+        "@effect/platform-bun": "catalog:",
+        "@effect/platform-node": "catalog:",
+        "alchemy": "workspace:*",
+        "effect": "catalog:",
+      },
+    },
     "examples/monorepo-multi-stack/backend": {
       "name": "@monorepo-multi-stack/backend",
       "version": "0.0.0",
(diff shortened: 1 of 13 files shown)


sam-goodwin (approved):

boarush:
This fix is incomplete. When returning from the stack, since the Redacted value is wrapped in an `Alchemy.Output`, it fails to type check correctly, it is required to pipe it to `Output.map` before passing to a `Redacted.value` to print the raw auth token.

PR #600, opened 2026-06-11
feat(build): Build.Exec — run a command for its side effects, no artifact contract

### LINKED ISSUE #599: Build.Command's required outdir forces an artifact contract on effectful commands (no v2 Exec equivalent)
## Problem

`Build.Command` requires `outdir` and treats it as an owned artifact, which imposes three obligations that don't fit effectful commands:

1. **The path must exist after the run** — reconcile dies with `Build completed but output path does not exist` otherwise (`Command.ts:184-187`).
2. **Memo-skip is gated on the artifact existing locally** (`Command.ts:166-171`) — on a fresh checkout or CI runner the command re-runs even with unchanged inputs.
3. **`delete` removes `outdir` recursively** (`Command.ts:202-208`).

For commands whose value is the execution itself — database seeds, `drizzle-kit migrate` for users who keep drizzle as the migration authority, cache warms — there is no artifact. The workaround is to fake one:

```ts
const seed = yield* Build.Command("Seed", {
  command: "bun run src/seed.ts && mkdir -p node_modules/.cache/alchemy/seed",
  outdir: "node_modules/.cache/alchemy/seed",
  // ...
})
```

This is ceremony that lies to the engine, and obligation 3 makes the natural-looking misuse destructive: we initially pointed `outdir` at our `migrations/` directory, and `alchemy destroy` recursively deleted committed source files from the working tree.

## Why the existing pieces don't cover this

- **`RemovalPolicy.retain()`** only governs the delete phase. It prevents the destroy incident but does nothing about obligations 1 and 2 — the stamp-dir hack is still required, and memo-skip still breaks on fresh machines.
- **`Drizzle.Schema` / `migrationsDir` / `Planetscale.PostgresMigrations`** cover one use case (migrations) for specific tools. Seeding, cache warming, and migrate flows that keep the tool's own journal as the applied-tracking authority have no home.
- **v1 had `Exec`** for exactly this kind; v2's `Build.Command` conflates "run a command when inputs change" with "produce a directory of files".

## Proposal

A `Build.Exec` sibling resource (mirroring how `Build.DevServer` split out the long-lived-process kind): no `outdir`, output is a single run-key hash, `delete` is a no-op that only forgets the key. The run key must cover the **command, `cwd`, and `env`** in addition to the memoized files — hashing files alone would silently skip a run when only the environment changed (e.g. a recreated database's connection URL), which is the headline use case.

PR with implementation + tests incoming.

agcty:
Closes #599.

## What

`Build.Exec` — a sibling of `Build.Command` for commands run for their side effects: database seeds, migrations, cache warms. No `outdir`, output is a single run-key `hash`, and `delete` is an explicit no-op — the command's side effects are not the engine's to reverse, so deleting the resource only forgets the run key. Registered next to `Build.CommandProvider()` in both the Cloudflare and AWS registries.

## Why a separate resource

Making `outdir` optional on `Build.Command` would turn its output type conditional (`outdir?: string`), and real consumers read it — `Cloudflare/Website/StaticSite.ts` (`directory: build.outdir`) and the AWS StaticSite. The Build family already splits by kind (`DevServer` is the long-lived-process sibling), and v1 had `Exec` for exactly this. Two existing mechanisms cover parts of the problem but not the kind itself:

- `RemovalPolicy.retain()` only governs delete. It doesn't touch the other two artifact obligations: `Command` dies when `outdir` doesn't exist after the run, and memo-skip is gated on the artifact existing locally — so the stamp-dir workaround re-runs effectful commands on every fresh checkout/CI machine even with unchanged inputs.
- `Drizzle.Schema` / `migrationsDir` are tool-specific. Seeding has no home at all, and migrate flows that keep the tool's own journal as the applied-tracking authority don't fit `migrationsDir` either.

## The run key covers props, not just files

`reconcile` and `diff` key on `sha256Object({ files, command, cwd, env })`, not the file hash alone. Hashing files alone silently skips a run when only the command or its environment changed — for the headline use case (`env: { DATABASE_URL: database.connectionUrl }`), replacing the database would skip the migration forever while the plan reports success. `DevServer` already keys respawn on whole props (`Hash.structure(news)`); `Exec` follows. Details:

- `env` values are folded in Redacted-unwrapped (one-way hash in state), so a rotated secret re-runs the command.
- `cwd` is hashed as given, not resolved — absolute paths in persisted state break across machines/checkouts (same reasoning as `Drizzle.Schema`'s relative-path note).
- The JSDoc states the idempotency requirement: per the reconciler doctrine the engine may re-run the command after a crash between the run and state persistence, and documents the default-memo foot-gun (exclude files the command itself writes inside `cwd`).

No `read` — a local command run leaves nothing observable to adopt or refresh; the engine only consults `read` when state is missing, and the correct answer there is "nothing to adopt".

## Tests

`test/Build/Exec.test.ts` (same harness as `Build.test.ts`): first deploy runs; unchanged redeploy skips; **env-only** change re-runs; **command-only** change re-runs; memoized file change re-runs; destroy leaves side effects untouched and forgets the run key, so an unchanged redeploy after destroy runs again. `bun tsc -b`, format, and the full Build suite pass.

## Verified beyond the suite

Deployed a production app's database stack through this branch (PlanetScale branch + role + `drizzle-kit migrate` + a seed script, the seed's env threading `MIGRATIONS_HASH: migrations.hash` for ordering):

- fresh stage deploy creates the branch/role and runs Migrate then Seed;
- unchanged redeploy noops everything (`Done: 0 succeeded`);
- touching the seed script re-runs only Seed (plan: `1 to update, 3 to noop`);
- destroy deletes branch/role and leaves the working tree untouched — replacing the stamp-dir workaround (and the incident where a source-dir `outdir` got recursively deleted on destroy) with the declaration meaning what it says.


sam-goodwin:
Should we instead just update `Command.outdir` to be optional?

agcty:
I considered that first — it's defensible, but three things pushed me to a sibling resource:

1. **The output type goes conditional.** `Command`'s Attributes are a fixed shape on the `Resource` declaration, so optional `outdir` means every consumer sees `outdir?: string` — and real consumers read it (`StaticSite` passes `build.outdir` as the assets directory, same on the AWS side).
2. **`outdir` isn't one prop, it's three behaviors**: the must-exist-after-run check, memo-skip gated on the artifact existing locally, and the recursive delete. Optional `outdir` means `reconcile` and `delete` branch on its presence throughout — one resource with two behavioral modes.
3. **The memo key genuinely differs.** `Exec` folds command/cwd/env into the run key — a recreated database's `DATABASE_URL` must re-trigger the migration with unchanged files. `Command`'s reuse legitimately consults the artifact's existence instead.

`DevServer` set the precedent of splitting the Build family by kind, and v1 had `Exec` as its own primitive.

That said, it's your surface — if you'd rather have optional `outdir`, happy to rework. The parts I'd want to preserve either way: props in the run key, no artifact gating on memo-skip, and a non-destructive delete for the artifact-less case.

john-royal:
Replacing with #673

PR #605, opened 2026-06-12
chore(test): fix Windows test failures and eager Cloudflare credential resolution

Mkassabov:
Fixes the 19 test failures / 2 suite failures from a full local `vitest run` on Windows. Most were Windows-specific test bugs, but two were real cross-platform source bugs the run surfaced.

### Source fixes

- **Eager Cloudflare credential resolution** (`Gateway/Rule.ts`, `Tunnel/Configuration.ts`): both providers did

  ```ts
  const { accountId } = yield* yield* CloudflareEnvironment; // at layer build!
  ```

  forcing auth (or an interactive login prompt) just to construct the provider collection — `Providers.test.ts`'s unknown-profile guard test caught it by hanging on the Clack prompt. Both now hold the env effect and resolve `accountId` inside lifecycle ops, per the providers convention.
- **`PurePlugin.resolvePackageInfo`** walked above the `node_modules/<pkg>` root, so a stray `package.json` higher up (e.g. a drive-root manifest on Windows) masked the path-derived package name. The walk now stops at the package's own root.
- **`Drizzle/Schema.diff`** forced `{ action: "update" }` whenever `output.out` was absolute (legacy-state migration shim). On Windows an `out` on a different drive than cwd has no relative form, so every deploy flagged an update. Now compares against today's canonical `relativeOut` form instead.
- **Cloudflare OAuth `DEFAULT_SCOPES`**: added `dns_records:read/edit` + `teams:read/write` so a default oauth login can manage the DnsRecord and Tunnel/Gateway resources alchemy ships.

### Test/fixture fixes

- `new URL(...).pathname` → `fileURLToPath(...)` (yields `/D:/...` on Windows) in `Lambda/Function.test.ts`, `Local/RpcServer.test.ts`
- `pidListeningOn` fixture: `netstat -ano` on win32 (`lsof` doesn't exist) — un-cascades all 7 RpcSpawner/RpcSpawnerCleanup failures
- `PurePlugin.test.ts`: junction instead of dir symlink on win32 (no elevation needed; type arg ignored on POSIX)
- `Build.test.ts`: normalize platform separators before comparing `outdir`

### Not fixed (credentials, not code)

`Zone`, `Ruleset`, `Dns`, `Tunnel` suites provision **zones**, and Cloudflare's OAuth client has no `zone:write` scope at all — those suites require API-token auth (`CLOUDFLARE_API_TOKEN`, what CI uses via `TEST_CLOUDFLARE_API_TOKEN`).

All fixed suites verified locally on Windows (PurePlugin 30/30, RpcSpawner 5/5, RpcSpawnerCleanup 4/4, RpcServer, Build, Drizzle 3/3, Providers, Lambda Function 2/2 incl. live deploys).

### DIFF
--- a/packages/alchemy/test/AWS/Lambda/Function.test.ts
+++ b/packages/alchemy/test/AWS/Lambda/Function.test.ts
@@ -7,9 +7,13 @@ import * as Effect from "effect/Effect";
 import * as Schedule from "effect/Schedule";
 import * as HttpClient from "effect/unstable/http/HttpClient";
 import { TestFunction, TestFunctionLive } from "./handler.ts";
+import { fileURLToPath } from "node:url";
 
-const timeoutHandlerPath = new URL("./timeout-handler.ts", import.meta.url)
-  .pathname;
+// `URL.pathname` yields `/D:/...` on Windows; fileURLToPath gives a real
+// filesystem path on every platform.
+const timeoutHandlerPath = fileURLToPath(
+  new URL("./timeout-handler.ts", import.meta.url),
+);
 
 const { test } = Test.make({ providers: AWS.providers() });
 
--- a/packages/alchemy/src/Bundle/PurePlugin.ts
+++ b/packages/alchemy/src/Bundle/PurePlugin.ts
@@ -282,6 +282,21 @@ export function packageNameFromId(id: string): string | null {
 }
 
 /**
+ * Returns the directory of the `node_modules/<pkg>` root that owns `id`,
+ * or `null` when the id has no `node_modules/` segment. Preserves the
+ * id's original separators so it compares cleanly against
+ * `path.dirname` results.
+ */
+function packageRootFromId(id: string): string | null {
+  const name = packageNameFromId(id);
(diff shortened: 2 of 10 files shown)


packages/alchemy/test/AWS/Lambda/Function.test.ts
```diff
@@ -7,9 +7,13 @@ import * as Effect from "effect/Effect";
...
 import { TestFunction, TestFunctionLive } from "./handler.ts";
+import { fileURLToPath } from "node:url";
 
-const timeoutHandlerPath = new URL("./timeout-handler.ts", import.meta.url)
-  .pathname;
+// `URL.pathname` yields `/D:/...` on Windows; fileURLToPath gives a real
+// filesystem path on every platform.
+const timeoutHandlerPath = fileURLToPath(
+  new URL("./timeout-handler.ts", import.meta.url),
+);
```
sam-goodwin, line 16: `pathe.resolve(import.meta.dirname, "timeout-handler.ts")`?

PR #607, opened 2026-06-12
Add GitHub Repository resource

just-be-dev:
I've been wanting to move my GitHub configuration over to an IaC solution for a bit now so I can stop continually flipping off merge commits and other random bits like that. Y'all don't have a GitHub resource, but I was hoping this would be useful. 

The implementation was agent driven, but I've done a review pass. I assume that eventually you'll actually just want coverage in distilled for GitHub and this'll be handled more like other providers. 

## AI Summary

Manages the full lifecycle of a GitHub repository — create, update, rename, and delete — with a reconciler that converges settings on every deploy.

- Split `archived` into its own PATCH (GitHub rejects other settings changes in the same archiving call)
- Guard `default_branch` against non-existent branches with a 422 retry that strips the field rather than hard-failing
- `topics` cleared when field is removed (was silently left behind)
- `read` implemented via stable numeric `repoId` so renames don't leave stale names in state
- `delete` resolves live owner/name by `repoId` before deleting, surviving state-persistence failures after a rename
- `createInOrg` uses `visibility` directly; `internal` repos now created with the correct visibility class
- `replaceAllTopics` uses `updated.name` (confirmed post-PATCH name)

### DIFF
--- a/packages/alchemy/src/GitHub/Repository.ts
+++ b/packages/alchemy/src/GitHub/Repository.ts
@@ -135,15 +135,6 @@ export interface RepositoryProps {
    * Only used at create time.
    */
   licenseTemplate?: string;
-
-  /**
-   * Whether to delete the repository when the resource is destroyed. By
-   * default the repository is preserved on destroy to protect its history
-   * (issues, pull requests, commits). Set to `true` to opt in to deletion —
-   * the token must have the `delete_repo` scope.
-   * @default false
-   */
-  allowDelete?: boolean;
 }
 
 export interface Repository extends Resource<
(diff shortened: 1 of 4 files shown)


packages/alchemy/src/GitHub/Repository.ts
```diff
@@ -0,0 +1,610 @@
...
+  nodeId: data.node_id,
+  fullName: data.full_name,
+  htmlUrl: data.html_url,
+  gitUrl: data.git_url,
+  sshUrl: data.ssh_url,
+  cloneUrl: data.clone_url,
+  defaultBranch: data.default_branch,
+  createdAt: data.created_at ?? new Date().toISOString(),
+  updatedAt: data.updated_at ?? new Date().toISOString(),
+});
```
sam-goodwin: nit: all helpers should be below the provider. Types and significant code comes first.

packages/alchemy/src/GitHub/Repository.ts
```diff
@@ -0,0 +1,610 @@
...
+ * @example Allow Repository Deletion
+ * ```typescript
+ * const repo = yield* GitHub.Repository("ephemeral", {
+ *   owner: "my-org",
+ *   name: "ephemeral-preview",
+ *   allowDelete: true,
+ * });
+ * ```
+ */
+export const Repository = Resource<Repository>("GitHub.Repository");
```
sam-goodwin: Set default removal policy to retain (and document, see Zone.ts for an example)

just-be-dev: Hmm, I thought I had done that. Will follow up on it.

just-be-dev: Ahh, okay, no, my implementation was going down the wrong path. I see what you mean. Should be updated now.

packages/alchemy/src/GitHub/Repository.ts
just-be-dev: Updated

PR #611, opened 2026-06-13
feat(cloudflare)!: expand workflows api

Gerbuuun:
Expand the Effect-native Cloudflare Workflow wrapper to cover the current Workers API surface, 1:1 with the native binding.

### Breaking

`create` now takes the native `WorkflowInstanceCreateOptions` object instead of the bare input payload:

```diff
-const instance = yield* workflow.create({ orderId: "abc" });
+const instance = yield* workflow.create({ params: { orderId: "abc" } });
```

`id` and `retention` are also accepted:

```ts
const instance = yield* workflow.create({
  id: "order-abc",
  params: { orderId: "abc" },
  retention: { successRetention: "1 day", errorRetention: "7 days" },
});
```

### New API

`task` takes an optional third argument for retries, timeout, and rollback; `WorkflowStepContext` exposes the current attempt:

```ts
const result = yield* Cloudflare.Workflows.task(
  "call-api",
  Effect.gen(function* () {
    const context = yield* Cloudflare.Workflows.WorkflowStepContext;
    return { attempt: context.attempt };
  }),
  {
    retries: { limit: 3, delay: "5 seconds", backoff: "linear" },
    rollback: ({ output }) => (output ? cleanup(output.id) : Effect.void),
  },
);
```

`waitForEvent` parks the instance until a matching `sendEvent`, resolving with the native `{ payload, timestamp, type }` event:

```ts
const approval = yield* Cloudflare.Workflows.waitForEvent<{ approved: boolean }>(
  "approval",
  { type: "approval", timeout: "1 day" },
);
approval.payload.approved;
```

```ts
yield* instance.sendEvent({ type: "approval", payload: { approved: true } });
```

- Add `createBatch`, `restart`, rollback status, and extended event metadata (`workflowName`, cron `schedule`)
- Update the workflow guides (`workflows.mdx`, `add-a-workflow.mdx`) for the new `create({ params })` signature and new primitives
- Verified live: `ALCHEMY_PROFILE=testing bun vitest run test/Cloudflare/Workers/Workflow.test.ts`


packages/alchemy/src/Cloudflare/Workers/Workflow.ts
```diff
@@ -252,17 +420,48 @@ export class WorkflowScope extends Context.Service<
...
+ *   effect: Effect.gen(function* () {
+ *     const context = yield* Cloudflare.WorkflowStepContext;
+ *     return { attempt: context.attempt };
+ *   }),
+ * });
+ * ```
+ *
+ * @example Registering rollback
+ * ```typescript
+ * yield* Cloudflare.task({
```
sam-goodwin: Should this be .do? Are all the docs wrong?

packages/alchemy/src/Cloudflare/Workers/Workflow.ts
```diff
@@ -252,17 +420,48 @@ export class WorkflowScope extends Context.Service<
...
+ * const result = yield* Cloudflare.task({
+ *   name: "process-order",
+ *   effect: Effect.succeed({ orderId: "abc", total: 42 }),
+ * });
+ * ```
+ *
+ * @example Configuring retries and reading step context
+ * ```typescript
+ * const result = yield* Cloudflare.task({
+ *   name: "call-api",
```
sam-goodwin: I think it wouldn't be better if the id was the first argument instead of an object property to align with the rest of alchemy's api design.

packages/alchemy/src/Cloudflare/Workers/Workflow.ts
```diff
@@ -252,17 +420,48 @@ export class WorkflowScope extends Context.Service<
...
  * ```typescript
  * yield* Cloudflare.sleep("cooldown", "30 seconds");
  * ```
  *
+ * @example Waiting for an external event
+ * ```typescript
+ * const event = yield* Cloudflare.waitForEvent<{ approved: boolean }>(
+ *   "approval",
+ *   { type: "approval", timeout: "1 day" },
+ * );
```
sam-goodwin: Should we support an Effect schema here?

packages/alchemy/src/Cloudflare/Workers/Workflow.ts
```diff
@@ -30,19 +30,99 @@ export class WorkflowEvent extends Context.Service<
...
+  rollback?: (
+    context: WorkflowRollbackContext<Output>,
+  ) => Effect.Effect<void, never, RollbackReq>;
+  rollbackConfig?: WorkflowStepConfig;
+}
+
+export interface WorkflowWaitForEventOptions {
+  type: string;
+  timeout?: string | number;
+}
```
just-be-dev, line 119: These could just be defined inline. It's how they do it. 

```ts
    export abstract class WorkflowStep {
        do<T extends Rpc.Serializable<T>>(name: string, callback: (ctx: WorkflowStepContext) => Promise<T>, rollbackOptions?: WorkflowStepRollbackOptions<T>): Promise<T>;
        do<T extends Rpc.Serializable<T>>(name: string, config: WorkflowStepConfig, callback: (ctx: WorkflowStepContext) => Promise<T>, rollbackOptions?: WorkflowStepRollbackOptions<T>): Promise<T>;
        sleep: (name: string, duration: WorkflowSleepDuration) => Promise<void>;
        sleepUntil: (name: string, timestamp: Date | number) => Promise<void>;
        waitForEvent<T extends Rpc.Serializable<T>>(name: string, options: {
            type: string;
            timeout?: WorkflowTimeoutDuration | number;
        }): Promise<WorkflowStepEvent<T>>;
    }

```

I do need this though, so thank you for putting up the PR!

just-be-dev: Ahh, nvm, I see why you separated it out.

Gerbuuun:
Thanks for the feedback. I'm in the middle of exam week so it'll be probably end of next week I'll continue working on this

packages/alchemy/src/Cloudflare/Workers/Workflow.ts
john-royal: Done — reshaped to id-first so the name leads and the Effect stays positional, with retries/timeout/rollback in an optional third config object:

```ts
yield* Cloudflare.task("process-order", Effect.succeed({ orderId: "abc" }));

yield* Cloudflare.task("call-api", effect, {
  retries: { limit: 3, delay: "5 seconds", backoff: "linear" },
  rollback: ({ output }) => output ? cleanup(output.id) : Effect.void,
});
```

This keeps the simple `task(name, effect)` call working unchanged.

packages/alchemy/src/Cloudflare/Workers/Workflow.ts
john-royal: Deferring this for now to keep the PR focused — `waitForEvent` (and `task`/`create`) stay on the unchecked `<T>` generic in this change. Adding optional Effect `Schema` decoding for event payloads is a good follow-up and can land separately.

sam-goodwin:
Merged latest `main` into this branch (the file moved to `Cloudflare/Workflows/` and picked up the async-Worker reference form, `RuntimeContext`, and the `Effect.fn` migration — all preserved alongside the expanded step API).

While verifying live against Cloudflare, two fixes on top of the merge (718b04b24):

- `step.waitForEvent` resolves with the full event object (`{ payload, timestamp, type }`), so the bridge now unwraps it — `waitForEvent<T>` returns the typed payload as the docs show:

```ts
const external = yield* Cloudflare.Workflows.waitForEvent<{ message: string }>(
  "external-event",
  { type: "test-event", timeout: "5 minutes" },
);
external.message; // was undefined before the unwrap
```

- Cloudflare reports an instance parked in `waitForEvent` as `running` (`waiting` is only used for sleeps — confirmed via the instances API, which shows the unfinished `waitForEvent` step while the instance status is `running`). The test now polls for `running` and re-sends the event until the workflow acknowledges it, instead of asserting `waiting`.

Both tests pass live: `ALCHEMY_PROFILE=testing bun vitest run test/Cloudflare/Workers/Workflow.test.ts` → 2 passed, 1 skipped (the pre-existing gated `list` test).

sam-goodwin:
Correction to the above: rather than unwrapping the payload, `waitForEvent` now stays 1:1 with the native binding and resolves with the full `WorkflowStepEvent` (378e27c9f):

```ts
const event = yield* Cloudflare.Workflows.waitForEvent<{ message: string }>(
  "external-event",
  { type: "test-event", timeout: "5 minutes" },
);
event.payload.message; // { payload, timestamp, type } — same shape as step.waitForEvent
```

`sleepUntil` also passes its `Date | number` through unchanged now instead of converting to an ISO string. Live suite re-verified green.

PR #614, opened 2026-06-15
feat(aws/lambda): support function URL config

mineiro:
Allow Lambda function URLs to configure:

```ts
  url: {
    authType: "AWS_IAM",
    cors: {
      AllowOrigins: ["https://example.com"],
    },
    invokeMode: "RESPONSE_STREAM",
  }
  ```

`url: true` remains the public default. When `authType` is `AWS_IAM`, Alchemy removes the public Function URL permission statements.

### DIFF
--- a/packages/alchemy/src/AWS/Lambda/Function.ts
+++ b/packages/alchemy/src/AWS/Lambda/Function.ts
@@ -17,7 +17,7 @@ import type * as rolldown from "rolldown";
 import { Unowned } from "../../AdoptPolicy.ts";
 import * as Bundle from "../../Bundle/Bundle.ts";
 import * as TempRoot from "../../Bundle/TempRoot.ts";
-import { isResolved } from "../../Diff.ts";
+import { deepEqual, isResolved } from "../../Diff.ts";
 import type { HttpEffect } from "../../Http.ts";
 import * as Output from "../../Output.ts";
 import { createPhysicalName } from "../../PhysicalName.ts";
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: packages/alchemy/src/AWS/Lambda/Function.ts [hunk 1]
```diff
@@ -1025,78 +1080,110 @@ export default await Effect.runPromise(handlerEffect)
...
+          }).pipe(
+            Effect.catchTag("ResourceNotFoundException", () => Effect.void),
+          ),
+          Lambda.removePermission({
+            FunctionName: functionName,
+            StatementId: publicUrlInvokeStatementId,
+          }).pipe(
+            Effect.catchTag("ResourceNotFoundException", () => Effect.void),
+          ),
+        ]);
```
sam-goodwin: Can these run in parallel?

packages/alchemy/src/AWS/Lambda/Function.ts
```diff
@@ -1025,78 +1080,110 @@ export default await Effect.runPromise(handlerEffect)
...
+            Action: "lambda:InvokeFunctionUrl",
+            Principal: "*",
+            FunctionUrlAuthType: "NONE",
+          }),
+          upsertPermission({
+            FunctionName: functionName,
+            StatementId: publicUrlInvokeStatementId,
+            Action: "lambda:InvokeFunction",
+            Principal: "*",
+            InvokedViaFunctionUrl: true,
```
sam-goodwin: Can these run in parallel?

packages/alchemy/src/AWS/Lambda/Function.ts
mineiro: Good call! Updated this to run with Effect.all(..., { concurrency: "unbounded" }).

packages/alchemy/src/AWS/Lambda/Function.ts
mineiro: Same here, updated this batch to run in parallel as well.

Issue #621, opened 2026-06-16
Cloudflare.Vite roadmap: Alchemy-owned Vite deploys and dev bindings


agcty:
## Short version

We started this work because RSC Vite apps could not be deployed cleanly through Alchemy's `Cloudflare.Vite` resource.

The architectural end state is:

> make `Cloudflare.Vite` follow the official Cloudflare Vite plugin model, but with Alchemy as the deploy/resource engine instead of Wrangler.

That means:

- Vite/framework owns the application entrypoint, framework plugins, RSC/SSR environment topology, dev server, HMR, and emitted Worker module graph.
- The distilled Cloudflare Vite plugin adapts the Vite output into an explicit Cloudflare deploy/runtime contract.
- Alchemy owns resources, bindings, compatibility settings, assets routing, Durable Object metadata/migrations, deploy diffs, local binding orchestration, and uploads.

The important boundary is that Alchemy should not wrap, merge, or replace the Vite Worker entrypoint. The Vite-built Worker remains the Worker. Alchemy attaches Cloudflare resources and deploys/runs that Worker.

## Why this started

The original problem was RSC support.

React Router RSC / `@vitejs/plugin-rsc` apps are not shaped like a simple single-environment SSR build. They usually have an entry environment, for example `rsc`, plus child environments such as `ssr`, and the entry Worker loads child-environment output at runtime.

Alchemy's old `Cloudflare.Vite` path did not have a reliable deploy contract for that shape. It could capture simple Vite output, but it could not robustly answer:

- which file is the actual Worker entry,
- which files belong to the deployable Worker module set,
- which child-environment files must be uploaded together,
- which compatibility flags/assets routing the build used,
- and how to attach Alchemy-owned bindings to that exact Vite-built Worker.

So the initial RSC issue turned into a clearer architecture question: how should Alchemy consume Vite-built Workers at all?

## What became clear

The official Cloudflare Vite plugin already has the right broad model:

```text
Vite/framework builds the app Worker
Cloudflare plugin records the deploy/runtime contract
Wrangler consumes that contract and deploys/configures the Worker
```

For Alchemy, the equivalent model should be:

```text
Vite/framework builds the app Worker
Distilled Cloudflare Vite plugin records the deploy/runtime contract
Alchemy consumes that contract and deploys/configures the Worker/resources
```

[...]

PR #615, opened 2026-06-15
feat(cloudflare): consume distilled Vite manifests

agcty:
## Status / stack

Draft until the distilled Cloudflare Vite plugin side lands or publishes an equivalent manifest contract.

Dependency chain:

- alchemy-run/cloudflare-tools#47 - RSC dev support via `viteEnvironment` child environments.
- alchemy-run/cloudflare-tools#50 - RSC production-build manifest (`__distilled-build.json`, v2 shape).
- This PR - Alchemy consumer for that manifest in `Cloudflare.Vite`.

This PR is intentionally scoped to consuming the distilled manifest and preserving Alchemy's existing resource/binding ownership model. It does not add remote Worker HMR; `alchemy dev` remains local Vite + local Cloudflare runtime, with supported bindings local or remote-backed depending on the binding type.

Broader roadmap context is tracked in #621: Vite/framework owns the app entrypoint and Worker module graph, while Alchemy owns resources, bindings, deploy diffs, and uploads. That follow-up roadmap covers the polished `alchemy dev` binding loop; this PR stays on the deploy-side manifest consumer.

## Summary

- Read distilled `__distilled-build.json` after Vite builds and upload the manifest-selected Worker module set with explicit module content types.
- Pass compatibility, asset routing, and advanced `viteEnvironment` topology into the distilled Vite build/dev plugin so RSC/custom Vite environment graphs use Alchemy's injected plugin consistently.
- Fail closed when a Vite Worker/server output does not emit the distilled manifest; pure assets-only builds can still proceed without one.
- Include Vite build-affecting deploy inputs in the stored `hash.input` so VITE env, asset routing, compatibility, and topology-only changes do not get skipped.
- Hash upload content types for manifest, Rolldown, script, and prebuilt Worker module bundles so deploy diffs reflect upload shape.
- Document the Vite-owned Worker + local Durable Object/RSC patterns and add one-worker Vite + Durable Object plus SPA manifest-spoof fixture coverage.
- Add an Alchemy-owned React Router RSC fixture: the app config uses only `@vitejs/plugin-rsc`; `Cloudflare.Vite` injects the distilled Cloudflare Vite plugin and deploys the resulting manifest module set.

## Reviewer-round fixes

- Removed the misleading internal `requireBuildManifest` flag; any Vite server/Worker output now requires the distilled manifest.
- Forwarded `viteEnvironment` through `alchemy dev`, not just deploy.
[...]

PR #622, opened 2026-06-16
feat(cloudflare): support Vite dev bindings

agcty:
## Status / stack

Draft follow-up to #615. This PR is intentionally stacked on top of the manifest-consumer work:

1. alchemy-run/cloudflare-tools#47 - RSC dev support via `viteEnvironment` child environments.
2. alchemy-run/cloudflare-tools#50 - distilled Vite build manifest.
3. #615 - Alchemy consumes the distilled Vite manifest for deploys.
4. This PR - hardens `alchemy dev` for Vite Workers with Alchemy-managed bindings.

Because the stack crosses forks, GitHub cannot cleanly target this PR at the fork branch from #615 while still opening it against `alchemy-run/alchemy-effect`. Until #615 lands, this PR's diff will include that base commit. After #615 merges, the diff should collapse to this follow-up commit.

Refs #621.

## Summary

- Recompute local Worker instance identity from the resolved Worker config, not unresolved props, so `alchemy dev` restarts/rebinds a Vite Worker when Alchemy-managed binding identity or env values change.
- Avoid duplicating native Cloudflare bindings as local text/json env values when `env` contains a bindable resource.
- Normalize local Worker URLs returned from the dev provider so path joins do not accidentally produce proxy-breaking `//` URLs.
- Log local binding mode per Worker start/update, distinguishing local runtime bindings from remote-backed Cloudflare bindings and unsupported local binding types.
- Document `Cloudflare.Vite` dev semantics: Vite owns the dev server/HMR loop; Alchemy owns resources and binding configuration.
- Add a TanStack Start fixture plus a live dev test that exercises Vite HMR, R2 remote binding access, and Alchemy rebinding on a second apply.

## Architecture

This keeps the same line as #615:

> Vite owns the app Worker. Alchemy owns resources, bindings, diffs, and deploy/dev orchestration.

In dev mode that means:

- Vite serves the framework app and owns HMR.
- Alchemy starts a local Worker/proxy around that Vite server.
- Alchemy resolves `env` into the runtime binding hooks consumed by the distilled Cloudflare Vite plugin.
- Local-capable bindings stay local; remote-capable account resources such as R2/KV/D1 use remote-backed bindings where the runtime supports them.
- When `alchemy.run.ts` changes the binding set, Alchemy re-applies the resource graph and restarts/rebinds the local Vite Worker behind the same local URL.

## Verification

- `bun install --frozen-lockfile`
- `bun tsgo -b packages/alchemy/tsconfig.test.json`
- `bunx oxfmt --check ...`
- `git diff --check`
- `bun run --filter alchemy test -- test/Cloudflare/Website/Vite.test.ts -t "Vite: ignores manifest-like files copied into client assets"`
- Live Cloudflare test: `bun run --filter alchemy test -- test/Cloudflare/Website/Vite.test.ts -t "Vite dev: TanStack Start keeps Alchemy-managed R2 bindings"`
  - renders the TanStack route through local Vite dev
  - edits a route file and observes Vite HMR without another Alchemy apply
  - writes/reads through an Alchemy-managed R2 bucket binding
  - re-applies with a different bucket and marker
  - verifies the same local URL now sees the new binding and not the old bucket data
  - destroys the temporary buckets/Worker
- Direct TanStack fixture build smoke through `Vite.viteBuild(...)`, producing client assets and an SSR bundle.
- Manual browser smoke at `http://localhost:1337`: verified route render, R2 route write/read, HMR route edit, then restored the file and destroyed the manual Cloudflare stage.
- `bun generate:api-reference`

## Notes

This does not add full remote Worker execution with Vite HMR. It hardens the local Vite HMR path with Alchemy-managed bindings, which matches the current official Cloudflare Vite plugin development model more closely.

## Reviewer-round follow-ups

- Fixed local dev instance hashing to use stable, serializable hash material instead of runtime binding hook closures. This avoids no-op `alchemy dev` applies restarting unchanged Workers while still detecting binding/env/Vite/crons changes.
- Included Vite topology/root and cron inputs in the local dev instance hash.
- Made live-test cleanup strict on success, so a passing run cannot silently leak R2 buckets; cleanup remains best-effort only after an assertion failure so the original failure is preserved.
- Kept `viteEnvironment.name` optional so Alchemy does not narrow the distilled plugin passthrough shape.
- Clarified the local-dev binding docs table as common examples rather than an exhaustive implementation matrix.
- Noted: the `bun.lock` distilled workspace version normalization reflects the already-checked-out `distilled/packages/*` versions (`0.25.1`), not a separate submodule pointer change in this PR.

## Final lifecycle review follow-ups

- QueueConsumer changes now invalidate/restart the target local Worker through LocalRuntime reload callbacks, with pending reloads covering startup and restart windows.
- Worker restarts preserve the browser-facing proxy but clear the backend pointer while the new Worker starts, avoiding stale proxy backends.
- `dev: false` and external-dev-string Workers now diff from persisted props/bindings, not an in-memory stub hash, so no-op applies remain stable across provider restarts.
- The TanStack dev-binding fixture intentionally does not set `build.rolldownOptions.external = ["cloudflare:workers"]`; that default belongs to the Cloudflare Vite integration path, not each app fixture. The fixture also does not advertise standalone `vite build`, because Alchemy-managed dev/build is the path that injects the integration.

## Separate stack note

The local Durable Object build-manifest test still depends on the lower Cloudflare Vite manifest work emitting `__distilled-build.json` for that fixture. That remains part of the #615 / cloudflare-tools #50 stack rather than this dev-bindings follow-up.


agcty:
Final review update after the extra lifecycle pass:

- QueueConsumer changes now invalidate the target local Worker directly through LocalRuntime reload callbacks, instead of relying on Worker diff ordering.
- Reloads that happen while a Worker is still starting or between stop/start are recorded as pending reloads and consumed after the Worker registers.
- Worker restarts preserve the browser-facing proxy but clear the backend pointer while the new Worker starts, so the proxy cannot keep pointing at a closed backend.
- `dev: false` and external-dev-string Workers now diff from persisted props/bindings, not an in-memory hash, so no-op applies stay stable across provider restarts.
- The TanStack dev-binding fixture no longer advertises standalone `vite build`; Alchemy-managed dev/build is the path that injects the Cloudflare Vite integration. App-level `build.rolldownOptions.external = ["cloudflare:workers"]` should not be required.

Validation rerun:

- `bunx oxfmt --check ...`
- `bun tsgo -b packages/alchemy/tsconfig.test.json`
- `git diff --check`
- `bun vitest run test/Cloudflare/Website/Vite.test.ts -t "Vite: ignores manifest-like files copied into client assets"`
- live Cloudflare-backed TanStack dev-binding test with R2 binding

One separate stack note: the local Durable Object build-manifest test still depends on the lower Cloudflare Vite manifest work emitting `__distilled-build.json` for that fixture. I did not fold that into this dev-bindings PR.

sam-goodwin:
Replaced by https://github.com/alchemy-run/alchemy-effect/pull/685

PR #627, opened 2026-06-17
feat(aws/lambda): add event invoke config

mineiro:
Add Lambda asynchronous invocation settings (retries, event age, success/failure destinations) as an `eventInvokeConfig` prop on `Function` and `Alias`.

```ts
const fn = yield* AWS.Lambda.Function("AsyncFn", {
  main: "./src/handler.ts",
  eventInvokeConfig: {
    maximumRetryAttempts: 0,
    maximumEventAgeInSeconds: 60,
    destinationConfig: {
      OnFailure: {
        Destination: queue.queueArn,
      },
    },
  },
});
```

Alias-scoped config uses the same shape on `AliasProps`:

```ts
const alias = yield* AWS.Lambda.Alias("LiveAlias", {
  functionName: fn.functionName,
  functionVersion: version,
  aliasName: "live",
  eventInvokeConfig: { maximumRetryAttempts: 0 },
});
```

Both providers converge through a shared `syncEventInvokeConfig` helper: observe the live config, diff, then put — or delete it when the prop is omitted. Alias delete clears its qualifier-scoped config since Lambda does not remove it with the alias.


sam-goodwin:
Any reason not to make this a property on `AWS.Lambda.FunctionProps`? Alchemy supports circular dependencies, so I think there's less reason to model this as a separate resource.

mineiro:
Makes sense for the function-level case. One catch I noticed is that Lambda async config can also be scoped to a qualifier, so aliases/versions can have their own config.

Would you prefer this PR to only cover `FunctionProps.eventInvokeConfig`, and leave alias-specific config for a follow-up on `AWS.Lambda.Alias`?

sam-goodwin:
> Would you prefer this PR to only cover FunctionProps.eventInvokeConfig, and leave alias-specific config for a follow-up on AWS.Lambda.Alias?

We can do it as part of this change since they're related.

PR #629, opened 2026-06-17
docs(github): add Repository examples for templates, rename, archive, secrets

just-be-dev:
I don't feel strongly about adding this, but thought a few extra examples could be helpful. Looks like y'all haven't deployed the docs yet though so this doesn't show up on the live site anyway. Feel free to close if you think it's not useful or to edit it as you see fit. 

---

Add four new JSDoc examples to the GitHub `Repository` resource, documenting behavior the provider already implements but didn't previously show. Edits are on the source `.ts` (the generated provider markdown is gitignored and overwritten by `bun generate:api-reference`).

- **Initialize from Templates** — `autoInit` / `gitignoreTemplate` / `licenseTemplate` (create-time only)
- **Renaming a Repository** — same logical ID + changed `name` renames in place; only `owner` triggers replacement
- **Archiving a Repository** — `archived: true` for read-only, reversible
- **Store a Secret in the Repository** — `GitHub.Secret` wiring with `Redacted.make`

```typescript
// Rename in place: same logical ID ("api"), new name → renames, preserving history
const repo = yield* GitHub.Repository("api", {
  owner: "my-org",
  name: "gateway",
});
```

### DIFF
--- a/packages/alchemy/src/GitHub/Repository.ts
+++ b/packages/alchemy/src/GitHub/Repository.ts
@@ -235,6 +235,20 @@ export interface Repository extends Resource<
  * });
  * ```
  *
+ * @example Initialize from Templates
+ * The `autoInit`, `gitignoreTemplate`, and `licenseTemplate` props seed the
+ * first commit. They are only honored at create time — changing them on a
+ * later deploy has no effect on an existing repository.
+ * ```typescript
+ * const repo = yield* GitHub.Repository("service", {
+ *   owner: "my-org",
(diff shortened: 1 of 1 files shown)


just-be-dev:
@sam-goodwin I originally kicked this off as a task to update the docs because I saw the repository was missing. I guess y'all just have to deploy the site for that though? 

These examples seem reasonable (except maybe the secrets one which is a bit duplicative, can remove that if you want. You can also remove it without waiting for me).

PR #630, opened 2026-06-17
feat(aws/logs): support log group config

mineiro:
Add CloudWatch Logs configuration support to `AWS.Logs.LogGroup`.

This adds:

- `logGroupClass`
- `deletionProtectionEnabled`
- KMS key association/disassociation during reconcile
- retries for transient log group mutation conflicts

`logGroupClass` is treated as immutable and replaces the log group when changed. Deletion disables deletion protection
before removing the log group.

This should make follow-up CloudWatch Logs resources easier to add.

Tests:

- `bunx oxfmt packages/alchemy/src/AWS/Logs/LogGroup.ts packages/alchemy/src/AWS/Logs/index.ts packages/alchemy/test/AWS/
Logs/LogGroup.test.ts --check`
- `bun --check packages/alchemy/src/AWS/Logs/LogGroup.ts packages/alchemy/src/AWS/Logs/index.ts packages/alchemy/test/AWS/
Logs/LogGroup.test.ts`
- `cd packages/alchemy && bun vitest run test/AWS/Logs/LogGroup.test.ts --reporter=verbose`
- Manual smoke: verified class, retention, deletion protection, replacement, and KMS association/disassociation


packages/alchemy/src/AWS/Logs/LogGroup.ts
```diff
@@ -61,6 +75,14 @@ export interface LogGroup extends Resource<
  */
 export const LogGroup = Resource<LogGroup>("AWS.Logs.LogGroup");
 
+const retryLogGroupMutation = Effect.retry({
+  while: (error: any) =>
+    error._tag === "OperationAbortedException" ||
+    error._tag === "ServiceUnavailableException",
+  schedule: Schedule.exponential(100),
+  times: 8,
+});
```
sam-goodwin: Prefer to do this inline or at least make `error` not be `any`. This can disrupt type-safe effect pipelines.

mineiro: Thanks, fixed. I kept the helper since it is reused in a few places, but typed the retry predicate to the specific CloudWatch Logs errors instead of `any`.

PR #631, opened 2026-06-17
feat(aws): expand resource surface area + add VpcOrigin, ListenerRule, TrustStore, HostedZone, HealthCheck

sam-goodwin:
Many AWS resources were narrow / opinionated and couldn't express the full config AWS supports. This expands them to the full surface area and adds the missing lower-level resources. All non-gated tests pass and `tsgo -b` is clean.

Depends on alchemy-run/distilled#351 (Route53 list-response schema fix); the submodule pointer is bumped to that commit.

### New resources

**`CloudFront.VpcOrigin`** — front a private ALB/NLB/EC2 without exposing it publicly.

```ts
const origin = yield* CloudFront.VpcOrigin("api-origin", {
  arn: loadBalancer.arn,
  httpPort: 80,
  originProtocolPolicy: "https-only",
});
```

**`ELBv2.ListenerRule`** — ALB path/host/header/query/method/source-IP routing.

```ts
yield* ELBv2.ListenerRule("api-route", {
  listenerArn: listener.listenerArn,
  priority: 10,
  conditions: [{ pathPattern: { values: ["/api/*"] } }],
  actions: [{ type: "forward", targetGroups: [{ targetGroupArn: tg.targetGroupArn }] }],
});
```

**`ELBv2.TrustStore`** — CA bundle for mTLS verify mode.

```ts
yield* ELBv2.TrustStore("client-ca", {
  caCertificatesBundleS3Bucket: "my-ca-bundles",
  caCertificatesBundleS3Key: "ca-bundle.pem",
});
```

**`Route53.HostedZone`** + **`Route53.HealthCheck`** — Records no longer need a hand-supplied zone id, and failover routing gets a health-check source.

```ts
const zone = yield* Route53.HostedZone("example.com", { name: "example.com" });

const health = yield* Route53.HealthCheck("primary", {
  type: "HTTPS",
  fqdn: "primary.example.com",
  resourcePath: "/health",
  failureThreshold: 3,
});
```

### Expanded resources

**`S3.Bucket`** — versioning, encryption, public-access-block, CORS, lifecycle, ownership/ACL, logging, acceleration, request-payer, website, replication, intelligent-tiering, object-lock, explicit policy.

```ts
yield* Bucket("assets", {
  versioning: "Enabled",
  encryption: { sseAlgorithm: "aws:kms", bucketKeyEnabled: true },
  publicAccessBlock: { blockPublicAcls: true, blockPublicPolicy: true },
  cors: [{ allowedMethods: ["GET"], allowedOrigins: ["https://example.com"] }],
  lifecycleRules: [{ id: "expire-old", status: "Enabled", expiration: { days: 90 } }],
});
```

**`SQS.Queue`** — redrive/DLQ, redrive-allow, KMS/SSE, policy, full FIFO surface.

```ts
yield* SQS.Queue("work", {
  redrivePolicy: { deadLetterTargetArn: dlq.queueArn, maxReceiveCount: 5 },
  kmsMasterKeyId: "alias/aws/sqs",
});
```

**`Route53.Record`** — weighted / latency / failover / geolocation / multivalue / alias routing.

```ts
yield* Route53.Record("blue", {
  hostedZoneId: zone.id,
  name: "app.example.com",
  type: "A",
  setIdentifier: "blue",
  weight: 90,
  aliasTarget: { dnsName: blueLb.dnsName, hostedZoneId: blueLb.canonicalHostedZoneId, evaluateTargetHealth: true },
});
```

**`ELBv2.Listener`** — full action union (redirect, fixed-response, weighted forward, OIDC/Cognito auth), mTLS, ALPN, SSL policy, SNI certs. `LoadBalancer` + `TargetGroup` gaps filled.

```ts
yield* ELBv2.Listener("https", {
  // ...
  defaultActions: [{ type: "redirect", redirectConfig: { protocol: "HTTPS", port: "443", statusCode: "HTTP_301" } }],
  certificates: [primaryCertArn, sniCertArn],
  sslPolicy: "ELBSecurityPolicy-TLS13-1-2-2021-06",
  mutualAuthentication: { mode: "verify", trustStoreArn: trustStore.trustStoreArn },
});
```

**`ECS.Task` / `ECS.Service`** — multi-container task definitions (sidecars + task-level config); de-opinionated Service: ALB is optional, plus capacity-provider strategy, deployment circuit breaker, placement, exec, with corrected replace-vs-in-place diff.

```ts
const task = yield* ECS.Task("web", {
  main: "nginx:latest",
  sidecars: [{ name: "log-router", image: "fluent/fluent-bit:latest", essential: false }],
  networkMode: "awsvpc",
  volumes: [{ name: "tmp" }],
});

yield* ECS.Service("web", {
  // ...
  desiredCount: 3,
  capacityProviderStrategy: [{ capacityProvider: "FARGATE_SPOT", weight: 1 }],
  loadBalancers: [/* optional — no longer force-created */],
});
```

**`RDS.DBInstance` / `RDS.DBCluster`** (passed through `Aurora`) — storage/IOPS, engine version, multi-AZ, performance insights, enhanced monitoring, backup/maintenance windows, parameter/option groups, encryption/KMS, IAM auth, log exports, serverless-v2 scaling. Diff replacement-sets corrected so in-place-updatable knobs no longer force replacement.

```ts
yield* RDS.DBInstance("db", {
  // ...
  allocatedStorage: 100,
  storageType: "gp3",
  multiAZ: true,
  performanceInsightsEnabled: true,
  deletionProtection: true,
  enabledCloudwatchLogsExports: ["postgresql"],
});
```

### Testing

Non-gated live tests pass for every resource and `tsgo -b` is clean. Lifecycles that need slow provisioning or external prerequisites are implemented and `skipIf`-gated behind env vars (`RDS_TEST_LIFECYCLE`, `CLOUDFRONT_TEST_VPC_ORIGIN`, `ELBV2_TEST_MTLS`, `S3_TEST_REPLICATION`); making those green is in progress, which is why this is a draft.


packages/alchemy/src/AWS/ECS/Task.ts
```diff
@@ -104,11 +109,74 @@ export interface TaskProps extends PlatformProps {
...
+   * Additional sidecar containers appended to the task definition after the
+   * primary bundled container. Each entry is a full, typed
+   * {@link ecs.ContainerDefinition} (image URIs supplied by the user, e.g.
+   * from an `ECR.Image` or an external registry).
+   *
+   * Use this to declare multi-container tasks: log routers (firelens),
+   * proxies (Envoy/App Mesh), metric agents (otel/cloudwatch), or any
+   * companion process that shares the task's network namespace.
+   */
+  sidecars?: ecs.ContainerDefinition[];
```
bbarks15, line 126: nice work on this. one thing on the Task container shape, feel free to ignore if you've already thought about it.

the main/container/sidecars setup feels a bit different from how task defs normally work everywhere else (aws api, terraform, cfn, cdk all just take a flat list of containers). ecs doesn't really have a "primary" container either, it's just the essential flag plus the service picking one by name, so the primary vs sidecar thing is kind of us inventing it.

wonder if it'd be cleaner to keep it as a flat array and have the bundling just be an image source, something like:

```ts
containers: [
    { name: "app", image: Bundle({ main: import.meta.filename }), essential: true },
    { name: "otel", image: "public.ecr.aws/.../otel:latest" },
]
```
then a prebuilt image task is just image: "..." with no main needed. There might be better DX then this just what came to mind. But anyway just a thought, not a blocker

PR #633, opened 2026-06-17
example(cloudflare-worker-auth): use the real @alchemy.run/pr-package library

Mkassabov:
Convert the `cloudflare-worker-auth` example from a hand-rolled worker to the real `@alchemy.run/pr-package` library, so the integration test exercises the actual library code path (the worker bearer-token check + the `AuthTokenValue` stack output) rather than a copy of it.

```diff
 // src/Api.ts
-export default class Api extends Cloudflare.Worker<Api>()("Api", { ... },
-  Effect.gen(function* () {
-    const authToken = yield* Cloudflare.Secret.bind(AuthToken);
-    return { fetch: /* hand-rolled bearer check */ };
-  }).pipe(Effect.provide(Cloudflare.SecretBindingLive)),
-) {}
+export default class Api extends Cloudflare.Worker<Api>()(
+  "Api",
+  { main: import.meta.filename, url: true },
+  PrPackage.handler(),
+) {}
```

The integ test drives the real bearer-protected routes:

```ts
// PUT /projects/:pkg/packages — auth runs before any body/binding work
valid token        -> 200 { resourceId }
invalid / missing  -> 401
"<redacted>"       -> 401   // the literal both old bugs produced
```

### Distilled pin

Bumps the `distilled` submodule to include the Cloudflare error-class tree-shaking fix ([distilled#352](https://github.com/alchemy-run/distilled/pull/352)). Without it the worker bundle keeps a dangling `import "@distilled.cloud/cloudflare/flagship"` (pulled in transitively by the `alchemy/Cloudflare` barrel) and Cloudflare rejects the upload with `ScriptModuleNotFound`.

The pin points at the merged commit on distilled `main`, which is one commit ahead of the `v0.25.2` tag — it is not yet a published release. Re-pin to the tagged release once distilled cuts `v0.25.3`, before the next `alchemy` npm publish.


Mkassabov:
Superseded by #638 — renamed the example from `cloudflare-worker-auth` to `cloudflare-pr-package` (it demonstrates `@alchemy.run/pr-package` specifically) and added upload/download + garbage-collection tests.

PR #644, opened 2026-06-18
fix(cloudflare/workers): support workflow env bindings

jonbeckman:
Allow Worker `env` to accept Cloudflare Workflow resources and named Worker entrypoint service bindings directly.

```ts
env: {
  WORKFLOW: workflow,
  WORKFLOW_ACTIONS: {
    BindingType: "Cloudflare.WorkerEntrypointBinding",
    service: worker.workerName,
    entrypoint: "WorkflowActions",
  },
}
```

- Convert `WorkflowResource` values into `workflow` bindings.
- Convert `WorkerEntrypointBinding` values into `service` bindings with `entrypoint`.
- Map `InferEnv` to runtime `Workflow` and RPC service-binding types.

### DIFF
--- a/packages/alchemy/src/Cloudflare/Workers/InferEnv.ts
+++ b/packages/alchemy/src/Cloudflare/Workers/InferEnv.ts
@@ -24,49 +24,58 @@ export type GetBindingType<T> =
       ? Flagship
       : T extends Cloudflare.Assets
         ? Service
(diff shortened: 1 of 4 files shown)


sam-goodwin:
Fixed in https://github.com/alchemy-run/alchemy-effect/pull/707

PR #652, opened 2026-06-19
fix(github): resolve lazy secret values

harrysolovay:
Resolve lazy inputs passed through `GitHub.Secrets` before wrapping them as `Redacted`.

```ts
secrets: {
  API_TOKEN: Config.redacted("API_TOKEN"),
}
```

Previously the bulk helper eagerly called `Redacted.make(input)`, producing an opaque `Redacted<Config | Effect | Output>` that the plan resolver could not evaluate. The helper now maps through lazy inputs and wraps the resolved value instead.

Adds a regression test proving a `Config` secret becomes a real `Redacted` value in the plan.


packages/alchemy/test/GitHub/Secret.test.ts
```diff
@@ -18,6 +24,36 @@ const logLevel = Effect.provideService(
 const owner = process.env.GITHUB_TEST_OWNER ?? "";
 const repository = process.env.GITHUB_TEST_REPOSITORY ?? "";
 
+test(
+  "Secrets resolves Config values before wrapping them as Redacted",
+  Effect.gen(function* () {
+    const plan = yield* Effect.gen(function* () {
+      yield* GitHub.Secrets({
+        owner: "alchemy-run",
+        repository: "alchemy-effect",
```
sam-goodwin: Use the `owner` and `repository`. Don't hard-code to `alchemy-run/alchemy-effect`

PR #654, opened 2026-06-19
fix(neon)!: replace branch when project changes

harrysolovay:
Replace Neon branches when the parent project ID changes.

This involves a workaround to correctly distinguish a changed project ID from a project that has been updated in-place without the project ID changing. Specifically, we use a `Proxy` to detect when `props.project` is a `Neon.Project` resource and map it into `{ projectId: props.project.projectId }` so that the value is stable across project updates while still accurately reflecting when the project ID has actually changed.

**Breaking Change:** The type of `NeonBranchProps["project"]` was `Project | { projectId: string } | { project: Project }`. To make the aforementioned transformation easier, the type is now `Project | { projectId: string }`.


packages/alchemy/test/Neon/Branch.test.ts
```diff
@@ -12,6 +13,39 @@ const logLevel = Effect.provideService(
...
+        protected: false,
+        expiresAt: undefined,
+        migrationsHashes: {},
+        importHashes: {},
+      },
+    );
+    expect(result).toEqual({ action: "replace" });
+  }),
+);
+
```
sam-goodwin: We don't test units this. Please write and end to end test with test.provider and multiple test.deploy and test.destroy steps to actually exercise this and ensure it works. It's not enough to just have diff return replace at the right time. We actually have to verify replace works.

john-royal (approved):
Approved contingent on #670 being merged first.

john-royal:
Woops, didn't mean to close this. Oh GitHub, why didn't you change the base branch to main like you normally do?

PR #675, opened 2026-06-23
fix(aws): Route53 change ID normalization

jperezrealini:
## Summary

- normalize Route53 change IDs before polling `GetChange`
- strip the `/change/` resource prefix returned by Route53 APIs

## Why

Route53 mutation responses can return change IDs in resource-path form, while `GetChange` expects the bare identifier. This mirrors the normalization discussed in [aws/aws-sdk-go-v2#1838](https://github.com/aws/aws-sdk-go-v2/issues/1838) and implemented in [aws/aws-sdk-go-v2#1852](https://github.com/aws/aws-sdk-go-v2/pull/1852).

## Impact

ACM DNS validation can reliably poll Route53 changes without sending a prefixed ID.

## Validation

- `bunx oxfmt --check packages/alchemy/src/AWS/ACM/Certificate.ts`
- `git diff --check`

### DIFF
--- a/packages/alchemy/src/AWS/ACM/Certificate.ts
+++ b/packages/alchemy/src/AWS/ACM/Certificate.ts
@@ -255,22 +255,26 @@ export const CertificateProvider = () =>
       });
 
       const waitForChange = Effect.fn(function* (changeId: string) {
-        return yield* route53.getChange({ Id: changeId }).pipe(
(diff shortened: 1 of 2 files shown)


sam-goodwin (commented):
Please add a test

Use Test.make and test.provider (an established pattern) to exercise the resource to test this works.

PR #677, opened 2026-06-23
feat(aws): support external packages in Lambda bundles

jperezrealini:
## Summary

- add `build.externalPackages` to Lambda functions
- exclude selected dependencies from Rolldown and materialize them under `node_modules` in the deployment package
- install dependencies from Bun or npm lockfiles inside Docker for the Lambda runtime architecture
- preserve executable permissions for packaged binaries
- cache materialized packages and include their contents, runtime, and architecture in deployment hashes
- enforce Lambda compressed and uncompressed package-size limits

## Why

Some Lambda dependencies cannot be bundled into a single JavaScript artifact. Native modules such as `sharp`, and packages that ship executables such as `ffmpeg-static`, need their platform-specific files and binaries available alongside the Lambda code.

This change lets users declare those dependencies explicitly while Alchemy builds them for Lambda Linux and packages their complete runtime dependency trees into the function ZIP.

## Example

```ts
const fn = yield* AWS.Lambda.Function("ImageProcessor", {
  main: "./src/handler.ts",
  runtime: "nodejs22.x",
  architecture: "arm64",
  build: {
    externalPackages: ["sharp", "ffmpeg-static"],
  },
});
```

## Validation

- `timeout 240 bun vitest run test/Bundle/ExternalPackages.test.ts test/zip.test.ts` — 17 passed, 2 Docker integration tests skipped behind `ALCHEMY_TEST_LAMBDA_EXTERNAL_PACKAGES=1`
- formatting check across changed TypeScript, JavaScript, and JSON files
- `git diff --check main...HEAD`

### DIFF
--- a/packages/alchemy/src/AWS/Lambda/Function.ts
+++ b/packages/alchemy/src/AWS/Lambda/Function.ts
@@ -13,9 +13,12 @@ import * as Option from "effect/Option";
 import * as Redacted from "effect/Redacted";
 import * as Schedule from "effect/Schedule";
 import * as Stream from "effect/Stream";
+import path from "node:path";
 import type * as rolldown from "rolldown";
 import { Unowned } from "../../AdoptPolicy.ts";
+import { AlchemyContext } from "../../AlchemyContext.ts";
 import * as Bundle from "../../Bundle/Bundle.ts";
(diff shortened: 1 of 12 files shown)


jperezrealini:
replaced by https://github.com/alchemy-run/alchemy-effect/pull/682

PR #680, opened 2026-06-23
chore(website): update to astro 7

john-royal:


john-royal:
This looks like it's actually slower - 7 minutes instead of 5. Shelving for now.

PR #682, opened 2026-06-23
feat(aws/lambda): add build.install for native packages

jperezrealini:
Add `build.install` for Lambda packages that must stay out of the Rolldown bundle but still ship inside the Lambda zip.

```ts
const func = yield* AWS.Lambda.Function("ImageProcessor", {
  main: "./src/handler.ts",
  architecture: "arm64",
  build: {
    install: ["sharp"],
  },
});
```

`build.install` accepts package roots as an array, or a record with explicit versions:

```ts
build: {
  install: {
    sharp: "^0.33.5",
  },
}
```

Packages listed in `build.install` are externalized during bundling, resolved from the source `package.json` / catalog metadata, npm-installed into a clean Linux-targeted artifact, and zipped with the full `node_modules` tree.

`build.external` remains a Rolldown escape hatch only. It does not install packages into the zip, so users can still externalize modules that are provided by Lambda layers or another runtime mechanism.

Change detection stays fast by hashing the source bundle, resolved external install versions, nearest package-manager lockfile, and architecture. The final S3 asset key still uses the true archive-byte hash.

### DIFF
--- a/packages/alchemy/src/AWS/Lambda/Function.ts
+++ b/packages/alchemy/src/AWS/Lambda/Function.ts
@@ -17,13 +17,13 @@ import type * as rolldown from "rolldown";
 import { Unowned } from "../../AdoptPolicy.ts";
 import * as Bundle from "../../Bundle/Bundle.ts";
 import {
-  hashExternalPackageIdentity,
+  hashPackageInstallIdentity,
   installResolvedPackages,
-  type InstallExternalPackages,
   matchesPackageRoot,
-  resolveExternalPackageIdentity,
-  validateInstallExternalTargets,
-} from "../../Bundle/ExternalPackages.ts";
+  type PackageInstall,
+  resolvePackageInstallIdentity,
+  validateInstallTargets,
+} from "../../Bundle/InstalledPackages.ts";
 import * as TempRoot from "../../Bundle/TempRoot.ts";
 import { deepEqual, isResolved } from "../../Diff.ts";
 import type { HttpEffect } from "../../Http.ts";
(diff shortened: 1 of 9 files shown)


packages/alchemy/src/AWS/Lambda/Function.ts
```diff
@@ -61,6 +69,24 @@ export const isFunction = (value: any): value is Function => {
 };
 
 export interface FunctionBuildOptions {
+  /**
+   * Native or Node-only packages to install into the Lambda artifact with npm,
+   * targeting Linux and the function's architecture.
```
sam-goodwin, line 74: Should it always use `npm` or use the user's package manager? Or should we let them configure it and default to npm?

packages/alchemy/src/Bundle/ExternalPackages.ts
```diff
@@ -0,0 +1,617 @@
...
+    `--cpu=${npmArchitecture}`,
+    ...(packageNames.includes("sharp") ? ["--libc=glibc"] : []),
+  ];
+}
+
+/**
+ * Validates a `build.installExternal` declaration and normalizes it to a
+ * package-root → requested-version map. Array entries default to `"*"`.
+ */
+export function validateInstallExternalTargets(
```
sam-goodwin: ~~nit: should we call it `resolveInstallExternalTargets` since it both validates and resolves them?~~

I find this confusing. There's a resolveInstallTargets below. What's the difference between InstallExternal and just Install?

This validate function looks more like a resolver.

packages/alchemy/test/Bundle/InstalledPackages.test.ts
sam-goodwin, line 1: Does this cover all of the cases in ExternalPackages.ts? There's a ton of code in there. I am particularly concerned about all the code for different package mangers

packages/alchemy/src/AWS/Lambda/Function.ts
jperezrealini: we use `npm` here because these packages are installed in a tmp dir using the `--os=linux --cpu=arm64 --libc=glibc` flags to match the lambda arch. other package managers use different flag names or don't support this.

packages/alchemy/src/Bundle/ExternalPackages.ts
jperezrealini: - `normalizeInstallTargets` converts `sharp` to `{ sharp: "*"}`
- `resolvePackageInstallIdentity` determines which version to install by looking at lockfiles

packages/alchemy/test/Bundle/InstalledPackages.test.ts
jperezrealini: just added the missing tests

PR #690, opened 2026-06-25
refactor(bindings)!: remove Binding.Policy and align all bindings to a single convention

sam-goodwin:
A coordinated naming-alignment release: every Cloudflare service is now a proper namespace, every binding across AWS **and** Cloudflare follows one `{Verb}{Resource}` convention, and the two-layer `Binding.Policy` split is gone — folded into a single runtime-guarded `Layer`. Almost no new behavior; a rename across ~900 symbols and ~1200 files so the API reads the same everywhere.

### Breaking changes

- **Every Cloudflare service is namespaced** — `Cloudflare.KVNamespace` → `Cloudflare.KV.Namespace`, `Cloudflare.Worker` → `Cloudflare.Workers.Worker` (common ones re-exported at the root).
- **`AiGateway` / `AiSearch` / `AiSecurity` collapse into `AI`** — `Cloudflare.AiGateway` → `Cloudflare.AI.Gateway`.
- **Bindings are directly callable, namespaced, least-privilege** — `R2Bucket.bind(b)` → `Cloudflare.R2.ReadBucket(b)`, `DynamoDB.GetItem.bind(t)` → `AWS.DynamoDB.GetItem(t)`.
- **`Binding.Policy` is removed** — deploy-time wiring moves behind an `if (!globalThis.__ALCHEMY_RUNTIME__)` guard in one Layer.
- **Impl layers renamed** — `*Live` → `*Binding` (Cloudflare native worker binding) or `*Http` (HTTP-client binding, all AWS + Cloudflare HTTP impls).
- **Event sources are `consume<Resource><Event>(resource, handler)`** — `Queues.messages(q).subscribe(fn)` → `Queues.consumeQueueMessages(q, fn)`.
- **Worker-only bindings take a string id** — `Images({ name: "X" })` → `Images("X")`; native layers drop `*Live`/`*Layer` (`BrowserBindingLive` → `BrowserBinding`).
- **`Artifacts.Store` → `Artifacts.Namespace`** (and `ReadWriteStore` → `ReadWriteNamespace`).

### Every Cloudflare service is a namespace

`Cloudflare/index.ts` was a flat `export *`; it now mirrors the AWS barrel — one `export * as <Service>` per service, with the redundant prefix stripped:

```diff lang="typescript"
- const kv = yield* Cloudflare.KVNamespace("Sessions");
- const bucket = yield* Cloudflare.R2Bucket("Assets");
- export default Cloudflare.Worker("Worker", { /* ... */ });
+ const kv = yield* Cloudflare.KV.Namespace("Sessions");
+ const bucket = yield* Cloudflare.R2.Bucket("Assets");
+ export default Cloudflare.Workers.Worker("Worker", { /* ... */ });
```

Common building blocks stay at the root (`Cloudflare.Worker`, `Container`, `DurableObject`, `Workflow`, `RateLimit`, `cron`). The DO base class loses its suffix (`DurableObjectNamespace` → `DurableObject`). Folder reshuffles: `Container/` → `Containers/`, `Queue` → `Queues`, workflow code → `Workflows/`, plus a new `Cloudflare.WebSocket` type for hibernatable handlers.

### `Ai*` becomes one `AI` namespace

The three AI services collapse into `Cloudflare.AI`, dropping the `Ai` prefix:

```diff lang="typescript"
- const gateway = yield* Cloudflare.AiGateway("Gateway");
+ const gateway = yield* Cloudflare.AI.Gateway("Gateway");
```

`AiGatewayDataset` → `AI.Dataset`, `AiSearchInstance` → `AI.SearchInstance`, `AiSecuritySettings` → `AI.SecuritySettings`.

### One binding convention, everywhere

AWS and Cloudflare bindings now read identically — the capability is the noun, prefixed by the verb (access level), called **directly** (`.bind` is gone):

```diff lang="typescript"
- const get = yield* AWS.DynamoDB.GetItem.bind(table);
- const bucket = yield* Cloudflare.R2Bucket.bind(Bucket);
+ const get = yield* AWS.DynamoDB.GetItem(table);
+ const bucket = yield* Cloudflare.R2.ReadWriteBucket(Bucket);
```

Where the API distinguishes access levels, the capability splits into `Read` / `Write` / `ReadWrite` for least privilege:

```typescript
yield* Cloudflare.KV.ReadNamespace(Sessions);      // get / list
yield* Cloudflare.KV.WriteNamespace(Sessions);     // put / delete
yield* Cloudflare.KV.ReadWriteNamespace(Sessions); // both
```

### `*Binding` vs `*Http`

Each capability ships two interchangeable impl Layers; the suffix is the transport:

```typescript
.pipe(Effect.provide(Cloudflare.R2.ReadWriteBucketBinding)) // native Worker binding (env.BUCKET)
.pipe(Effect.provide(Cloudflare.R2.ReadWriteBucketHttp))    // scoped HTTP token — runs anywhere
```

- **AWS `*Http`** — emits a least-privilege IAM policy statement on the Lambda role, then calls the distilled HTTP API with it. (AWS has no native binding, so every AWS impl was renamed `*Live` → `*Http`.)
- **Cloudflare `*Http`** — mints a scoped `AccountApiToken` with only the needed permission groups, binds its value as a secret, and calls the REST API through distilled.

The client interface is identical regardless of which Layer you provide — a Worker over its native binding and a Container over a token run the same code.

### More Cloudflare `*Http` bindings

R2, KV, and Queues already had `*Http` impls; DNS joins them this release, all sharing one mechanism and the verb-first naming:

```typescript
yield* Cloudflare.DNS.ReadDns(Zone);      // list / get records
yield* Cloudflare.DNS.WriteDns(Zone);     // create / update / delete
yield* Cloudflare.DNS.ReadWriteDns(Zone); // both
```

An `*Http` Layer provisions its own credential: at deploy time it mints an `AccountApiToken` scoped to only the permission groups that access level needs (a `ReadNamespace` mints a read-only KV token, a `WriteNamespace` a write-only one) and binds its value into the host as a secret, behind the `__ALCHEMY_RUNTIME__` guard. At runtime the client reads the token's `value`/`accountId` and calls the REST API — no account-wide key, a different scoped token per capability.

### `Binding.Policy` is gone

Bindings were two pieces: a runtime `Binding.Service` plus a separate deploy-time `Binding.Policy` (with its own `*PolicyLive` layer). They collapse into a single `Binding.Service` whose setup Effect does the deploy-time wiring inline, guarded so it only runs at plan time:

```typescript
if (!globalThis.__ALCHEMY_RUNTIME__) {
  yield* host.bind`${resource}`(/* register IAM / native binding / env */);
}
// always: return the typed runtime client
```

At plan time the guard is open and `bind` records what the function needs; at runtime the branch is skipped, so the runtime bundle stays small. Every binding — including event sources — now follows this. `Binding.Policy`, `Binding.ServiceClass`, and the `Policy` / `PolicyLike` types are removed from the core.

### Worker-only bindings take a string id

The bindings with no backing resource (`Images`, `Browser`, `VersionMetadata`, `RateLimit`) take a string id first instead of a `{ name }` object, and their native layers drop `*Live`/`*Layer`:

```diff lang="typescript"
- const images = yield* Cloudflare.Images.Images({ name: "PIPELINE" });
- .pipe(Effect.provide(Cloudflare.Workers.BrowserBindingLive))
+ const images = yield* Cloudflare.Images.Images("PIPELINE");
+ .pipe(Effect.provide(Cloudflare.Workers.BrowserBinding))
```

`Artifacts.Store` → `Artifacts.Namespace` (access via `ReadNamespace` / `WriteNamespace` / `ReadWriteNamespace`).

### Event sources: `consume<Resource><Event>`

Event sources are bindings too. The old `X(resource, props?).subscribe(handler)` shape becomes a single verb-prefixed, resource-and-event-named callable taking the handler last:

```diff lang="typescript"
- yield* Cloudflare.Queues.messages(inbound).subscribe((records) => …);
+ yield* Cloudflare.Queues.consumeQueueMessages(inbound, (records) => …);
```

| Before | After |
| --- | --- |
| `SQS.messages(q).subscribe` | `SQS.consumeQueueMessages(q, fn)` |
| `S3.notifications(b).subscribe` | `S3.consumeBucketEvents(b, fn)` |
| `SNS.notifications(t).subscribe` | `SNS.consumeTopicNotifications(t, fn)` |
| `Kinesis.records(s).process` | `Kinesis.consumeStreamRecords(s, fn)` |
| `DynamoDB.stream(t).process` | `DynamoDB.consumeTableChanges(t, fn)` |
| `GitHub.events(r).subscribe` | `GitHub.consumeRepositoryEvents(r, fn)` |

Dropping `.subscribe` also drops the handler's runtime requirements at the source boundary, so explicit `Body` type arguments no longer break inference.


packages/alchemy/src/AWS/SQS/DeleteMessageHttp.ts
```diff
@@ -0,0 +1,39 @@
...
+
+export const DeleteMessageHttp = Layer.effect(
+  DeleteMessage,
+  Effect.gen(function* () {
+    const deleteMessage = yield* sqs.deleteMessage;
+
+    return Effect.fn(function* (queue: Queue) {
+      const QueueUrl = yield* queue.queueUrl;
+      if (!globalThis.__ALCHEMY_RUNTIME__) {
+        const host = yield* Binding.host;
```
sam-goodwin: nit: rename this to Binding.Host

packages/alchemy/src/AWS/SQS/DeleteMessageHttp.ts
```diff
@@ -0,0 +1,39 @@
...
+              {
+                Effect: "Allow",
+                Action: ["sqs:DeleteMessage"],
+                Resource: [Output.interpolate`${queue.queueArn}`],
+              },
+            ],
+          });
+        }
+      }
+      return Effect.fn(function* (request: DeleteMessageRequest) {
```
sam-goodwin: I think each of the binding functions should add an identifier

Effect.fn(`AWS.SQS.DeleteMessage(${host.LogicalId}), ${queue.LogicalId}`)(function*(...)

Whatever makes sense for the binding?

examples/cloudflare-git-artifacts/src/Worker.ts
```diff
@@ -37,7 +37,7 @@ export default class Worker extends Cloudflare.Worker<Worker>()(
     },
   },
   Effect.gen(function* () {
-    const artifacts = yield* Cloudflare.Artifacts.bind(Repos);
+    const artifacts = yield* Cloudflare.Artifacts.Artifacts(Repos);
```
sam-goodwin: This should be Cloudflare.Artifacts.ReadStore, Cloudflare.Artifacts.WriteStore, ReadWriteStore, etc.

examples/cloudflare-worker-async/alchemy.run.ts
```diff
@@ -4,15 +4,15 @@ import { Config } from "effect";
...
 
-export const Bucket = Cloudflare.R2Bucket("Bucket");
+export const Bucket = Cloudflare.R2.R2Bucket("Bucket");
 
 // Queue producer + consumer wiring (both sides exercised by the same worker).
 // The Worker sends a message via `env.QUEUE.send(...)` from POST /queue/send,
 // then receives and persists it via its `queue(batch)` handler — end-to-end
 // regression guard for the Queue, QueueWrite, and QueueConsumer resources.
-export const Queue = Cloudflare.Queue("Queue");
+export const Queue = Cloudflare.Queue.Queue("Queue");
```
sam-goodwin: It should be Cloudflare.Queues.Queue, Cloudflare.Queues.Consumer, Cloudflare.Queues.ReadQueue(queue)

examples/cloudflare-worker-async/alchemy.run.ts
```diff
@@ -63,7 +63,7 @@ export default Alchemy.Stack(
 
     // Register the same worker script as a consumer of Queue. The worker's
     // `queue(batch)` handler (see src/worker.ts) receives each message batch.
-    yield* Cloudflare.QueueConsumer("QueueConsumer", {
+    yield* Cloudflare.Queue.QueueConsumer("QueueConsumer", {
```
sam-goodwin: rename to Cloudflare.Queue.Consumer

examples/cloudflare-worker/src/Agent.ts
```diff
@@ -38,14 +38,14 @@ export default class Agent extends Cloudflare.DurableObjectNamespace<Agent>()(
             return yield* response.text;
           }).pipe(Effect.orDie),
         fetch: Effect.gen(function* () {
-          const [response, socket] = yield* Cloudflare.upgrade();
+          const [response, socket] = yield* Cloudflare.Workers.upgrade();
```
sam-goodwin: This should be Cloudflare.WebSocket.upgradeConnection(). There'll be a bunch of places to update docs and examples and tests here

examples/cloudflare-worker/src/Agent.ts
```diff
@@ -74,7 +74,7 @@ export default class Agent extends Cloudflare.DurableObjectNamespace<Agent>()(
     });
   }).pipe(
     Effect.provide(
-      Cloudflare.layerContainer(Sandbox, {
+      Cloudflare.Containers.layerContainer(Sandbox, {
```
sam-goodwin: Cloudflare.Containers.layer(Sandbox, { .. })

examples/cloudflare-worker/src/Api.ts
```diff
@@ -42,15 +42,15 @@ export default class Api extends Cloudflare.Worker<Api>()(
...
+    const queue = yield* Cloudflare.Queue.WriteQueue(queueResource);
+    const repos = yield* Cloudflare.Artifacts.ReadWriteStore(Repos);
+    const aiGateway = yield* Cloudflare.AiGateway.Inference(Gateway);
 
     // Effect-style queue consumer. Each batch is piped through the
     // handler; success ack()s every message in the batch, failure
     // retry()s. The persisted JSON at /queue/<id> on R2 lets the
     // integ test verify the producer→consumer round-trip.
-    yield* Cloudflare.messages<QueueMessageBody>(queueResource).subscribe(
+    yield* Cloudflare.Queue.messages<QueueMessageBody>(queueResource).subscribe(
```
sam-goodwin: Cloudflare.Queues.Messages<QueueMessageBodu>(queueResource)

Also, let's take the opportunity to refactor how event sources work. Instead of .subscribe, pass the function in as the second argument. Fan that change out to al event sources and guides/docs on them

examples/cloudflare-worker/src/Api.ts
```diff
@@ -118,7 +118,7 @@ export default class Api extends Cloudflare.Worker<Api>()(
               ),
             );
           } else if (request.method === "POST" || request.method === "PUT") {
-            // const request = yield* Cloudflare.Request
+            // const request = yield* Cloudflare.Workers.Request
```
sam-goodwin, line 122: Hoist Request to Cloudflare.Request

examples/cloudflare-worker/src/Bucket.ts
```diff
@@ -1,3 +1,3 @@
 import * as Cloudflare from "alchemy/Cloudflare";
 
-export const Bucket = Cloudflare.R2Bucket("Bucket");
+export const Bucket = Cloudflare.R2.R2Bucket("Bucket");
```
sam-goodwin: Rename R2Bucket to Bucket so it's Cloudflare.R2.Bucket(..)

examples/cloudflare-worker/src/KV.ts
```diff
@@ -1,3 +1,3 @@
 import * as Cloudflare from "alchemy/Cloudflare";
 
-export const KV = Cloudflare.KVNamespace("KV");
+export const KV = Cloudflare.KV.KVNamespace("KV");
```
sam-goodwin: Rename to Cloudflare.KV.Namespace

examples/cloudflare-worker/src/NotifyWorkflow.ts
```diff
@@ -25,7 +25,7 @@ export default class NotifyWorkflow extends Cloudflare.Workflow<NotifyWorkflow>(
     return Effect.fn(function* (input: { roomId: string; message: string }) {
       const { roomId, message } = input;
 
-      const stored = yield* Cloudflare.task(
+      const stored = yield* Cloudflare.Workers.task(
```
sam-goodwin: Should be Cloudflare.Workflows.task(..). Move workflow stuff to its own Workflows/ folder

examples/cloudflare-worker/src/Room.ts
```diff
@@ -13,7 +13,7 @@ export default class Room extends Cloudflare.DurableObjectNamespace<Room>()(
   Effect.gen(function* () {
     const state = yield* Cloudflare.DurableObjectState;
     return Effect.gen(function* () {
-      const sessions = new Map<string, Cloudflare.DurableWebSocket>();
+      const sessions = new Map<string, Cloudflare.Workers.DurableWebSocket>();
```
sam-goodwin: Rename to Cloudflare.WebSocket.Socket

examples/cloudflare-worker/src/Bucket.ts
sam-goodwin: Same for D1Database, rename to Cloudflare.D1.Database

Apply consistent suffixless convention for each of the namespaces. You'll have to audit each of the cloudflare resources for this problem. All files with a redundant suffix in the name or in the class name need to be removed

D1Database.ts -> Database.ts
class D1Database -> class Database

examples/cloudflare-worker/src/NotifyWorkflow.ts
sam-goodwin: I also want each of these namespaces to be importable directly, just liek effect

two options:
```ts

Cloudflare.Workflows.task
```
or
```ts
import * as Workflows from "alchemy/Cloudflare/Workflows";

Workflows.task
```

Consistent for all namespaces in src/Cloudflare and src/AWS

PR #692, opened 2026-06-26
fix(engine): honor deleteFirst on replace

john-royal:
`deleteFirst` was plumbed from a provider's replace diff into persisted state but the apply engine never read it — every replacement was create-first, with the old generation reclaimed afterwards by Phase-2 GC. That silently broke any resource whose replacement can't coexist with the original (fixed physical name, singleton): the new create collided with the not-yet-deleted original (Docker `network ... already exists`, no-op `volume create`, etc.).

When a provider returns:

```ts
diff: () => ({ action: "replace", deleteFirst: true })
```

the engine now tears the previous generation(s) down **before** creating the replacement, then commits a terminal `created` state so Phase-2 GC has no old chain left to drain:

```diff
+if (node.deleteFirst) {
+  yield* deleteOldGenerations(replState.old); // walk + delete the old chain
+}
 // ...create the new generation...
-yield* commit<ReplacedResourceState>({ status: "replaced", old: replState.old, ... });
+yield* commit<CreatedResourceState>({ status: "created", ... }); // deleteFirst path
```

`delete` is already required to be idempotent, so an interrupted apply re-converges on re-run.

### Tests

Three cases in `apply.test.ts`, backed by a new `DeleteFirstResource` + `CollisionRegistry` fixture:

- `deleteFirst` replacement deletes-then-creates (asserts hook order `["delete", "create"]`) and lands in terminal `created` with no leftover chain.
- default replacement is unchanged — still create-first (`["create", "delete"]`).
- regression: a same-identity (`name: "singleton"`) replacement now succeeds where create-first previously died with a collision.

### DIFF
--- a/packages/alchemy/src/Apply.ts
+++ b/packages/alchemy/src/Apply.ts
@@ -38,6 +38,7 @@ import {
   type PersistedState,
   type RanActionState,
   type ReplacedResourceState,
+  type ReplacementOldResourceState,
   type ReplacingResourceState,
   type ResourceState,
   type RunningActionState,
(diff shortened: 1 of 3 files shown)


sam-goodwin (commented):
Do we have tests for when we fail to delete them and have to redeploy to retry? I am curious if the way we store these pending items is durable and idempotent to ensure that they are eventually deleted and aren't lost or orphaned by accident

PR #699, opened 2026-06-29
feat: add alchemy action and gh env helper

harrysolovay:
Add a root GitHub Action for Alchemy deploys and expose GitHub Actions metadata to stack code.

```yml
- uses: alchemy-run/alchemy-effect@v1
```

The composite action resolves the lifecycle from the GitHub event:

```
pull_request opened/reopened/synchronize  -> deploy staging-{number}
pull_request closed                       -> destroy staging-{number}
push to main                              -> deploy prod
other events                              -> skip
```

It also exports PULL_REQUEST for PR stages so stacks can conditionally create resources.

```ts
import * as GitHub from "alchemy/GitHub";

const github = yield* GitHub.GitHubEnv;

if (github?.pr) {
  yield* GitHub.Comment("preview-comment", {
    owner: github.owner,
    repository: github.repository,
    issueNumber: github.pr,
    body: `Preview for ${github.sha}`,
  });
}
```

Docs now show GitHub.GitHubEnv in the CI guide and include root action usage in the README.


packages/alchemy/src/GitHub/Env.ts
```diff
@@ -0,0 +1,54 @@
...
+ *
+ * Resolves to `undefined` outside GitHub Actions. In GitHub Actions, it reads
+ * `GITHUB_SHA`, `GITHUB_REPOSITORY_OWNER`, `GITHUB_REPOSITORY`, and the optional
+ * `PULL_REQUEST` variable exported by the Alchemy GitHub Action.
+ */
+export const GitHubEnv: Config.Config<GitHubEnv | undefined> = Config.boolean(
+  "GITHUB_ACTIONS",
+).pipe(
+  Config.withDefault(false),
+  Config.flatMap((enabled) =>
```
harrysolovay: I'll fix this––but before I take another pass, question: can the GitHubEnv be provided via the gh providers? Or does that not quite make sense given its moreso an actions env concern than it is a gh resource concern?

website/src/content/docs/guides/ci.mdx
```diff
@@ -224,28 +224,29 @@ By the end you'll have:
      { providers: /* ... */ },
```
sam-goodwin: We should update this to include `Layer.mergeAll(GitHub.providers(), Cloudflare.providers())` since that's required below

packages/alchemy/src/GitHub/Env.ts
```diff
@@ -0,0 +1,54 @@
...
+export interface GitHubEnv {
+  /** The commit SHA that triggered the workflow run. */
+  readonly sha: string;
+  /** The repository owner from `GITHUB_REPOSITORY_OWNER`. */
+  readonly owner: string;
+  /** The repository name parsed from `GITHUB_REPOSITORY`. */
+  readonly repository: string;
+  /** The pull request number exported by the Alchemy GitHub Action. */
+  readonly pr: number | undefined;
+}
```
sam-goodwin, line 17: I don't think we should add this and instead rely on `Config.string` implicit connection. This context doesn't make much sense when running locally, for example. Or maybe provide these as exported `Config` values like 
```ts
export const GITHUB_SHA = Config.string("GITHUB_SHA")
// etc.
```

That way we don't couple `GitHub.providers()` to the GitHub action.

harrysolovay: Not sure I fully understand, but I'll lay out my rationale:

It'd be good to have a `GitHubEnv` service of type `GitHubEnv | undefined` (currently, it's represented as just large config below, but I can service-ify it). This would let one could easily check if in the actions env through an undefined check.

```ts
const github = yield* GitHub.GitHubEnv;
if (github) {
  // Do PR environment-specific things here.
  // ...
}
```

website/src/content/docs/guides/ci.mdx
harrysolovay:
```suggestion
     { providers: Layer.mergeAll(GitHub.providers(), Cloudflare.providers()) },
```

packages/alchemy/src/GitHub/Env.ts
sam-goodwin: It's too specific to PRs. Alchemy is not jsut ran in the context of PRs or branches, etc.

You can also do github stuff from outside github (like gitlab or local). I think these are better as environment variables exported as Config than as a GitHubEnv as part of the provider layer.

sam-goodwin: I think i misunderstood what was goin on here. I was saying it should just be a Config effect and then i look again and that's exactly what you have. So yeah, your implementation is right. Disregard my comment. I was wrong.

PR #700, opened 2026-06-29
fix(cloudflare): detect and recover cf state store partial deploy

Mkassabov:
its possible to deploy a state store where only the pre-create stub for the worker is create but our state store check only checks if it exists so we can never create the actual worker with the script it needs.

this results in errors where we try and connect to the state store because we think it exists, but its not a real state store.

the fix is to just have our state store check hit the `/version` endpoint to verify its not just a stub

### DIFF
--- a/packages/alchemy/src/Cloudflare/StateStore/State.ts
+++ b/packages/alchemy/src/Cloudflare/StateStore/State.ts
@@ -153,8 +153,9 @@ export const state = () =>
         if (credentials) {
           return yield* ensureLatest(credentials);
         }
-        const workerExists = yield* isStateStoreAvailable(scriptName);
-        if (workerExists) {
+        const { accountId } =
+          yield* yield* CloudflareEnvironment.CloudflareEnvironment;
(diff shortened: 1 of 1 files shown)


packages/alchemy/src/Cloudflare/StateStore/State.ts
```diff
@@ -619,6 +619,26 @@ const isStateStoreAvailable = (scriptName: string = "alchemy-state-store") =>
...
+const isStateStoreServing = (accountId: string) =>
+  Effect.gen(function* () {
+    const url = yield* workers.getSubdomain({ accountId }).pipe(
+      Effect.map(({ subdomain }) =>
+        subdomain
+          ? `https://${STATE_STORE_SCRIPT_NAME}.${subdomain}.workers.dev`
+          : undefined,
+      ),
+      Effect.catch(() => Effect.succeed(undefined)),
+    );
```
sam-goodwin, line 636: Don't we have this already? We should resolve it once/re-use resolution code

PR #717, opened 2026-06-30
feat(planetscale): expose pooledOrigin on PostgresRole

### LINKED ISSUE #716: Planetscale.PostgresRole: expose a structured pooled (PSBouncer) origin for Hyperdrive bindings
### Summary

`Planetscale.PostgresRole` materializes a structured `origin: PostgresOrigin` (the direct, port 5432 connection components) so callers can wire role credentials straight into `Cloudflare.Hyperdrive.Connection({ origin: role.origin })`. The pooled PSBouncer endpoint, however, is only exposed as the `connectionUrlPooled` URL *string*. To get the same structured shape for pooling, a consumer has to re-parse that URL back into `{ scheme, host, port, user, password, database }` themselves.

### Why it matters (the failure mode)

Hyperdrive is bypassed in local dev — the worker connects to the origin database directly. Against a real per-stage PlanetScale Postgres branch, that means every worker invocation opens a fresh **direct** (5432) connection. Local connection churn quickly exhausts PlanetScale's direct-connection limit and the branch starts refusing connections (effectively taking the dev database down).

The fix is to route dev through PSBouncer (port 6432):

```ts
const db = Cloudflare.Hyperdrive.Connection("db", Effect.map(dbRole, role => ({
  origin: role.origin,        // direct 5432 — Hyperdrive pools at the edge in prod
  dev: role.pooledOrigin,     // PSBouncer 6432 — pooled for the direct local connection
})))
```

But `role.pooledOrigin` doesn't exist today, so consumers either hand-build `{ ...componentsOf(origin), port: 6432 }` or re-parse `connectionUrlPooled`. The resource already knows all of this; it should hand back the structured pooled origin the same way it hands back `origin`.

### Proposal

Add a `pooledOrigin: PostgresOrigin` attribute to `PostgresRole`, identical to `origin` but on port 6432. It's a pure derived field — the single `buildAttributes` site (used by both create/update and `list`) populates it.

PR: agcty/alchemy-effect@feat/planetscale-postgres-role-pooled-origin (opening against this repo).

### Related

`Planetscale.PostgresDefaultRole` currently exposes neither `origin` nor `pooledOrigin` (only the URL strings). It would benefit from the same structured `origin` + `pooledOrigin` for symmetry with `PostgresRole`, but that's a larger change and out of scope for the linked PR — happy to follow up if the direction is welcome.

agcty:
Closes #716.

### What

Add a `pooledOrigin: PostgresOrigin` attribute to `Planetscale.PostgresRole`, identical to the existing `origin` but on PSBouncer's pooled port (6432) instead of the direct port (5432).

```ts
const role = yield* Planetscale.PostgresRole("DbRole", { database, inheritedRoles: ["postgres"] })

const db = Cloudflare.Hyperdrive.Connection("db", {
  origin: role.origin,        // direct 5432 — Hyperdrive pools at the edge in prod
  dev: role.pooledOrigin,     // PSBouncer 6432 — pooled connection for local dev
})
```

### Why

`PostgresRole` already exposes a structured `origin` so role credentials drop straight into `Cloudflare.Hyperdrive.Connection`. The pooled endpoint is only available as the `connectionUrlPooled` **string** today, so anyone wanting the pooled components for a binding has to re-parse that URL.

The motivating case is the Hyperdrive `dev` origin. In dev Hyperdrive is bypassed and the worker connects to the origin directly; against a real per-stage PlanetScale branch that opens a fresh **direct** (5432) connection per invocation and quickly exhausts the direct-connection limit, taking the dev database down. Pointing `dev` at the pooled (6432) origin fixes it — but only if the resource hands back a structured pooled origin. See #716 for the full failure description.

### Change

- New `pooledOrigin: PostgresOrigin` field on `PostgresRoleAttributes`.
- Populated in the single `buildAttributes` helper (used by both create/update and `list`), mirroring the existing `origin` block with `port: 6432`. Pure derived field — no new API calls, no behavior change to existing attributes.

### Notes

- `PostgresOrigin` is unchanged; `pooledOrigin` reuses it.
- Existing `PostgresRole` tests use `toMatchObject`, so the added field doesn't disturb them. I couldn't run the full local suite from a standalone fork clone (the workspace resolves private `@distilled.cloud/*` packages), so CI here is the real gate — the change is a pure, type-trivial mirror of `origin`.
- Out of scope: `PostgresDefaultRole` currently exposes neither `origin` nor `pooledOrigin` (only URL strings) and could get the same treatment for symmetry — happy to follow up separately if wanted.

### DIFF
--- a/packages/alchemy/src/Planetscale/Postgres/PostgresRole.ts
+++ b/packages/alchemy/src/Planetscale/Postgres/PostgresRole.ts
@@ -113,8 +113,10 @@ export interface PostgresRoleAttributes {
   database: string;
   /** The Postgres database name inside the branch. */
   databaseName: string;
-  /** Parsed connection components ready to feed into Cloudflare Hyperdrive. */
+  /** Parsed direct (port 5432) connection components ready to feed into Cloudflare Hyperdrive. */
   origin: PostgresOrigin;
+  /** Parsed pooled (PSBouncer, port 6432) connection components, e.g. for a Hyperdrive `dev` origin. */
+  pooledOrigin: PostgresOrigin;
   /** Direct connection URL for the database (Redacted). */
   connectionUrl: Redacted.Redacted<string>;
   /** Pooled connection URL via PSBouncer (port 6432, Redacted). */
@@ -528,6 +530,14 @@ const buildAttributes = (
       scheme: "postgres",
       host: role.access_host_url,
       port: 5432,
+      database: role.database_name,
+      user: role.username,
+      password,
+    },
+    pooledOrigin: {
+      scheme: "postgres",
+      host: role.access_host_url,
+      port: 6432,
       database: role.database_name,
       user: role.username,
       password,
--- a/packages/alchemy/test/Planetscale/Postgres/Hyperdrive.test.ts
+++ b/packages/alchemy/test/Planetscale/Postgres/Hyperdrive.test.ts
@@ -14,8 +14,18 @@ import HyperdriveWorker from "./fixtures/hyperdrive-worker.ts";
 import type { Widget } from "./fixtures/schema.ts";
 import { Hyperdrive, PlanetscaleDb } from "./fixtures/Stack.ts";
 
+const providers = Layer.mergeAll(
+  Cloudflare.providers(),
+  Planetscale.providers(),
+);
+
 const { test } = Test.make({
-  providers: Layer.mergeAll(Cloudflare.providers(), Planetscale.providers()),
+  providers,
+});
+
+const { test: devTest } = Test.make({
+  providers,
+  dev: true,
 });
 
 const logLevel = Effect.provideService(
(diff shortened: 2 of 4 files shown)


sam-goodwin:
Looks great. Please add a fixture test that uses it and verifies it works end to end

sam-goodwin:
What about MySQL? Does Neon also support this?

agcty:
Added the fixtures and also created a Neon PR.

This is what the agent said about MySql:

PlanetScale MySQL’s password API only gives us the normal MySQL connection shape we already expose as origin on port 3306. I didn’t find a connectionUrlPooled, pooled: true, or alternate pooled port equivalent. The API has MySQL-specific fields like direct_vtgate, direct_vtgate_addresses, regional access hosts, and mysql_edge_address, but that looks like a different feature surface, not the same pooled-origin concept.

🤔

PR #722, opened 2026-07-02
feat(core): per-resource credential metadata + scope-aware login/create-token

sam-goodwin:
Providers now declare the credentials their lifecycle operations require via a `metadata` field on `ProviderService`, and the CLI uses it to request or mint exactly the grants a stack needs.

```ts
// core: empty AlchemyProviderMetadata interface, augmented per cloud
metadata: {
  cloudflare: {
    scope: "account",
    auth: {
      oauth: { scopes: ["workers:write", "workers_scripts:write"], readScopes: ["workers:read"] },
      token: {
        permissionGroups: [
          { id: "e086da7e2179491d91ee5f35b3ca210a", name: "Workers Scripts Write" },
        ],
      },
    },
  },
}
// AWS: metadata: { aws: { iam: { actions, readActions, notes } } }
```

- All 380 AWS + Cloudflare resource providers annotated (per-resource lifecycle→API mapping researched from the provider source; 878 permission-group `{id, name}` pairs validated against the static catalog). A static test fails CI when a new provider lands without metadata (`metadata: {}` marks local-only providers).
- `collectProviderMetadata(context, usedTypes)` keeps core cloud-agnostic; compiling the stack (`yield* stackEffect` — no cloud calls, no credentials) yields the exact resource types in use. Cloud-specific analysis lives with each auth provider (`Cloudflare/Auth/Requirements.ts`, `AWS/Requirements.ts`).
- `alchemy login`: the Cloudflare OAuth prompt preselects exactly the scopes the stack's resources need (9 instead of the 26 `DEFAULT_SCOPES` for `examples/cloudflare-worker-async`) and warns up front when resources can't be managed over OAuth at all (158 of 238 Cloudflare resources are API-token-only).
- `alchemy cloudflare create-token --from-stack alchemy.run.ts`: mints a least-privilege token from the stack's permission groups, id-resolved against the live `/user/tokens/permission_groups` endpoint with a name fallback and a loud warning for anything unresolvable.
- Deploy preflight (`Auth/Preflight.ts`): before `Plan.make`, each registered auth provider diffs required-vs-granted for the active profile and renders warnings with resource attribution. It never blocks the run.
- `AWS/Requirements.ts` includes `toPolicyDocument` (deploy or plan-only tiers) for a future `alchemy aws create-policy`.
- `bun docs:gen` now emits a "Required permissions" section per resource page (372 pages): OAuth scopes, token permission groups with ids, IAM actions, and plan-only tiers.

Also replaces two raw NUL bytes in `Cloudflare/Web3/ContentList.ts` with `\0` escapes (same semantics; the file was being treated as binary by search tools).

### DIFF
--- a/packages/alchemy/src/AWS/ACM/Certificate.ts
+++ b/packages/alchemy/src/AWS/ACM/Certificate.ts
@@ -288,6 +288,27 @@ export const CertificateProvider = () =>
       });
 
       return {
+        metadata: {
+          aws: {
+            iam: {
+              actions: [
+                "acm:RequestCertificate",
+                "acm:DeleteCertificate",
+                "acm:AddTagsToCertificate",
+                "acm:RemoveTagsFromCertificate",
+                "route53:ChangeResourceRecordSets",
+                "route53:GetChange",
+              ],
+              readActions: [
+                "acm:ListCertificates",
+                "acm:DescribeCertificate",
+                "acm:ListTagsForCertificate",
+              ],
+              notes:
+                "ACM calls are pinned to us-east-1. route53:ChangeResourceRecordSets/GetChange are needed only when hostedZoneId is set for automatic DNS validation. RequestCertificate with inline Tags requires acm:AddTagsToCertificate.",
+            },
+          },
+        },
         stables: ["certificateArn"],
         list: () =>
           Effect.gen(function* () {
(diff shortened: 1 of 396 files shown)


sam-goodwin:
@BlankParticle you were interested in this one but closing in favor of a new impleentation later

PR #727, opened 2026-07-02
fix(cloudflare): persist dev image for containers

john-royal:
Fixes `alchemy dev` after a live deploy failing with "Container has no dev image".

The live provider hardcoded `dev: undefined`, so once a container was deployed the persisted state carried no dev image and the next `alchemy dev` died in `LocalWorkerProvider`:

```ts
// before — toAttributes in ContainerProvider
dev: undefined,
```

Now the live provider computes and persists `dev` for all three image variants, and `read` preserves it:

```ts
// effectful / external → build context the runtime can `docker build`
dev: { context, dockerfile }
// remote → pull directly
dev: { imageUri: props.image }
```

The build-context materialization is extracted into a shared `prepareContainerBuildContext` so the live and local providers agree on the deterministic context path.

### Verified

- `deploy` then `alchemy dev` on `examples/cloudflare-dev`: reaches `Done: succeeded` with no "no dev image" error.
- `GET /sandbox` returns the container's response, confirming the locally built dev image serves traffic.


sam-goodwin:
What's outstanding here?

john-royal:
Confirmed that the alternate path for environment variables will work in cloudflare-runtime. Reverting the environment variables part in this PR, will merge and then environment/secrets will be a separate PR.

PR #737, opened 2026-07-03
fix(cloudflare): fix state-store errors

Mkassabov:
Triaged 30 days of Axiom `prod-traces` (~400k `state_store.*` spans, ~60 distinct error messages) and fixed every state-store error that was an actual code defect. Each fix ships with a hermetic test that reproduces the exact production failure. `bun tsc -b` is clean and the live `test/Cloudflare/StateStore/State.test.ts` suite passes end-to-end.

Counts below are distinct users over the 30-day window unless noted.

### Errors fixed

**1. Empty-message `StateStoreError`** — 31 spans / 14 users (beta.57–59), thrown at `HttpStateStore.ts:236`.
The store's no-content `401` (and other tag-only client errors) produced a blank message, so users saw `StateStoreError:` with nothing to act on. `mapStateStoreError` now derives a message from tag → status → cause via a new `describeStateStoreFailure` helper; `Unauthorized` maps to an actionable hint. This also fixes the opaque `Decode error (500 PUT …)` bodies and the `SyntaxError: JSON Parse error: Unexpected identifier "Alchemy"` crash (non-JSON `200` body).

```ts
// before: message: e instanceof Error ? e.message : String(e)   // "" for no-content 401
// after:
new StateStoreError({ message: describeStateStoreFailure(e), cause: ... })
// → "State store rejected the request as unauthorized … run 'alchemy login' … (HTTP 401)"
```

Tests: `test/State/HttpStateStore.test.ts` — `describeStateStoreFailure` (empty message, Unauthorized→login hint, status appended, cause appended, non-Error) + `makeHttpStateStore` (401→non-empty actionable, non-JSON 200→non-empty).

**2. Malformed account-id URLs** — `Could not route to /accounts/<empty>/workers/scripts/alchemy-state-store/script-settings` — 242 users (the single biggest error). Empty or placeholder account IDs (`""`, `"-"`, `"dummy"`, `"test"`, `"mock"`) were interpolated straight into the API path; the account-id prompt is skippable, so stored credentials could carry `accountId: ""`. New `validateAccountId` rejects them (32-hex check) with an actionable `AuthError` before any call, wired into the env / stored / oauth credential paths and the interactive prompt.

Tests: `test/Cloudflare/Auth/AccountId.test.ts` — valid id, trim/lowercase, undefined, empty, whitespace, each placeholder (`-`, `dummy`, `test`, `mock`, `abc123`), source named in message.

**3. No retry on rate limiting / transient network** — `Please wait and consider throttling your request speed` (39 spans, 429), `fetch failed` (42), `Transport error (GET …/state/stacks)`, `internal_error`, `Unable to connect`. The `state()` layer never received the blanket Cloudflare retry policy that `providers()` applies, so init-time probes ran on the SDK's short default. Exported it as `CloudflareRetryLive()` and provided it to `state()`. Also **bounded two previously-unbounded / uncapped retries** (latent hangs): `checkHttpStateStoreAuth` retried `404`/`5xx` forever (now `+75` cap, and it now retries transport-level failures), and the edge-secret-read exponential backoff had no cap (now 2s ceiling).

Tests: `test/State/HttpStateStore.test.ts` — 5xx retry-then-succeed, persistent 5xx fails after exactly 6 attempts (bounded), transport-error retry-then-succeed; `test/Cloudflare/StateStore/TransientErrors.test.ts` — `isTransientBootstrapWriteError` driven through the real client.

**4. Secrets Store binding not-found race** — `Secrets Store binding 'StateStoreEncryptionKey'/'AlchemyStateStoreToken' references store … which were not found` — a just-created secret is still `pending` when the worker deploy that binds it runs. Added a typed `SecretsStoreBindingNotFound` distilled patch on `putScript` + a bounded retry, and the `Secret` reconciler now waits (bounded ~10s) for the secret to report `active` before returning.

Tests: `test/Cloudflare/StateStore/TransientErrors.test.ts` covers the transient-write predicate; live `State.test.ts` bootstrap exercises the deploy path.

**5. Edge-preview secret read flakiness** — `Failed to create edge preview session` (77 spans), `Secret probe returned 400/502: <!DOCTYPE html>…` (Cloudflare HTML error pages). The retry only matched one narrow config error; now retries non-200 probe responses, session-create flakes and transport errors (bounded), while still excluding permanent auth causes (`Unauthorized`/`Forbidden`/`InvalidRoute`).

Tests: `test/Cloudflare/StateStore/TransientErrors.test.ts` — `isTransientEdgeSessionError` (HTML probe page, transient cause, permanent-auth exclusion, non-EdgeSessionError).

**6. Lock file failures** — `EINVAL: … mkdir 'C:\Users\anton\.alchemy\lock\${ALCHEMY_PROFILE:-default}-Cloudflare.lock.lock'` (unexpanded shell placeholder → invalid Windows filename), `EROFS: read-only file system` (CI containers), `Lock file is already being held` (49 spans). Lock keys are now sanitized; read-only / permission-denied filesystems degrade to lock-less with a warning instead of crashing; the held-lock timeout explains what to do.

Tests: `test/Auth/Lock.test.ts` — `sanitizeLockKey` (plain keys untouched, `${VAR}` neutralised, path-separator escape), plus a `withLock` roundtrip using the exact production EINVAL key and an in-process serialisation check.

**7. Misc** — `This Worker has no versions …` (interrupted deploy) is now typed (`WorkerHasNoVersions` distilled patch on `getScriptSetting`) and treated as store-absent so bootstrap redeploys. `wait_for_version` budget raised 10s → 30s (traces showed redeploys serving the old version past the old window).

Tests: covered by the live `State.test.ts` version/bootstrap cases; `WorkerHasNoVersions` handling verified by the typed `catchTag` compiling against the patched union.

### Distilled patches

The two new typed errors (`WorkerHasNoVersions` on `getScriptSetting`, `SecretsStoreBindingNotFound` on `putScript`) landed in alchemy-run/distilled#363 and the submodule pin here is bumped to it.

The `SecretsStoreBindingNotFound` matcher (`{ "code": 10182 }`) is verified against the live API: a `secrets_store_secret` binding referencing a missing secret or store fails with code `10182`, and the code is unique to Secrets Store bindings (KV `10041`, R2 `10085`, D1 `10181`, Queue `11000`, Service `10144` all use their own not-found codes).

### Not fixed (not code bugs)

The highest-volume message — `Cloudflare State store not found. Run 'alchemy bootstrap cloudflare'` in CI (311 spans / 46 users) — is by-design (CI refuses to auto-bootstrap), but the volume suggests a UX decision worth making. `You need a workers.dev subdomain`, `verify your email address`, and genuine bad-credential `Authentication error` are user-account states with already-actionable messages.


packages/alchemy/src/Auth/Lock.ts
```diff
@@ -13,6 +30,11 @@ import { rootDir } from "./Profile.ts";
  * holders by path (so same-process callers wait via `retries`) and uses
  * an OS file lock for cross-process coordination, with stale-lock
  * detection at 60s.
+ *
+ * Locking is best-effort: on file systems where the lock directory
+ * cannot be created at all (read-only home in containers/CI), the
+ * effect runs unserialised with a warning instead of failing — a
+ * missed lock only risks a redundant credential refresh.
```
sam-goodwin, line 37: The better solution is to not use `~/.alchemy/` in CI/CD? Or just skip local locking in CI/CD?

packages/alchemy/src/Cloudflare/StateStore/State.ts
```diff
@@ -190,6 +190,12 @@ export const state = () =>
     Layer.provideMerge(Access.AccessLive),
     Layer.provideMerge(ProfileLive),
     Layer.provideMerge(CredentialsStoreLive),
+    // Same blanket retry policy `providers()` applies — without it the
+    // init-time subdomain/script/secrets probes run on the SDK default
+    // and give up early under Cloudflare rate limiting. `provide` (not
+    // `provideMerge`) so the distilled Retry tag stays out of this
+    // layer's public type.
+    Layer.provide(Cloudflare.CloudflareRetryLive()),
```
sam-goodwin: Should we fix this in distilled?

Otherwise fix it in src/Cloudflare/Providers.ts (we already have this kind of thing for AWS/Cloudflare)

Mkassabov: move to be for all providers

PR #742, opened 2026-07-06
fix(cloudflare): cold-recovery read for Access Application

microagi-andy:
`AccessApplication.read` bails out when there is no persisted `applicationId`, so after state loss (or a stage migration rebuilding state via the adoption probe) the engine plans a blind `create`. Cloudflare accepts a second application on the same domain — with a fresh `aud` — which silently breaks JWT validation for anything pinned to the original application's audience.

Add a cold-recovery fallback: when no id is persisted, scan the account's applications and match by the desired `domain` (the probe passes resolved props as `olds`).

```ts
if (output?.applicationId) {
  observed = yield* observeById(accountId, output.applicationId);
} else if (olds?.type !== "warp" && typeof olds?.domain === "string") {
  observed = yield* findByDomain(accountId, olds.domain);
} else {
  return undefined;
}
```

- Recovered **by id** → plain attrs (positively ours, same as before).
- Recovered **by domain scan** → `Unowned(attrs)` — existence is certain but ownership is not (Access applications carry no alchemy marker), so takeover is gated behind `--adopt` per the AdoptPolicy contract.
- `warp` apps are excluded from the fallback: they are a per-account singleton that `reconcile` already recovers via `findWarpApp`.

Hit in production while rebuilding a ~100-resource stack's state under a new stage: three `self_hosted` apps planned as duplicate creates; with this fallback they adopt cleanly (verified live — plan converges to noop, applicationIds/AUDs preserved).

### DIFF
--- a/packages/alchemy/src/Cloudflare/Access/Application.ts
+++ b/packages/alchemy/src/Cloudflare/Access/Application.ts
@@ -3,6 +3,7 @@ import * as Effect from "effect/Effect";
 import * as Schedule from "effect/Schedule";
 import * as Stream from "effect/Stream";
 
+import { Unowned } from "../../AdoptPolicy.ts";
 import { createPhysicalName } from "../../PhysicalName.ts";
 import * as Provider from "../../Provider.ts";
 import { Resource } from "../../Resource.ts";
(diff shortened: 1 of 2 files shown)


sam-goodwin:
How can we add a regression test?

microagi-andy:
Added in 20e875dc — the scratch-stack harness makes the state-loss scenario directly testable since the test body shares the in-memory state store with its deploys:

1. deploy a `self_hosted` app on `alchemy-test-cold-read.${zoneName}` (standing test zone, same pattern as the existing cases)
2. `state.delete({ stack, stage, fqn: "ColdReadApp" })` — the app still exists in Cloudflare, the engine has no `applicationId` (exactly the post-state-loss shape)
3. redeploy **without** adopt → `Effect.flip` must yield `OwnedBySomeoneElse` (the domain-matched app surfaces as `Unowned`; on `main` this step instead creates a duplicate)
4. redeploy **with** `AdoptPolicy.adopt(true)` → asserts the **same** `applicationId` and `aud` come back
5. out-of-band `listAccessApplicationsForAccount` → asserts exactly **one** app exists on the domain, then `stack.destroy()`

Steps 3 and 5 are the regression teeth: on `main`, 3 succeeds (duplicate created with a fresh `aud`) and 5 finds two apps.

One honest caveat: I don't have credentials for the testing account / `alchemy-test-2.us`, so I couldn't execute it — it mirrors the existing suites in this file structurally, but please give it a CI run.

packages/alchemy/src/Cloudflare/Access/Application.ts
```diff
@@ -544,6 +567,24 @@ const findWarpApp = (accountId: string) =>
...
+    Effect.map((chunk) =>
+      Array.from(chunk).find(
+        (a) => (a as { domain?: string | null }).domain === domain,
+      ),
+    ),
+    Effect.map((found) =>
+      found === undefined
+        ? undefined
+        : narrowApp(found as Parameters<typeof narrowApp>[0]),
+    ),
```
sam-goodwin, line 585: Nit: it's better to use `Stream.filter`, `Stream.map` and `Stream.runHead` at the end so that it's more optimistic. This current version scans all and then filters.

PR #743, opened 2026-07-06
feat(cloudflare): pin the state-store account via state({ accountId })

microagi-andy:
The Cloudflare account hosting a stack's state store is decided implicitly by the auth profile. A `login` against the wrong account — easy for users who belong to several — resolves an **empty** state store: the engine then cold-probes live cloud resources and fails with confusing `OwnedBySomeoneElse` errors on the first plan, or (worse) `--adopt` silently forks the stack's state into the wrong account, arming a `destroy` that can delete live infra. We hit exactly this onboarding a new team member this week.

Let stacks pin the account:

```ts
Alchemy.Stack("my-stack", {
  providers: ...,
  state: Cloudflare.state({ accountId: "c8eeff0e..." }),
}, ...)
```

- `StateOptions.accountId` (and the same on `BootstrapOptions`) is asserted against the profile-resolved account **before** anything is read, bootstrapped, or upgraded — an account mismatch can no longer reach the "State Store not available. Do you want to deploy it?" prompt.
- On mismatch it dies with an `AuthError` naming both accounts and the fix:

```
Cloudflare account mismatch: profile 'default' resolves account <actual>, but this
stack pins its state store to account <expected>. Re-run 'alchemy login --profile
default' and select the expected account.
```

- No pin → no behavior change (`state()` remains zero-config).

The `assertAccountId` helper carries an explicit return type — letting it infer joins the self-referential `typeof init` context inference inside `state()` and breaks the `deployStateStore` providers-layer type downstream.

### DIFF
--- a/packages/alchemy/src/Cloudflare/StateStore/State.ts
+++ b/packages/alchemy/src/Cloudflare/StateStore/State.ts
@@ -56,7 +56,54 @@ const CI = Config.boolean("CI").pipe(Config.withDefault(false));
 /** Filename used for stored credentials under the profile directory. */
 const CREDENTIALS_FILE = "cloudflare-state-store";
 
-export const state = () =>
+export interface StateOptions {
+  /**
+   * Pin the Cloudflare account this stack's state store must live in.
+   *
+   * The account is otherwise decided implicitly by the auth profile —
+   * a `login` against the wrong account (easy for users who belong to
+   * several) resolves an **empty** state store, so the engine
+   * cold-probes live cloud resources and fails with confusing
+   * `OwnedBySomeoneElse` errors (or, with `--adopt`, silently forks
+   * state into the wrong account). When set, the layer asserts the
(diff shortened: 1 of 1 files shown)


packages/alchemy/src/Cloudflare/StateStore/State.ts
```diff
@@ -56,7 +56,54 @@ const CI = Config.boolean("CI").pipe(Config.withDefault(false));
...
+): Effect.Effect<
+  void,
+  never,
+  CloudflareEnvironment.CloudflareEnvironment
+> =>
+  Effect.gen(function* () {
+    if (expected === undefined) return;
+    const { accountId } =
+      yield* yield* CloudflareEnvironment.CloudflareEnvironment;
+    if (accountId !== expected) {
```
sam-goodwin, line 93: This looks like it just asserts its the same account id, but doesn't enable you to have the state store in one account and deploy to another?

PR #746, opened 2026-07-06
fix(cloudflare): preserve explicit globalOutbound: null in WorkerLoader

agcty:
## Bug

`unwrapWorkerLoader` coerces an explicit `globalOutbound: null` to `undefined`:

```ts
globalOutbound: loader.globalOutbound?.raw,
```

`null` is the documented "block all outbound network access" signal — this same file advertises it three times (the type on `WorkerLoaderWorkerCode`, the module JSDoc, and the `sandbox` example). But `null?.raw` evaluates to `undefined`, which the runtime treats as "no override" — so a worker loaded with the sandbox explicitly enabled silently gets default outbound network access.

## Fix

Preserve the null:

```ts
globalOutbound:
  loader.globalOutbound === null ? null : loader.globalOutbound?.raw,
```

One line; `undefined`/absent and `Fetcher` cases behave exactly as before. We've been running this in our fork in production.


sam-goodwin:
Is there a test we can add to catch regressions?

agcty:
Added one in 936120e: `unwrapWorkerLoader` is now exported (`@internal`, mirroring `normalizeStateDomains`) with a pure unit test in `WorkerLoaderUnwrap.test.ts` — kept separate from `WorkerLoader.test.ts` since that file's `beforeAll(deploy(...))` would force a live deploy for a unit assertion.

It covers the three `globalOutbound` states (explicit `null` preserved, omitted stays `undefined`, wrapped fetcher unwraps to `.raw`) plus `tails`/`streamingTails` unwrapping. Verified the null-preservation test fails against the pre-fix `?.raw` coercion and passes with this PR.

packages/alchemy/test/Cloudflare/Workers/WorkerLoaderUnwrap.test.ts
sam-goodwin, line 1: I usually like to avoid tests like this and instead test behavior. Is there a way to test the behavior rather than the properties?

PR #747, opened 2026-07-06
fix(cloudflare): include worker metadata in the update diff

### LINKED ISSUE #745: Worker metadata-only changes (compat flags, observability, placement, env literals) don't trigger an update
## Summary

The Worker provider's update decision is:

```ts
if (domainsChanged || cronsChanged || (yield* hasChanged(id, news, output))) {
  return { action: "update", ... };
}
```

and `hasChanged` compares only content hashes: script hash, vite input hash, or `prepareBundle` hash, plus assets. Nothing hashes the deploy **metadata** — so changes that only affect the `PutScriptRequest` metadata never mark the worker stale and silently noop:

- `compatibilityDate` / `compatibilityFlags`
- observability settings (e.g. enabling `observability.traces`)
- placement / logpush
- plain-string env var values (not resource-backed bindings, which the engine's binding diff catches)
- assets config (`html_handling`, `not_found_handling`, headers/redirects)

Same failure mode as #744, different trigger: deploy exits 0, plan says noop, the worker keeps running with the old configuration. Enabling tracing on an already-deployed worker and seeing nothing happen is how this class of bug typically gets discovered.

## Fix we carry in our fork

We added a `resolveWorkerMetadataHash` step: build the same metadata object the upload path would send (bindings resolved, env folded in, assets config, compat fields, observability, placement, tags), hash it, store it in the resource state alongside the bundle hash, and check it first in `hasChanged`:

```ts
if (bindings) {
  const metadataHash = yield* resolveWorkerMetadataHash(props, bindings, accountId);
  if (metadataHash !== output.hash?.metadata) {
    return true;
  }
}
// ...existing bundle/vite/assets hash checks
```

This has been running in production for us across many deploys. It has no fork-specific dependencies — happy to adapt it into a PR against main if you want it (our WorkerProvider has drifted from yours in the Vite build path, but this piece is self-contained).

agcty:
Fixes #745.

**The bug.** The Worker update decision is `domainsChanged || cronsChanged || hasChanged(...)`, and `hasChanged` compared only the bundle, vite-input, and asset-content hashes. Metadata-only changes — compatibility flags/date, observability, placement, limits, logpush, env literals, bindings, subdomain config, and tags — never touch those hashes, so the plan reported a noop and the change silently never deployed.

**The fix.** Hash the deploy-time metadata surface and store it as `hash.metadata` in output state, then compare it first in `hasChanged`. The materializer runs nested Effects, unwraps `Redacted` secrets by value (identical secrets hash identically regardless of reference identity), ISO-stringifies `Date`s, and drops functions and class instances, so the hash is stable and value-based. The local asset `directory` path and precomputed asset `hash` are excluded (machine-specific / already compared via `hash.assets`). The metadata check is skipped when bindings are still unresolved during planning — it's computed and stored on the resolving apply.

**Tests.** Pure unit tests in `test/Cloudflare/Workers/WorkerProvider.test.ts` (mirroring the exported-helper style of `normalizeStateDomains`): a changed compatibility flag changes the hash; a changed observability config changes the hash; identical props are stable; Redacted env affects the hash by value not reference; materialization drops functions and undefined-valued fields.

**Migration note.** State written before this change has no `hash.metadata`. The first diff after upgrading will see a differing (undefined) metadata hash and plan a one-time update for each worker, which backfills the hash; subsequent diffs are stable. This is self-healing and requires no action.

### DIFF
--- a/packages/alchemy/test/Cloudflare/Workers/WorkerProvider.test.ts
+++ b/packages/alchemy/test/Cloudflare/Workers/WorkerProvider.test.ts
@@ -1,5 +1,12 @@
-import { normalizeStateDomains } from "@/Cloudflare/Workers/WorkerProvider";
+import type { WorkerProps } from "@/Cloudflare/Workers/Worker";
+import {
+  normalizeStateDomains,
+  resolveMetadataHashValue,
+  resolveWorkerMetadataHash,
+} from "@/Cloudflare/Workers/WorkerProvider";
 import { describe, expect, test } from "@effect/vitest";
+import * as Effect from "effect/Effect";
+import * as Redacted from "effect/Redacted";
 
 describe("WorkerProvider", () => {
   describe("normalizeStateDomains", () => {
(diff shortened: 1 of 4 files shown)


packages/alchemy/test/Cloudflare/Workers/WorkerProvider.test.ts
```diff
@@ -59,4 +66,74 @@ describe("WorkerProvider", () => {
...
+    const stack = { name: "test-stack", stage: "test" };
+    const hashFor = (props: WorkerProps) =>
+      Effect.runPromise(
+        resolveWorkerMetadataHash({
+          props,
+          bindings: [],
+          accountId: "acct-1",
+          stack,
+        }),
+      );
```
sam-goodwin: Don't test like this. Use `test.provider` and do subsequent `test.deploy` with stacks you expect to actually make changes and assert they were applied.

PR #764, opened 2026-07-07
fix(cloudflare/workers): declare binding-hosted DO classes in the worker precreate stub

### LINKED ISSUE #763: Worker precreate stub omits DO classes that only exist as env bindings, breaking worker<->container cycles on first deploy
## Summary

A Worker's very first `create` fails with `Worker did not expose Durable Object namespace <class>`
whenever that Durable Object class is declared only as an `env` **binding** — e.g. a bare
`Cloudflare.DurableObject` fronting a `Cloudflare.Container` image — rather than as an Effect-native
export from the Worker entry. A retry of the same deploy succeeds, because by then the Worker script
already exists and the resource takes the adopt path instead of precreate.

## Repro

```
class Sandbox {} // no export from the worker entry -- this class is container-fronted only

const sandboxDo = Cloudflare.DurableObject<undefined>("sandbox-do", {
  className: "Sandbox",
});
const sandboxContainer = Cloudflare.Container("sandbox", {
  context: "./containers/sandbox",
});

class Website extends Cloudflare.Vite<Website>()("website", {
  env: { SANDBOX: sandboxDo },
}) {}

// worker <-> container binding cycle
const worker = yield* Website;
const sandbox = yield* sandboxContainer.Application;
const namespaceId = Output.map(worker.durableObjectNamespaces, (namespaces) => {
  const id = namespaces["Sandbox"];
  if (!id) throw new Error(`Worker did not expose Durable Object namespace Sandbox.`);
  return id;
});
yield* sandbox.bind("sandbox-do", { durableObjects: { namespaceId } });
```

On a fresh stage's first deploy, this throws the "Worker did not expose Durable Object namespace
Sandbox" error above, from inside the resolution of `worker.durableObjectNamespaces`. Re-running the
exact same deploy immediately after succeeds.

## Root cause

`Cloudflare.Vite`/`Cloudflare.Worker`'s `precreate` builds a placeholder `Worker["Attributes"]` before
the real Cloudflare-assigned Durable Object namespace ids exist, so dependents in a binding cycle
(like the container above) can resolve `worker.durableObjectNamespaces` during the same apply pass.
That placeholder derives its Durable Object class list only from `news.exports` — Effect-native DO
classes defined in the Worker's own entry module:

https://github.com/alchemy-run/alchemy-effect/blob/main/packages/alchemy/src/Cloudflare/Workers/WorkerProvider.ts#L1746-L1752
(current `main`, function `precreate` inside `LiveWorkerProvider`)

```
const exportMap = news.exports ?? {};
const durableObjects = Object.keys(exportMap)
  .filter((logicalId) => isDurableObjectExport(exportMap[logicalId]))
  .map((logicalId) => ({
    logicalId,
    className: logicalId,
  }));
```

It never looks at DO classes that only exist as `env` **bindings** — the shape a bare
`Cloudflare.DurableObject` fronting a `Container` takes. `reconcile` already has a helper for exactly
this,
[`getDurableObjectBindings`](https://github.com/alchemy-run/alchemy-effect/blob/main/packages/alchemy/src/Cloudflare/Workers/WorkerProvider.ts#L2256-L2288),
but `precreate` doesn't call it, and doesn't even destructure `bindings` from its own input even
though `Provider.precreate`'s type signature already passes them
([`Provider.ts`](https://github.com/alchemy-run/alchemy-effect/blob/main/packages/alchemy/src/Provider.ts#L194-L200)):

```
precreate?(input: {
  id: string;
  news: Props<Res>;
  instanceId: string;
  session: ScopedPlanStatusSession;
  bindings: BindingData<Res>;
}): Effect.Effect<Res["Attributes"], any, PrecreateReq>;
```

So a binding-only DO class is entirely absent from the stub's `durableObjectNamespaces`, and any
resource resolving `worker.durableObjectNamespaces[className]` against that stub (rather than the
final `reconcile` output) fails to find it.

## Impact

Any Worker that hosts a Durable Object solely as an `env` binding (the pattern used to front a
Cloudflare Container with a DO namespace, since the container needs to bind to that namespace by id)
hits this on the Worker's first-ever create. It's transient — retrying the deploy resolves it, since
the Worker script now exists and future runs take the adopt path — but it's a guaranteed failure on
every fresh stage's first deploy for this pattern, with no warning in the error message that a retry
is the fix.

## Fix

A fix exists and is attached as a draft PR (see linked PR). It derives `precreate`'s Durable Object
class list from both `news.exports` and the binding-derived classes (via the same
`getDurableObjectBindings` helper `reconcile` already uses), deduping by class name with the
binding-derived entry winning on collision, and scopes the placeholder's container metadata to only
the classes actually fronted by a `Container` binding (mirroring `reconcile`'s own
`containerClassNames` computation) instead of assuming every DO class is container-backed.

danielgangl:
Fixes #763.

The precreate placeholder derived its Durable Object class list only from `news.exports`, but a
worker also hosts DO classes declared as env bindings — e.g. a bare `Cloudflare.DurableObject`
fronting a Container image. Such a class was absent from the stub, so its namespace was never
created there, and a resource caught in a worker<->container dependency cycle — which resolves
`worker.durableObjectNamespaces` against the precreate stub, not the final reconcile output — failed
with `Worker did not expose Durable Object namespace <class>` on every fresh-stage deploy (reruns hit
the adopt branch and passed, since the script now exists).

```
// before: only exports
const durableObjects = Object.keys(exportMap)
  .filter((logicalId) => isDurableObjectExport(exportMap[logicalId]))
  .map((logicalId) => ({ logicalId, className: logicalId }));

// after: exports + binding-derived classes, deduped by class name
const durableObjects = mergeDurableObjectClasses(
  exportDerived,
  getDurableObjectBindings(bindings, name),
);
```

`getDurableObjectBindings` is the same helper `reconcile` already uses to derive DO classes from
bindings — `precreate` just wasn't calling it, and wasn't even destructuring `bindings` from its
input, even though `Provider.precreate`'s type signature already passes them.

`mergeDurableObjectClasses` is a small new exported helper: it dedupes the export-derived and
binding-derived lists by class name, letting the binding-derived entry win on a collision so the
`alchemy:do:` tag keys off the same logical id (the binding sid) that `reconcile` writes.

Also scoped the placeholder's container metadata to only the classes actually fronted by a Container
binding (mirroring `reconcile`'s own `containerClassNames`), instead of assuming every DO class is
container-backed — the old code mapped every export-derived DO class straight to a container entry,
which would have wrongly marked plain DOs (no container at all) as container-backed in the
placeholder once binding-derived classes were added to the same list.

### Tests

`test/Cloudflare/Workers/WorkerProvider.test.ts`: unit coverage for `mergeDurableObjectClasses` —
includes a binding-only class, dedupes with the binding-derived entry winning, passes exports through
unchanged with no bindings, and the empty case. Followed the existing `normalizeStateDomains` pattern
in the same file (pure exported helper, hermetic unit tests, no live provider needed).

Verified locally against the real toolchain (`distilled` submodule initialized, `bun install`):
`tsc -b` clean, and `vitest run test/Cloudflare/Workers/WorkerProvider.test.ts` passes 9/9 (the 5
pre-existing `normalizeStateDomains` tests plus the 4 new ones). I have not run the full live-provider
suite (needs Cloudflare credentials).

### DIFF
--- a/packages/alchemy/test/Cloudflare/Workers/WorkerProvider.test.ts
+++ b/packages/alchemy/test/Cloudflare/Workers/WorkerProvider.test.ts
@@ -1,4 +1,7 @@
-import { normalizeStateDomains } from "@/Cloudflare/Workers/WorkerProvider";
+import {
+  mergeDurableObjectClasses,
+  normalizeStateDomains,
+} from "@/Cloudflare/Workers/WorkerProvider";
 import { describe, expect, test } from "@effect/vitest";
 
 describe("WorkerProvider", () => {
@@ -57,6 +60,52 @@ describe("WorkerProvider", () => {
 
     test("returns an empty array for undefined state", () => {
       expect(normalizeStateDomains(undefined)).toEqual([]);
+    });
+  });
+
+  describe("mergeDurableObjectClasses", () => {
+    // A worker's precreate stub must declare every Durable Object class it
+    // hosts so each namespace is created there — not just classes defined as
+    // Effect-native exports, but also classes that only appear as an env
+    // binding (e.g. a bare `Cloudflare.DurableObject` fronting a Container
+    // image). Missing a binding-only class from the stub means its namespace
+    // never gets created there, so a resource caught in a worker<->container
+    // dependency cycle (which resolves `worker.durableObjectNamespaces`
+    // against the stub, not the final reconcile output) fails with
+    // "Worker did not expose Durable Object namespace <class>" on every
+    // fresh-stage deploy.
+    test("includes a class that only exists as a binding", () => {
+      const exportDerived = [{ logicalId: "Chat", className: "Chat" }];
+      const bindingDerived = [
+        { logicalId: "sandbox-do", className: "Sandbox" },
+      ];
+      expect(
+        mergeDurableObjectClasses(exportDerived, bindingDerived),
+      ).toEqual([
+        { logicalId: "Chat", className: "Chat" },
+        { logicalId: "sandbox-do", className: "Sandbox" },
+      ]);
+    });
+
+    test("dedupes by class name, letting the binding-derived entry win", () => {
+      const exportDerived = [{ logicalId: "Chat", className: "Chat" }];
+      const bindingDerived = [
+        { logicalId: "chat-do-binding", className: "Chat" },
+      ];
+      expect(
+        mergeDurableObjectClasses(exportDerived, bindingDerived),
+      ).toEqual([{ logicalId: "chat-do-binding", className: "Chat" }]);
+    });
+
+    test("returns export-derived classes unchanged when there are no bindings", () => {
+      const exportDerived = [{ logicalId: "Chat", className: "Chat" }];
+      expect(mergeDurableObjectClasses(exportDerived, [])).toEqual(
+        exportDerived,
+      );
+    });
+
+    test("returns an empty array when there are no classes at all", () => {
+      expect(mergeDurableObjectClasses([], [])).toEqual([]);
     });
   });
 });
--- a/packages/alchemy/test/Cloudflare/Workers/WorkerProvider.test.ts
+++ b/packages/alchemy/test/Cloudflare/Workers/WorkerProvider.test.ts
@@ -1,7 +1,4 @@
-import {
-  mergeDurableObjectClasses,
-  normalizeStateDomains,
-} from "@/Cloudflare/Workers/WorkerProvider";
+import { normalizeStateDomains } from "@/Cloudflare/Workers/WorkerProvider";
 import { describe, expect, test } from "@effect/vitest";
 
 describe("WorkerProvider", () => {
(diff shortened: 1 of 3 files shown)


packages/alchemy/test/Cloudflare/Workers/WorkerProvider.test.ts
```diff
@@ -59,4 +62,50 @@ describe("WorkerProvider", () => {
...
+      const bindingDerived = [
+        { logicalId: "sandbox-do", className: "Sandbox" },
+      ];
+      expect(
+        mergeDurableObjectClasses(exportDerived, bindingDerived),
+      ).toEqual([
+        { logicalId: "Chat", className: "Chat" },
+        { logicalId: "sandbox-do", className: "Sandbox" },
+      ]);
+    });
```
sam-goodwin: Don't test internals. Use test.provider to actually deploy the stack to verify end to end to deploy and destroy

sam-goodwin: Done in 2ff8032 — dropped the `mergeDurableObjectClasses` unit tests (and un-exported the helper) and replaced them with a `test.provider` case in `DurableObjectNamespace.test.ts` that deploys and destroys a real worker↔worker binding cycle on a scratch stack: the host declares its `Counter` DO class only as an env binding (no Effect-native exports), the consumer resolves `host.durableObjectNamespaces.Counter` from the precreate stub, and the host binds the consumer's name back to close the cycle — the same shape as the worker↔container cycle in #763, without needing a Docker build.

Verified both directions live: against the unfixed provider the deploy fails with `Worker did not expose Durable Object namespace Counter.` (the exact #763 error), and with the fix it deploys, the consumer's live env carries the same namespace id the host's final reconcile reports, and the stack destroys cleanly.

PR #774, opened 2026-07-08
feat(cloudflare/workers)!: build layers once per isolate, scope resources per event

sam-goodwin:
The Worker bridge previously rebuilt the user's entire layer stack on every request. Layers now build **once per instance**, and each event runs as an effect against the built context with a fresh request `Scope`:

```
instance scope:  layer build (once, first event) — assembly + one-shot I/O, closed at shutdown at best
request scope:   fresh Scope.Scope per event — disposables, finalizers, per-request memoization
```

- `getWorkerExport` memoizes one `Layer.buildWithMemoMap` under a never-closed instance scope, shared across the default worker and every Durable Object / Workflow class (WeakMap keyed on the common `meta.entrypoint`). Success-only memo (transient init failure heals), `Layer.CurrentMemoMap` stripped so in-handler `Effect.provide` still builds per event, and every awaiting event pins the in-flight build with `ctx.waitUntil` — workerd *drops* cross-request continuations whose origin context ended.
- `ExecutionContext` (scope + cache) is deleted. The request scope is the plain `effect/Scope` service; `Drizzle.postgres` memoizes its pool in a module `WeakMap` keyed on `yield* Effect.scope`, so memo key and finalizer target can never disagree:

```ts
const caches = new WeakMap<Scope.Scope, Map<symbol, Effect.Effect<any, any, any>>>();
```

- Init may perform one-shot I/O that caches a plain value (fetch a secret for a client) — `InitIO.test.ts` pins a real fetch inside a `Layer.effect` running exactly once per instance on both Workers and Lambda. What init must not do is acquire disposables (connections, pools, anything with a finalizer) or retain I/O-backed objects across events.
- On workerd, init-level finalizers **never run** (no isolate-teardown hook). `Platform` builds class layers against the ambient scope **at runtime only** (`Effect.provide`'s `scopedWith` would tear them down the moment init returns) and pins init's ambient `Scope` to the build scope; plan/deploy keeps `Effect.provide` so each `stack.deploy` in a session re-registers Output sources.
- RPC exits whose value is a genuine `Stream` now eject the request scope and close it via `Stream.onExit` when the stream drains — previously the close raced the drain.
- `handle_cross_request_promise_resolution` is forced for effect workers pinning a compatibility date before 2024-10-14 (Cloudflare rejects specifying a flag the date already defaults on); passing the `no_` flag is a deploy-time error.
- DurableObjectBridge shares the isolate build across activations (previously one full layer build per activation) and provides the per-call scope — `Drizzle.postgres` now works inside DO methods.

### Lambda graceful shutdown

Lambda's execution environment has a real Shutdown phase, but only for functions with a registered extension — with none, the sandbox is killed with **0 ms and no signal**; a registered *internal* extension gets **SIGTERM + 500 ms**.

What's implemented (`src/AWS/Lambda/RuntimeExtension.ts` + the generated entry in `Function.ts`):

```ts
// generated entry
await registerLambdaExtension();          // POST /extension/register {"events": []}
                                          // + parked GET /event/next (readiness signal)
const instanceScope = Scope.makeUnsafe(); // layers build under it (Layer.buildWithScope)
process.on("SIGTERM", () =>
  Effect.runPromise(Scope.close(instanceScope, Exit.void)).finally(() => process.exit(0)),
);
```

- **Instance scope**: layers build once per sandbox under an explicit scope (not a transient `Effect.scoped` region); init-level finalizers run in the SIGTERM window at sandbox spin-down.
- **Request scope**: each invocation gets a fresh `Scope`, settled **inline before the response** — there is **deliberately no `waitUntil` on Lambda** (see findings). Keep request finalizers cheap; must-not-lose work is written durably inside the handler.
- Cost: two localhost HTTP calls at cold start (~1–2 ms), zero per-invoke overhead (no `INVOKE` subscription → Lambda never waits on it).
- Tests: `Shutdown.test.ts` (registration + no Init hang + request-finalizer marker always on; idle spin-down half gated behind `AWS_LAMBDA_TEST_SHUTDOWN=1`), `InitIO.test.ts` (one-shot async I/O in a layer build runs once per sandbox).

Findings pinned live (the interesting bits for a writeup):

- A registered extension that never calls `/event/next` **wedges Init** — every invoke 502s with only `INIT_START` in the logs. For an extension subscribed to no events the parked call never resolves, which is exactly the ready-signal Lambda wants.
- SIGTERM **is** delivered at idle spin-down (~10 min observed; marker flushed to CloudWatch from inside the 500 ms window). It is **not** delivered on a timeout-forced environment reset.
- The obvious "Lambda `waitUntil`" — subscribe the extension to `INVOKE` and hold `/event/next` until queued cleanup settles — was implemented and then **falsified by measurement**: a buffered Function URL does not release the response until the entire Invoke phase completes (a 2 s finalizer produced a ~3 s response, same as inline), and environments whose invoke phase was held open were condemned and recycled (no `END`/`REPORT`, SIGTERM within seconds, fresh `INIT_START` on the next request). Reverted; the post-mortem lives in `RuntimeExtension.ts`.
- Dangling promises were considered and rejected: they resume on a later thaw at best (TCP rarely survives the freeze) and are dropped entirely on crash/timeout resets — silent data loss for durability flushes.

### Breaking changes

- init closures are evaluated once per instance (previously per event) and their finalizers no longer fire per event; `WaitUntil.test.ts` pins `initRuns === 1 && initFinalizerRuns === 0`, and on Lambda they now fire at shutdown.
- `ExecutionContext` is removed from `PlatformServices` / `WorkflowRunServices` (replaced by `Scope`), and `PlatformServices` no longer includes `Scope`, so scope-requiring init fails to typecheck.

Test fixture note: Hyperdrive's default SELECT caching (~60s TTL) can serve a pre-insert empty result across retries; the Drizzle fixtures now create their `Hyperdrive.Connection` with `caching: { disabled: true }`, and the DrizzleWorkflow suite gained a sequential + concurrent fetch-handler Drizzle test pinning the cross-request regression.


packages/alchemy/src/Drizzle/Postgres.ts
```diff
@@ -60,16 +72,29 @@ export const postgres = <
...
+          memo = yield* Effect.gen(function* () {
+            const pgCtx = yield* Layer.buildWithScope(
+              PgClient.layer({ url: yield* connectionString }),
+              scope,
+            );
+            return yield* PgDrizzle.makeWithDefaults(config).pipe(
+              Effect.provideContext(pgCtx),
+            );
+          }).pipe(Effect.cached);
+          cache.set(symbol, memo);
```
sam-goodwin: This has a race. We should pipe to Effect.cached and put that synchronously in the cache and evaluate it afterwards so that another fiber can't call cache.get while it's running.

website/src/content/docs/aws/compute/lambda.mdx
```diff
@@ -267,6 +267,43 @@ bun test test/integ.test.ts
...
 
+## Sandbox scope vs invocation scope
+
+The init closure runs **once per sandbox**, at cold start —
+everything it builds (bindings, SDK clients) is reused by every
+invocation. Each invocation then runs with a **fresh `Scope`** that
+is settled *inline* before the handler returns: Lambda freezes the
+sandbox after the response, so unlike Workers there is no
+`waitUntil` to detach cleanup to — `Effect.addFinalizer` in a
+handler runs before the response leaves.
```
sam-goodwin: Are we sure there's no way to do work after returning the response? 

Blocking on async cleanup is bad.

website/src/content/docs/cloudflare/compute/containers.mdx
```diff
@@ -154,6 +154,23 @@ return {
...
+## Process scope vs request scope
+
+Unlike Workers, a Container is a **real process** — and that changes
+what the instance scope means. The bundled program runs under a root
+scope that closes when the process shuts down gracefully, so
+resources acquired at init (a connection kept warm across requests,
+a background consumer) are genuinely released on exit — something no
+serverless runtime can offer, because workerd and Lambda have no
+teardown hook. A hard kill still skips finalizers, as in any
+process, so treat instance-level cleanup as best-effort.
```
sam-goodwin: This is out of date. We support it in AWS lambda

website/src/content/docs/infrastructure-as-effects/functions-and-servers.mdx
```diff
@@ -277,6 +277,76 @@ Pick whichever fits your team. The Effect style unlocks
...
+| Scope              | Covers                                           | Closes                                                    |
+| ------------------ | ------------------------------------------------ | --------------------------------------------------------- |
+| **Instance scope** | the constructor (init) and its Layers            | workerd: never · Lambda: 500 ms at shutdown · servers: on exit |
+| **Request scope**  | one event — a request, RPC call, run, invocation | when the event settles                                     |
+
+The constructor runs **once per instance** — a Worker isolate, a
+Lambda sandbox, a server process. Everything it builds (bindings,
+SDK clients, Layers) is assembled once and reused by every event.
+Treat it as pure assembly: resolve services, build clients, return
+handlers. Don't perform I/O or acquire connections there.
```
sam-goodwin: I/O should be ok, just as long as it doesn't need disposable resources. Eg fetching and caching a secret for a client.

packages/alchemy/src/Drizzle/Postgres.ts
sam-goodwin: Fixed: the cached effect is now constructed and inserted synchronously — `Effect.runSync(Effect.cached(build))` then `cache.set` with no yield between the miss and the set, so a concurrent first query joins the memo instead of building a second pool. (`Effect.cached` construction only allocates the memo cell, so `runSync` can't block.) Verified live via the DrizzleWorkflow suite incl. the concurrent-fetch test.

website/src/content/docs/cloudflare/compute/containers.mdx
sam-goodwin: Fixed — now reads: workerd never closes its instance scope, Lambda gets a best-effort 500 ms SIGTERM window at sandbox shutdown, servers get a real graceful exit.

website/src/content/docs/aws/compute/lambda.mdx
sam-goodwin: There is a way, and it's now implemented. The internal extension subscribes to `INVOKE` events; Lambda returns the function response to the client as soon as the runtime produces it ("even if extensions are still running"), but keeps the environment thawed until every subscribed extension signals done via `/event/next`. The entry holds that call until the invocation's queued work settles — effectively `ctx.waitUntil` for Lambda.

The handler dispatch now enqueues the request-scope close into that window instead of awaiting it before returning, so finalizers (e.g. pg pool `end`) run after the response. Queued work is raced against the invocation's `deadlineMs` minus a 500 ms margin so a hung finalizer can't become a function-timeout reset. Caveats: the window extends billed duration, and where the extension isn't active (local dev) it degrades to the old inline settle.

Implementation in `src/AWS/Lambda/RuntimeExtension.ts`; `Shutdown.test.ts` pins it live (2 s request finalizer, response asserted < 2 s, marker asserted in CloudWatch) — final live run pending an `aws sso login` (token expired).

website/src/content/docs/infrastructure-as-effects/functions-and-servers.mdx
sam-goodwin: Reworded. Init guidance is now: one-shot I/O that produces a plain cached value (fetching a secret for a client, reading config) is exactly what once-per-instance is for; what doesn't belong at init is anything *disposable* — connections, pools, streams, anything with a finalizer — plus (workerd-specific) retaining I/O-backed objects across events, since they're pinned to the creating request's IoContext. Same change applied to the AGENTS.md doctrine and the Worker JSDoc.

website/src/content/docs/aws/compute/lambda.mdx
sam-goodwin: Correction after live measurement: the extension-window approach doesn't deliver what I claimed. Two empirical findings killed it:

1. A **buffered Function URL does not release the response until the entire Invoke phase completes** — a 2 s finalizer held in the extension's `/event/next` produced a ~3 s response, same as inline.
2. Environments whose invoke phase was extended this way were **condemned and recycled**: no `END`/`REPORT` for the invocation, SIGTERM within seconds, fresh `INIT_START` on the next request.

So I reverted to `events: []` (registration still buys the verified SIGTERM shutdown window) and per-invoke cleanup settles inline. The honest position for buffered Lambda: post-handler work is response latency no matter where it runs, so request finalizers must be fast. The genuine escape hatch for latency-sensitive trailing work is **response streaming** (`invokeMode: RESPONSE_STREAM` — bytes flush as written, execution continues after `end`), which our HTTP path doesn't use yet — worth a follow-up if we want true waitUntil semantics on Lambda. Docs updated to say exactly this.

PR #778, opened 2026-07-08
fix(docker): unit-suffix Container healthcheck durations

Butch78:
## Problem

Any `Docker.Container` with a `healthcheck` fails to create. `normalizeDuration` renders durations as a bare nanosecond count:

```ts
// Docker/Container.ts
return Duration.toNanosUnsafe(duration).toString();   // "1000000000"
```

Docker parses `--health-interval` / `--health-timeout` / `--health-start-period` / `--health-start-interval` with Go's `time.ParseDuration`, which **requires a unit suffix**, so `docker container create` rejects the bare number:

```
invalid argument "1000000000" for "--health-interval" flag:
time: missing unit in duration "1000000000"
```

Repro:

```ts
Docker.Container("pg", {
  image: "postgres:17-alpine",
  healthcheck: { cmd: "pg_isready", interval: "5 seconds" },
  start: true,
});
// → deploy fails: docker container create ... missing unit in duration
```

## Fix

Append `ns` — the lossless Go-duration rendering of the nanosecond count. `Duration` keeps nanosecond precision, and Docker accepts `1000000000ns`.

```ts
return `${Duration.toNanosUnsafe(duration).toString()}ns`;
```

`normalizeDuration` is used only for the four `--health-*` duration flags, so this covers the whole surface.

## Test

Adds a docker-gated regression test (`test.provider.skipIf(!isDockerReady)`) that deploys an `nginx:alpine` container with a healthcheck and asserts the durations round-trip through `docker inspect`:

```
Config.Healthcheck = { Interval: 1000000000, Timeout: 2000000000, StartPeriod: 1000000000, Retries: 3 }
```

Before the fix the deploy throws on `container create`; after, it succeeds and the values match.

## Verification

I couldn't run the full workspace test suite in my environment (the `@distilled.cloud/*` workspace packages don't resolve without the private submodule/registry). I validated the fix directly against a real Docker daemon (29.6.1):

- `docker create --health-interval 1000000000` → `missing unit in duration` (the current bug).
- `docker create --health-interval 1000000000ns` → accepted; `docker inspect .Config.Healthcheck` returns exactly the values the new test asserts.

Found this while adopting the new `Docker` namespace (beta.61) to manage a local-dev Postgres — a `healthcheck` block made every `db:up` fail until I dropped it.

Co-authored-by: Claude <noreply@anthropic.com>

### DIFF
--- a/packages/alchemy/src/Docker/Container.ts
+++ b/packages/alchemy/src/Docker/Container.ts
@@ -418,5 +418,8 @@ const normalizeDuration = (
 ): string | undefined => {
   if (!input) return undefined;
   const duration = Duration.fromInputUnsafe(input);
-  return Duration.toNanosUnsafe(duration).toString();
+  // Docker parses `--health-*` durations with Go's `time.ParseDuration`, which
+  // requires a unit suffix — a bare nanosecond count is rejected with "missing
+  // unit in duration". `ns` is the lossless Go-duration rendering of the nanos.
+  return `${Duration.toNanosUnsafe(duration).toString()}ns`;
 };
--- a/packages/alchemy/test/Docker/Container.test.ts
+++ b/packages/alchemy/test/Docker/Container.test.ts
@@ -251,4 +251,38 @@ describe("Docker.Container", { concurrent: false }, () => {
       }),
     { timeout: 240_000 },
   );
+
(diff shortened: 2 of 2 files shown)


Mkassabov (approved):
LGTM; @john-royal will pull down and merge if they pass

PR #779, opened 2026-07-08
feat(cloudflare/website): forward a custom worker entry (main) to the Vite plugin

danielgangl:
## Problem

`@distilled.cloud/cloudflare-vite-plugin` accepts a `main` option that overrides the entry environment's entry module, but `Website.Vite` provides no way to reach it: `ViteProps` omits `main` from `WorkerProps` and the wrapper hardcodes `main: undefined`, so the deployed Worker entry is always the framework's own server bundle.

That blocks Workers that must export more than the framework's fetch handler — most commonly Durable Object classes exported from a custom module that wraps the framework handler:

```ts
// worker/index.ts
import handler from "virtual:react-router/unstable_rsc/server"; // framework handler
export { MyDurableObject } from "./my-durable-object";
export default handler;
```

For single-environment builds the `vite-do-fixture` workaround (setting `environments.ssr.build.rollupOptions.input` in `vite.config.ts`) works, but React Server Components setups (e.g. React Router RSC) can't use it: the framework plugin owns the entry environment's input.

## Change

Adds `main?: string` to `ViteOptions` and threads it through both paths:

- **deploy**: `WorkerProvider` → `viteBuild` pluginOptions
- **dev**: `LocalWorkerProvider` worker config → `viteDev` pluginOptions

`Website.Vite` forwards its (previously omitted) flat `main` prop into `vite.main`, so usage reads like wrangler:

```ts
const app = yield* Cloudflare.Website.Vite("App", {
  main: "worker/index.ts",
  viteEnvironments: { entry: "rsc", children: ["ssr"] },
});
```

The plugin already gives `pluginOptions.main` precedence over config-derived entries for the entry environment (`cloudflare-rolldown-plugin` `plugins/options.ts`), so no plugin change is needed.

## Test

New live test `Vite: main overrides the worker entry from the Vite config": deploys the DO fixture with `main: "src/worker-main.ts"` — an alternate entry that re-exports the Durable Object and answers `/api/entry` — while the fixture's `vite.config.ts` still points at `src/worker.ts`. Asserts the deployed Worker serves the alternate entry and the DO binding still works.

`bun tsc -b` and `tsconfig.test.json` typecheck clean; touched files pass `oxfmt --check`.

## Context

This is the missing piece for driving a React Router RSC app that hosts Durable Objects entirely through stock alchemy + the stock Vite plugin, without maintaining a forked plugin to inject the entry.

### DIFF
--- a/packages/alchemy/src/Cloudflare/Website/Vite.ts
+++ b/packages/alchemy/src/Cloudflare/Website/Vite.ts
@@ -90,6 +90,25 @@ export interface ViteProps<Bindings extends WorkerBindingProps = {}>
  * });
  * ```
  *
+ * @section Custom Worker Entry
+ * By default the deployed Worker entry is the server bundle the
+ * framework produces. When the Worker must export more than the
+ * framework's fetch handler — Durable Object classes, additional
+ * handlers — point `main` at your own module that wraps the framework
+ * handler and re-exports the extras. `main` takes precedence over any
+ * entry configured in the Vite config.
+ *
+ * @example Custom entry hosting Durable Objects
+ * ```typescript
+ * const app = yield* Cloudflare.Website.Vite("App", {
+ *   main: "worker/index.ts",
+ *   viteEnvironments: {
+ *     entry: "rsc",
+ *     children: ["ssr"],
+ *   },
+ * });
+ * ```
+ *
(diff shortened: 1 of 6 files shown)


agcty:
+1

PR #467, opened 2026-05-29
feat(planetscale): OAuth login + shared landing pages

Mkassabov:
Adds browser-based OAuth login for PlanetScale, mirroring the Cloudflare flow, plus the shared OAuth landing pages on the website.

### PlanetScale OAuth

`alchemy login` now offers OAuth alongside env vars and service tokens.

```
authorize → browser → localhost:9976/auth/callback → exchange → store → /auth/success
```

- `Planetscale/OAuthClient.ts` — authorize/exchange/refresh/callback over a localhost loopback
- `Planetscale/AuthProvider.ts` — `oauth` method through configure/login/resolve (proactive refresh)/logout/prettyPrint
- `Planetscale/Credentials.ts` — resolves OAuth creds to a `Bearer` token

PlanetScale differs from Cloudflare: **no PKCE**, so the app `client_secret` ships in the CLI (same exposure posture as a public `client_id`; rotate via release). Scopes are configured on the OAuth app, not requested per-authorization, and there's no revocation endpoint, so logout just drops local tokens.

Bearer support landed in `@distilled.cloud/planetscale` via alchemy-run/distilled#311 and is in the distilled version pinned on `main` — no catalog changes needed.

### Website landing pages

```
website/src/pages/auth/
├── success.astro    → /auth/success    credentials saved, return to the CLI
└── error.astro      → /auth/error      authorization failed
```

Provider-agnostic — the CLI loopback already knows which provider it started, so one shared pair covers Cloudflare, PlanetScale, and future providers. The Cloudflare flow's redirect target moves from `alchemy.run` to `v2.alchemy.run` (where this website deploys) via the shared `AUTH_SUCCESS_URL`/`AUTH_ERROR_URL` constants.

PR #782, opened 2026-07-08
docs(planetscale): cite planetscale/cli shipping its OAuth client_secret

Mkassabov:
Follow-up to #467: strengthen the JSDoc justifying the shipped OAuth `client_secret`.

- Cite that PlanetScale's own CLI ships its OAuth `client_secret` in source the same way: https://github.com/planetscale/cli/blob/main/internal/auth/authenticator.go
- Sharpen the "no PKCE flow" wording — PlanetScale's discovery doc does advertise PKCE (`code_challenge_methods_supported: ["plain","S256"]`), but the token endpoint still rejects secret-less clients (`invalid_client`), so the accurate claim is "no public-client flow".

### DIFF
--- a/packages/alchemy/src/Planetscale/OAuthClient.ts
+++ b/packages/alchemy/src/Planetscale/OAuthClient.ts
@@ -25,12 +25,17 @@ export interface Authorization {
 /**
  * Registered PlanetScale OAuth application credentials.
  *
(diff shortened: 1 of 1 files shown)


Mkassabov:
Pushed directly to main as d520ac74.

PR #786, opened 2026-07-08
fix(engine): resolve field-level Effect inputs in resolveInput

### LINKED ISSUE #784: bug: resolveInput doesn't resolve field-level Effect inputs — providers receive walked raw Effect objects
## What happened

Declaring a stage-dependent **field-level** input on a Cloudflare Worker (the `Input<T>` type explicitly allows `Effect`s per field):

```ts
export class Website extends Cloudflare.Website.Vite<Website>()("website", {
  rootDir: import.meta.dirname,
  url: Stack.useSync(stack => stack.stage !== "prod"),
  domain: Stack.useSync(stack => (stack.stage === "prod" ? "v2.example.com" : undefined)),
  // ...
}) {}
```

crashes every deploy in the custom-domain reconciler:

```
SchemaError: Expected string | undefined, got {"~effect/Effect/args":(fiber) => ...}
  at buildRequestParts … GET /accounts/{account_id}/workers/domains
```

## Root cause

`resolveInput` in `Plan.ts` (~380, `alchemy@2.0.0-beta.61`) has branches for `Output.isExpr`, `Config.isConfig`, `Duration`/`Redacted`, arrays, resources, and plain objects — but no `Effect.isEffect` branch. A raw field-level Effect falls through to the `typeof input === "object"` branch and gets **walked as a plain object** (`Object.fromEntries(Object.entries(effect))`), which is exactly the mangled `{"~effect/Effect/args": …}` value the provider then receives.

The asymmetry is subtle: **whole-props Effects work** because the resource declaration path funnels them through `asOutput` (`Output.ts:39`), which wraps an Effect into an `EffectExpr` that `resolveOutput` knows how to resolve. Field-level Effects inside a plain props object never get that wrapping, so they reach `resolveInput` raw. The type system accepts both shapes; only one is actually resolved.

Two failure modes, one worse than the other:

- **Loud:** `normalizeDomains(news.domain)` gets the mangled object → schema decode crash on every stage (the object is truthy, so even stages whose lambda resolves to `undefined` enter the domain reconciler).
- **Silent:** boolean-ish fields are only *accidentally* correct — a raw Effect is truthy, so `url: Stack.useSync(stack => stack.stage !== "prod")` enables the workers.dev subdomain on **every** stage, including prod where the lambda resolves to `false`.

## Expected

`resolveInput` resolves field-level Effects the same way the whole-props path does — e.g. an `Effect.isEffect(input)` branch that yields the Effect (mirroring the `Config.isConfig` branch) or wraps it via `asOutput` before recursing.

## Environment / workaround

- `alchemy@2.0.0-beta.61`, Cloudflare Worker via `Cloudflare.Website.Vite`
- Repro: any field-level `Stack.useSync(...)` on a prop the provider reads directly (e.g. `domain`, `url`)
- Workaround: lift the stage-dependent fields into a single whole-props `Stack.useSync`

Happy to send a PR adding the branch to `resolveInput` if you'd take it.

danielgangl:
Fixes #784.

## Problem

`resolveInput` (Plan.ts) handles `Output` exprs, `Config`, `Duration`/`Redacted`, arrays, resources, and plain objects — but a raw field-level `Effect` (which `Input<T>` explicitly allows, e.g. `domain: Stack.useSync(...)` inside a Worker's props) falls through to the plain-object walk and reaches providers as a mangled `{"~effect/Effect/args": …}` object. Whole-props Effects are unaffected because the resource declaration path wraps them into an `EffectExpr` via `asOutput`; field-level ones never get that wrapping.

Observed in the wild: the Cloudflare Worker custom-domain reconciler crashes with `SchemaError: Expected string | undefined, got {"~effect/Effect/args":(fiber) => ...}`, and boolean-ish props like `url` are silently truthy on every stage.

## Fix

An `Effect.isEffect` branch after the `Config` branch: run the effect in the plan environment (same environment `EffectExpr` thunks resolve in) and recurse on the result, so an Effect yielding an object with further Configs/Outputs resolves normally. Wrapped whole-props Effects still take the existing `Output.isExpr` path.

## Tests

New `field-level Effect inputs` suite in `test/plan.test.ts` (probe-provider pattern): field-level Effect, nested Effect inside an object, Effect resolving to `undefined`, and Effect yielding an object with further inputs — asserting the provider's `diff` sees resolved primitives. Fails without the Plan.ts change (verified by reverting), passes with it; full `test/plan.test.ts` 96 tests green; `tsc -b` clean.


sam-goodwin:
This hides the error and requirements channel of those inner effects. We'd need to extract it in the Resource effect and also make sure it doesn't negatively affect auto complete and other type driven IDE features. 

`Output<A, Req>` also has the same problem, so it's something. We do need to solve.

PR #788, opened 2026-07-09
feat(cloudflare/workers)!: workersDev + domains props; url/allUrls/domains outputs

sam-goodwin:
Replaces the Worker's overlapping `url: boolean` + `subdomain: { enabled, previewsEnabled }` + `domain` props with a single `workersDev` toggle and a `domains` list, and makes the URL outputs self-describing. Applies to `Cloudflare.Worker` and everything built on it (`Website.Vite`, `Website.StaticSite`, `RpcWorker`, ...).

### Props

```diff
 Cloudflare.Worker("Api", {
   main: "./src/api.ts",
-  url: true,
-  subdomain: { enabled: true, previewsEnabled: true },
-  domain: ["api.example.com"],
+  workersDev: true, // or { url: boolean, previews: boolean }
+  domains: ["api.example.com"],
 });
```

`workersDev: true` (the default) enables the stable workers.dev URL and version previews; `false` disables both; the object form toggles them independently. The old `subdomain` prop was hashed but never applied — `previews_enabled` was always forced to follow `enabled`. It now actually reconciles.

### Outputs

```ts
worker.url;     // always allUrls[0] — the most relevant URL
worker.allUrls; // [previewUrl?, workersDevUrl?, localhostUrl?, ...customDomains, ...devNetworkUrls]
worker.domains; // the hostnames of allUrls, aligned 1:1
```

`url` selection falls out of that ordering:

- `alchemy dev` → `http://localhost:<port>` (plus one LAN URL per interface when `dev.host` binds a wildcard, mirroring Vite's Local/Network output)
- deployed with workers.dev enabled → the stable workers.dev URL
- `workersDev: { url: false, previews: true }` → the current version's preview URL (`https://<version-prefix>-<name>.<sub>.workers.dev`, read from the latest deployment)
- `workersDev: false` + `domains` → `https://<domains[0]>`, and reordering `domains` follows

The preview URL only enters `allUrls` when it is the primary workers.dev surface (`url: false, previews: true`) — in the default config it would change on every deploy and defeat the diff's `url`-stability optimization that keeps downstream resources from spuriously re-updating.

### Notes

- Breaking: the `url`, `subdomain`, and `domain` props are removed, and the `domains` output changed from `https://` URLs to bare hostnames (`allUrls` carries the full URLs).
- All three persisted-state generations (`{ hostname }` objects ≤ beta.44, `https://` URL strings beta.45–57, bare hostnames now) are still read by the diff; the first deploy after upgrading rewrites state to the new shape. A live test writes both legacy shapes into state and redeploys over them.
- Local dev URLs are now bare origins (`http://localhost:1337`, no trailing slash), matching the deployed form.


sam-goodwin:
Doing this in a new change

PR #800, opened 2026-07-10
fix(cloudflare/workers): guard Durable Object class move across scripts

### LINKED ISSUE #799: Cloudflare: moving a DO class to another Worker fails — deleted_classes migration ships in the same upload as the new cross-script binding
## Summary

Moving a Durable Object class from one Worker to another (previously-local class becomes a cross-script `Cloudflare.DurableObject(..., { scriptName })` reference) makes the next deploy of the *former host* Worker fail: the provider emits a `deleted_classes` migration for the class **and** ships the new cross-script binding referencing the same class name in the same script upload. Cloudflare rejects that combination:

```
ERROR (#3): DurableObjectClassNotFound: Cannot apply --delete-class migration to class 'MyDOClass' without also removing the binding that references it.
```

Cloudflare requires binding removal and class deletion to happen in **two separate uploads**, and its validator appears to match bindings by `class_name` alone — a `script_name` pointing at the new (foreign) host does not exempt the binding.

## Reproduction

Starting topology — `worker-b` hosts the class locally:

```ts
// worker-b alchemy.run.ts
const editorDo = Cloudflare.DurableObject<MyRpc>("editor-do", {
  className: "MyDOClass",
})
// bound into worker-b's env; class exported from worker-b's entrypoint
```

Target topology — the class moves to `worker-a`, `worker-b` keeps a reference-only binding:

```ts
// worker-a alchemy.run.ts (new host): declares the DO locally, exports the class
// worker-b alchemy.run.ts:
const editorDo = Effect.gen(function* () {
  const stage = yield* Alchemy.Stage
  const a = yield* Output.stackRef<AStackOutput>("stack-a", { stage })
  return Cloudflare.DurableObject<MyRpc>("editor-do", {
    className: "MyDOClass",
    scriptName: Output.map(a, o => o.workerName),
  })
})
```

Deploy order `worker-a` → `worker-b` (enforced). `worker-a` deploys fine and hosts the class in a fresh namespace. `worker-b`'s deploy fails with the error above. Fresh stages (first deploy of `worker-b`) are unaffected — only stages where `worker-b` previously hosted the class.

Observed on `alchemy@2.0.0-beta.61`; the relevant code on `main` is unchanged.

## Root cause

`WorkerProvider.ts`, migration computation (~line 1371 on current `main`):

```ts
// Compute deleted classes
const deletedClasses: string[] = [];
for (const [logicalId, className] of Object.entries(oldDoClassNameByLogicalId)) {
  if (!currentDoClassNameByLogicalId[logicalId]) {
    deletedClasses.push(className);
  }
}
```

`currentDoClassNameByLogicalId` is built from `getDurableObjectBindings(bindings, name)`, which (correctly) excludes cross-script bindings from migration ownership:

```ts
if (item.scriptName !== undefined && item.scriptName !== workerName) {
  return [];
}
```

So when a logical id transitions from locally-hosted to cross-script, it vanishes from the "current" map while its `alchemy:do:` tag is still in the "old" map → the class lands in `deletedClasses`, while the binding list of the same upload still contains a `durable_object_namespace` binding with that `class_name` (now with `script_name`). Cloudflare rejects the upload, and there is no single-upload sequence it would accept.

(The backward-compat branch right below already knows cross-script bindings are "NEVER candidates for delete-class migrations" — the tagged path just has no equivalent awareness of the local→cross-script *transition*.)

## Possible fixes

> **Edit:** an earlier version of this section recommended doing the two-phase split automatically. On reflection that would be dangerous: today's rejection is accidentally a safety interlock. A user who moves a DO class cross-script and assumes the data moves with it (it does not — the new host gets a fresh, empty namespace) is loudly stopped. Silently splitting the deploy would turn an innocent-looking topology refactor into irreversible deletion of the old namespace's data. Hence the safety-first shape below.

1. **Fail loudly by default, split only on explicit opt-in.** When the conflict is detected (`deletedClasses` ∩ class names of current cross-script DO bindings), fail *before any upload* with an actionable error: the class moved to another script, Durable Object data does **not** transfer, applying the delete migration would destroy the old namespace irreversibly — set the explicit opt-in to confirm, or handle it manually. With the opt-in set, the provider performs the two-phase upload Cloudflare demands: upload №1 = bindings *minus* the conflicting cross-script bindings + `deleted_classes` migration; upload №2 = full binding set, tags reflecting the final state. This is strictly better than today's opaque CF error, and makes `{ scriptName }` adoption safe on live stages instead of only fresh ones.
2. **Skip the delete on transition** (rejected): don't emit `deleted_classes` when the class is still referenced cross-script. Avoids the error but leaks the old namespace forever and hides the data question instead of surfacing it.
3. **Longer term — `transferred_classes`:** Cloudflare has a data-*preserving* migration for exactly this move. Alchemy doesn't drive it today; we're investigating feasibility and will file a follow-up issue if it pans out.

## Workarounds we used

- Disposable stage (DO data re-derivable): delete the former host script once (`DELETE /workers/scripts/:name?force=true`); the next deploy takes the fresh-create path.
- Otherwise: manual two-phase deploy (commit removing the binding, then a commit restoring it as cross-script).

danielgangl:
Fixes #799.

## Problem

When a Durable Object class moves from one Worker to another and the former host keeps a cross-script reference (`Cloudflare.DurableObject(..., { scriptName })`), the former host's next deploy fails. The class disappears from the worker's locally-owned DO bindings (`getDurableObjectBindings` excludes cross-script bindings from migration ownership), so it lands in `deletedClasses` — while the same script upload still carries a `durable_object_namespace` binding referencing that `class_name`. Cloudflare rejects the combination:

```
DurableObjectClassNotFound: Cannot apply --delete-class migration to class 'MyDOClass' without also removing the binding that references it.
```

Cloudflare requires binding removal and class deletion in **two separate uploads**, and matches bindings by `class_name` alone — a `script_name` pointing at the new host does not exempt the binding. There is no single-upload sequence it accepts.

## Why this isn't silently auto-fixed

That rejection is effectively a safety interlock. **Durable Object storage does not move with the class** — the new host gets a fresh, empty namespace, and the old namespace on the former host still holds all its data. A delete-class migration permanently destroys it. Silently splitting the upload would turn an innocent-looking topology refactor into irreversible data loss.

## Change

- **Default — fail loudly, before any upload.** When the computed `deletedClasses` intersects the class names of current cross-script DO bindings, the deploy now fails with a new `DurableObjectClassMoved` error that states (a) the class moved to another script, (b) the data does **not** transfer and the delete is irreversible, and (c) how to proceed. This is strictly better than the opaque raw Cloudflare error users hit today.

- **Opt-in — two-phase upload.** Setting the new `deleteMovedDurableObjectClasses: true` prop on the Worker confirms the deletion. The provider then splits the deploy:
  1. **Upload #1** — the same code/modules with the conflicting cross-script DO binding(s) omitted, carrying the `deleted_classes` migration. Cloudflare accepts it (the binding is gone) and drops the old namespace and its data.
  2. **Upload #2** — the full binding set (cross-script bindings included), no migration. Upload #1 already advanced Cloudflare's migration tag to the bumped value and the metadata tags carry the same `alchemy:migration-tag:` value, so the recorded tag stays consistent.

  Non-DO bindings and local DO bindings are present in both uploads; the two-phase path only triggers when the conflict set is non-empty — every other deploy stays single-upload. A `session.note` records when the split activates.

  Note: between the two uploads there is a brief window where the moved cross-script binding is absent — in-flight requests that use it will fail during that window. This is inherent to Cloudflare's two-upload requirement; a manual two-phase deploy has the same window. It is documented on the `deleteMovedDurableObjectClasses` prop.

## Test coverage

The conflict detection is extracted into `getConflictingCrossScriptDoBindings` (exported for unit testing, matching the existing `normalizeStateDomains` seam) and covered by unit tests, alongside tests asserting the `DurableObjectClassMoved` guidance. The live local→cross-script move is only exercisable by the integration suite against real Cloudflare, so no new live test was invented here; `bun tsgo` typecheck, oxlint, oxfmt, and `packages/alchemy` `WorkerProvider.test.ts` (14 passing) all run offline and are green.

### DIFF
--- a/packages/alchemy/src/Cloudflare/Workers/Worker.ts
+++ b/packages/alchemy/src/Cloudflare/Workers/Worker.ts
@@ -435,7 +435,10 @@ export interface WorkerProps<
    * Set this to `true` to confirm the deletion: Alchemy then performs the move
    * as two uploads (Cloudflare rejects doing both at once) — first removing the
    * cross-script binding while applying the delete-class migration, then
-   * re-attaching the full binding set.
+   * re-attaching the full binding set. Between the two uploads there is a
+   * brief window where the moved binding is absent — in-flight requests that
+   * use it will fail; the same window exists when performing the move
+   * manually across two deploys.
    *
    * Only affects the transition where the class leaves this Worker; it never
    * deletes classes the Worker still hosts locally.
--- a/packages/alchemy/src/Cloudflare/Workers/Worker.ts
+++ b/packages/alchemy/src/Cloudflare/Workers/Worker.ts
@@ -420,6 +420,30 @@ export interface WorkerProps<
   limits?: WorkerLimits;
   placement?: WorkerPlacement;
   /**
+   * Opt in to deleting a Durable Object class (and its stored data) when it
+   * moves from this Worker to another script.
+   *
+   * When a Durable Object class that this Worker previously hosted becomes a
+   * cross-script reference (a `Cloudflare.DurableObject(..., { scriptName })`
+   * binding pointing at another Worker), Cloudflare requires the class to be
+   * removed from this Worker with a delete-class migration. **Durable Object
+   * storage does not move with the class** — the new host starts with a fresh,
+   * empty namespace, and the delete-class migration permanently destroys the
+   * old namespace and all of its data on this Worker. This is irreversible.
+   *
+   * Because of that, the deploy fails by default with an explanatory error.
+   * Set this to `true` to confirm the deletion: Alchemy then performs the move
+   * as two uploads (Cloudflare rejects doing both at once) — first removing the
+   * cross-script binding while applying the delete-class migration, then
+   * re-attaching the full binding set.
+   *
+   * Only affects the transition where the class leaves this Worker; it never
+   * deletes classes the Worker still hosts locally.
+   *
+   * @default false
+   */
+  deleteMovedDurableObjectClasses?: boolean;
+  /**
    * Tracks Durable Object and Workflow exports for Effect-native Workers only.
    * Populated automatically from bindings; do not set manually.
    * @internal
--- a/packages/alchemy/src/Cloudflare/Workers/WorkerProvider.ts
+++ b/packages/alchemy/src/Cloudflare/Workers/WorkerProvider.ts
@@ -50,12 +50,14 @@ export class DurableObjectClassMoved extends Data.TaggedError(
   "DurableObjectClassMoved",
 )<{
   scriptName: string;
-  classNames: string[];
+  movedClasses: { className: string; scriptName: string }[];
 }> {
   override get message(): string {
-    const classes = this.classNames.map((c) => `'${c}'`).join(", ");
+    const moves = this.movedClasses
+      .map((m) => `'${m.className}' (now hosted on '${m.scriptName}')`)
+      .join(", ");
     return [
-      `Durable Object class ${classes} moved from Worker '${this.scriptName}' to another script (now a cross-script binding with 'scriptName' set), and is being removed from '${this.scriptName}'.`,
+      `Durable Object class ${moves} moved from Worker '${this.scriptName}' to another script — the binding is now a cross-script reference — and the class is being removed from '${this.scriptName}'.`,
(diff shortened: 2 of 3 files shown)


sam-goodwin:
transfer would be better, right? Put together that here https://github.com/alchemy-run/alchemy-effect/pull/803

PR #816, opened 2026-07-13
feat(cloudflare/workflows): add per-workflow step limits

filipfalcon:
Thread an optional `limits: { steps }` through both Workflow forms — the async ref form (`Workflow(name, { limits })` in a Worker `env`) and the Effect-native form (`Workflow(name, { limits }, impl)`) — down to the Workflows definition API (`putWorkflow`). Omitting `limits` is unchanged.

```ts
// async ref form
env: {
  LIMITS_WORKFLOW: Cloudflare.Workflow("LimitsWorkflow", { limits: { steps: 100 } }),
}

// effect-native form (props second, impl third — consistent with Worker/DurableObject)
export default class LimitsWorkflow extends Cloudflare.Workflow<LimitsWorkflow>()(
  "LimitsWorkflow",
  { limits: { steps: 250 } },
  Effect.gen(function* () { ... }),
) {}
```

Verified live: the limit is applied on deploy and read back from the Workflows versions API. Both forms are covered by tests that deploy a workflow with a step limit and assert the applied limit via the versions API.

### DIFF
--- a/packages/alchemy/src/Cloudflare/Workflows/Workflow.ts
+++ b/packages/alchemy/src/Cloudflare/Workflows/Workflow.ts
@@ -288,6 +288,16 @@ export const isWorkflowExport = (value: unknown): value is WorkflowExport =>
   (value as any).kind === "workflow";
 
 /**
+ * Limits applied to the workflow on create or update.
+ */
+export interface WorkflowLimits {
+  /**
+   * Maximum number of steps a single workflow instance may execute.
+   */
+  steps?: number;
+}
+
+/**
  * Props for the reference (async) form of {@link Workflow}. Used when binding
  * a Workflow class to a plain async Worker (one without an Effect runtime) via
  * the Worker's `env`. Mirrors `DurableObjectProps`.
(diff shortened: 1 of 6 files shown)


packages/alchemy/test/Cloudflare/Workers/WorkflowAsync.test.ts
sam-goodwin, line 1: Please add an effectful test too

filipfalcon: I've added it.

packages/alchemy/src/Cloudflare/Workflows/Workflow.ts
```diff
@@ -500,6 +531,19 @@ export class WorkflowScope extends Context.Service<
...
+ * @example Setting a step limit
+ * ```typescript
+ * export default class MyWorkflow extends Cloudflare.Workflow<MyWorkflow>()(
+ *   "MyWorkflow",
+ *   Effect.gen(function* () {
+ *     return Effect.fn(function* (input: { name: string }) {
+ *       return { received: input.name };
+ *     });
+ *   }),
+ *   { limits: { steps: 25000 } },
```
sam-goodwin: I think props should be second argument and the effect third. That would be more consistent with alchemy design.

filipfalcon: Yeah, swapped

PR #817, opened 2026-07-13
test: tail in tests

john-royal:
Deploying with `Test.make({ log: true })` now streams deployed Workers' `console.*` output into the test console with sub-second latency — Cloudflare's tail API is too slow to be useful in tests.

```ts
const { test, beforeAll, afterAll, deploy, destroy } = Test.make({
  state: Cloudflare.state(),
  log: true, // patch console.* in deployed Workers + auto-tail the stack
});
```

With logging enabled, the Worker bundle's virtual entry patches `console.*` to mirror lines into an account-level `alchemy-test-logger` singleton hosting one SQLite-buffered Durable Object per `{stack}/{stage}`. `WorkerProvider.tail` streams from that DO over a websocket (buffered replay on connect + live push, deduped by row id across reconnects) when the worker's new `testLogger` attribute is set, and falls back to the Cloudflare tail API otherwise.

- `src/Tail.ts`: tail orchestration extracted from the CLI (`alchemy tail` is now a thin wrapper). `Deploy.deploy({ tail: true })` forks it after apply; the test harness ties the fiber to its shared scope, so logs flow from `beforeAll` until `destroy`/`afterAll`.
- The logger singleton is ensured lazily in `reconcile` instead of via an `enableTestLogging` layer in `Cloudflare.providers()`: memoized per account per process, content-hash versioned, and guarded by the new `Util/FileSemaphore.ts` (a lockfile-backed, key-partitioned cross-process mutex) so concurrent vitest processes don't race `putScript`.
- `Worker` gains a `testLogger` output attribute, set during reconcile and reconstructed by `read` from the observed `ALCHEMY_TEST_LOGGER` DO binding. Only set where the bundle is patchable — `script:`, `bundle: false`, and dispatch-namespace workers keep the standard tail path.

### DIFF
--- a/packages/alchemy/src/Test/Core.ts
+++ b/packages/alchemy/src/Test/Core.ts
@@ -59,8 +59,14 @@ export interface MakeOptions<ROut = any> {
    */
   dev?: boolean;
   /**
-   * Enable test logging. When `true`, logs are written to the test logger
-   * worker and can be viewed in the test harness's console.
+   * Enable test logging. When `true`:
+   *
+   * - deployed Cloudflare Workers get `console.*` patched to mirror logs
+   *   into the account's test-logger Durable Object (see
+   *   `Cloudflare/Workers/TestLoggerWorker.ts`), and
+   * - `deploy(Stack)` forks a background tail of every tailable resource in
+   *   the stack, printing logs to the test console until `destroy(Stack)`
+   *   (or the file's `afterAll`) closes the shared scope.
    */
   log?: boolean;
 }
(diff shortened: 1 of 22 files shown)


packages/alchemy/src/Test/Core.ts
```diff
@@ -112,6 +126,7 @@ export const toEffect = <A>(
       Effect.provide(Layer.succeed(ConfigProvider, configProvider)),
     );
   }).pipe(
+    Effect.provideService(TestLoggingPolicy, options.log ?? true),
     Effect.provideService(AdoptPolicy, options.adopt ?? false),
     Effect.provide(overrideAlchemyContext({ dev: resolveDev(options) })),
```
sam-goodwin: Let's remove the double Effect.provide*. Use Layer. Not your fauklt, but let's fix it while we're here

PR #818, opened 2026-07-13
chore(deps): bump rolldown 1.0.1 -> 1.0.3 to fix dual-native-binding SIGSEGV

filipfalcon:
rolldown@1.0.1's native addon segfaults (SIGSEGV, exit 139) when two physical copies are loaded into one process — the real-world trigger is any setup where two Alchemy copies coexist (monorepo, bun link shadow, version skew across workspaces). The crash is at native binding import, before any API call, and takes down alchemy dev/deploy.

Fixed upstream in 1.0.3. Bumping the catalog pin moves alchemy's bundler off the broken version; verified 1.0.3 dual-copy loads cleanly and a worker still bundles.

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -1029,7 +1029,7 @@
     "oxfmt": "^0.36.0",
     "oxlint": "^1.51.0",
     "pathe": "^2.0.3",
-    "rolldown": "1.0.1",
+    "rolldown": "1.0.3",
     "solid-js": "latest",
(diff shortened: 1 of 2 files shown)


sam-goodwin:
This broke tests, i am reverting

sam-goodwin:
We are going to update distilled to fix the bug preventing us from upgrading to rolldown 1.1

PR #830, opened 2026-07-15
feat(cloudflare/containers): keep a container warm on a schedule, warm a DO pool ahead of first request

Butch78:
### Why

`Container.start()` already eagerly boots a container the instant its Durable Object constructs — so "warm on wake" already exists. What's missing is warmth *between* real requests within a long DO session (Cloudflare can put a container back to sleep during an idle stretch, and the next request pays a full restart), and warmth *before* a DO has ever been touched at all.

### What

- `keepWarm` on `Cloudflare.Containers.layer(Container, options)`: after the initial eager start, re-runs the existing readiness check on a schedule for the life of the DO. A tick where the container is already warm costs one local state read (no Cloudflare API call, no billing impact); only an actually cold container pays the restart, off the request's critical path. This keeps *one* container warm — not a pool of spares. A container is always bound 1:1 to the DO that owns it, so there's no platform concept of a pre-warmed instance handed to a *different* DO.
- `Cloudflare.Containers.warmPool(namespace, { names, concurrency })`: a generic one-pass fan-out that wakes a caller-supplied list of DO names (this repo can't know a caller's traffic prediction) by firing a lightweight `fetch` at each stub. Waking a DO is enough on its own — any inbound call forces it to construct if it isn't already resident, which is exactly where a container's eager start already fires as a side effect.
- Deliberately no schedule on `warmPool` itself — call it from a Cron Trigger's handler (which already supplies the recurring cadence) rather than wrapping it in `Effect.repeat`. A self-repeating fan-out invoked by something that's itself repeatedly invoked would fork one forever-running loop *per fire*.

### DX

**`keepWarm`** — one field on the layer you already write. No new call sites, no fibers to manage, no teardown; the schedule bounds itself.

```diff lang="typescript"
// src/Agent.ts
import * as Cloudflare from "alchemy/Cloudflare";
+import * as Duration from "effect/Duration";
import * as Effect from "effect/Effect";
+import * as Schedule from "effect/Schedule";
import { Sandbox } from "./Sandbox.ts";

export default class Agent extends Cloudflare.DurableObject<Agent>()(
  "Agents",
  Effect.gen(function* () {
    const sandbox = yield* Sandbox;
    // ... unchanged ...
  }).pipe(
    Effect.provide(
      Cloudflare.Containers.layer(Sandbox, {
        enableInternet: true,
+       keepWarm: Schedule.spaced(Duration.minutes(4)).pipe(
+         Schedule.upTo({ duration: Duration.hours(1) }),
+       ),
      }),
    ),
  ),
) {}
```

**`warmPool`** — one call from a Cron Trigger handler. You supply the names (your own recent-activity list or prediction); it fans out and returns. No schedule to own — the trigger *is* the cadence.

```typescript
  "PoolWorker",
  { main: import.meta.url },
  Effect.gen(function* () {
    const agents = yield* Agent;

    yield* Cloudflare.Workers.cron("* * * * *", () =>
      Cloudflare.Containers.warmPool(agents, {
        names: recentlyActiveAgentIds(), // your own prediction/heuristic
        concurrency: 10,
      }),
    );

    return { fetch: Effect.succeed(HttpServerResponse.text("ok")) };
  }).pipe(Effect.provide(Cloudflare.Workers.CronEventSourceLive)),
);
```

The two compose: `warmPool` gets a DO (and its container) booting *before* anyone touches it; `keepWarm` keeps that same container warm for the rest of the DO's life once it has.

### Tests

Per @sam-goodwin's ask. `ContainerWarm.test.ts` + a `fixtures/warm` stack, modelled on `ContainerRestart.test.ts` — same fixture shape, same readiness-retry helpers, same live-deploy style. Now **run live** against a real Cloudflare account (not this repo's test account — I still don't have creds for that one):

| Test | With this PR | Feature disabled |
|---|---|---|
| `keepWarm restarts a stopped container with no requests in between` | ✓ pass (29.8s) | ✗ **fail** — `running=false, want true`, container stays down for the whole 400s poll |
| `warmPool wakes every named DO and boots its container` | ✓ pass (46.9s) | — |

**`keepWarm`**: start the container, hard-stop it, confirm it's observably down, then assert it comes back with **zero** requests touching it. `/running` only reads a flag, so the schedule is the only thing that can flip it back; a per-process boot id then proves it really restarted rather than never having gone down. The right-hand column is the point — with the `keepWarm` fork disabled, the container never comes back, so the test is testing the feature and not the platform. (`ContainerRestart`'s keepWarm-less container is the standing negative control for the same claim.)

**`warmPool`**: fans out to three DO names nothing has ever addressed. Each records its own wake in durable storage, so the pass demonstrably reached every name — not just the first, and not only what the concurrency window admitted first. The test drives `warmPool` from a fetch route rather than a Cron Trigger: the unit under test is one pass, and this keeps it deterministic instead of waiting out cron's once-a-minute floor.

### What the live run caught

Worth recording, since three real bugs only showed up under a real deploy — all in my test code, none in the feature:

1. **`crypto.randomUUID()` at module scope failed the entire Worker at startup** — this fixture module is pulled into the Worker bundle, and workerd forbids generating random values in global scope (`Disallowed operation called within global scope`). Now minted on first ask.
2. **No readiness retry on the JSON helper** — a request landing before the deploy propagated got Cloudflare's 404 HTML page and died on `Unrecognized token '<'`.
3. **The `warmPool` assertion overclaimed.** It demanded the container be *already running* at the instant the wake handler ran. It isn't: `ensureRunning` calls `container.start()`, which asks for a boot rather than waiting one out, so `running` is still false while the wake handler is on the stack. Warming ahead of the first request only promises the boot is *under way* by then — so the test now polls `running` instead. Worth being precise about in the docs too, and it's why `warmPool` pairs with a cron cadence rather than being called inline before traffic.

Also rebased onto current main and migrated to the `alchemy-test` harness, which the container tests moved to while this branch sat behind.

### Docs

Added a `keepWarm` example + bonus section to the container guides, and a new "Warm containers ahead of the first request" section covering `warmPool`.

### DIFF
--- a/packages/alchemy/src/Cloudflare/Containers/StartContainer.ts
+++ b/packages/alchemy/src/Cloudflare/Containers/StartContainer.ts
@@ -66,9 +66,55 @@ const getStartCoordination = (key: object) => {
   return coordination;
 };
 
+/**
+ * Options accepted by {@link layer} — Cloudflare's own container startup
+ * options plus Alchemy-only scheduling knobs layered on top.
+ */
+export interface ContainerLayerOptions extends ContainerStartupOptions {
+  /**
+   * Keep this Durable Object's container warm for its lifetime: after the
+   * initial eager start, periodically re-run the same readiness check that a
+   * real request would (restart the container and re-probe its default
+   * port if it has gone idle/cold) on this schedule — instead of waiting for
+   * the next real request to discover it needs a fresh boot.
+   *
+   * This keeps *one* container warm; it does not create a pool of spares.
+   * Cloudflare binds a container 1:1 to the Durable Object that owns it, so
+   * there is no such thing as a pre-warmed instance handed off to a
+   * different DO. To warm containers ahead of their first real request
+   * (across many DOs), see {@link warmPool} instead.
+   *
+   * A tick where the container is already warm costs a single local state
+   * read (no Cloudflare API round trip, no billing impact) plus a
+   * ready-port cache hit — both lock-free, so it never contends with real
+   * traffic. Only a container that has actually gone cold pays the restart
+   * + readiness cost, and it pays it on the schedule's cadence rather than
+   * on the critical path of the next real request.
+   *
+   * Typed `Schedule<unknown, void>` because only the schedule's *cadence* is
+   * used: its output is never consumed (`unknown` accepts any schedule the
+   * caller already has), and each tick is fed the ignored `void` result of
+   * the readiness check, so combinators like `Schedule.upTo` that inspect
+   * their input still compose.
+   *
+   * @example Keep warm every 4 minutes, bounded to a 1-hour session window
+   * ```typescript
+   * import * as Duration from "effect/Duration";
+   * import * as Schedule from "effect/Schedule";
+   *
+   * Cloudflare.Containers.layer(Sandbox, {
+   *   keepWarm: Schedule.spaced(Duration.minutes(4)).pipe(
+   *     Schedule.upTo({ duration: Duration.hours(1) }),
+   *   ),
+   * });
+   * ```
+   */
+  keepWarm?: Schedule.Schedule<unknown, void>;
+}
+
 export const layer = <Image extends Container.Decl.Any>(
   container: Image,
-  options?: ContainerStartupOptions,
+  options?: ContainerLayerOptions,
 ) => {
   const id = (container as any)["~alchemy/Id"] as string;
   // Provide the *started* instance under the same tag `yield* MyContainer`
@@ -84,7 +130,18 @@ export const layer = <Image extends Container.Decl.Any>(
  */
 export const startContainer = Effect.fn(function* <
   Image extends Container.Decl.Any,
->(containerEff: Image, options?: ContainerStartupOptions) {
+>(containerEff: Image, options?: ContainerLayerOptions) {
+  // `keepWarm` is an Alchemy-only scheduling knob layered on top of
+  // Cloudflare's real container startup options — split it off here so
+  // `container.start()` below only ever sees the keys Cloudflare declares,
+  // rather than relying on the runtime tolerating an unknown extra key.
+  const keepWarm = options?.keepWarm;
+  let startupOptions: ContainerStartupOptions | undefined;
+  if (options !== undefined) {
+    const { keepWarm: _stripped, ...rest } = options;
+    startupOptions = rest;
+  }
+
   const bindEff = (containerEff as any)["~alchemy/Container/Binding"] as
     | Effect.Effect<Effect.Effect<Container>, never, any>
     | undefined;
@@ -202,7 +259,7 @@ export const startContainer = Effect.fn(function* <
       // A (re)start means nothing is listening yet — clear any stale ready
       // marks from a previous run so the next request re-probes this instance.
       yield* Effect.sync(() => readyPorts.clear());
-      yield* container.start(options).pipe(
+      yield* container.start(startupOptions).pipe(
         Effect.catchDefect((defect) => Effect.fail(classifyDefect(defect, -1))),
         // A rate limit is transient — back off well clear of the per-second
         // window and try again a few times before surfacing it.
@@ -399,6 +456,30 @@ export const startContainer = Effect.fn(function* <
   if (phase === "runtime") {
     // erase the RuntimeContext color (we are applying it eagerly as an optimization only during runtime)
     yield* ensureRunning as Effect.Effect<void>;
+
+    if (keepWarm) {
+      // Proactively re-run the full readiness check (not just `ensureRunning`)
+      // on the given schedule, so a container that went idle/cold between
+      // real requests is already warm again by the time the next one
+      // arrives — port 3000 mirrors the same convention `base.fetch` below
+      // uses for the container's default RPC/fetch port. Forked into the
+      // layer's scope (the DO's runtime): the fiber outlives any single
+      // request but is interrupted with the runtime instead of leaking, and
+      // an erroring tick never fails the DO — the eager start above already
+      // covers the first run, so a transient failure here just means the
+      // next tick tries again. This loop is why the start can't simply be
+      // awaited: the initial start *is* awaited just above; the re-check
+      // repeats for the DO's whole life.
+      yield* Effect.forkScoped(
+        ensureReady(3000).pipe(
+          Effect.tapError((err) =>
+            Effect.logDebug(`keepWarm tick failed: ${err}`),
+          ),
+          Effect.ignore,
+          Effect.repeat(keepWarm),
+        ),
+      );
+    }
   }
 
   // The container exposes its non-`fetch` shape methods (declared on the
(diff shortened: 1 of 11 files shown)


sam-goodwin:
Let's expand the Container tests to test this

packages/alchemy/src/Cloudflare/Containers/StartContainer.ts
```diff
@@ -66,9 +66,49 @@ const getStartCoordination = (key: object) => {
...
+   * import * as Schedule from "effect/Schedule";
+   *
+   * Cloudflare.Containers.layer(Sandbox, {
+   *   keepWarm: Schedule.spaced(Duration.minutes(4)).pipe(
+   *     Schedule.upTo({ duration: Duration.hours(1) }),
+   *   ),
+   * });
+   * ```
+   */
+  keepWarm?: Schedule.Schedule<unknown, void>;
```
sam-goodwin, line 112: what does unknown mean here?

packages/alchemy/src/Cloudflare/Containers/StartContainer.ts
```diff
@@ -399,6 +446,26 @@ export const startContainer = Effect.fn(function* <
...
+    if (keepWarm) {
+      // Proactively re-run the full readiness check (not just `ensureRunning`)
+      // on the given schedule, so a container that went idle/cold between
+      // real requests is already warm again by the time the next one
+      // arrives — port 3000 mirrors the same convention `base.fetch` below
+      // uses for the container's default RPC/fetch port. Detached: this
+      // fiber outlives any single request and must not fail the DO if a
+      // tick errors (the eager start above already covers the first run, so
+      // a transient failure here just means the next tick tries again).
+      yield* Effect.forkDetach(
```
sam-goodwin: is forkDetached the right thing? Why can't we just start them up here and wait? Or Effect.forkScoped?

Butch78:
Done — `ContainerWarm.test.ts` + a `fixtures/warm` stack, modelled on `ContainerRestart.test.ts` (same fixture shape, same readiness-retry helpers).

**`keepWarm`** — start the container, hard-stop it, confirm it's observably down, then assert it comes back with **zero** requests touching it. `/running` only reads a flag, so the schedule is the only thing that can flip it back. A per-process boot id then proves it really restarted rather than never having gone down.

The negative control is already in the repo: `ContainerRestart`'s container carries no `keepWarm`, and its tests only pass while a stopped container *stays* stopped until the next real request. So "it came back on its own" is a property of `keepWarm`, not of the platform.

**`warmPool`** — fan out to three DO names nothing has ever addressed. Each records its own wake in durable storage, so the pass demonstrably reached every name (not just the first, and not only what the concurrency window admitted first), and `runningAtWarm` proves waking the DO had already booted its container. The test drives `warmPool` straight from a fetch route rather than a Cron Trigger: the unit under test is one pass, and this keeps it deterministic instead of waiting out cron's once-a-minute floor.

Both are live-deploy tests in this repo's existing style, so I still can't run them against the test account — the logic is the same as what I verified by hand against my own account (per the PR description). Shout if the fixtures should be folded into an existing stack instead of a new one to save a deploy.

sam-goodwin:
This one fell off my radar. Checking back in. Is this in a good status?

packages/alchemy/src/Cloudflare/Containers/StartContainer.ts
Butch78: It is the schedule's output type. `keepWarm` only uses the schedule for its cadence and never consumes what it emits, so `unknown` in the output slot accepts any schedule a caller already has (`Schedule.spaced`, `Schedule.exponential`, anything composed). The `void` input is what each tick gets fed: the (ignored) result of the readiness check, so input-inspecting combinators like `Schedule.upTo` still compose. The docstring now spells this out.

packages/alchemy/src/Cloudflare/Containers/StartContainer.ts
Butch78: Switched to `Effect.forkScoped`, agreed it is the better fit: `startContainer` runs under `Layer.effect`'s scope, so the fiber is now tied to the DO's runtime and interrupted with it, instead of living as an unowned detached fiber.

On "start them up here and wait": the initial start *is* awaited, just above (the eager `ensureRunning`). What gets forked is only the repeating re-check loop, which runs for the DO's whole life on the `keepWarm` cadence, so awaiting it would block layer construction forever. A failing tick is `Effect.ignore`d either way, so it cannot fail the DO; the next tick just retries.

Butch78:
Yes, good status now. Just pushed a refresh:

- Rebased onto current main (clean, no conflicts).
- Fixed the two things that drifted while it sat: the warm fixture now uses `Dockerfile.inline` (the `dockerfile` prop became `InlineDockerfile` on main, which is what was failing `check`), and the keepWarm example in `run-a-container.mdx` now satisfies the new diff-indent docs check.
- Took the `Effect.forkScoped` suggestion from the review thread (details there), and documented what `Schedule<unknown, void>` means on the `keepWarm` option.

`bun tsc -b` and `bun run docs:check` both pass locally on the rebased branch. Live `ContainerWarm` runs were against my own account as before; happy to re-run them if you want a fresh live pass post-rebase.

packages/alchemy/src/Cloudflare/Containers/WarmPool.ts
```diff
@@ -0,0 +1,97 @@
...
+ * Options accepted by {@link warmPool}.
+ */
+export interface WarmPoolOptions {
+  /**
+   * The Durable Object names to warm on this pass, evaluated fresh on every
+   * call. Alchemy has no way to know which DOs you expect traffic for —
+   * this is your own prediction, recent-activity list, or fixed roster.
+   */
+  names: Effect.Effect<Iterable<string>>;
+  /**
```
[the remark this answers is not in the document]
Butch78: Good catch, the type and its own examples disagreed. Widened to `Iterable<string> | Effect.Effect<Iterable<string>>` with an `Effect.isEffect` branch inside `warmPool`, so the plain-list examples typecheck as written and effectful name resolution (a KV read, a query) still works. The fixture now exercises the plain-iterable form.

packages/alchemy/src/Cloudflare/Containers/StartContainer.ts
```diff
@@ -84,7 +130,14 @@ export const layer = <Image extends Container.Decl.Any>(
  */
 export const startContainer = Effect.fn(function* <
   Image extends Container.Decl.Any,
->(containerEff: Image, options?: ContainerStartupOptions) {
+>(containerEff: Image, options?: ContainerLayerOptions) {
+  // `keepWarm` is an Alchemy-only scheduling knob layered on top of
+  // Cloudflare's real container startup options — read it here, but pass
+  // `options` through to `container.start()` unchanged below (it structurally
+  // satisfies `ContainerStartupOptions`; Cloudflare's runtime ignores the
+  // extra key rather than rejecting it).
```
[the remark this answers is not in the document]
Butch78: Agreed, done: `keepWarm` is now split off before the call, so `container.start()` only ever receives the keys Cloudflare declares (`ContainerStartupOptions` stays `undefined` when no options were given, since `enableInternet` is required on the type). The comment no longer claims the runtime tolerates unknown keys.

PR #832, opened 2026-07-15
feat(cost): generic pricing registry on ProviderService, Cloudflare as the first example

Butch78:
### Why

`alchemy plan` only ever showed a resource diff — no way to see what a deploy costs before committing to it.

### What

- `Cost.ts`: a platform-neutral `ResourceCost<Props>` — `floorMonthlyUsd` (deterministic floor, mostly `0`: nothing here has a nonzero cost at zero usage, not even Containers under scale-to-zero), `rates` (usage-based reference lines, intentionally never summed into a total since actual usage is unknowable at plan time), and `requiresPaidPlan`.
- **Plan-time Outputs, same contract as a provider `diff`** (review feedback): pricing runs during plan, so its functions receive `Input<Props>` — props that may still be, or contain, unresolved `Output`s. `Cost.ts` ships `planProp` (built on `Diff.isResolved`, the same helper `diff` handlers use) as the safe accessor; a price-determining prop that is unknown at plan time degrades to a labeled default (e.g. `Container active time (instance type unresolved at plan time — lite rates shown)`) instead of silently mis-pricing — before this, an `Output` `instanceType` indexed into the rate table would have priced the container at $0. Unresolved is distinguished from absent, so a genuine default (Cloudflare provisions `lite` when you don't set an instance type) still reads as chosen.
- `Provider.ts`: one new optional `pricing?` field on `ProviderService`. Fully backward compatible — every existing provider across every cloud keeps working unchanged; any of them can adopt the same field later with no further core changes.
- `Cloudflare/CloudflarePricing.ts`: the first real implementation, for the resources that actually have a deterministic, publishable per-unit rate — Workers, Containers, KV, R2, D1, Queues, Vectorize. Deliberately **not** all 245 Cloudflare resource types this framework implements: the other ~235 are either free, bundled into an account-level plan tier this framework has no visibility into, or seat-billed independent of resource count — pricing those would just print "$0/included in your plan" noise.
- `Cli/LoggingCli.ts`: `formatPlanLines` sums touched resources' floors (plus the $5/mo Workers Paid base fee exactly once, however many paid-plan resources are touched together) into one "Estimated minimum monthly cost" line, followed by the deduplicated rate card.

### Durable Objects are not priced here — deliberately, not an oversight

Every other resource in this PR (Workers, Containers, KV, R2, D1, Queues, Vectorize) has its own `Provider.succeed(...)`/`Provider.effect(...)` construction, which is where `pricing` attaches. `Cloudflare.DurableObject` (`Workers/DurableObject.ts`) has no independent provider of that shape — a DO is bound and deployed as part of its *hosting Worker*, not through its own create/reconcile/delete lifecycle against a dedicated API resource. There's nowhere on a DO's own `ProviderService` to attach this to. Pricing DOs needs either a different attachment point (e.g. surfaced through the hosting Worker's own `pricing` entry) or a small CLI-side special case — worth doing, but as a follow-up once this mechanism proves out, not blocking this PR.

### A generic-registry side effect

Attaching `pricing` (a plain function, not Effect/Stream-shaped) to `ProviderService` broke the dev-mode RPC provider proxy generically — `RpcSerialization.RpcWrapped<T>`'s recursive type only recognizes Effect-returning and Stream-returning handlers, so a plain function collapsed to `{}`. Fixed by stripping `pricing` before it goes over the wire in `Local/RpcServer.ts` rather than teaching the generic wrapper about a fourth shape — pricing is pure math, identical regardless of which process evaluates it, so it was never meaningful to proxy anyway.

### Verification

`bun tsc -b packages/alchemy` clean. The pricing/cost unit tests (30, including the unresolved-Output cases and the plan-summary aggregation) pass; the rest of `bun vitest run` shows the same pre-existing "Missing required env: CLOUDFLARE_ACCOUNT_ID" failures as an unmodified checkout (live-account integration tests I don't have credentials for).

Manually confirmed live against a real Cloudflare account (not this repo's own CI/test account): `alchemy plan` with a Worker + Container + KV namespace + D1 database showed a correct `$5.00` total — deduped, not doubled across the two paid-plan resources — and a clean rate card:

```
Estimated minimum monthly cost: $5.00
  includes the $5.00/mo Workers Paid plan base fee
  usage-based rates below are not included above — actual usage is unknown at plan time
  Cloudflare.Worker
    Workers requests: $0.3 / million requests (10M/mo free)
    Workers CPU time: $0.02 / million ms (30M ms/mo free)
  Cloudflare.D1Database
    D1 rows read: $0.001 / million rows (25B/mo free)
    D1 rows written: $1 / million rows (50M/mo free)
    D1 storage: $0.75 / GB-month (5 GB free)
  Cloudflare.KV.Namespace
    KV reads: $0.5 / million reads (10M/mo free)
    KV writes: $5 / million writes (1M/mo free)
    KV deletes: $5 / million deletes (1M/mo free)
    KV storage: $0.5 / GB-month (1 GB free)
  Cloudflare.Container
    Container active time (lite): $0.002754 / active hour (memory + disk baseline) (25 GiB-hr memory / 200 GB-hr disk /mo free)
    Container vCPU: $0.072 / vCPU-hour actually used (375 vCPU-min/mo free)
```

Didn't add a live integration test for the same reason as #830 — every existing Container/provider test in this repo deploys to a real account I don't have credentials for. Marking this **draft** for that reason.

~~Stacked on #830~~ **Now rebased directly onto `main`** (review feedback): this PR and #830 are fully independent — this one is cost estimation only, #830 is warm containers only.


packages/alchemy/src/Cloudflare/Containers/Container.ts
```diff
@@ -364,6 +364,33 @@ export type Container<Id extends string = string> = Named<Id> & {
  * ) {}
  * ```
  *
+ * @example Keeping the container warm between requests
```
sam-goodwin: Does this belong in this PR or in a dedicated PR for warm containers?

Butch78: I thought it could be useful for everything. Containers are probably the most expensive things on CloudFlare so I thought I would start with that example first. 

So it can be seperate, you don't need to know the cost to implement a container

sam-goodwin: Stack the PRs. One for warm containers and one to support cost estimation in the provider.

packages/alchemy/src/Provider.ts
```diff
@@ -110,6 +111,13 @@ export interface ProviderService<
    * provider via {@link tryFindProviderByType}.
    */
   aliases?: readonly string[];
+  /**
+   * Cost model for this resource, shown in `alchemy plan`'s cost summary.
+   * Optional and platform-neutral — any provider (Cloudflare, AWS, GCP, ...)
+   * can attach one; omit it entirely for resources with no meaningful
+   * per-unit billing dimension (the common case).
+   */
+  pricing?: ResourceCost<Props<Res>>;
```
sam-goodwin, line 127: Nice I like how it takes the props   

Look at how diff is defined and works. If it's called during plan then it has to handle Outputs and if a value isn't known it won't be able to estimate. 

I wonder if we should also call this after deployment to show a cost estimation of the deployed stack too?

packages/alchemy/src/Cost.ts
```diff
@@ -0,0 +1,47 @@
...
+   * Deterministic floor: what this resource costs per month with zero
+   * usage. Mostly `0` — a provisioned-capacity resource (e.g. a Cloudflare
+   * Container's memory/disk baseline) is the main exception.
+   */
+  floorMonthlyUsd: (props: Props) => number;
+  /**
+   * Per-unit rates for the usage-based dimensions of this resource —
+   * reference only, intentionally never summed into a total.
+   */
+  rates: (props: Props) => RateLine[];
```
sam-goodwin: These will have to take care like diff does because it runs during plan. There may be Outputs in the props still. Look at how diff is defined and works.

packages/alchemy/src/Cloudflare/Containers/Container.ts
Butch78: Done. Rebased this branch directly onto `main`, so the warm-container commit (docs example included) is out of this PR entirely; it lives only in #830 now. It was showing here because the branch was cut from #830's commit and GitHub can't base a cross-fork PR on another fork branch, so both diffs rendered. The two PRs are now fully independent: this one is cost estimation only, #830 is warm containers only.

packages/alchemy/src/Provider.ts
Butch78: Good catch, and you're right that I'd missed it: `diff` deliberately takes `news: Input<Props>` for exactly this reason, and `pricing` was typed against plain resolved `Props`, which lies about plan time. Concretely, an `instanceType` wired to another resource's output would have been indexed into the rate table as an `Output` object and silently priced the container at $0.

Fixed by adopting diff's contract wholesale: `ResourceCost`'s functions now receive `Input<Props> | undefined`, and `Cost.ts` ships `planProp` (built on the same `Diff.isResolved` the diff handlers use) as the safe accessor for any price-determining prop. When a value isn't known at plan time it can't be estimated, so it degrades to a labeled default, e.g. `Container active time (instance type unresolved at plan time — lite rates shown)`; unresolved is distinguished from absent so a genuine default (Cloudflare provisions lite when you don't set an instance type) still reads as chosen. Unit tests cover Output-valued props, whole-props-as-Output, Effect-valued props, and nested unresolved leaves.

On calling it after deployment: agreed, and it's arguably the better home. Post-deploy every prop is fully resolved, so the estimate gets strictly more accurate (no unresolved fallbacks), and the same `formatCostSummaryLines` aggregation can be reused over the applied resources. I'd like to do that as a follow-up PR rather than grow this one, in the same spirit as splitting off #830.

packages/alchemy/src/Cost.ts
Butch78: Addressed (details in the sibling thread on `Provider.ts`): these signatures now take `Input<Props> | undefined` exactly like `diff`'s `news`, the docstring states the plan-time rule, and `planProp` (on top of `Diff.isResolved`) is the required accessor for any prop that changes the price. The three implementations that read props (R2 `storageClass`, Vectorize `dimensions`, Container `instanceType`) all go through it and fall back to a labeled default when the value is unresolved, with tests for each.

sam-goodwin:
Merged main back into this PR. Really like this feature. Should merge if it's safe to do so

sam-goodwin:
Should this be gated behind a flag to opt-in to see it as part of the plan? Since it adds a lot of extra information.

also, how do we plan to keep this up to date. Do we need to watch for pricing changes and propagate? What about regional cost differences?

sam-goodwin:
Pushed two commits to this branch:

### 2ffeda3b7 — merge `main`, resolve conflicts

Main's local-provider/mode machinery had landed since this PR was opened. Resolution:

- `Provider.ts`: `ProviderService` keeps both the new `pricing?` field and main's `mode`/`modes` fields.
- `Local/RpcServer.ts`: the pricing strip is folded into main's `RpcProviderService`:

```ts
export type RpcProviderService<R extends ResourceLike> = Omit<
  ProviderService<R>,
  "mode" | "modes" | "pricing"
>;
```

- Cloudflare D1/KV/R2/Queues/Workers: kept both the pricing imports and main's `LocalRuntime` imports; Queue takes main's `stables: ["queueId", "accountId"]`.
- The three new test files migrated from `@effect/vitest` to `alchemy-test` (main's runner) via `scripts/codemod-alchemy-test.ts`.

Note: with `ProviderLayer.dual`, `pricing` is attached to the live provider variants only, so dev-mode plans show no cost for locally-emulated resources.

### da9b0b57e — pricing for 21 more resources

Audited all 236 Cloudflare provider files and extended the registry beyond the original 7 to every resource with a published, deterministic, resource-attributable rate. Each priced service gets its own `Cloudflare/{Service}/Pricing.ts` (source URLs cited, rates verified against the published pricing pages 2026-08) plus a pure unit-test file — 172 tests total.

- Usage-metered: Workflows, R2 Data Catalog, Stream, Images, Calls (App + TurnKey), RealtimeKit, AI Gateway logs, Cache Reserve, Zaraz, Page Shield, Workers Logpush, Pipelines (Pipeline + Sink), Email sends, Pages Functions, Custom Hostname ($0.10/hostname-mo after 100 free)
- Deterministic monthly floors: Workers for Platforms DispatchNamespace ($25/mo), ACM CertificatePack ($10/mo), Argo Smart Routing ($5/mo + $0.10/GB), Load Balancer ($5/mo)

The other ~215 provider files are deliberately left unpriced (free / plan-bundled / seat-billed / enterprise-contract / at-cost passthrough), per this PR's no-$0-noise doctrine — absence of a `pricing` field is the representation.

Every implemented rate was independently re-verified against the cited pricing pages; two issues were caught and fixed (Argo's missing 1 GB/mo free allotment + unit wording, Custom Hostname misclassifying `customOriginServer` as Enterprise-only).

Caveat, documented in-file: DispatchNamespace's $25 and ACM's $10 are account/zone-scoped subscription fees expressed as per-resource floors, so two such resources in one plan double-count until `ResourceCost` grows an account-level dedupe hook like `requiresPaidPlan`'s.

`bun tsc -b` clean; all pricing tests green.

Butch78:
I think it will align well with this:
https://blog.cloudflare.com/billable-usage-api/

I hope they do Enterprise billing soon, I have no idea how much I'm spending or if I'll be charged haha

PR #834, opened 2026-07-15
fix(cloudflare): hash workspace dependency closure for Vite worker diffs

### LINKED ISSUE #744: Vite worker diff hashes only rootDir — workspace-package changes silently skip deploy
## Summary

For Vite-based Workers, `hasChanged` decides staleness by hashing the Vite root directory:

```ts
// packages/alchemy/src/Cloudflare/Workers/WorkerProvider.ts
if (props.vite) {
  const input = yield* hashDirectory({
    cwd: props.vite.rootDir,
    memo: props.vite.memo,
  });
  return input !== output.hash?.input;
}
```

`hashDirectory` matches `**/*` under `rootDir` with gitignore exclusions (so `node_modules` is excluded as well). In a workspace monorepo (bun/pnpm/yarn workspaces), the worker bundle typically contains many workspace packages that live **outside** `rootDir`. A change that only touches a workspace dependency produces an identical input hash → plan says `noop` → deploy exits 0 having uploaded nothing.

This is a nasty failure mode because nothing errors: you ship a fix in a shared package, the deploy reports success, and production still runs the old code.

## Repro sketch

1. Monorepo: `apps/web` (Vite worker, `rootDir: apps/web`) importing `packages/lib` via `workspace:*`.
2. Deploy once.
3. Change only `packages/lib/src/*`.
4. Deploy again → `noop`, worker not updated, even though the bundle output differs.

`--force` works as a workaround (promotes noop → update).

## Why the current design can't see it

The input-hash-instead-of-bundle-hash approach is a deliberate optimization (avoids building at plan time — the comment near the reconcile path explains the history of scoping it away from `process.cwd()`). But `rootDir` is the wrong closure: the true input set is `rootDir` **plus the workspace dependency closure**. `memo` today can only narrow the set further, never widen it beyond `rootDir`.

## Suggested fix

Resolve the app's workspace dependencies recursively — start from `rootDir/package.json`, follow `workspace:` ranges through the workspace root's `workspaces`/`pnpm-workspace.yaml` globs — and fold those package directories (their `src/`, or their built `dist/` which changes iff outputs change) into the hashed file set. Keeps plan-time cheap, needs no per-app configuration, and fixes every monorepo user at once.

An alternative is hashing the Vite build output instead of inputs (truthful by construction), at the cost of requiring the build at diff time.

We run a fork of alchemy in a large Bun-workspace monorepo, got bitten by this in production this week, and are planning to implement the workspace-closure fix — happy to adapt it into a PR here if the direction sounds right to you.

agcty:
Fixes #744.

## Problem

The Vite worker's rebuild diff hashes only the files under `props.vite.rootDir`. In a monorepo, the app at `rootDir` imports workspace packages that live outside it — changing only those packages produces a "noop" plan and the deploy silently ships nothing (`Done: 0 succeeded`, exit 0, worker unchanged; this bit us in production). The `memo.include` workaround can't actually cover it either: those globs are relative to `rootDir` and can never reach packages outside it.

## Fix

`hashViteBuildInputs({ cwd, memo })` replaces `hashDirectory` at both Vite hash sites (apply + `hasChanged`). It hashes `rootDir` exactly as before (still scoped by `memo`) and additionally folds in the app's **workspace-dependency closure**:

- Walks up from `rootDir` to the nearest ancestor declaring a workspace (bun/npm/yarn `workspaces` — array or `{ packages }` — or `pnpm-workspace.yaml`), expands the globs once, and reads each `package.json` once.
- Resolves the closure from `rootDir`'s manifest: the root follows `dependencies` + `devDependencies` + `optionalDependencies` + `peerDependencies` (Vite plugins and shared config live in devDeps); transitive steps follow runtime edges only, so the closure stays scoped to what the app bundles rather than ballooning to the whole monorepo. Cycle-safe; matched by package *name* against the workspace's own package map, so `workspace:*` vs hoisted-semver specifiers don't matter.
- Each closure package directory is hashed with the same gitignore-respecting semantics as `rootDir` (`lockfile: false` per package — the root hash already folds in the shared lockfile), keyed by package **name** so the stored `hash.input` stays machine-independent.

With no workspace dependencies the result is byte-identical to the previous `hashDirectory(rootDir)` — non-monorepo workers see zero hash churn. Monorepo workers take one expected redeploy on upgrade.

## Measured on the monorepo that hit the bug

An app depending on **55 workspace packages**: closure resolution 15ms; input hash 137ms → 391ms (+254ms per plan). Both hash sites had this same shape before (the patch widens each call, it doesn't add one); if that cost ever matters, the closure hash is a natural candidate for `Artifacts.cached`-style memoization within a plan run — happy to follow up.

## Deliberate choices (flagging for your call)

- **Closure is always unioned; `memo` scopes only the `rootDir` portion.** Over-inclusion errs toward a spurious rebuild, never a missed deploy — the right bias for a rebuild detector. An opt-out could be added if someone needs it.
- **A malformed sibling `package.json` drops out of the closure instead of failing the plan** (matches the file's existing NotFound-tolerant convention; non-NotFound IO errors still propagate). Say the word if you'd rather fail loud.
- Scoped strictly to the two `WorkerProvider` Vite sites the issue names; non-Vite workers untouched.

Tests: 6 new unit tests in `test/Command/Memo.test.ts` (closure resolution incl. cycles and self-exclusion, hash sensitivity to transitive dep changes, insensitivity to unrelated packages, `memo.include` still respected, empty-closure identity with `hashDirectory`); `WorkerProvider.test.ts` 15/15 green; `tsc -b` clean.

### DIFF
--- a/packages/alchemy/src/Cloudflare/Workers/Worker.ts
+++ b/packages/alchemy/src/Cloudflare/Workers/Worker.ts
@@ -579,9 +579,15 @@ export interface ViteOptions {
    */
   rootDir?: string;
   /**
-   * Controls which files are hashed to decide whether a rebuild is needed.
-   * By default every non-gitignored file in `cwd` is hashed, plus the nearest
-   * lockfile. Provide explicit globs to narrow the scope.
+   * Controls which files under `rootDir` are hashed to decide whether a
+   * rebuild is needed. By default every non-gitignored file in `rootDir` is
+   * hashed, plus the nearest lockfile. Provide explicit globs to narrow the
+   * scope.
+   *
+   * The workspace packages `rootDir` depends on are always hashed in addition
+   * to this — changing an imported monorepo package triggers a rebuild even
+   * when nothing under `rootDir` changed. These options only scope the
+   * `rootDir` portion of the hash.
    *
    * @see {@link MemoOptions}
    */
--- a/packages/alchemy/src/Cloudflare/Workers/WorkerProvider.ts
+++ b/packages/alchemy/src/Cloudflare/Workers/WorkerProvider.ts
@@ -12,7 +12,7 @@ import * as crypto from "node:crypto";
 import { Unowned } from "../../AdoptPolicy.ts";
 import * as Artifacts from "../../Artifacts.ts";
 import type { ScopedPlanStatusSession } from "../../Cli/Cli.ts";
-import { hashDirectory } from "../../Command/Memo.ts";
+import { hashViteBuildInputs } from "../../Command/Memo.ts";
 import { isResolved } from "../../Diff.ts";
 import * as ProviderLayer from "../../Local/ProviderLayer.ts";
 import * as Provider from "../../Provider.ts";
@@ -1118,14 +1118,13 @@ export const LiveWorkerProvider = () =>
             const [{ assets, bundle }, input] = yield* Effect.all(
               [
                 viteBuild(props),
-                // hashDirectory expects `{ cwd, memo }`. The vite props
-                // store the project root under `rootDir`, so map it
-                // here. Without this, `cwd` falls back to
-                // `process.cwd()` and the input hash is computed over
-                // the wrong directory tree (often the entire monorepo
-                // root), making it both slow and unable to detect
-                // changes scoped to the actual Vite project.
-                hashDirectory({
+                // Hashes the Vite root (`rootDir`) plus the workspace
+                // packages it depends on. Without `rootDir`, `cwd` would
+                // fall back to `process.cwd()` and hash the wrong tree
+                // (often the whole monorepo); without the workspace
+                // closure, changes to imported packages outside `rootDir`
+                // would look like a no-op and silently skip the deploy.
+                hashViteBuildInputs({
                   cwd: props.vite.rootDir,
                   memo: props.vite.memo,
                 }),
@@ -1732,7 +1731,7 @@ export const LiveWorkerProvider = () =>
           return assetsHash !== output.hash?.assets;
         }
         if (props.vite) {
-          const input = yield* hashDirectory({
+          const input = yield* hashViteBuildInputs({
             cwd: props.vite.rootDir,
             memo: props.vite.memo,
           });
--- a/packages/alchemy/src/Command/Memo.ts
+++ b/packages/alchemy/src/Command/Memo.ts
@@ -3,6 +3,7 @@ import * as FileSystem from "effect/FileSystem";
 import * as Path from "effect/Path";
 import type { PlatformError } from "effect/PlatformError";
 import fg from "fast-glob";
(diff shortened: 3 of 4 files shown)


sam-goodwin:
@john-royal please take a look at this

sam-goodwin:
Is this related to https://github.com/alchemy-run/alchemy/pull/822

agcty:
Yes — same bug, different discovery mechanism. alchemy-run/alchemy#822 discovers the workspace directories from the build module graph (`writeBundle` → module IDs that resolve outside the app root → their nearest `package.json`), then hashes those directories. This PR resolves the closure from package manifests (workspace declaration + dependency edges) instead.

Trade-offs as I see them:

- **Graph-based (#822)** is precise — it only hashes packages the bundle actually pulls in — but the graph only exists *after* a build, while the memo hash has to answer "should we rebuild?" *before* one. That means persisting the discovered workspace list in state and trusting the previous build's graph (a newly added cross-package import isn't in the persisted list until after the rebuild it should have triggered — though in practice adding the import also edits a file under `rootDir`, which busts the hash anyway).
- **Manifest-based (this PR)** is available statically with no state to persist and no build required; the cost is over-inclusion (a declared-but-unbundled workspace dep still triggers a rebuild). Over-inclusion errs toward a spurious rebuild rather than a missed deploy, which felt like the right bias for a diff detector.

The hashing side is basically the same idea in both (per-directory gitignore-aware hash + lockfile). If #822's direction wins for both repos I'm happy to rework this to match — the two approaches could also compose (manifest closure as the baseline, graph pruning as a later refinement).

john-royal:
@agcty Can you try the preview package on 822 and let me know if that works for you? Was going to merge that today.

For Vite specifically, I'm leaning towards 822 because:
- It's more surgical, reducing unnecessary rebuilds as well as overhead from hashing (e.g. if you have, say, a shared package for lint configs, that doesn't need to trigger a rebuild)
- It doesn't need to re-implement any package resolution logic because it can piggyback on what Vite already does
- You're right that it requires a Vite build to identify dependencies, but you're also right that anything that involves pulling in a new workspace package dependency would necessarily mean changes in your Vite root directory
- Persisting additional workspace paths is a low-cost tradeoff for only hashing what's necessary IMO

I think the `package.json`-based approach could make sense for `Command` providers because we don't get to peer inside the module graph in the same way. If you're interested in reworking this for `Command` specifically, I'd be happy to merge that - the only thing I'd consider a blocker is that this current implementation appears to be coupled to pnpm specifically by reading the `pnpm-workspace.yaml`; Alchemy needs to support all package managers.

agcty:
Withdrawing — superseded by #822, which shipped the workspace-package input-hash coverage in 2.0.0-beta.63 (verified the `externalWorkspaces` collection + workspace-directory hashing are in the published package). Adopting beta.63 downstream now.

PR #844, opened 2026-07-16
feat(vite-dev): route vite's logger through Effect and surface the startup banner

Cyberistic:
Cloudflare.Website.Vite dev mode was silent because vite's programmatic API doesn't print the banner, Local/Network URLs, or help shortcut on its own — the `vite dev` CLI does that explicitly after `server.listen()` resolves. This mirrors that here so the apply session shows the same output v1 did.

**What this PR does**

- Passes a `customLogger` to `vite.createServer` that funnels every vite `info` / `warn` / `error` / `warnOnce` call through Alchemy's Effect logger via `Effect.runSync(Effect.logInfo|logWarning|logError)`. Transform errors, plugin warnings, resolve failures, and the SSR `vite:transform` failures now land in the terminal as `[ERROR] Error [ERR_UNSUPPORTED_ESM_URL_SCHEME]: …` instead of being swallowed by Vite's default `console.*` writer.
- Wraps `createServer` / `listen` failures in a typed `ViteDevError` (`phase: "createServer" | "listen"`, preserves the underlying defect) and threads it through `LocalWorkerProvider`'s fiber error union so the engine can render a clean hint instead of crashing.
- If `listen()` fails after `createServer()` succeeded, eagerly closes the half-initialized dev server so the engine can retry without holding the port.

**Verification**

In tmux with the example running:

```
[22:44:36.034] INFO (#93): VITE v8.1.5  ready in 882 ms

✓ Site (Cloudflare.Worker) updated
{
  url: "http://localhost:1337/",
}
```

When SSR throws (the cloudflare:workers / `__CLOUDFLARE_MODULE__CompiledWasm__` paths), the actual error message now appears as a structured Effect log instead of just an opaque `{"status":500,"unhandled":true,"message":"HTTPError"}` JSON.

### DIFF
--- a/packages/alchemy/src/Cloudflare/Workers/Vite.ts
+++ b/packages/alchemy/src/Cloudflare/Workers/Vite.ts
@@ -2,6 +2,7 @@ import cloudflare, {
   type CloudflareVitePluginOptions,
 } from "@distilled.cloud/cloudflare-vite-plugin";
 import * as ConsoleService from "effect/Console";
+import * as Duration from "effect/Duration";
 import * as Effect from "effect/Effect";
 import * as Redacted from "effect/Redacted";
 import { createRequire } from "node:module";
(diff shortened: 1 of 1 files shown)


sam-goodwin:
Can you share a screenshot of what this looks like ?

Cyberistic:
> Can you share a screenshot of what this looks like ?

image

Not the prettiest, but this info is critical


for deploy:
image

image

sam-goodwin:
Are all those warnings alchemy's fault or something on your end?

Cc @john-royal

workspaces/fix-vite-ssr-bundler/packages/cloudflare-rolldown-plugin/src/plugins/additional-modules.ts
sam-goodwin, line 1: why all this?

Cyberistic: accident hold on

packages/alchemy/src/Cloudflare/Workers/Vite.ts
```diff
@@ -10,29 +11,34 @@ import { pathToFileURL } from "node:url";
 import type * as vite from "vite";
 import { viteBuildOutputPlugin } from "../../Bundle/Vite.ts";
 
+const VITE_TAG = "[vite] ";
```
john-royal, line 14: Should the tag be the resource ID instead of just `[vite]`?

packages/alchemy/src/Cloudflare/Workers/Vite.ts
john-royal, line 1: Right now, we take the first URL returned by Vite, proxy to it (so it's stable between reloads of your `alchemy.run.ts` file), and the proxy URL is returned on the Worker/Vite resource as `worker.url`. Should we remove that since the URL is being logged here?

packages/alchemy/src/Cloudflare/Workers/Vite.ts
Cyberistic: I think if you’re going to do that, then better abstract away the logging and do it for all services? otherwise why vite only.

packages/alchemy/src/Cloudflare/Workers/Vite.ts
Cyberistic: I don’t know, just matched default vite. Logging url is optional, vite banner is not.

sam-goodwin:
Is there a reason why `return { url: website.url }` doesn't work here? Why do we need the banner? All the same information is on the worker

Cyberistic:
> Is there a reason why `return { url: website.url }` doesn't work here? Why do we need the banner? All the same information is on the worker

Just matching what vite _would_ show you (along with the kb shortcuts), no reason other than that. My initial issue with logs has already been solved by another PR.

sam-goodwin:
I'm not sure we should try and show the vite banner. Although I could see it being important in a TUI where we show the logs of all external processes. 

If you feel strongly that we should show this, please re-open and I'll re-consider.

PR #851, opened 2026-07-17
feat(cloudflare/workers): place a Durable Object in a region with locationHint (+ implement get/idFromName)

Butch78:
### Why

A Durable Object is created wherever its **first-ever** request came from, and it lives there for good. That's the right default when the name tracks the caller — `getByName(userId)` lands next to that user by construction.

It's the wrong default the moment the name stops tracking geography. A pool sharded by hash (`pool:7`), a fixed roster, a shard index — any of these gets created by whichever request happened to arrive first, and every later caller pays for that accident. A Sydney user hashed onto an instance a Frankfurt request created is making a cross-planet round trip on every call, forever. (Not hypothetical: ~300-600ms of a measured 0.3-1.3s warm path on a hash-sharded pool of mine, which is what sent me looking for the knob.)

Cloudflare's `getByName` takes an options bag whose `locationHint` fixes exactly this. Alchemy dropped it on the floor:

```typescript
getByName: (name: string) => DurableObjectStub<Shape>;
//                        ^ Cloudflare's `options` never reaches the runtime
```

And the other way in — `get(id, options)` — was **declared on the interface but never implemented**, so it wasn't a fallback: it was a `TypeError` waiting to happen (see below). Between them there was no way to reach the hint at all.

### What

```typescript
const shard = shards.getByName("pool:apac:3", { locationHint: "apac" });
```

- threads `DurableObjectGetDurableObjectOptions` through `getByName` on the DO namespace and its rpc twin (`bindEffectRpc` / the typed `RpcDurableObject`)
- exports `DurableObjectLocationHint` so callers can type their own region mapping without importing `@cloudflare/workers-types`
- **fixes `get` / `idFromName` / `idFromString` / `newUniqueId`** — see below

Hints only apply at *creation*, so this is safe to add to a live call site: nothing you pass can move an instance that already exists.

### The `get` bug

The namespace interface declares `get`, `idFromName`, `idFromString`, `newUniqueId` and `jurisdiction`. The implementation returns an object literal carrying only `getByName` — [the rest sit commented out right below it](https://github.com/alchemy-run/alchemy/blob/main/packages/alchemy/src/Cloudflare/Workers/DurableObject.ts). TypeScript says `namespace.get(id)` exists; at runtime it's `undefined`. The interface lies and the call site finds out in production.

I fixed it here rather than in a follow-up because `get(id, { locationHint })` is the path **Cloudflare's own docs use** to pass a hint — threading the options bag through `getByName` alone would have left the documented route still broken. The id constructors come along because they're the only way to obtain an `id` to hand `get`. All four are plain pass-throughs.

`jurisdiction` stays commented: it returns a whole namespace rather than a stub (needs the object literal factored out first), and it's about data residency rather than placement. Happy to do it as a follow-up if you want it.

### Verification

Live against a real Cloudflare account (not this repo's test account — still no creds for that one), reusing the already-deployed `do-rpc` stack, no new stack. Each DO reports its own colo via a subrequest trace.

| Test | With this PR | Reverted to old behaviour |
|---|---|---|
| `unhinted instances are created in the caller's colo` (control) | ✓ pass | ✓ pass — baseline holds either way |
| `locationHint places new instances in different regions` | ✓ pass | ✗ **fail** — `AssertionError: NOT: expected "ARN" to be "ARN"`, both created in the caller's colo |
| `locationHint applies through get(idFromName(name), options)` | ✓ pass | ✗ **fail** — `Worker not ready: 500`, no such method at runtime |

The middle row is the point: mutating `getByName` back to dropping `options` fails the hint test while the **control keeps passing**, so the failure is the hint and not noise in edge routing. Without that control, "two colos differ" would prove nothing — it could just be this box's two requests landing on different edges, which is why I added it.

The bottom row was written test-first: it 500s against the deployed worker before the `get` fix, passes after.

Assertions are "different colos", never an exact colo — a hint is best-effort, and Cloudflare places the instance in the nearest location it *can* to the hinted region.

### Docs

A "Place an instance in a region" section on the Durable Objects guide plus a matching `@section` in the JSDoc. Both lead with *when the default is wrong* rather than the API surface, and both flag the trap:

> Hint the region you name. If two regions can address one name, they race to create the single instance behind it and the loser is stuck talking across the planet — the hint cannot fix it afterwards.

Also noted that `alchemy dev` accepts and ignores hints (one location to place anything in), so placement is only observable once deployed.

### One thing worth your call: the `workers-types` pin

The catalog range is `^4.20250805.0` and resolves to `4.20260616.1`, which predates the `apac-ne` / `apac-se` hints (they land in `4.20260623.1`; latest 4.x is `4.20260702.1`, and `5.x` is now `latest` on npm). So this PR documents the **nine** hints the pinned version accepts, not the eleven Cloudflare currently offers — typing the fixture's hint list against `DurableObjectLocationHint` is what surfaced that.

I've deliberately left the pin alone: refreshing it re-types every Cloudflare surface in the repo, which isn't this PR's business and isn't my call. Say the word if you'd like a separate PR to move it (4.x lock refresh, or the 5.x major).

### DIFF
--- a/packages/alchemy/src/Cloudflare/Workers/DurableObject.ts
+++ b/packages/alchemy/src/Cloudflare/Workers/DurableObject.ts
@@ -48,6 +48,12 @@ export type DurableObjectId = cf.DurableObjectId;
 export type DurableObjectJurisdiction = cf.DurableObjectJurisdiction;
 export type DurableObjectGetDurableObjectOptions =
   cf.DurableObjectNamespaceGetDurableObjectOptions;
+/**
+ * The regions a Durable Object can be *hinted* toward at creation —
+ * `"wnam"`, `"enam"`, `"sam"`, `"weur"`, `"eeur"`, `"apac"`, `"oc"`, … See
+ * the "Placing a Durable Object in a Region" section on {@link DurableObject}.
+ */
+export type DurableObjectLocationHint = cf.DurableObjectLocationHint;
 
 export type AlarmInvocationInfo = cf.AlarmInvocationInfo;
 
@@ -78,7 +84,10 @@ export interface DurableObject<
   Type: TypeId;
   name: string;
   namespaceId: Output.Output<string>;
-  getByName: (name: string) => DurableObjectStub<Shape>;
+  getByName: (
+    name: string,
+    options?: DurableObjectGetDurableObjectOptions,
+  ) => DurableObjectStub<Shape>;
   newUniqueId: () => DurableObjectId;
   idFromName: (name: string) => DurableObjectId;
   idFromString: (id: string) => DurableObjectId;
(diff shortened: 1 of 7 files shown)


Butch78:
Strengthened the verification, and ran it live.

**The test had a hole.** `expect(wnam).not.toBe(apac)` only means something if two *unhinted* instances would have landed in the same colo. That assumption was unstated and untested — if this box's requests ever got routed to two different edges, the test would have passed even with the hint dropped on the floor. So it added a control:

```
✓ unhinted instances are created in the caller's colo
✓ locationHint places new instances in different regions
```

The first pins the baseline (two unhinted instances, created by two requests from the same test, must report the same colo). With that pinned, the hinted pair can only come apart if the hint reached the runtime.

**Verified live** against a real Cloudflare account (not this repo's test account — still no creds for that one). Both pass. And mutating `getByName` back to the old behaviour — dropping `options` before it reaches the runtime — fails the hint test while the control keeps passing:

```
✓ unhinted instances are created in the caller's colo
✗ locationHint places new instances in different regions
  AssertionError: NOT: expected "ARN" to be "ARN"
```

Both instances created in `ARN`, the colo this box's requests land in — exactly the bug, and the test fails closed on it.

Two other things fell out of this:

- the `/colo` fixture no longer casts the hint query param (or non-null-asserts the name); it matches against the hints Cloudflare accepts, so a typo reads as "no hint" rather than reaching the runtime.
- that list caught `apac-ne` / `apac-se` in my docs — they're in newer `workers-types` but not the version pinned here, so the docs were advertising hints that don't typecheck. Now listing the nine this repo actually accepts. Worth knowing the pin is behind if `apac-ne`/`apac-se` matter to anyone.

PR #855, opened 2026-07-17
fix(lambda): preserve installed package archive metadata

jperezrealini:
Preserve Unix file modes when packaging installed Lambda dependencies so executables remain runnable and npm-created symlinks survive archive extraction.

- Capture regular-file modes and represent symlinks as ZIP link entries.
- Pass Unix permissions through `zipCode`.
- Cover executable and symlink metadata in the installed-package tests.


packages/alchemy/src/Bundle/InstalledPackages.ts
```diff
@@ -587,11 +589,25 @@ const readArtifactFiles = (directory: string) =>
...
+      const linkTarget = yield* fs
+        .readLink(absolutePath)
+        .pipe(Effect.catch(() => Effect.succeed(undefined)));
+      if (linkTarget !== undefined) {
+        files.push({
+          path: relativePath.replaceAll("\\", "/"),
+          content: yield* Effect.sync(() =>
+            new TextEncoder().encode(linkTarget),
+          ),
+          mode: 0o120777,
```
sam-goodwin: why hard coded?

jperezrealini: replaced the literal with a named `symbolicLinkMode` derived from `fsConstants.S_IFLNK | 0o777`. Effect’s `FileSystem` exposes `readLink` but does not have `lstat`, and `stat` follows symlinks, so the ZIP symlink type bits must be synthesized after detecting a link

PR #856, opened 2026-07-17
chore(local): tests for round-trip RPC success values through marker serialization

asterikx:
## Bug

`alchemy dev` (since the capnweb RPC split) crashes on any provider whose **result** contains an Effect `Redacted`:

```
TypeError: Cannot serialize value: <redacted>
    at evaluateImpl (capnweb/dist/index.js:1636:48)
    at provider.update (alchemy/src/Apply.ts:123:14)
```

**Repro:** any Cloudflare Container with a `Redacted` env value (e.g. `Config.redacted(...)` or an `Output`-mapped `Redacted.make`) under `alchemy dev` @ 2.0.0-beta.63 — create/update of the container fails, `alchemy dev: apply failed`. The container provider's result embeds `dev.env: Record<string, string | Redacted>` (`LocalContainerProvider`), which capnweb (correctly) refuses to serialize.

## Cause

`Local/RpcSerialization.ts` marker-encodes RPC **arguments** (`serializeRpcArgs`/`deserializeRpcArgs` turn `Redacted`/`Output` into `{_tag}` markers) — but `wrapRpcEffectHandler` returns `{ _tag: "Success", value: exit.value }` with the value **raw**, and `unwrapRpcEffectHandler` does `Exit.succeed(exit.value)` without decoding.

## Fix

Success values now round-trip the same marker (de)serialization as arguments, on both sides. Two return-value tests added — they fail on current main and pass with the fix (14/14 in `test/Local/RpcSerialization.test.ts`).

## Note

`wrapRpcStreamHandler`/`unwrapRpcStreamHandler` pipe **stream elements** raw as well — if streamed values can ever carry `Redacted`/`Output`, the same treatment is needed there. Left out of this PR since I couldn't confirm a live path that streams such values; happy to extend if you want it symmetric.

We hit this in production dev-loop usage (zero-cache + Python containers with secret env under `alchemy dev`) and are running exactly this change as a local patch meanwhile.


sam-goodwin:
Is there a test we can add?

john-royal (approved):

asterikx:
There are two in the PR already — `packages/alchemy/test/Local/RpcSerialization.test.ts` gains a `return value serialization` block:

- **round-trips a Redacted return value** — a handler returning `Redacted.make("hush")`; the existing `roundTrip` harness threads the JSON wire hop, so on `main` the value comes back corrupted instead of a reconstructed `Redacted`
- **round-trips Redacted values nested in a returned object** — the container-provider shape (`{ env: { TOKEN: Redacted } }`) that crashes `alchemy dev` in practice

Red/green verified: on `main` (fix stashed) exactly these two fail — `Tests: 2 failed | 12 passed`; with the fix `14 passed`.

Happy to also cover the stream-element path (`wrapRpcStreamHandler` has the same gap), but that needs the symmetric stream fix first — say the word and I will extend the PR to both.

sam-goodwin:
I accidentally fixed this in https://github.com/alchemy-run/alchemy/pull/1010 but your change added tests.

PR #858, opened 2026-07-17
fix(lambda): pin external installs from lockfiles

jperezrealini:
Lambda `build.install` pinned only the direct requested packages; npm re-resolved every transitive dependency at install time, so the deployed artifact could drift from the project lockfile. Aliases, per-context duplicate resolutions, Bun workspace importers, and legacy binary lockfiles were also mishandled.

`build.install` now resolves the requested packages against the nearest lockfile (`package-lock.json` v1/v3, `pnpm-lock.yaml`, `bun.lock`, legacy `bun.lockb`, `yarn.lock` v1/berry) and pins the reachable dependency graph through generated npm `overrides`:

```json
{
  "dependencies": { "sharp": "0.33.5" },
  "overrides": { "sharp@0.33.5": { "semver": "7.6.3" } }
}
```

- Overrides come from the deduplicated graph: one flat rule per uniquely-resolved name, nested rules only where the lockfile resolves the same name differently per branch. A 2^20-path diamond graph emits 0.5KB of overrides; `jest`'s full graph emits 209KB and installs in ~6s.
- npm aliases (`npm:real@x`) and context-specific duplicate resolutions are preserved; resolutions that npm overrides cannot express (same override path, different versions) are left unpinned instead of failing the deploy.
- With overrides, install runs `npm install --package-lock-only --ignore-scripts` then `npm ci`; with nothing to pin it stays a single `npm install`.
- `bun.lock` is parsed with a runtime-agnostic JSONC parser (no `Bun.*` globals, works under Node); reading a legacy `bun.lockb` shells out to `bun` and says so if the binary is missing.
- Resolved versions, overrides, and the lockfile hash fold into the artifact identity hash, so lockfile-only changes redeploy the function.

### Behavior changes

- A stale lockfile (declared version no longer resolvable from the lock) fails the deploy with "refresh the package-manager lockfile" instead of silently installing different versions.
- Existing functions using `build.install` rebuild their artifact once on upgrade because the identity hash now includes overrides.


sam-goodwin:
What's the status here? Is there an issue we need to fix with lock files?

PR #876, opened 2026-07-18
fix(aws/rds): fingerprint-guard MasterUserPassword — stop re-sending it on every reconcile

zeyuri:
`DBInstance` re-sends `MasterUserPassword` on every reconcile of an instance whose props carry a password. AWS never returns the password, so there is nothing to observe-and-diff — the modify branch just sends it unconditionally:

```ts
} else if (news.masterUserPassword !== undefined) {
  core.MasterUserPassword = news.masterUserPassword; // fires on EVERY deploy
  coreDirty = true;
}
```

Each of those sends puts the live instance into a `resetting-master-credentials` modify cycle: existing connections can be reset, and `waitForInstance` blocks the whole deploy until the instance settles again. For anyone deploying regularly against a production database, every routine deploy pays that cost for a password that hasn't changed.

### Context — how we hit this

We run a production Postgres `DBInstance` behind Cloudflare Hyperdrive (a better-auth session store, so connection resets are user-visible). Our stack deploys several times a day. We had carried a vendored copy of this resource specifically because of this behavior; this PR upstreams that fix so we can drop the vendored copy.

### What this enables

Declaring the password in props becomes deploy-neutral: reconcile only touches the master password when the configured value actually changed.

```ts
const db = yield* DBInstance("Db", {
  engine: "postgres",
  dbInstanceClass: "db.t4g.micro",
  masterUserPassword: Redacted.make(process.env.DB_PASSWORD!),
});
// deploy N times → zero resetting-master-credentials cycles
// change DB_PASSWORD → exactly one modify on the next deploy
```

### How

The password can't be diffed against cloud state, so reconcile fingerprints the configured value (SHA-256, via the existing `Util/sha256`) and persists the fingerprint — never the secret — as a new `masterUserPasswordFingerprint` attribute. The modify branch sends `MasterUserPassword` only when the fingerprint differs from the stored one. `read` carries the stored fingerprint forward, since it isn't observable from AWS.

States that predate this attribute have no fingerprint, so their next reconcile sends the password once (exactly what happens today) and records it; every deploy after that skips.

The `manageMasterUserPassword`/`rotateMasterUserPassword` (Secrets Manager) branch is untouched.

### DIFF
--- a/packages/alchemy/src/AWS/RDS/DBInstance.ts
+++ b/packages/alchemy/src/AWS/RDS/DBInstance.ts
@@ -11,6 +11,7 @@ import * as Provider from "../../Provider.ts";
 import { Resource } from "../../Resource.ts";
 import type { Providers } from "../Providers.ts";
 import { createInternalTags, diffTags } from "../../Tags.ts";
+import { sha256 } from "../../Util/sha256.ts";
 
 export interface DBInstanceProps {
   /**
@@ -379,6 +380,13 @@ export interface DBInstance extends Resource<
      */
     finalDBSnapshotIdentifier: string | undefined;
     /**
+     * SHA-256 hex fingerprint of the last `masterUserPassword` this provider
+     * sent to RDS — never the secret itself. Lets reconcile skip the
+     * `MasterUserPassword` modify (and the `resetting-master-credentials`
+     * cycle it triggers) when the configured password has not changed.
+     */
+    masterUserPasswordFingerprint: string | undefined;
+    /**
      * ARN of the Secrets Manager secret holding master credentials.
      */
     masterUserSecretArn: string | undefined;
@@ -476,14 +484,17 @@ const toAttrs = ({
   tags,
   skipFinalSnapshot,
   finalDBSnapshotIdentifier,
+  masterUserPasswordFingerprint,
 }: {
   instance: rds.DBInstance;
   tags: Record<string, string>;
   skipFinalSnapshot?: boolean | undefined;
   finalDBSnapshotIdentifier?: string | undefined;
+  masterUserPasswordFingerprint?: string | undefined;
 }): DBInstance["Attributes"] => ({
   skipFinalSnapshot,
   finalDBSnapshotIdentifier,
+  masterUserPasswordFingerprint,
   dbInstanceIdentifier: instance.DBInstanceIdentifier ?? "",
   dbInstanceArn: instance.DBInstanceArn ?? "",
   dbClusterIdentifier: instance.DBClusterIdentifier,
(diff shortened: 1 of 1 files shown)


packages/alchemy/src/AWS/RDS/DBInstance.ts
```diff
@@ -350,6 +351,13 @@ export interface DBInstance extends Resource<
      * Master username.
      */
     masterUsername: string | undefined;
+    /**
+     * SHA-256 hex fingerprint of the last `masterUserPassword` this provider
+     * sent to RDS — never the secret itself. Lets reconcile skip the
+     * `MasterUserPassword` modify (and the `resetting-master-credentials`
+     * cycle it triggers) when the configured password has not changed.
+     */
+    masterUserPasswordFingerprint: string | undefined;
```
sam-goodwin: Should this be `Redacted`? A sha of a password is still sensitive

packages/alchemy/src/AWS/RDS/DBInstance.ts
```diff
@@ -649,11 +660,27 @@ export const DBInstanceProvider = () =>
...
             return undefined;
           }
-          return toAttrs({ instance, tags: toTagRecord(instance.TagList) });
+          return toAttrs({
+            instance,
+            tags: toTagRecord(instance.TagList),
+            // Not observable from AWS — carry the last-sent fingerprint
+            // forward so a refresh doesn't force a password re-send.
+            masterUserPasswordFingerprint:
+              output?.masterUserPasswordFingerprint,
```
sam-goodwin: Makes sense

packages/alchemy/src/AWS/RDS/DBInstance.ts
```diff
@@ -649,11 +660,27 @@ export const DBInstanceProvider = () =>
...
             output?.dbInstanceIdentifier ?? (yield* toIdentifier(id, news));
+          // AWS never returns the master password, so there is nothing to
+          // observe-and-diff — fingerprint the configured value instead and
+          // only send `MasterUserPassword` when the fingerprint changed.
+          // Without this, every reconcile of an instance whose props carry a
+          // password triggers a live `resetting-master-credentials` modify.
+          const passwordFingerprint =
+            news.masterUserPassword !== undefined
+              ? yield* sha256(Redacted.value(news.masterUserPassword))
+              : undefined;
```
sam-goodwin, line 746: I think we should keep this Redacted, it's sensitive. Vulnerable to rainbow tables, for example

sam-goodwin (commented):
Looks good, some minor changes required

packages/alchemy/src/AWS/RDS/DBInstance.ts
zeyuri: Done — the fingerprint is now `Redacted.Redacted<string>` end to end (attr type, `toAttrs`, the reconcile computation, and the comparison). It round-trips through state the same way `Random.text` and the Planetscale/Neon password attrs do: `StateEncoding` wraps Redacted values in a `__redacted__` envelope on write and rebuilds them on read.

packages/alchemy/src/AWS/RDS/DBInstance.ts
zeyuri: Agreed, and went one further: the digest is now salted with the stable `dbInstanceIdentifier` (`sha256(identifier + ":" + password)`), which defeats precomputed tables outright, while the Redacted wrapper keeps the digest out of logs/state dumps. Comparison unwraps both sides with `Redacted.value` and treats an absent stored fingerprint as "changed" so a pre-existing instance sends once and then records. No migration concern since nothing has been persisted yet.

sam-goodwin (commented):
Is there a test case we can add that exercises this?

PR #877, opened 2026-07-18
feat(aws/rds): skipFinalSnapshot option — final snapshot on deliberate teardown

zeyuri:
`DBInstance.delete` always calls `deleteDBInstance` with `SkipFinalSnapshot: true` — there is no way to have RDS take a final snapshot when a stack tears the instance down. `deletionProtection` guards against *accidental* deletion, but a deliberate `destroy` of an instance holding real data currently discards that data with no recovery point.

### Context — how we use it

Our production Postgres runs as a `DBInstance` with `deletionProtection: true` *and* (in a vendored copy of this resource) `skipFinalSnapshot: false` — two independent layers: protection stops the accident, the final snapshot makes even a deliberate teardown recoverable. This PR upstreams the second layer so we can drop the vendored copy.

### What this enables

```ts
const db = yield* DBInstance("Db", {
  engine: "postgres",
  dbInstanceClass: "db.t4g.micro",
  deletionProtection: true,
  skipFinalSnapshot: false, // destroy → RDS takes a final snapshot first
});
```

On destroy, RDS snapshots the instance before deleting it. The snapshot identifier defaults to `<instance-identifier>-final-<timestamp>` so repeated destroy/create cycles never collide on snapshot names; `finalDBSnapshotIdentifier` overrides it.

### How

- Two new props: `skipFinalSnapshot` (default `true` — existing behavior unchanged) and `finalDBSnapshotIdentifier`.
- Both persist into state as attributes, because `delete` receives only stored attributes, never live props — the snapshot decision has to ride in state. `read` carries them forward (not observable from AWS).
- `delete` passes them through, and the delete-completion wait doubles its budget on the snapshot path (the snapshot serializes before the delete, so `deleting` lasts correspondingly longer).

### DIFF
--- a/packages/alchemy/src/AWS/RDS/DBInstance.ts
+++ b/packages/alchemy/src/AWS/RDS/DBInstance.ts
@@ -204,6 +204,20 @@ export interface DBInstanceProps {
    * User-defined tags.
    */
   tags?: Record<string, string>;
+  /**
+   * Skip the final snapshot when the instance is deleted. Set `false` to
+   * have RDS take a final snapshot on teardown — belt-and-suspenders beyond
+   * `deletionProtection` for databases whose data must survive a deliberate
+   * destroy. Persisted into state so `delete` honors it without props.
+   * @default true
+   */
+  skipFinalSnapshot?: boolean;
+  /**
+   * Identifier for the final snapshot taken when `skipFinalSnapshot` is
+   * `false`. Defaults to `<instance-identifier>-final-<timestamp>` so
+   * repeated destroy/create cycles never collide on snapshot names.
+   */
+  finalDBSnapshotIdentifier?: string;
 }
 
 export interface DBInstance extends Resource<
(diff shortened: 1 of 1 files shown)


packages/alchemy/src/AWS/RDS/DBInstance.ts
```diff
@@ -350,6 +364,13 @@ export interface DBInstance extends Resource<
      * Master username.
      */
     masterUsername: string | undefined;
+    /**
+     * Deletion behavior persisted from props — `delete` receives only the
+     * stored attributes, never live props, so the snapshot decision must
+     * ride in state.
+     */
+    skipFinalSnapshot: boolean | undefined;
+    finalDBSnapshotIdentifier: string | undefined;
```
sam-goodwin, line 380: add docs

zeyuri: Added — each attribute now has its own docstring. `skipFinalSnapshot` and `finalDBSnapshotIdentifier` both note that they are persisted because `delete` receives only stored attributes (never live props), and the identifier docstring documents the default `<instance-identifier>-final-<timestamp>` naming.

PR #884, opened 2026-07-20
refactor(cloudflare/workers): source provider interface

john-royal:
Replace the per-bundler `if/else` ladders in `WorkerProvider` (`prepareAssetsAndBundle`, `hasChanged`) and `LocalWorkerProvider` (`runVite` vs `runWorker`) with a uniform `SourceProvider` interface, implemented by the five existing arms and pluggable by framework packages (Next/Astro/SvelteKit/Waku). Implements `cloudflare-tools/research/source-provider-design.md`, steps 1–3.

```ts
export interface SourceProvider {
  readonly ownsAssets: boolean; // vite/frameworks: assets come out of the build
  readonly build: (ctx: SourceContext) => Effect<SourceBuildOutput, SourceError, SourceServices>;
  readonly hash: (ctx: SourceContext, previous: SourceHash | undefined) => Effect<Partial<SourceHash>, SourceError, SourceServices>;
  readonly dev: (ctx: DevContext) => Effect<SourceDevHandle, SourceError, SourceDevServices | Scope>;
}

// dev is either "host runs workerd, source streams bundles" or "source serves itself"
export type SourceDevHandle =
  | { mode: "bundle"; bundles: Stream<BundleWatchEvent, SourceError, SourceDevServices> }
  | { mode: "server"; url: URL };
```

- `resolveSource(props)` maps the legacy props to built-ins with the historical precedence (`script` → `vite` → `.py` → `bundle: false` → rolldown), so nothing changes for existing workers: same hash slots (`bundle`/`assets`/`input`/`additionalWorkspaces`), same `Artifacts.cached("build")` reuse across diff→reconcile, same AssetsWithHash/`keepAssets` fast path, #745 metadata hash untouched and still compared first.
- New `source` prop for external providers, resolved by dynamic `import()` (props stay serializable across state and the local-provider RPC boundary):

```ts
Worker("Site", {
  source: { provider: "@alchemy.run/cloudflare-next", options: { rootDir: "." } },
})
```

  `loadSource` validates the module's default export (`WorkerSourceModule`: `{ make(options) }`) and wraps failures in a typed `SourceProviderError` naming the package to install. The descriptor participates in the metadata hash, so switching providers or changing options always deploys.
- Fixes local dev of `bundle: false` workers: the local provider re-bundled prebuilt entries through rolldown, breaking the byte-for-byte contract the deploy path documents. `PrebuiltSource.dev` now fs-watches the entry directory and re-reads the module graph (`watchPrebuiltWorkerBundle`); pinned by `PrebuiltWorkerLocal.test.ts`.
- `props.script` workers now also work under `alchemy dev` (single-emit bundle stream; previously they fell into the rolldown watcher with `main: undefined`).

### Deviations from the design doc

- `SourceContext.env`/`assets` carry the raw props values instead of pre-resolved ones — resolving env Effects centrally would execute plan-phase-only Effects for non-vite arms; the vite source resolves them itself exactly as before.
- `ownsAssets` is a provider field (the doc implied it structurally); `skipAssetsRead` stays a `putWorker` concern instead of a ctx field, since props-assets reading never moved out of the WorkerProvider.
- `additionalWorkspaces` is explicitly documented as auxiliary (not compared as a change signal), matching today's diff exactly.
- `source` + `script`/`vite`/`main` mutual exclusion fails with `SourceProviderError` rather than `WorkerValidationError` (avoids a module cycle with `LocalWorkerProvider`).

### DIFF
--- a/packages/alchemy/src/Cloudflare/Workers/LocalWorkerProvider.ts
+++ b/packages/alchemy/src/Cloudflare/Workers/LocalWorkerProvider.ts
@@ -66,11 +66,16 @@ import { CloudflareEnvironment } from "../CloudflareEnvironment.ts";
 import { LOCAL_ENTRY_URL, LocalRuntimeState } from "../LocalRuntime.ts";
 import type { WorkerAssetsConfig, WorkerProps } from "../Workers/Worker.ts";
 import { getCompatibility } from "./Compatibility.ts";
-import { isPythonMain, watchPythonWorkerBundle } from "./PythonWorkerBundle.ts";
+import {
+  makeSourceContext,
+  resolveSource,
+  type DevContext,
+  type SourceDevServices,
+  type SourceError,
+} from "./Source.ts";
 import { Worker } from "./Worker.ts";
 import { getCronBindings } from "./WorkerAsyncBindings.ts";
 import type { WorkerBinding } from "./WorkerBinding.ts";
-import { WorkerBundle, type WorkerBundleOptions } from "./WorkerBundle.ts";
 import { createWorkerName } from "./WorkerName.ts";
 
 type WorkerPropsWithDev = Omit<WorkerProps, "dev"> & {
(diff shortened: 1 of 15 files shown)


john-royal:
Going to use #886 for this instead of stacking

PR #886, opened 2026-07-21
feat(cloudflare/website): Waku, Astro, SvelteKit, Nuxt resources

### LINKED ISSUE #772: Question: About SvelteKit adapter + local dev philosophy shift from v1
Hey! I am migrating a SvelteKit project from Alchemy v1 to v2, ran into a couple things I'd love to understand better:

## 1. No SvelteKit adapter/resource in v2?
v1 had alchemy/cloudflare/sveltekit (the SvelteKit adapter) and a SvelteKit resource — both worked great. In v2 I see dedicated resources for SolidStart, TanStack Start, etc., but no SvelteKit equivalent. Is SvelteKit support planned, or is the expectation that SvelteKit projects use Cloudflare.Worker with bundle: false pointing at the adapter output?

## 2. Why the shift from local emulation to remote tunneling in dev?
In v1, alchemy dev with SvelteKit ran everything locally via miniflare — D1 was SQLite-backed, KV was emulated, zero network calls. It was fast and worked offline.
In v2, I notice D1/KV/R2 tunnel through a deployed Cloudflare preview worker even during dev, which means every dev session hits the network. Durable Objects still run local (workerd), which is nice. Curious about the reasoning behind this shift — was miniflare's emulation not accurate enough? Was it a maintenance burden? Something else entirely?

Mostly trying to understand whether this is the long-term direction or if local emulation (plus SvelteKit support) might return.

Thanks!

### LINKED ISSUE #820: vocs SSR deployment on Cloudflare
For example, the bootstrapped template + the MCP-enabled config:

- https://vocs.dev/
- https://vocs.dev/reference/site-config#mcp

There're many layers of vite/cloudflare/react/bundling madness.

john-royal:
Framework website resources for Waku, Astro, SvelteKit, and Nuxt, built on a `source`-descriptor dispatch in the Worker provider. (`Website.Nextjs` is split out to #923.)

Each resource sets `source: { provider: "@distilled.cloud/<fw>/source", options }` on the Worker. The framework package's source module does the build (programmatic, no wrangler binary, no wrangler.json), asset manifesting, and rebuild-free input hashing for diff memoization. All four return a plain `Worker<Bindings>` — `env` takes the full binding vocabulary, `url`/`domain`/`compatibility`/asset routing all apply — and all four run under `alchemy dev` through the same dispatch (server-mode dev handle → dev proxy).

### Waku

```ts
const app = yield* Cloudflare.Website.Waku("App", {
  main: "src/worker-entry.ts", // optional custom entry — can host Durable Objects
  env: { MESSAGE: "hello", COUNTER: Cloudflare.DurableObject("Counter") },
});
```

- RSC SSR at request time; SSG pages prerender at build time **inside workerd**, served from assets
- Zero-config defaults: `nodejs_als` auto-injected, `assets.htmlHandling: "drop-trailing-slash"` (waku links SSG pages without trailing slashes)
- `main` escape hatch wraps `virtual:waku/server-entry`, so the deployed Worker can export DO classes alongside waku's fetch handler
- Dev: waku's vite dev server with the rsc environment in workerd; real bindings in-process

### Astro

```ts
const site = yield* Cloudflare.Website.Astro("Site", {
  env: { CACHE: kv, UPLOADS: bucket },
});
```

- `output: "server"` by default (astro's own default is `static`, which would prerender "SSR" pages inside workerd where bindings don't exist); `astro: { output: "static" }` deploys assets-only — no server bundle at all
- `nodejs_compat` auto-injected
- Zero-config sessions: a KV namespace is auto-provisioned and bound as `SESSION` (`sessionKVBindingName: false` opts out; skipped for static output); destroyed with the stack
- Your `astro.config.*` still loads natively; the resource's inline config merges over it; prerendered routes render inside workerd by default
- Dev: astro's dev server with the ssr environment in workerd; real bindings in-process

### SvelteKit

```ts
const site = yield* Cloudflare.Website.SvelteKit("Site", {
  env: { API_KEY: Alchemy.secret("API_KEY") },
});
```

- In-memory kit v3 adapter (no `svelte.config.js`, no `@sveltejs/adapter-cloudflare`) + rolldown finishing pass re-bundles `_worker.js` for workerd
- `nodejs_compat` auto-injected; bindings reach server routes via `platform.env`
- Dev: kit's own vite dev server (Node SSR, full HMR); `platform.env`/`caches`/`cf` served by cloudflare-runtime's platform proxy with real bindings, literal `env` overlaid
- Ships with a fix for kit resolving root-relative runtime ids against `process.cwd()` — dev now works from any working directory

### Nuxt

```ts
const site = yield* Cloudflare.Website.Nuxt("Site", {
  main: "worker-entry.ts", // optional custom entry — can host Durable Objects
  nuxt: { routeRules: { "/about": { prerender: true } } },
});
```

- nitro `cloudflare_module` preset; the only resource that **natively loads your `nuxt.config.ts`** — the `nuxt` prop merges over it (override wins)
- `nodejs_compat` auto-injected; bindings reach handlers via `event.context.cloudflare.env` (+ `cf`, `waitUntil`)
- Known limits: nitro `isr` route rules are ignored; custom-entry DO classes are not servable in dev (the platform proxy can't host them)
- Dev: nitro dev in a Node worker thread; the platform reconstructed over HTTP from the cloudflare-runtime platform proxy with real bindings

### Tests, examples, docs

- Live suites `test/Cloudflare/Website` green: create → memoized noop → edit-redeploy → destroy per framework (also fixes the StaticSite noop plan assertion that fails on main)
- Dev-mode `{Waku,Astro,SvelteKit,Nuxt}.local.test.ts` (`Test.make({ dev: true })`): local identity, SSR markers, env bindings, static assets
- `examples/cloudflare-website-{waku,astro,sveltekit,nuxt}` with integ tests, wired into `test:examples` and root tsconfig references — all four verified green against the real cloud
- Docs: rewritten `frontends`/`astro`/`nuxt` guides, new `sveltekit` + `waku` pages, sidebar registration

Submodule: pins cloudflare-tools to alchemy-run/cloudflare-tools#74 (framework packages, DeployTarget architecture, fixtures, and the kit runtime-id re-rooting fix).

### DIFF
--- a/packages/alchemy/test/Cloudflare/Website/Astro.local.test.ts
+++ b/packages/alchemy/test/Cloudflare/Website/Astro.local.test.ts
@@ -1,12 +1,22 @@
+import { CloudflareEnvironment } from "@/Cloudflare/CloudflareEnvironment.ts";
 import * as Cloudflare from "@/Cloudflare/index.ts";
+import { isLocalId } from "@/Cloudflare/LocalRuntime";
+import * as Alchemy from "@/index.ts";
 import * as Test from "@/Test/Alchemy";
+import * as kv from "@distilled.cloud/cloudflare/kv";
 import { expect } from "alchemy-test";
+import * as Data from "effect/Data";
 import * as Effect from "effect/Effect";
+import * as Option from "effect/Option";
 import { MinimumLogLevel } from "effect/References";
+import * as Schedule from "effect/Schedule";
+import * as Stream from "effect/Stream";
 import * as pathe from "pathe";
 import { cloneFixture } from "../Utils/Fixture.ts";
 import { expectUrlContains } from "../Utils/Http.ts";
 
+// `dev: true` runs local providers behind the RPC sidecar proxy by default,
+// matching the process topology of the real `alchemy dev` command.
 const { test } = Test.make({ providers: Cloudflare.providers(), dev: true });
 
 const logLevel = Effect.provideService(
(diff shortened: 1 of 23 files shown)


AmanVarshney01:
all eyes on this pr 👀

PR #893, opened 2026-07-21
ci: debug mysql2 check failure (temporary)

sam-goodwin:
TEMPORARY debug PR — do not merge. The check job fails on every post-#867 branch with `Cannot find module mysql2/promise` in Planetscale MySQLMigrations; a clean install repro on macOS passes, so this instruments the Linux CI install state (store entries, workspace link, bun pm why, lockfile drift) to find the divergence.

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -37,5 +37,17 @@ jobs:
           restore-keys: bun-${{ runner.os }}-
 
       - run: bun install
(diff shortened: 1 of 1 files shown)


sam-goodwin:
Root cause found and fixed in #897 (mysql2 dependency dropped by #887 while MySQLMigrations still imports it). Debug instrumentation no longer needed.

PR #895, opened 2026-07-21
fix(aws): AWS_REGION env overrides SSO profile/stored region

sam-goodwin:
`resolveCredentials` returned the `~/.aws/config` profile region verbatim for `sso` (and `stored`) auth, silently ignoring an explicit `AWS_REGION` env var — e.g. `AWS_REGION=us-east-1` still deployed to the profile's `us-west-2`.

The fix applies the env override as the final step of `resolveCredentials`, for every auth method:

```ts
export const applyEnvRegionOverride = <C extends { region: string }>(
  creds: C,
): Effect.Effect<C> =>
  getEnv("AWS_REGION").pipe(
    Effect.map((envRegion) =>
      envRegion ? { ...creds, region: envRegion } : creds,
    ),
  );
```

Precedence: `AWS_REGION` > profile/stored region. `AWS_DEFAULT_REGION` deliberately does **not** override — it is a default, not an override, and the profile's region is explicit configuration (review feedback). When `AWS_REGION` is unset, behavior is unchanged (profile region wins). `Region.fromEnvironment` derives from `AWSEnvironment.region`, so the engine's `Region` service picks up the override everywhere.

Verified live against the testing SSO profile (profile region `us-west-2`):

```
no env                         -> resolvedRegion us-west-2  (unchanged default)
AWS_REGION=us-east-1           -> resolvedRegion us-east-1
pre-fix + AWS_REGION=us-east-1 -> resolvedRegion us-west-2  (the bug)
```

Offline unit coverage in `test/AWS/AuthProvider.test.ts` via `ConfigProvider.fromEnv({ env })`, including profile-beats-`AWS_DEFAULT_REGION`.

### DIFF
--- a/packages/alchemy/src/AWS/AuthProvider.ts
+++ b/packages/alchemy/src/AWS/AuthProvider.ts
@@ -79,17 +79,15 @@ export interface AwsResolvedCredentials {
 }
 
 /**
- * An explicitly-set region env var always wins over the region recorded in an
- * SSO profile (`~/.aws/config`) or in stored credentials:
- * `AWS_REGION` > `AWS_DEFAULT_REGION` > profile/stored region.
+ * An explicitly-set `AWS_REGION` env var wins over the region recorded in an
+ * SSO profile (`~/.aws/config`) or in stored credentials. `AWS_DEFAULT_REGION`
+ * deliberately does NOT override — it is a *default* for when no region is
+ * configured anywhere, and the profile's region is explicit configuration.
  */
 export const applyEnvRegionOverride = <C extends { region: string }>(
   creds: C,
 ): Effect.Effect<C> =>
   getEnv("AWS_REGION").pipe(
-    Effect.flatMap((region) =>
-      region ? Effect.succeed(region) : getEnv("AWS_DEFAULT_REGION"),
-    ),
     Effect.map((envRegion) =>
       envRegion ? { ...creds, region: envRegion } : creds,
     ),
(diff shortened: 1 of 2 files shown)


packages/alchemy/src/AWS/AuthProvider.ts
```diff
@@ -78,6 +78,23 @@ export interface AwsResolvedCredentials {
...
+ * An explicitly-set region env var always wins over the region recorded in an
+ * SSO profile (`~/.aws/config`) or in stored credentials:
+ * `AWS_REGION` > `AWS_DEFAULT_REGION` > profile/stored region.
+ */
+export const applyEnvRegionOverride = <C extends { region: string }>(
+  creds: C,
+): Effect.Effect<C> =>
+  getEnv("AWS_REGION").pipe(
+    Effect.flatMap((region) =>
+      region ? Effect.succeed(region) : getEnv("AWS_DEFAULT_REGION"),
```
sam-goodwin: Should AWS_DEFAULT_REGION override the one in the auth provider? I don't think that makes sense

sam-goodwin: Agreed — dropped it. `AWS_DEFAULT_REGION` no longer overrides the profile/stored region (it's a default, not an override); only an explicit `AWS_REGION` wins. Test flipped to assert the profile beats `AWS_DEFAULT_REGION`.

PR #908, opened 2026-07-21
feat(cloudflare/workers): support entrypoint on raw service binding specs in Worker env

apostolos-geyer:
## Motivation

A service binding declared as a `Cloudflare.Worker` resource always targets the bound script's **default export**. That leaves no way to address a **named `WorkerEntrypoint` class** on the target script — the standard Cloudflare pattern for exposing several RPC surfaces from one Worker (`export class BillingRpc extends WorkerEntrypoint { … }`). A multi-worker Cloudflare deployment using named entrypoint RPC targets therefore cannot express its fleet's bindings in Alchemy today.

The Workers platform already supports this: the script-upload service-binding metadata accepts `entrypoint` and `environment`, and the `@distilled.cloud/cloudflare` `service` binding variant already types both as `entrypoint?: string | null` / `environment?: string | null`. Nothing in the Worker `env` surface emits them.

## Change

Two additive pieces, both in `src/Cloudflare/Workers`:

1. **`WorkerBinding.ts`** — a new exported `ServiceBindingSpec` interface (`{ $binding: "service"; service: string; entrypoint?: string; environment?: string }`), an `isServiceBindingSpec` type guard, and the spec added to the `WorkerBindingResource` union. A Worker's `env` can now carry a raw service-binding descriptor directly:

   ```ts
   const worker = yield* Cloudflare.Worker("api", {
     env: {
       BILLING: {
         $binding: "service",
         service: "billing-worker",
         entrypoint: "BillingRpc",
       },
     },
   });
   ```

2. **`WorkerAsyncBindings.ts`** — a new `isServiceBindingSpec` branch in `toBinding`, right after the existing `isWorker` branch, emitting `{ type: "service", name, service, entrypoint, environment }`. `entrypoint` and `environment` pass straight through (both optional); omitting `entrypoint` preserves today's default-export behavior.

## Notes

- **Backwards-compatible / additive.** No existing binding path changes; the `Cloudflare.Worker`-as-binding shape (`isWorker`) is untouched and still emits `{ type: "service", service }` with no entrypoint.
- **`environment` passthrough.** The spec also forwards the wire schema's `environment` field for parity, defaulting to the bound service's default environment when omitted.

## Tests

Added `test/Cloudflare/Workers/ServiceBindingSpec.test.ts` — a pure unit test for the `isServiceBindingSpec` guard (matches bare / entrypoint+environment specs, rejects missing `service`, other `$binding` discriminants, and non-objects), in the offline `describe`/`it` style of `test/Cloudflare/Fetcher.test.ts` (no cloud credentials required).

The existing service-binding integration tests (`test/Cloudflare/Workers/WorkerBinding.test.ts`) deploy live Workers and hit `workers.dev` URLs, so an end-to-end test that a named entrypoint resolves at runtime would need live Cloudflare credentials; I've left that out of this PR. `toBinding` is not exported, so the guard test is the isolated-unit seam for this change.

`bun tsc -b packages/alchemy` and the test tsconfig both typecheck clean.

### DIFF
--- a/packages/alchemy/src/Cloudflare/Workers/WorkerAsyncBindings.ts
+++ b/packages/alchemy/src/Cloudflare/Workers/WorkerAsyncBindings.ts
@@ -34,7 +34,11 @@ import { isRateLimit } from "./RateLimit.ts";
 import { isVersionMetadata } from "./VersionMetadata.ts";
 import type { WorkerBindingProps } from "./Worker.ts";
 import { isWorker, type Worker, type WorkerProps } from "./Worker.ts";
-import type { WorkerBinding, WorkerBindingResource } from "./WorkerBinding.ts";
+import {
+  isServiceBindingSpec,
+  type WorkerBinding,
+  type WorkerBindingResource,
+} from "./WorkerBinding.ts";
 import { isWorkerLoader } from "./WorkerLoader.ts";
 
 export const bindWorkerAsyncBindings = Effect.fn(function* (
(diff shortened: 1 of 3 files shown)


sam-goodwin:
We'll solve this soon but in a different way.

Issue #926, opened 2026-07-23
Cloudflare.Worker read fans out unused API calls and repeatedly fetches the account subdomain, causing 429/code 971


benpsnyder:
### Version

- `alchemy@2.0.0-beta.64`
- `effect@4.0.0-beta.100`
- Cloudflare Workers

### What happened

Repeated `alchemy plan` / `alchemy deploy` operations for a Worker using its `workers.dev` URL were blocked by a Cloudflare throttling response:

```text
HTTP 429
code 971
Please wait and consider throttling your request speed
```

The observed account-level request was:

```text
GET /accounts/{account_id}/workers/subdomain
```

Cloudflare documents this endpoint as returning the account Workers subdomain. The account subdomain is stable during a deployment, but `WorkerProvider` currently fetches it again whenever it computes a Worker URL.

### Request fan-out

For a regular Worker, `WorkerProvider.read` concurrently starts:

1. the per-script subdomain read;
2. script/version settings;
3. the account custom-domain listing filtered to the Worker; and
4. route discovery, which first enumerates account zones and then lists routes across those zones.

It then separately reads the account Workers subdomain and cron schedules. The domain, route, and cron reads occur even when the corresponding `domain`, `routes`, and `crons` props are omitted.

This means a simple Worker with no custom domains, routes, or crons can generate multiple avoidable API calls, including account/zone-wide fan-out. Multiple Workers or closely repeated plan/deploy operations amplify that traffic. The beta.64 Cloudflare retry layer still exhausted its bounded retry budget in the observed case.

### Expected behavior

A Worker read should avoid unnecessary Cloudflare calls and should not independently repeat account-level invariant lookups for every Worker.

Possible approaches:

- memoize/deduplicate `getSubdomain({ accountId })` for the provider execution;
- optionally allow the known, non-secret account subdomain to be configured at an appropriate provider boundary;
- skip domain, route, and cron observation when those surfaces are not managed, while preserving the intended adoption/drift semantics;
- bound the concurrency of the remaining Worker read calls and honor the existing retry policy.

### Downstream workaround

A downstream pnpm patch currently:

- wraps the account-subdomain lookup with `cachedFunction`;
- permits a validated account-subdomain value to bypass the lookup;
- retries typed `TooManyRequests` failures with bounded exponential backoff;
- avoids domain/route/cron reads when those props are absent; and
- changes the top-level Worker read `Effect.all` concurrency from `"unbounded"` to `1`.

[...]

PR #932, opened 2026-07-23
fix(cloudflare/workers): cut Worker read API fan-out that trips 429/971

john-royal:
Cut Cloudflare Worker read/reconcile API fan-out that was tripping account-level `429` / code `971` on repeated `plan`/`deploy` for plain `workers.dev` Workers ([#926](https://github.com/alchemy-run/alchemy/issues/926)).

- Memoize `GET /accounts/{id}/workers/subdomain` for the provider lifetime via `cachedFunction`
- Allow `CLOUDFLARE_WORKERS_SUBDOMAIN` to bypass that lookup when the account subdomain is already known
- Skip `listDomains`, account-wide route discovery, and `getScriptSchedule` when those surfaces are unmanaged (empty-array props still observe so deletions/drift converge)
- Bound Worker `read` `Effect.all` concurrency to `1`
- Same domain/cron gating on reconcile so deploys don't pay for unused calls either

```ts
// plain workers.dev Worker — read no longer fans out domains/routes/crons
yield* Cloudflare.Worker("Api", { main: "./src/api.ts" });

// optional: skip the account-subdomain GET entirely
process.env.CLOUDFLARE_WORKERS_SUBDOMAIN = "my-team";
```


benpsnyder:
Thanks for turning this around so quickly. Before this lands I want to correct the root cause I gave you in #926, because I got it wrong and this PR inherits the error.

### The account-subdomain lookup was the symptom, not the cause

I filed #926 pointing at `GET /accounts/{id}/workers/subdomain` because that is the endpoint the `429` / code `971` surfaced on. Reading the beta.64 provider more carefully, that endpoint is called **twice** per plan+deploy — once in `read` and once in `reconcile`. Two calls cannot exhaust an account-level throttle.

The dominant cost is `readWorkerRoutes`:

```
read
  └─ readWorkerRoutes(scriptName)
       ├─ listAllZones(accountId)                 // paginated over every zone in the account
       └─ listWorkerRoutesInZones(...)
            └─ Effect.all(zones.map(listRoutes), { concurrency: "unbounded" })
```

Routes are zone-scoped with no account-level enumeration API, so every Worker `read` enumerates all account zones and then issues one `GET /zones/{id}/workers/routes` per zone, all at once. On the account in #926 that is 328 zones, so a single Worker read is on the order of 330+ requests fired concurrently. A couple of plan/deploy cycles clears Cloudflare's global per-user API budget, and whichever call lands next takes the `971` — in my case the subdomain GET.

So the memoization and `CLOUDFLARE_WORKERS_SUBDOMAIN` bypass in this PR are worth having, but they are not what was causing the throttle.

### The consequence for this PR

`shouldObserveWorkerRoutes` **gates** the fan-out; it never **bounds** it. For a Worker with `olds.routes !== undefined` or any persisted routes — i.e. exactly the Workers that use zone routes — the account-wide sweep at `concurrency: "unbounded"` is unchanged. Several of my stacks declare `routes`, and they will still trip this after the PR merges.

The fix is already in the file. `reconcileRoutes` doesn't sweep the account — it derives its zone set from `desired ∪ previous`:

```ts
const zoneIds = Array.from(new Set([
  ...desired.map((route) => route.zoneId),
  ...previous.map((route) => route.zoneId),
]));
const liveAll = yield* listWorkerRoutesInZones(scriptName, zoneIds);
```

`read` can do the same thing with `output.routes[].zoneId` plus whatever zones `olds.routes` resolves to. That preserves drift detection for routes Alchemy manages, and it drops the account-wide sweep to zero calls for a Worker with no routes and a handful for a Worker with them. Bounding the `concurrency: "unbounded"` in `listWorkerRoutesInZones` would be worth doing independently — the same helper is on the reconcile path.

Discovering routes attached entirely out of band in a zone Alchemy has never touched is the only thing lost, and that is an adoption concern that probably shouldn't cost a full account sweep on every read anyway.

### Second point: the domain gating re-opens #942

Separately I hit #942 — an omitted `domain` deleting a custom hostname that was attached out of band. This PR closes the clean case, which I appreciate. But:

```ts
const reconcileCustomDomains =
  news.domain !== undefined || previousCustomDomains.length > 0;
```

The second disjunct re-arms it. If `output.domains` holds any non-`workers.dev` URL — true for any state written by a build whose `read` observed domains unconditionally, i.e. everything before this PR — then reconcile runs with `desiredDomains === []` and `reconcileDomains` deletes **every** live domain, out-of-band ones included. That is the adopted / partially-managed case #942 is actually about, and it is the state most existing users' state files are in right now.

### On my own workaround

The downstream patch I described in #926 had the same misattribution, and one part of it is worse than what you have written here: I gated the domain read on `olds?.domain === undefined`. On adoption `olds` is undefined, so it hides an adopted Worker's live domains from `read` while `reconcile` still deletes them. `shouldObserveWorkerDomains` consulting `output.domains` is the right call. I am dropping my version.

### Evidence quality

To be clear about what this is: static analysis of the beta.64 provider plus the account inventory I posted in #926. I do not have captured request-count telemetry from the original failure, so I am not claiming an exact number for the run that broke. You asked earlier for a preferred test shape — if it would help, I will put together a reproduction that stubs the Cloudflare operations and asserts request counts for (a) a plain `workers.dev` Worker and (b) a Worker declaring one route on a many-zone account, which would make the before/after concrete for both halves of this.

sam-goodwin:
Thanks for the correction on the root cause — you were right on both counts. Both are addressed as of the latest push.

### Routes

Your #944 landed on `main` first, so this branch now merges it rather than re-implementing: `readWorkerRoutes` takes its zone set from the routes already recorded in state (no account-wide zone sweep), and `listWorkerRoutesInZones` is bounded at `WORKER_ROUTE_LIST_CONCURRENCY = 8` — that same helper serves the reconcile path, so the bound applies there too. This PR's `shouldObserveWorkerRoutes` gating sits on top: a plain `workers.dev` Worker makes zero route calls on read, and a Worker declaring routes only lists the zones it's actually in.

### Domains

Agreed that deriving intent from persisted state was the bug. The gate is now keyed purely on declared intent:

```ts
const manageCustomDomains = news.domain !== undefined;
```

- `domain: []` — explicit detach-all.
- `domain` omitted — unmanaged: `reconcileDomains` never runs, so out-of-band attachments survive, **including** hostnames that pre-#932 reads persisted into `output.domains`. Those carry forward in the returned attributes so `read` keeps observing them, but they can never arm a deletion.

The contract is now documented on the `domain` prop.

### Regression coverage

Rather than the stubbed request-count harness, I added the live shape #942 suggested (`test/Cloudflare/Workers/WorkerDomain.test.ts`): deploy with a declared domain, attach a second hostname out of band via the API, update with `domain` omitted and assert both attachments survive, then `domain: []` and assert both are detached. It runs green against the real API in ~10s. Happy to take the request-count repro as a follow-up if you still think the counting half is worth pinning down.

PR #935, opened 2026-07-23
fix(aws/cloudfront): merge observed config into UpdateDistribution requests

bweis:
### Problem

`UpdateDistribution` replaces the **entire** distribution config — CloudFront rejects a request with any missing member as `IllegalUpdate`. The update path builds its request purely from props (`toConfig`), so updating a live distribution fails on every member the props don't express. Hit in production while managing adopted distributions:

```
IllegalUpdate: Default root object is missing for the resource
IllegalUpdate: The 'OriginCustomHeaders' field is missing.
```

Pinning members one-by-one in props can't close this: some required members (per-origin `CustomHeaders`, per-behavior `TrustedSigners`/`GrpcConfig`, …) live inside collection items and aren't expressible in `DistributionProps` at all.

### Fix

The standard CloudFront update pattern is read-modify-write — which the **delete** path here already uses (`{ ...current.config, Enabled: false }`). This PR applies the same idea to updates: members the desired config leaves `undefined` are filled from the freshly observed config (already fetched for the ETag), with item-wise merging for `Origins` (by `Id`) and `CacheBehaviors` (by `PathPattern`). Desired values always win, so declared props still fully control drift; desired removals are never resurrected from the observed config.

### Testing

Unit tests for `mergeWithObservedConfig` with a `GetDistributionConfig`-shaped fixture (desired-wins, top-level + nested fill, partial-object completion, removal/new-origin edge cases). Also validated live: this change (as a runtime patch on `2.0.0-beta.64`) unblocked real `UpdateDistribution` calls against three production distributions that previously failed with the errors above.

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -206,7 +206,7 @@
     },
     "distilled/packages/aws": {
       "name": "@distilled.cloud/aws",
-      "version": "0.30.1",
+      "version": "0.30.2",
       "dependencies": {
         "@aws-crypto/crc32": "catalog:",
         "@aws-crypto/util": "catalog:",
@@ -238,7 +238,7 @@
     },
     "distilled/packages/axiom": {
       "name": "@distilled.cloud/axiom",
-      "version": "0.30.1",
+      "version": "0.30.2",
       "dependencies": {
         "@distilled.cloud/core": "workspace:*",
       },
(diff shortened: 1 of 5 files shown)


sam-goodwin (changes requested):
Please add regression tests that actually run a deployment

bweis:
> Please add regression tests that actually run a deployment

Added.

PR #936, opened 2026-07-23
fix(aws/ecs): attach taskRoleManagedPolicyArns in Service reconcile

bweis:
### Problem
The image-owning `ServiceProps` form inherits `TaskDefinitionConfig`, which declares both `taskRoleManagedPolicyArns` and `executionRoleManagedPolicyArns`. `Service`'s reconcile honors the execution-role prop but silently drops the task-role one — services deploy with task roles missing their declared permissions, and nothing fails until the workload's first AWS call is denied at runtime. (Hit in production: a worker's `ses:Send*`/`bedrock:InvokeModel*` policy was declared but never attached.)

### Fix
Attach `taskRoleManagedPolicyArns` in `Service`'s reconcile right after the task role is resolved:

```ts
yield* Effect.all(
  (news.taskRoleManagedPolicyArns ?? []).map((policyArn) =>
    iam.attachRolePolicy({ RoleName: taskRoleName, PolicyArn: policyArn }),
  ),
  { concurrency: "unbounded" },
);
```

Per review:

- Attaches run in parallel via `Effect.all` instead of a sequential loop.
- `LimitExceededException` is no longer swallowed — it means the policy was NOT attached, so it now fails the deploy instead of silently shipping a role with missing permissions. `attachRolePolicy` is idempotent for already-attached ARNs, so reconcile stays idempotent without the catch.

### Regression test
New live test in `Service.test.ts`: deploys an image-form service (`desiredCount: 0`, no ingress) declaring an AWS-managed policy, proves attachment out of band via `iam.listAttachedRolePolicies`, redeploys to prove the attach is idempotent (same `serviceArn`), and gone-proofs that destroy detaches the policy and deletes the task role.

### DIFF
--- a/packages/alchemy/src/AWS/ECS/Service.ts
+++ b/packages/alchemy/src/AWS/ECS/Service.ts
@@ -2792,6 +2792,20 @@ export const ServiceProvider = () =>
         const taskRoleArn =
           output?.taskRoleArn ??
           (yield* createTaskRoleIfNotExists({ id, roleName: taskRoleName }));
+
+        // `taskRoleManagedPolicyArns` is part of the inherited
+        // `TaskDefinitionConfig` surface — attach it exactly like the
+        // standalone Task provider does (its sibling
+        // `executionRoleManagedPolicyArns` is already honored below).
+        for (const policyArn of news.taskRoleManagedPolicyArns ?? []) {
+          yield* iam
+            .attachRolePolicy({
+              RoleName: taskRoleName,
+              PolicyArn: policyArn,
+            })
+            .pipe(Effect.catchTag("LimitExceededException", () => Effect.void));
+        }
+
         const executionRoleArn =
           output?.executionRoleArn ??
           (yield* ensureTaskExecutionRole({
--- a/packages/alchemy/src/AWS/ECS/Service.ts
+++ b/packages/alchemy/src/AWS/ECS/Service.ts
@@ -2794,17 +2794,21 @@ export const ServiceProvider = () =>
           (yield* createTaskRoleIfNotExists({ id, roleName: taskRoleName }));
 
         // `taskRoleManagedPolicyArns` is part of the inherited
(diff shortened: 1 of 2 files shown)


packages/alchemy/src/AWS/ECS/Service.ts
```diff
@@ -2792,6 +2792,20 @@ export const ServiceProvider = () =>
         const taskRoleArn =
           output?.taskRoleArn ??
           (yield* createTaskRoleIfNotExists({ id, roleName: taskRoleName }));
+
+        // `taskRoleManagedPolicyArns` is part of the inherited
+        // `TaskDefinitionConfig` surface — attach it exactly like the
+        // standalone Task provider does (its sibling
+        // `executionRoleManagedPolicyArns` is already honored below).
+        for (const policyArn of news.taskRoleManagedPolicyArns ?? []) {
```
sam-goodwin: Use Effect.all to run in parallel

packages/alchemy/src/AWS/ECS/Service.ts
```diff
@@ -2792,6 +2792,20 @@ export const ServiceProvider = () =>
...
+          yield* iam
+            .attachRolePolicy({
+              RoleName: taskRoleName,
+              PolicyArn: policyArn,
+            })
+            .pipe(Effect.catchTag("LimitExceededException", () => Effect.void));
```
sam-goodwin: Why swallow an error?

sam-goodwin (commented):
Missing regression test

packages/alchemy/src/AWS/ECS/Service.ts
bweis: Updated

packages/alchemy/src/AWS/ECS/Service.ts
bweis: LimitExceededException propagates instead of being swallowed now

bweis:
> Missing regression test

Added.

PR #941, opened 2026-07-24
fix(core): keep branded primitives opaque in ToOutput

sam-goodwin:
Branded primitives used in Action outputs exploded into `ObjectExpr` mapped over every `String`/`Number` method, because `string & Brand<"...">` is assignable to `object` through the brand intersection and `ToOutput` classified with `Extract<A, object>`.

```ts
type UserId = string & Brand.Brand<"UserId">;

const MintId = Action("MintId", Effect.fn(function* (_: { seed: string }) {
  return { userId: "u1" as UserId };
}));

const minted = yield* MintId({ seed: "s" });
minted.userId; // was: ObjectExpr<string & Brand<"UserId">> with charAt, slice, ...
               // now: Output<UserId>
```

Fix mirrors the existing `AttrOutput` guard in `Resource.ts` — short-circuit before the object check:

```diff
-export type ToOutput<A, Req = never> = [Extract<A, object>] extends [never]
-  ? Output<A, Req>
+export type ToOutput<A, Req = never> =
+  [A] extends [Primitive | Date]
+    ? Output<A, Req>
+    : [Extract<A, object>] extends [never]
+      ? Output<A, Req>
```

- Objects still upgrade to `ObjectExpr`, arrays to `ArrayExpr`; branded props reached through a nested `ObjectExpr` also stay opaque since it recurses through `ToOutput`.
- `Input<T>` needed no change — its `T extends Primitive` guard already matches branded primitives.
- Type test at `packages/alchemy/test/types/BrandedTypes.ts` covers `effect/Brand` strings, `Schema.brand` strings, branded numbers, nested access, arrays, `Input` assignability, and the Action scenario; all explosion assertions fail without the fix.

### DIFF
--- a/packages/alchemy/src/Output.ts
+++ b/packages/alchemy/src/Output.ts
@@ -12,7 +12,7 @@ import { RuntimeContext, sanitizeKey } from "./RuntimeContext.ts";
 import { Stack } from "./Stack.ts";
 import { Stage } from "./Stage.ts";
 import * as State from "./State/State.ts";
-import { isPrimitive } from "./Util/data.ts";
+import { isPrimitive, type Primitive } from "./Util/data.ts";
 
 const inspect = Symbol.for("nodejs.util.inspect.custom");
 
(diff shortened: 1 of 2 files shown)


kay-is:
@sam-goodwin thanks, that fixed my issue!

PR #944, opened 2026-07-25
fix(cloudflare/workers): scope Worker read route discovery to known zones

benpsnyder:
Worker `read` discovered zone routes by enumerating every zone on the account and then listing routes in each one. Routes are zone-scoped with no account-level enumeration API, so that is O(zones) requests on every read, issued at `concurrency: "unbounded"`. On an account with a few hundred zones a couple of plan/deploy cycles exhausts Cloudflare's per-user API budget and unrelated calls start coming back `429` / code `971` ([#926](https://github.com/alchemy-run/alchemy/issues/926)).

`reconcileRoutes` already avoids the sweep — it scopes itself to the zones implied by `desired ∪ previous`. `read` now uses the same bounded zone set.

```diff
-const readWorkerRoutes = (scriptName: string) =>
-  Effect.gen(function* () {
-    const { accountId } = yield* yield* CloudflareEnvironment;
-    const accountZones = yield* listAllZones(accountId).pipe(
-      Effect.catch(() => Effect.succeed([])),
-    );
-    return yield* listWorkerRoutesInZones(
-      scriptName,
-      accountZones.map((zone) => zone.id),
-    );
-  });
+const readWorkerRoutes = (
+  scriptName: string,
+  knownRoutes: Worker["Attributes"]["routes"] | undefined,
+) =>
+  listWorkerRoutesInZones(
+    scriptName,
+    (knownRoutes ?? []).map((route) => route.zoneId),
+  );
```

`listWorkerRoutesInZones` is also bounded now, since it still fans out one request per zone on the reconcile path.

This is complementary to #932 rather than a replacement. That PR gates the sweep on whether Alchemy manages routes for the Worker, which removes the cost for a plain `workers.dev` Worker; this removes the cost for a Worker that actually declares `routes`, where the gate passes and the account-wide fan-out still runs unchanged.

### Trade-off

A route attached out of band in a zone this Worker has never had a route in is no longer discovered by `read`. That is the blind spot `reconcileRoutes` already has, so the two paths are now consistent rather than one of them paying for account-wide discovery on every read.

### DIFF
--- a/packages/alchemy/src/Cloudflare/Workers/WorkerProvider.ts
+++ b/packages/alchemy/src/Cloudflare/Workers/WorkerProvider.ts
@@ -22,7 +22,7 @@ import { Stack } from "../../Stack.ts";
 import { sha256Object } from "../../Util/sha256.ts";
 import { CloudflareEnvironment } from "../CloudflareEnvironment.ts";
 import { CloudflareLogs } from "../Logs.ts";
-import { listAllZones, resolveZoneId } from "../Zone/lookup.ts";
+import { resolveZoneId } from "../Zone/lookup.ts";
 import {
   mergeAssetsConfigFiles,
   readAssets,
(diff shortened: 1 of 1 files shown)


packages/alchemy/src/Cloudflare/Workers/WorkerProvider.ts
```diff
@@ -810,28 +818,39 @@ export const LiveWorkerProvider = () =>
...
-        });
+      // Observe the routes attached to `scriptName` in the zones Alchemy
+      // already associates with this Worker.
+      //
+      // This used to enumerate every zone on the account and then list routes
+      // in each one. Routes are zone-scoped with no account-level enumeration
+      // API, so that costs O(zones) requests on *every* Worker read — on an
+      // account with a few hundred zones a couple of plan/deploy cycles
+      // exhausts Cloudflare's per-user API budget and unrelated calls start
+      // coming back 429 / code 971 (#926).
```
sam-goodwin, line 838: LOL holy shit.

benpsnyder: Did someone order a stress test?

PR #945, opened 2026-07-25
fix(aws): wait for ECS service eventual consistency

samnan-kradle:
## Summary

- make ECS service reconciliation wait for the exact task definition, completed rollout, settled desired/running/pending counts, old-task drain, and healthy ALB targets
- make deletion wait for task and target drain
- reconcile ECR tag mutability and scan-on-push for existing/adopted repositories
- harden AWS.state() buckets with versioning, encryption, bucket-owner ownership, and full public-access blocking
- keep every poll bounded, including a 30-minute ECS ceiling

## Validation

- `bun tsc -b`
- `bun run format:check`
- `bun run lint:providers`
- live AWS ECR reconciliation: 3/3 passed
- live AWS state-store convergence: 4/4 passed
- live AWS ECS service reconciliation/recovery: 4/4 passed across bounded sequential runs
- post-run inventory: zero `alchemy-test` ECR repositories, ECS clusters, or security-test S3 buckets

The live suites use an explicit `AWS_ACCOUNT_ID` resolved from STS because the current default AWS provider layer can deadlock while deriving it from the surrounding environment.

### DIFF
--- a/packages/alchemy/test/AWS/ECS/Task.smoke.test.ts
+++ b/packages/alchemy/test/AWS/ECS/Task.smoke.test.ts
@@ -3,6 +3,7 @@ import { SecurityGroup } from "@/AWS/EC2";
 import { Cluster } from "@/AWS/ECS/Cluster.ts";
 import { Service } from "@/AWS/ECS/Service.ts";
 import * as Test from "@/Test/Alchemy";
+import * as ecs from "@distilled.cloud/aws/ecs";
 import * as elbv2 from "@distilled.cloud/aws/elastic-load-balancing-v2";
 import { expect } from "alchemy-test";
 import * as Effect from "effect/Effect";
@@ -56,7 +57,13 @@ test.provider.skipIf(!process.env.AWS_TEST_SLOW || !!process.env.FAST)(
       }
       const publicSubnetIds = subnetIds.slice(0, 2);
 
-      const { url, targetGroupArn } = yield* stack.deploy(
+      const {
+        url,
+        targetGroupArn,
+        clusterArn,
+        serviceName,
+        taskDefinitionArn,
+      } = yield* stack.deploy(
         Effect.gen(function* () {
           // The default VPC supplies public routing and subnets; this managed
           // security group remains isolated to the test and is swept exactly.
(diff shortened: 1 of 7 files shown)


packages/alchemy/test/AWS/ECS/Task.smoke.test.ts
```diff
@@ -115,37 +122,39 @@ test.provider.skipIf(!process.env.AWS_TEST_SLOW || !!process.env.FAST)(
...
-          }),
-          Effect.tapError((error) => Effect.logError(error)),
-          // ~12 min budget: image pull + container boot + the ALB health check
-          // ramp (default 5 × 30s consecutive successes) can take several
-          // minutes for a cold Fargate task.
-          Effect.retry({ schedule: Schedule.spaced("12 seconds"), times: 60 }),
-        );
+      // Service.deploy now blocks until the exact task definition is the sole
+      // completed deployment, ECS counts have settled, and every registered
+      // target is healthy. Prove that contract immediately out-of-band.
```
sam-goodwin, line 137: This test change does a good job at indicating the problem solved by your change. We typically do this intentionally and prefer to resolve stabilization as far downstream as we can (for speed). But some times it is necessary to ensure a resource is stable before moving on. What did you run into specifically that triggered this change?

Geektrovert: Hey it's me the same person, I was just using my corpo acc.

Anyway, I was trying to get both health and live connectivity stats for all the services everytime I deploy to make sure "everything is working together" as the goal of IaC systems -  I tried to connect from my ECS server with third party services like Trigger, Neon, PostHog, etc. I'm currently using a patch of this on my local server.

I'm new to both Effect and Alchemy, I love the concept of both Effect & Alchemy, I'm trying to migrate from SST with a hell lotta custom CI/CD scripts, so that I can spend my time on some actual features not just infra, and if I'm doing something wrong please don't mind me directing to the better way of doing this.

samnan-kradle: The concrete failure was a release verifier reporting success against the previous ECS task: the replacement task started one second after verification had already finished. That allowed the workflow to attest the wrong image/task revision even though the service eventually converged.

For the ALB-backed API, the downstream capability probes must start only after ECS is serving the exact requested task definition, all previous tasks are drained, desired == running, pending == 0, and registered targets are healthy. The delete wait addresses the symmetric teardown race. This PR keeps the wait in the ECS resource boundary so every caller receives a truthful converged result; application-level probes still remain downstream and separate.

PR #963, opened 2026-07-27
feat(engine): first-class local vs live provider modes

sam-goodwin:
Make provider mode (local vs live) a first-class engine concept: the mode that reconciled a resource is persisted in state, switching modes plans a **replacement**, and every delete resolves the provider variant of the mode that created the row — so `alchemy deploy` right after `alchemy dev` cleanly replaces local instances instead of updating over them (and vice versa). On top of that engine, this PR brings the Cloudflare dev experience to parity: KV/R2/D1/Queues are emulated locally, `Alchemy.remote()` opts individual resources onto real Cloudflare — including queues, in both directions — and the Node-side capability clients reach the same local simulators the workers see.

### Mode resolution

- every state commit stamps `providerMode: "live" | "local"`; a mismatch with the resolved mode forces a replace (the provider diff is not consulted — diffing props across runtimes is meaningless)
- the mode comes from the command (`alchemy dev` → local, `alchemy deploy` → live); the one user knob is `Alchemy.remote()` — "run this live even in dev", captured at registration like `adopt()` (a no-op during deploy):

```ts
const queue = yield* Queue("Jobs", {}).pipe(Alchemy.remote());
```

- mode-agnostic providers (a single implementation) satisfy any requested mode, so dev runs over constructs that mix emulatable and live-only resources just work — they never stamp and never replace on a switch
- conflicting decorations on the same FQN die with `ConflictingProviderModeError`
- rows without a stamp are assumed to be the current run's mode (re-stamped on next write)

### `ProviderLayer.dual` (replaces `select`)

```ts
export const WorkerProvider = () =>
  ProviderLayer.dual(Worker, {
    live: () => LiveWorkerProvider(),
    local: () =>
      LocalWorkerProvider().pipe(Layer.provide(localRuntimeServices())),
  });
```

Registers **both** variants: the run-default eagerly, the other lazily (memoized via `Layer.fromBuildMemo` against the build's MemoMap), so a live deploy only constructs workerd machinery when a local-mode row actually needs deleting. `deleteOldGenerations`, GC chain draining, orphan deletes, `sync`, `tail`, and `logs` all resolve by the row's stamped mode.

### `LocalProvider` — standard long-running local providers

```ts
LocalProvider.make(Dev, serverEntryUrl, Effect.gen(function* () {
  return {
    // plain-data restart surface; canonically hashed for noop-vs-restart
    resolveConfig: (ctx) => Effect.succeed({ ... }),
    // boot in the provided Scope; return Attributes at readiness
    start: Effect.fn(function* ({ config, invalidate }) { ... }),
    // delete-only cleanup that intentionally outlives restarts
    stop: ({ id }) => Effect.void,
  };
}));
```

Generates `diff`/`reconcile`/`delete`/`list` from the instance registry (per-id lock, config-hash comparison, instanceId-guarded delete for replacement safety, `invalidate` for self-dying processes) and wraps `RpcProvider` so the registry lives in the dev sidecar across user-code reloads. `Command.Dev` and `LocalWorkerProvider` are rewritten on it.

### Local emulation for KV, R2, D1 (and inline-script workers)

`Namespace`, `Bucket`, and `Database` are now `ProviderLayer.dual`: in dev they fabricate `dev:`-prefixed ids with no cloud calls, and the worker's bindings lower onto the local workerd simulators. The providers carry **no promotion logic** — mode transitions are engine-orchestrated replacements (a live provider never sees a `dev:` id). D1 applies `migrationsDir` and `importFiles` against the local simulator through the platform proxy (multi-statement SQL over the raw `d1` service). Workers defined with inline `script:` (no `main`) bypass the bundler and serve/restart on script change; assets-only Workers serve locally through an ASSETS-delegating stub; new `dev: { cache, cf }` props control the local Cache API and `request.cf` blob.

### Node-side capability clients reach the local simulators

The `*Local` KV/R2 capability layers (and `QueryDatabaseLocal`) previously spoke only the cloud REST API. For `dev:` resources they now drive the local simulators through the runtime's **platform proxy** (cloudflare-tools#88, `getPlatformProxy` extracted from cloudflare-tools#74): the `*Local` layers reuse the same native-binding client builders the `*Binding` variants use, over per-operation proxied bindings — real ids keep the HTTP clients, dispatched per call from the lazily-resolved id. R2 rich objects (`R2ObjectBody` bodies, metadata) ride the proxy's `r2-object` encoding:

```ts
const Seed = Action("Seed", Effect.gen(function* () {
  const kv = yield* Cloudflare.KV.ReadWriteNamespace(ns);
  return Effect.fn(function* () {
    yield* kv.put("seeded", "from-action");   // lands in the local simulator
  });
}).pipe(Effect.provide(Cloudflare.KV.ReadWriteNamespaceLocal)));
```

Shared scaffolding in `Cloudflare/LocalGateway.ts` (`localGatewayRuntime` + the per-call mode dispatcher).

### `Alchemy.remote()` queues work from local dev — both directions

Cloudflare's preview/remote-binding sessions reject queue bindings at the platform level (cloudflare/workers-sdk#9929), so:

- **Producing**: when a LOCAL worker binds a LIVE queue, the eval registers a real **shim worker** (+ an `Alchemy.Random` bearer token) as ordinary engine-managed resources — the same idiom as `AccountApiToken`. The local binding relays workerd's queue wire protocol to the shim over HTTPS; the shim decodes it back into `send()`/`sendBatch()` on its real queue binding. Output-based ordering, stable-name updates, deletion on destroy, and orphan GC on promotion all come from the engine. v8-serialized messages are rejected with guidance to pass an explicit `contentType`.
- **Consuming**: the Consumer's local provider attaches an `http_pull` consumer to the real queue, and the runtime's pull loop drains it into the local broker, which delivers to the local `queue()` handler with the usual local batching/retry/DLQ semantics.

```
local worker ──▶ shim ──▶ real queue ──▶ pull loop ──▶ local queue() handler
```

The e2e test round-trips a message: produced through the binding, observed in the REAL queue, received by the local handler.

### Also in this PR

- the opt-out decorator is named `Alchemy.remote()` (matching wrangler's "remote bindings" vocabulary; the persisted mode enum stays `"live" | "local"`)
- `alchemy-test` runner: a timed-out test's hanging teardown no longer stalls the run — the body is interrupted with a bounded 10s grace period and abandoned (with a clear report) if it doesn't settle
- new `.local.test.ts` convention for dev-mode suites, plus `ModeCycle.test.ts`: the full local → hybrid (`remote()` within dev) → live → local cycle against the real cloud, including the demotion leg (live → local with real cloud deletions)
- `ProviderModePolicy` is a tri-state `Context.Reference` (unset / `remote()` / `remote(false)`), simplifying its consumers while keeping conflict detection
- `DurableObjectNamespace` live tests run concurrently (~450s → ~80s)

### Tests

23 plan/apply mode tests plus `LocalProvider.make` lifecycle unit tests. The 8 dev-mode suites (KV/R2/D1/Queues/CacheCf/Script/AssetsOnly/ModeCycle, 19 tests incl. `Alchemy.remote()` and Action-gateway coverage) run green repeatedly with retries disabled; companion `cloudflare-runtime` suite green (267 tests).


packages/alchemy/src/Cloudflare/D1/Database.ts
```diff
@@ -248,12 +258,19 @@ export type Database = Resource<
...
     diff: Effect.fn(function* ({ id, olds = {}, news = {}, output }) {
       const { accountId } = yield* yield* CloudflareEnvironment;
       if (!isResolved(news)) return undefined;
+      // A `dev:` id means the live database doesn't exist yet — update
+      // (create) rather than replace, even if name/account changed. The dev
+      // id is about to be replaced by a real one, so it must not be
+      // advertised as stable for this update.
+      if (output?.databaseId && !isLiveId(output.databaseId)) {
+        return { action: "update", stables: ["accountId"] } as const;
+      }
```
john-royal: I thought we wanted the engine to see these as distinct entities with a `replace` in between, orchestrated by the engine, so that way the `isLiveId` bit is no longer necessary. Did that not work out?

packages/alchemy/src/ProviderMode.ts
```diff
@@ -0,0 +1,118 @@
...
+ * resources registered while it is in context resolve the **live**
+ * provider even during `alchemy dev`.
+ *
+ * Apply it with the {@link live} combinator. During `alchemy deploy`
+ * everything is live anyway, so the policy only has an effect in dev.
+ */
+export class ProviderModePolicy extends Context.Service<
+  ProviderModePolicy,
+  boolean
+>()("ProviderModePolicy") {}
```
john-royal: Have you considered using `Context.Reference` instead of `Context.Service`? This way the default value is returned automatically if not specified in context.

packages/alchemy/src/Cloudflare/D1/LocalD1Gateway.ts
john-royal, line 1: This can probably be simplified using the [platform proxy](https://github.com/alchemy-run/cloudflare-tools/pull/74/changes#diff-72bed67ae201304bdb9fc17d43374a8763c882e6b8ee3c603729525d02ec8090) - implementation in that PR isn't perfect but can be adapted.

packages/alchemy/src/Cloudflare/D1/Database.ts
sam-goodwin: You were right — it did work out; this was vestigial. The engine plans a replacement on any stamped-mode mismatch (without consulting the provider diff) and routes each generation's delete to the provider of the mode that created it, so a live provider can never actually see a `dev:` id. Removed every `isLiveId` branch from the KV/R2/D1/Queue live providers in 427b5644f and replaced `DevPromotion.test.ts` with `ModeCycle.test.ts` — the full local → hybrid (`live()` within dev) → live → local cycle against the real cloud, including the previously-untested demotion leg. All green with zero promotion logic in the providers.

packages/alchemy/src/ProviderMode.ts
sam-goodwin: Done in 427b5644f — with one nuance: the some/none distinction was load-bearing (an explicit policy that disagrees with a prior registration dies with `ConflictingProviderModeError`; an absent one inherits). So it's a tri-state `Context.Reference<boolean | undefined>` defaulting to `undefined`, which keeps that distinction while collapsing the `serviceOption`/`Option` juggling at the consumers.

packages/alchemy/src/Cloudflare/D1/LocalD1Gateway.ts
sam-goodwin: Adopted — and it reached further than D1. I extracted the platform proxy from #74 into alchemy-run/cloudflare-tools#88 (adapted for post-#85 main; the main fix was a layer-lifetime bug where `Effect.provide` tore down the scoped services as soon as `open()` resolved), then migrated all three gateways in bccbdad1b:

- **D1**: `withLocalD1Query` opens a proxy with a raw `d1` service binding and POSTs the D1 HTTP protocol through the fetch passthrough — multi-statement migrations keep working, bespoke gateway worker gone.
- **KV/R2**: the REST-API emulation gateways are deleted entirely; for `dev:` ids the `*Local` layers now reuse the `*Binding` client builders over proxied native bindings (R2 rich objects needed a new `r2-object` protocol tag in #88 — body buffering + Node-side rehydration with bodyUsed semantics).

PR #968, opened 2026-07-28
fix(auth): update clack and pass cancel signal

BlankParticle:
Clack prompts were not being cancel properly at cleanup
update them and pass in abort signal

### DIFF
--- a/package.json
+++ b/package.json
@@ -180,5 +180,12 @@
     "turbo": "^2.9.16",
     "typescript": "catalog:",
     "yaml": "^2.8.2"
+  },
+  "devEngines": {
+    "packageManager": {
+      "name": "bun",
+      "version": "1.3.14",
+      "onFail": "download"
+    }
   }
 }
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -180,5 +180,12 @@
     "turbo": "^2.9.16",
     "typescript": "catalog:",
     "yaml": "^2.8.2"
+  },
+  "devEngines": {
+    "packageManager": {
+      "name": "bun",
+      "version": "1.3.14",
+      "onFail": "download"
+    }
```
BlankParticle, line 189: this accidently snuck in since vite-plus and others uses it, so vite plus auto added this
i guess we should keep it?

sam-goodwin: What does it do

BlankParticle: tells supported node/bun version managers/corepack or equivalent which bun version to use, if that version is not installed download before running
https://viteplus.dev/guide/env#overview

sam-goodwin: 1.3.14 has been notorious for seg faults, i am stuck on 1.3.13

PR #971, opened 2026-07-28
feat(aws/lambda): layer support

aryasaatvik:
Add an `AWS.Lambda.LayerVersion` resource and a `layers` prop on `AWS.Lambda.Function`. `Layers` was previously forwarded to `UpdateFunctionConfiguration` as an always-`undefined` field, so any layer configured out of band was cleared on the next deploy.

```ts
const deps = yield* AWS.Lambda.LayerVersion("Deps", {
  // a directory (packaged into a zip) or an existing .zip
  path: "./layers/deps",
  compatibleRuntimes: ["nodejs22.x"],
});

const fn = yield* AWS.Lambda.Function("Handler", {
  main: import.meta.resolve("./handler.ts"),
  layers: [deps.layerVersionArn],
});
```

Content comes from either `path` or `s3` (required above Lambda's 50 MB direct-upload limit):

```ts
const layer = yield* AWS.Lambda.LayerVersion("BigLayer", {
  s3: { bucket: yield* bucket.bucketName, key: "layers/big-layer.zip" },
});
```

### Notes

Layer versions are immutable, so `diff` returns `replace` on any content or publish-setting change and an explicit `noop` otherwise. The archive bytes are hashed into a `sourceHash` attribute rather than the source path, so identical content at a different path does not republish. `reconcile` reuses the version recorded in state when it still exists and still matches that hash, so an interrupted apply does not leak a duplicate version.

`Layers` is always sent explicitly:

```diff
+ // `UpdateFunctionConfiguration` treats an omitted `Layers` as "leave as-is",
+ // so removing the prop needs an explicit empty array to detach.
+ Layers: news.layers ?? [],
```

`Util/zip.ts` gains a `zipFiles` helper (sorted entries, fixed timestamp, per-file unix mode so executables keep `+x` under `/opt`). The existing `zipCode` is unchanged.

Layers carry no tags, so `list` enumerates every version of every layer in the ambient account/region without an ownership filter — same as the existing `Permission` provider.


packages/alchemy/src/AWS/Lambda/Function.ts
```diff
@@ -195,6 +195,13 @@ export interface FunctionProps extends PlatformProps {
    */
   architecture?: FunctionArchitecture;
   memorySize?: number;
+  /**
+   * Lambda layers to attach, as layer version ARNs — pass
+   * `layer.layerVersionArn` from an `AWS.Lambda.LayerVersion` (or a literal
+   * ARN, e.g. an AWS-managed layer). Layers are extracted into `/opt` in the
+   * order given. Omit or pass `[]` to detach every layer.
+   */
+  layers?: string[];
```
sam-goodwin: Can this be `(string | AWS.Lambda.LayerVersion)[]`

packages/alchemy/src/AWS/Lambda/LayerVersion.ts
```diff
@@ -0,0 +1,452 @@
...
+import { zipFiles, type ZipFile } from "../../Util/zip.ts";
+import type { Providers } from "../Providers.ts";
+
+/**
+ * Layer content stored in S3. Use this instead of {@link LayerVersionProps.path}
+ * for archives larger than the 50 MB direct-upload limit.
+ */
+export interface LayerVersionS3Content {
+  /** Name of the S3 bucket holding the layer archive. */
+  bucket: string;
```
sam-goodwin: `string | S3.Bucket`

packages/alchemy/src/AWS/Lambda/Function.ts
aryasaatvik: Done in f4c2613 — `layers?: LayerRef[]`, so `layers: [myLayer]` works alongside a raw ARN for AWS-managed layers.

```ts
export type LayerRef = string | { layerVersionArn: string };
```

I used the structural form to match the existing `AccessPointRef` convention just above it in this file. It accepts the `LayerVersion` resource (verified by the test, which now passes the resource directly) plus anything else exposing the attribute. Happy to spell out the nominal `AWS.Lambda.LayerVersion` instead if you prefer it for documentation value — it type-checks either way.

packages/alchemy/src/AWS/Lambda/LayerVersion.ts
aryasaatvik: Done in f4c2613 — `bucket: BucketRef`, so you can pass the bucket resource or a raw name.

```ts
export type BucketRef = string | { bucketName: string };
```

Structural for the same reason as the `layers` thread, plus it keeps the Lambda module from importing S3.

One thing that fell out of this: the resolved bucket name is now what gets hashed into `sourceHash`, not the raw props. Otherwise switching a bucket between resource form and name form would look like a content change and republish the layer version for no reason.

sam-goodwin: Add tests for both forms

packages/alchemy/src/AWS/Lambda/Function.ts
sam-goodwin: Add tests for both forms. Also cycle between them

aryasaatvik: Added in 2afb1b5 — a live test deploys both forms and cycles between them in both directions.

Cycling surfaced a real wart, which I think is what you were pointing at. `layers: [layer]` → `layers: [layer.layerVersionArn]` was planning as an **update**: the default props comparison sees a resource object and an ARN string as different props, so it re-issued `UpdateFunctionConfiguration` with a byte-identical `Layers` array. Function`s diff now compares `layers` normalized, so switching forms plans as a noop:

```
before   Plan: 1 to update, 2 to noop
after    Plan: 3 to noop
```

Everything else still falls through to the default comparison.

packages/alchemy/src/AWS/Lambda/LayerVersion.ts
aryasaatvik: Added in 2afb1b5 — the test publishes a layer from S3 with `bucket` as the resource, cycles to the raw name, and back. Both directions plan as a noop and keep the same `layerVersionArn`, since the resolved bucket name is what feeds `sourceHash`.

Also, retracting part of my earlier reasoning: I said structural "keeps the Lambda module from importing S3", which is wrong — `Lambda/BucketEventSource.ts` already imports `Bucket` from `../S3/Bucket.ts`, and `Function.ts` imports IAM. Type-only cross-service imports are normal here, so that was not a real constraint. The precedent argument stands on its own (`AccessPointRef` in `Function.ts`), and the union accepts the resource either way — but happy to switch to the nominal types if you prefer them.

PR #972, opened 2026-07-28
feat(docker): introduce docker swarm resources

sbekrin:
Adds [Docker Swarm](https://docs.docker.com/engine/swarm/) support, from provisioning the swarm to running effectful services on it:

- `Docker.Swarm` — idempotent [`docker swarm init`](https://docs.docker.com/engine/swarm/swarm-mode/) on a target engine; adoptable when the engine is already a swarm manager
- `Docker.Context` — a named [Docker CLI context](https://docs.docker.com/engine/manage-resources/contexts/) (local socket, SSH host, TCP endpoint); existing Docker resources accept it via their `context` prop
- `Docker.Service` — a replicated [swarm service](https://docs.docker.com/engine/swarm/services/), promoted to an **effectful platform** like `AWS.ECS.Service`

### Provision the swarm, deploy into it

```typescript
const swarm = yield* Docker.Swarm("swarm"); // idempotent `docker swarm init`

const web = yield* Docker.Service("web", {
  context: swarm, // orders the service after the swarm + inherits its engine
  image: "nginx:alpine",
  replicas: 2,
  ports: [{ external: 8080, internal: 80 }],
});
```

`Service` and overlay `Network` take the swarm (or a `Docker.Context` / context name) as `context`. An engine that is already in swarm mode is foreign until `adopt(true)`; destroying the swarm resource runs `docker swarm leave --force`. Multi-node membership (join tokens) remains host-level CLI setup for now.

### Effectful services

Pass `main` and an Effect impl instead of a pre-built image. Alchemy bundles the program, builds a content-addressed image against the swarm's engine, and deploys it as a swarm service:

```typescript
import * as Docker from "alchemy/Docker";
import { ServerHost } from "alchemy/Server/Process";

const Swarm = Docker.Swarm("swarm");

export default class Api extends Docker.Service<Api>()(
  "Api",
  // Class props may be an Effect: yield the swarm the service deploys into.
  Effect.gen(function* () {
    const swarm = yield* Swarm;
    return {
      context: swarm,
      main: import.meta.url,
      port: 3000,
      ports: [{ external: 8080, internal: 3000 }], // published via the routing mesh
      replicas: 2,
    };
  }),
  Effect.gen(function* () {
    const host = yield* ServerHost;
    yield* host.run(myBackgroundLoop); // long-running loops inside each replica

    return {
        return yield* HttpServerResponse.json({ ok: true });
      }),
    };
  }),
) {}
```

- the `{ fetch }` handler is served by a Bun HTTP server on `port`; `ports` publishes it through the [swarm routing mesh](https://docs.docker.com/engine/swarm/ingress/)
- the image is content-addressed (`<name>:<hash>`): unchanged code redeploys as a noop, changed code replaces the service with the new image
- built on the target engine's local store, so single-node swarms need no registry; multi-node swarms build with `Docker.Image` + `registry` and pass the pushed ref as `image`

### Tests

All deploy-through against real engines: the `Docker.Swarm` lifecycle (greenfield init, adopt-vs-refuse, destroy) runs on a disposable `docker:dind` engine so the host swarm is never touched; the `Docker.Service` suites self-provision a local single-node swarm and cover create/replace/adopt plus an end-to-end effectful service (`fetch` served, `ServerHost.run` loop observed over HTTP, noop redeploy, clean destroy).

See the step-by-step [Docker Swarm guide](https://github.com/sbekrin/alchemy/blob/docker-swarm/website/src/content/docs/docker/swarm.mdx) covering local dev through remote SSH deployment.


sam-goodwin:
I updated your PR to include Effectful support for Docker.Service

PR #976, opened 2026-07-28
feat(aws/ses): SESv2 coverage — identity policies, contacts, tenants, dedicated IPs, account settings

aryasaatvik:
Follows #939 (receiving) with the SESv2 sending half: 9 resources and 5 runtime bindings, plus the `EmailIdentity` and `ConfigurationSet` props that were declared but never synced.

```ts
const identity = yield* SES.EmailIdentity("Sender", {
  emailIdentity: "mail.example.com",
  dkimSigningEnabled: true,
  mailFromDomain: "bounce.mail.example.com", // newly synced
});

yield* SES.EmailIdentityPolicy("AllowPartner", {
  emailIdentity: identity.emailIdentity,
  policy: {                                  // typed PolicyDocument, not a JSON string
    Version: "2012-10-17",
    Statement: [{
      Effect: "Allow",
      Principal: { AWS: "arn:aws:iam::111122223333:root" },
      Action: ["ses:SendEmail"],
      Resource: identity.identityArn,
    }],
  },
});

const list = yield* SES.ContactList("Newsletter", {
  topics: [{ TopicName: "product", DisplayName: "Product updates", DefaultSubscriptionStatus: "OPT_IN" }],
});

yield* SES.Contact("Subscriber", {
  contactListName: list.contactListName,
  emailAddress: "reader@example.com",
  topicPreferences: [{ TopicName: "product", SubscriptionStatus: "OPT_IN" }],
  attributes: { plan: "pro" },               // object; drift compared on canonical JSON
});
```

Props that SES only accepts together are nested, so half-specified state is unrepresentable:

```ts
yield* SES.Tenant("CustomerA", {
  suppression: { reasons: ["BOUNCE"], scope: "TENANT" },
});

yield* SES.ConfigurationSet("Tracked", {
  tracking: { customRedirectDomain: "links.example.com", httpsPolicy: "REQUIRE" },
  vdm: { dashboardEngagementMetrics: "ENABLED" },
});

yield* SES.AccountSettings("Account", { vdm: { enabled: "ENABLED" } });
yield* SES.MultiRegionEndpoint("Global", { regions: ["eu-west-1"] });
```

### Resources

`EmailIdentityPolicy`, `ContactList`, `Contact`, `CustomVerificationEmailTemplate`, `DedicatedIpPool`, `Tenant`, `TenantResourceAssociation`, `MultiRegionEndpoint`, `AccountSettings`.

- Only props you set are managed — omitted props leave the cloud setting untouched, and the JSDoc says so per prop.
- `EmailIdentity` syncs `dkimSigningEnabled`, `feedbackForwardingEnabled`, and `mailFromDomain` + `mailFromBehaviorOnMxFailure`; `ConfigurationSet` gains `tracking` and `vdm`. `putConfigurationSetVdmOptions` replaces `VdmOptions` wholesale, so the member the caller doesn't manage is backfilled from observed state.
- `ContactList` replaces **delete-first** — SES allows one contact list per account, so create-then-delete would exceed the limit.
- `DedicatedIpPool`, `MultiRegionEndpoint`, and `CustomVerificationEmailTemplate` brand ownership tags on create, and `read` returns `Unowned` for an existing resource that lacks them — taking one over requires `--adopt`. All three expose a `tags` prop. `MANAGED → STANDARD` pool scaling has no API and replaces.
- `MultiRegionEndpoint` provisions asynchronously: reconcile returns at `CREATING`, and `delete` retries `ConcurrentModificationException` until the endpoint settles.
- `AccountSettings` is the account/region singleton (`nuke: { singleton: true }`). Delete is a documented no-op — there is no safe default to reset account-wide toggles to.
- Child resources enumerate through their parent — `Contact` walks contact lists, `EmailIdentityPolicy` walks identities, `TenantResourceAssociation` walks tenants — and declare `nuke.dependsOn` so parents outlive them (associations also order before the config sets, identities, and templates they reference, which SES refuses to delete while associated).

### Bindings

`SendCustomVerificationEmail`, `GetMessageInsights`, `BatchGetMetricData`, `GetDomainStatisticsReport`, `GetBlacklistReports` — each with an `*Http` layer.

`SendCustomVerificationEmail(identity, configurationSet?)` scopes `ses:SendCustomVerificationEmail` to the identity being verified — SES authorizes this action against the recipient address the call creates, so the bound identity (a domain, or a single address; an `EmailIdentity` resource or a plain `{ emailIdentity }` reference) bounds who the function may start verification for, rather than `Resource: "*"`. The other four are account-level read-only operations.

### Not implemented

Export/import jobs and deliverability test reports (job-shaped, not resource-shaped), `putAccountDetails` (needs an AWS support review), `putDeliverabilityDashboardOption` ($1,250/month subscription), the dedicated-IP data plane, and reputation-entity ops.

Guide: `/aws/email/sending`.


sam-goodwin:
badass contributions man!

sam-goodwin:
> Test suites are authored per repo conventions (deterministic names, destroy-first-and-last, replacement coverage, exclusive capture-and-restore for the account singleton, out-of-band distilled verification) but have not yet run against a live account — draft until they do.

Do you plan to test this?

aryasaatvik:
> > Test suites are authored per repo conventions (deterministic names, destroy-first-and-last, replacement coverage, exclusive capture-and-restore for the account singleton, out-of-band distilled verification) but have not yet run against a live account — draft until they do.
> 
> Do you plan to test this?

Yes. planning on running suites run against a sandbox test account; the motivating use case is a multi-tenant email platform on a production-access account, which is where the production-gated paths get exercised for real.

| Surface | Verification | Why |
|---|---|---|
| `EmailIdentityPolicy`, `ContactList`/`Contact`, `CustomVerificationEmailTemplate`, `EmailIdentity` + `ConfigurationSet` aspects, `Tenant`, `TenantResourceAssociation`, `AccountSettings` | Live, sandbox | Free control-plane; tenants bill $0.005/mo each |
| `MultiRegionEndpoint`, `DedicatedIpPool` (`STANDARD`, no IPs) | Live, sandbox — ungating these | No provisioning charge; the $24.95/mo is per IP |
| `AccountSettings` sending toggle | Live, sandbox — pinned to an otherwise-unused region, `exclusive`, restore in a finalizer | `sendingEnabled` is per-region, so region isolation bounds the blast radius |
| `BatchGetMetricData`, `GetMessageInsights` | Live, sandbox with VDM enabled | VDM has no fixed fee — $0.07/1k emails is nothing at test volume. Insights lag a few minutes, so that one may gate on latency |
| `DedicatedIpPool` (`MANAGED`) | Gated `AWS_TEST_SES_DEDICATED_IP=1` | Managed IPs bill $15/mo/account once enabled |
| `SendCustomVerificationEmail` | Gated `AWS_TEST_SES_PRODUCTION=1`, production account | Blocked in sandbox; sends a real email, so it goes to an address we own and cleans up the pending identity |
| `GetDomainStatisticsReport`, `GetBlacklistReports` | Probe only, never enabled | Deliverability Dashboard: $1,250/mo per region per account |

Error unions for the probe cases already carry `BadRequestException`/`NotFoundException`, so no distilled patch expected — I'll pin whichever tag comes back.

aryasaatvik:
> > Test suites are authored per repo conventions (deterministic names, destroy-first-and-last, replacement coverage, exclusive capture-and-restore for the account singleton, out-of-band distilled verification) but have not yet run against a live account — draft until they do.
> 
> Do you plan to test this?

ran tests one sandbox and live account and updated the description

website/src/content/docs/aws/email/sending.mdx
```diff
@@ -0,0 +1,409 @@
...
+    Version: "2012-10-17",
+    Statement: [
+      {
+        Effect: "Allow",
+        Principal: { AWS: "arn:aws:iam::111122223333:root" },
+        Action: "ses:SendEmail",
+        Resource: identity.identityArn,
+      },
+    ],
+  }),
```
sam-goodwin: This should not be a string. It should be a well typed object. Look on IAM module. Add a new type if the policy is different

website/src/content/docs/aws/email/sending.mdx
```diff
@@ -0,0 +1,409 @@
...
+
+`AccountSettings` is an account/region singleton for account-wide
+sending status, the suppression list, and Virtual Deliverability
+Manager (VDM). Only the aspects you specify are managed — omit the
+rest to leave them untouched:
+
+```typescript
+const settings = yield* SES.AccountSettings("Account", {
+  vdm: {
+    vdmEnabled: "ENABLED",
```
sam-goodwin: should this be `enabled` since it's already in a `vdm` property.

aryasaatvik: Agreed — renamed it to `vdm.enabled`.

website/src/content/docs/aws/email/sending.mdx
aryasaatvik: Updated to use IAM `PolicyDocument`, serialized only at the SES API boundary.

packages/alchemy/src/AWS/SES/Contact.ts
```diff
@@ -0,0 +1,214 @@
...
+          );
+      });
+
+      return Contact.Provider.of({
+        stables: ["contactListName", "emailAddress"],
+
+        // Contacts are keyed by their parent list; there is no flat
+        // account-level enumeration, and deleting the list removes its
+        // contacts, so nuke handles them via the parent contact list.
+        list: () => Effect.succeed([]),
```
sam-goodwin: Can't you list all parent lists and walk them all?

packages/alchemy/src/AWS/SES/CustomVerificationEmailTemplate.ts
```diff
@@ -0,0 +1,199 @@
...
+      });
+
+      return CustomVerificationEmailTemplate.Provider.of({
+        stables: ["templateName"],
+
+        // Account/region-scoped: enumerate every template so leaked test
+        // resources are cleaned by nuke. Custom verification email templates
+        // carry no ownership signal, so existence is treated as ownership.
+        list: () =>
+          Effect.gen(function* () {
```
sam-goodwin: nit: use Effect.fn for better stack traces

packages/alchemy/src/AWS/SES/EmailIdentityPolicy.ts
```diff
@@ -0,0 +1,203 @@
...
+          );
+      });
+
+      return EmailIdentityPolicy.Provider.of({
+        stables: ["emailIdentity", "policyName"],
+
+        // Policies are keyed entirely by their parent identity; there is no
+        // flat account-level enumeration, and deleting the identity removes
+        // its policies, so nuke handles them via the parent identity.
+        list: () => Effect.succeed([]),
```
sam-goodwin: I'm concerned about this. We use list for our nuke command that wipes accounts. 

Why can't we list these? How will they be mass deleted in nuke?

packages/alchemy/src/AWS/SES/Tenant.ts
```diff
@@ -0,0 +1,268 @@
...
+ * const tenant = yield* SES.Tenant("CustomerA", {});
+ * ```
+ *
+ * @example Tenant with a Scoped Suppression List
+ * ```typescript
+ * const tenant = yield* SES.Tenant("CustomerA", {
+ *   suppressedReasons: ["BOUNCE", "COMPLAINT"],
+ *   suppressionScope: "TENANT",
+ * });
+ * ```
```
sam-goodwin, line 104: Are these examples comprehensive?

packages/alchemy/test/AWS/SES/Bindings.test.ts
```diff
@@ -164,39 +181,35 @@ describe("SES Bindings", () => {
...
-          const response = (yield* send(
-            HttpClientRequest.post(
-              `${baseUrl}/send-plain?from=${encodeURIComponent(VERIFIED_FROM!)}`,
-            ),
-          ).pipe(Effect.flatMap((r) => r.json))) as {
-            messageId?: string;
-            error?: string;
-          };
-          expect(response.error).toBeUndefined();
-          expect(response.messageId).toBeTruthy();
```
sam-goodwin, line 199: Are these tests no longer relevant?

packages/alchemy/test/AWS/SES/Bindings.test.ts
```diff
@@ -425,4 +438,159 @@ describe("SES Bindings", () => {
...
+          } else if (response.error === undefined) {
+            expect(typeof response.results).toBe("number");
+          } else {
+            expect(typeof response.error).toBe("string");
+          }
+        }),
+    );
+  });
+
+  describe("GetDomainStatisticsReport", () => {
```
sam-goodwin, line 667: I notice each block only has one test. Are we testing all cases or only happy paths?

sam-goodwin (changes requested):

sam-goodwin:
Checking in here.

aryasaatvik:
On it, addressing the feedback today.

packages/alchemy/src/AWS/SES/Contact.ts
aryasaatvik: Done — `list` now walks `ListContactLists` → `ListContacts` per list, and Contact declares `nuke.dependsOn: ["AWS.SES.ContactList"]` so nuke drains contacts before their list.

packages/alchemy/src/AWS/SES/CustomVerificationEmailTemplate.ts
aryasaatvik: Done, and swept the PR — all nine SES providers' `list` ops converted (the one lifecycle member still arrow-wrapped).

packages/alchemy/src/AWS/SES/EmailIdentityPolicy.ts
aryasaatvik: They are enumerable — `ListEmailIdentities` → `GetEmailIdentityPolicies` per identity, same walk for TenantResourceAssociation, plus `nuke.dependsOn` on the parents. Also fixed the adjacent pool case: `DedicatedIpPool.list` reads the real `ScalingMode` now, and `read` returns `Unowned` for pools without our tags — adopting a foreign (billable) pool now requires `--adopt`.

packages/alchemy/src/AWS/SES/Tenant.ts
aryasaatvik: They weren't — added resource association across all three `ResourceType`s and tenant-scoped suppression, and applied the same bar to the other resources (pool→config-set attachment, `sendingEnabled`, the new identity/config-set props).

packages/alchemy/test/AWS/SES/Bindings.test.ts
aryasaatvik: The two removed tests were `AWS_TEST_SES_FROM`-gated and unrunnable by construction: `SendEmail(identity, configSet)` scopes IAM to the bound identity's ARN, so an outside sender is denied before SES sees the message. Replaced by an ungated test asserting exactly that denial.

On happy paths — agreed. The binding tests now pin exact typed tags plus `not("AccessDeniedException")` (which is what proves each scoped policy), with a failure-mode case per binding. Two findings from live probing: VDM-disabled accounts return `NotFoundException` before input validation on GetMessageInsights/BatchGetMetricData (those cases are VDM-gated with the reason inline), and GetBlacklistReports has no rejection path at all — any string returns an empty report — so that's documented instead of faked.

packages/alchemy/test/AWS/SES/Bindings.test.ts
aryasaatvik: Addressed together with the thread above — typed tags + a failure-mode case per binding.

PR #981, opened 2026-07-29
docs(cloudflare): dedicated gradual deployments guide

sam-goodwin:
Add a dedicated guide for Worker versions & gradual rollouts at `/cloudflare/compute/gradual-deployments`, covering the `version` prop shipped in #948 plus the surrounding platform workflow from [Cloudflare's gradual deployments docs](https://developers.cloudflare.com/workers/configuration/versions-and-deployments/gradual-deployments/):

- gradual rollout of a Worker's own deploys (`version: { traffic: N }`), ramping, promotion, rollback
- canarying another stage's Worker via `version.parent` + `Worker.ref`
- traffic-free preview versions with stable aliased preview URLs
- smoke testing with `Cloudflare-Workers-Version-Overrides` (using the `versionId` output)
- version affinity with `Cloudflare-Workers-Version-Key`
- runtime verification via the `version_metadata` binding and `version.tag`
- platform limits: two versions per deployment, 100-version window, Durable Object version assignment, version-scoped cache (`crossVersionCache`)

Also registers the guide in the Compute sidebar and cross-links it from the Workers page.

### DIFF
--- a/website/src/content/docs/cloudflare/compute/gradual-deployments.mdx
+++ b/website/src/content/docs/cloudflare/compute/gradual-deployments.mdx
@@ -1,0 +1,220 @@
+---
+title: Gradual deployments
+description: Roll out Worker deploys incrementally — upload preview versions, canary a percentage of live traffic, ramp to 100%, pin users to a version during the rollout, and verify which version served each request.
+---
+
+Every deploy of a Worker uploads an immutable
+[version](https://developers.cloudflare.com/workers/configuration/versions-and-deployments/)
+— a snapshot of code, bindings, and compatibility settings — and a
+*deployment* routes traffic across up to two versions by percentage.
+By default alchemy deploys each new version at 100%, an immediate
+cutover. The `version` prop on `Cloudflare.Worker` unlocks the other
+shapes: gradual rollouts of a Worker's own deploys, canary versions of
+another stage's Worker, and traffic-free preview versions.
+
+## Roll out a deploy gradually
+
+On a Worker that owns its script, set `version.traffic` below 100. The
+deploy uploads the new version and creates a deployment that routes
+that share of traffic to it, while the previously-live version keeps
+the remainder:
+
+```typescript
+// alchemy.run.ts
+yield* Cloudflare.Worker("Api", {
+  main: "./src/api.ts",
+  version: {
+    traffic: 10,
+    message: "new search index",
+  },
+});
+```
+
+Each request is routed independently according to the percentages.
+Raise the number and re-deploy to ramp — 10, then 25, then 50. Remove
+the prop (or set `traffic: 100`) to promote the version to all
+traffic.
+
+The very first deploy of a script always goes to 100%, since there is
+no previous version to split with. `traffic: 0` uploads the version
+without touching the live deployment at all — the equivalent of
+`wrangler versions upload`.
+
+## Roll back
+
+A rollout is just a deployment pointing at two versions, so rolling
+back is another deploy. Re-deploy the previous code — alchemy uploads
+it as a new version at 100% (or at a `traffic` split of its own).
+Because versions are immutable snapshots, the previous behavior comes
+back exactly, including its bindings and compatibility settings.
+
+## Canary another stage's Worker
+
+Set `version.parent` to upload this Worker's code as a version of
+*another* Worker's script instead of creating a script of its own. The
+parent is usually the same Worker deployed in another stage,
+referenced with `Worker.ref`:
+
+```typescript
+const parent = yield* Cloudflare.Worker.ref("Api", { stage: "prod" });
+
+yield* Cloudflare.Worker("ApiCanary", {
+  main: "./src/api.ts",
+  version: { parent, traffic: 10 },
+});
+```
+
+The canary takes 10% of the parent's live traffic and the live version
+keeps the rest. Destroying the canary restores 100% of traffic to the
+live version. Cloudflare deployments split between at most two
+versions, so one canary can be active per script at a time.
+
+## Preview versions
+
+With `parent` set and no `traffic` (it defaults to 0), the version
+receives no traffic at all — it is reachable only at its preview URL.
+This is the PR-preview workflow:
+
+```typescript
+const parent = yield* Cloudflare.Worker.ref("Api", { stage: "staging" });
+
+const preview = yield* Cloudflare.Worker("Api", {
+  main: "./src/api.ts",
+  version: { parent, message: `PR #${process.env.PR_NUMBER}` },
+});
+```
+
+`preview.url` is the version's *aliased* preview URL
+(`<alias>-<name>.<subdomain>.workers.dev`). Alchemy derives a stable
+alias from the stack, stage, and logical id (override it with
+`version.alias`), so the link stays the same across re-deploys and
+always serves the latest uploaded version. The version carries its own
+bindings, so a preview stage binds its own KV namespaces, D1
+databases, and secrets for a fully isolated environment.
+
+Preview URLs require the script's workers.dev subdomain, which is
+enabled by default. To keep the stable workers.dev URL off while each
+version stays reachable at its preview URL, set
+`workersDev: { enabled: false, previewsEnabled: true }` on the parent.
+
+## Smoke test before shifting traffic
+
+The `Cloudflare-Workers-Version-Overrides` request header pins a
+request to a specific version, as long as that version is part of the
+current deployment. Deploy the new version at a small percentage, then
+drive test traffic at it directly using the `versionId` output:
+
+```typescript
+const api = yield* Cloudflare.Worker("Api", {
+  main: "./src/api.ts",
+  version: { traffic: 1 },
+});
+
+return { url: api.url.as<string>(), versionId: api.versionId };
+```
+
+```sh
+curl -s https://api.example.com/health \
+  -H 'Cloudflare-Workers-Version-Overrides: my-api="<versionId>"'
+```
+
+The header is a dictionary keyed by Worker *name* (the physical script
+name, not the logical id) with the version id as a quoted string. If
+the version is not in the current deployment or the header is
+malformed, the request falls back to the deployment's percentages.
+Overrides also apply to `fetch()`-based service binding calls —
+forward the incoming request (or set the header on the subrequest) to
+keep cooperating Workers on matching versions during a rollout.
+
+## Keep users on one version
+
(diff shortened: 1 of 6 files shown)


website/src/content/docs/cloudflare/compute/gradual-deployments.mdx
```diff
@@ -0,0 +1,320 @@
...
+With `parent` set and no `traffic` (it defaults to 0), the version
+receives no traffic at all — it is reachable only at its preview URL.
+This is the PR-preview workflow:
+
+```typescript
+const parent = yield* Cloudflare.Worker.ref("Api", { stage: "staging" });
+
+const preview = yield* Cloudflare.Worker("Api", {
+  main: "./src/api.ts",
+  version: { parent, message: `PR #${process.env.PR_NUMBER}` },
```
ericclemmons, line 130: Can this be omitted for preview deployments entirely?

The current behavior for preview deployments is to be their own `stage`, right?

I'm just wanting to understand when I'd do a "normal preview deploy" (with no traffic) vs. a "preview deploy with a parent".

website/src/content/docs/cloudflare/compute/gradual-deployments.mdx
```diff
@@ -0,0 +1,320 @@
...
+        default: "100"
+
+jobs:
+  deploy:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - run: bun alchemy deploy
+        env:
+          API_TRAFFIC: ${{ inputs.traffic }}
```
ericclemmons, line 88: This is a *fantastic* use-case and directly matches the need for health-mediated deployments.

website/src/content/docs/cloudflare/compute/gradual-deployments.mdx
```diff
@@ -0,0 +1,320 @@
...
+that share of traffic to it, while the version from the previous
+deploy keeps the remainder:
+
+```typescript
+// alchemy.run.ts
+yield* Cloudflare.Worker("Api", {
+  main: "./src/api.ts",
+  version: {
+    traffic: 10,
+    message: "new search index",
```
ericclemmons, line 27: Message is optional, right?  May be worth calling out what `message` is used for and how it appears. (This is just what we see in the dashboard, right?  Usually it's a commit message, but in this case it's overridden here)

website/src/content/docs/cloudflare/compute/gradual-deployments.mdx
```diff
@@ -0,0 +1,320 @@
...
+  },
+});
+```
+
+Each request is routed independently according to the percentages.
+Raise the number and re-deploy to ramp — 10, then 25, then 50. Remove
+the prop (or set `traffic: 100`) to promote the version to all
+traffic.
+
+The very first deploy of a script always goes to 100%, since there is
```
ericclemmons: Very good default of `100%`.  This is actually a gotcha with `wrangler versions upload` - you'll get an error unless you `deploy` 1st!

website/src/content/docs/cloudflare/compute/gradual-deployments.mdx
```diff
@@ -0,0 +1,320 @@
...
+majority — so deploying at 10 and then at 20 does not stack two
+canaries:
+
+1. Deploy at `traffic: 10` → new version **v2** at 10%, stable **v1**
+   at 90%.
+2. Deploy at `traffic: 20` → new version **v3** at 20%, split against
+   the majority holder **v1** at 80%. **v2** drops out of routing
+   entirely.
+
+The abandoned version still exists as an upload, but it receives no
```
ericclemmons: TIL!

website/src/content/docs/cloudflare/compute/gradual-deployments.mdx
```diff
@@ -0,0 +1,320 @@
...
+
+The `Cloudflare-Workers-Version-Overrides` request header pins a
+request to a specific version, as long as that version is part of the
+current deployment. Deploy the new version at a small percentage, then
+drive test traffic at it directly using the `versionId` output:
+
+```typescript
+const api = yield* Cloudflare.Worker("Api", {
+  main: "./src/api.ts",
+  version: { traffic: 1 },
```
ericclemmons, line 163: This would be 1%, right?  Because not setting `version` would be 100% of the _new_ version. What would 0% do?  Only respond to my `curl` requests (like the preview examples)?

I'm thinking of this example being a manual deploy or HMD for smoke testing.

```shell
API_PERCENTAGE=1 alchemy deploy
# outputs versionId

curl -s https://api.example.com/health \
  -H 'Cloudflare-Workers-Version-Overrides: my-api="<versionId>"'
# OK

API_PERCENTAGE=50 alchemy deploy

curl -s https://api.example.com/health \
  -H 'Cloudflare-Workers-Version-Overrides: my-api="<versionId>"'
# UNHEALTHY

alchemy rollback # or whatever the guidance would be. Setting it to 0%?
```

website/src/content/docs/cloudflare/compute/gradual-deployments.mdx
```diff
@@ -0,0 +1,320 @@
...
+`Cloudflare-Workers-Version-Key` request header to a stable identifier
+for that user — Cloudflare hashes the key and assigns a version
+deterministically from the configured percentages:
+
+```sh
+curl -s https://api.example.com -H 'Cloudflare-Workers-Version-Key: user-123'
+```
+
+On a zone you control, a
+[Transform Rule](https://developers.cloudflare.com/rules/transform/request-header-modification/)
```
ericclemmons: Is there an Alchemy equivalent of this?  For gradual deploys, "sticky deploys" and "version affinity" is a tricky technical topic for a lot of users.

Being able to easily say, "oh yeah, and sticky by cookie/claim, or fallback to sticky IP" would _really_ complete this story.

website/src/content/docs/cloudflare/compute/gradual-deployments.mdx
```diff
@@ -0,0 +1,320 @@
...
+      fetch: Effect.gen(function* () {
+        const { id, tag } = yield* versionMetadata;
+        return yield* HttpServerResponse.json({ ok: true, version: id, tag });
+      }),
+    };
+  }).pipe(Effect.provide(Cloudflare.Workers.VersionMetadataBinding)),
+);
+```
+
+Set `version.tag` at deploy time (a git SHA is the usual choice) and
```
ericclemmons: Really good usage of `VersionMetadata()` 👏

website/src/content/docs/cloudflare/compute/gradual-deployments.mdx
```diff
@@ -0,0 +1,320 @@
...
+  on the parent and reference them
+  [cross-script](/cloudflare/compute/cross-worker-durable-object)
+  instead.
+- Workers that implement Durable Objects do not get preview URLs.
+- [Workers Cache](/cloudflare/compute/cache) is scoped to a single
+  version by default, so every version starts cold. Set
+  `crossVersionCache: true` to share cached responses across versions
+  during a rollout.
+
+## Containers roll out by instance, not by request
```
ericclemmons, line 295: May be worth a similar callback for things like Workflows, logpush, Durable Object migrations, etc.

These need to basically be deployed in a backwards/forwards-compatible way:
1. Deploy the new service (e.g. Workflow) or new handler in a DO.
2. Gradually rollout the worker that uses the new call paths **while the old version continues calling the current paths**.
3. Once at 100%, you can safely remove the old call paths.

ericclemmons (approved):
Absolutely no blockers, this is great!

Only clarifying questions that AI can probably fix up.

This is a confusing topic with several gotchas, but this is a very, _very_ good addition for Alchemy and closes a parity gap with `wrangler`.

website/src/content/docs/cloudflare/compute/gradual-deployments.mdx
sam-goodwin: Yes — `message` is optional. It is the human-readable note shown next to the version in the Cloudflare dashboard and `wrangler versions list` (a commit message is the usual choice); `tag` is its machine-readable sibling. The example now shows both as comments: `message: "new search index"` and `tag: process.env.GITHUB_SHA`, and the tag round-trips at runtime through the version metadata binding.

website/src/content/docs/cloudflare/compute/gradual-deployments.mdx
sam-goodwin: Right — a plain preview stage needs no `version` at all and stays the default workflow. Added a "Preview version or preview stage?" section: a stage is fully independent infrastructure (own script, own copies of every resource), while a preview version rides the parent's script, settings, and version history without adding scripts to the account. Rule of thumb in the guide: stage for infrastructure changes, version for cheap shareable code previews.

website/src/content/docs/cloudflare/compute/gradual-deployments.mdx
sam-goodwin: Yes, exactly 1% — and your flow works as written. `traffic: 0` is upload-only: the version never enters the deployment, so an override header cannot reach it (overrides only resolve versions in the current deployment); it is reachable only at its preview URL. `traffic: 1` is the smallest override-addressable deployment. The guide now says this inline in the smoke-test snippet, and the abort guidance is: re-deploy the last good ref at the default 100 — the bad version drops out of the deployment entirely.

website/src/content/docs/cloudflare/compute/gradual-deployments.mdx
sam-goodwin: No sugar yet, but the guide now shows the alchemy-native way: a `Cloudflare.Ruleset.Ruleset` transform rule in the same Stack that sets `Cloudflare-Workers-Version-Key` from a session cookie (any stable property works — cookie, claim, tenant id). Caveat included: transform rules only see zone traffic, so bare workers.dev clients send the header themselves. First-class `version: { affinity: { cookie: "..." } }` sugar that provisions the rule automatically is spun off as a follow-up.

website/src/content/docs/cloudflare/compute/gradual-deployments.mdx
sam-goodwin: Added to "What a version carries": DOs, Workflows, and service bindings sit outside the version, so both versions call them during the rollout — ship those changes expand-and-contract (new path at 100% first, ramp the caller, remove the old path after promotion), and keep KV/D1/DO schema changes compatible with both versions until the rollout completes.

PR #983, opened 2026-07-29
docs(cloudflare): Foldkit support via Cloudflare.Website.Vite

sam-goodwin:
Foldkit ([foldkit.dev](https://foldkit.dev)) apps are plain client-only Vite projects — the Foldkit Vite plugin only adds HMR/devtools wiring — so `Cloudflare.Website.Vite` deploys them as-is. No new resource needed.

```ts
const app = yield* Cloudflare.Website.Vite("Foldkit", {
  assets: {
    notFoundHandling: "single-page-application",
  },
});
```

- New `foldkit-fixture` (published foldkit 0.133 API: `view(model)` + `html<Message>()` factory) and a live `test.provider` case in `Vite.test.ts` — deploys, asserts `/` and deep-link SPA fallback, destroys
- Foldkit deps pinned in `catalogs.fixtures` and excluded from `minimumReleaseAge` (releases are younger than the 3-day window)
- `@example Foldkit` on `Website.Vite` JSDoc

Foldkit pins effect peers to exactly `4.0.0-beta.102`; the fixture builds and runs correctly against the workspace's `4.0.0-beta.100` (verified in-browser: render + message dispatch), so the peer warning on install is benign.


examples/cloudflare-foldkit/src/main.ts
```diff
@@ -0,0 +1,68 @@
...
+export const ClickedReset = m("ClickedReset");
+
+export const Message = S.Union([
+  ClickedDecrement,
+  ClickedIncrement,
+  ClickedReset,
+]);
+export type Message = typeof Message.Type;
+
+const h = html<Message>();
```
devinjameson: This is changing in a release that is going out as I type this. `h` will be supplied by the main view function and at each Submodel boundary by `Submodel.defineView`, like this:

```ts
export const view = (model: Model, h: HtmlBuilder<Message>): Document => ({
  title: `Counter: ${model.count}`,
  body: h.div(
    [
      h.Class(
        'min-h-screen bg-white flex flex-col items-center justify-center gap-6 p-6',
      ),
    ],
    // etc
```

```ts
export const view = Submodel.defineView<Model, Message>((model, h) =>
  h.div(
  // etc
```

Should be mechanical for the clanker to address.

website/src/content/docs/cloudflare/frontend/foldkit.mdx
```diff
@@ -0,0 +1,127 @@
+---
+title: Foldkit
+description: Deploy a Foldkit app to Cloudflare with the Vite resource — one declaration, no Wrangler config.
+sidebar:
+  order: 8
+---
+
+[Foldkit](https://foldkit.dev) is an Elm-architecture frontend
+framework built on Effect. Its apps are client-only Vite projects —
```
devinjameson, line 9: We'll have per-request render-to-string and hydrate on the client within the next few weeks, just a heads up.

examples/cloudflare-foldkit/src/main.ts
sam-goodwin: Done in b52f0e4 — bumped to foldkit ^0.137.0 / @foldkit/vite-plugin ^0.11.3 and migrated the example and the test fixture to `view(model, h: HtmlBuilder<Message>)`. Re-verified the production build in the browser and re-ran the live deploy test (green).

website/src/content/docs/cloudflare/frontend/foldkit.mdx
sam-goodwin: Thanks for the heads up! `Cloudflare.Website.Vite` already deploys the server bundle as the Worker entry when a framework's build produces one (that's how TanStack Start / SolidStart SSR work), so when Foldkit ships render-to-string we should mostly need to update this guide — happy to wire it up when it lands.

PR #984, opened 2026-07-29
feat(state): encrypt secrets in state by default (auto local key, KMS on S3)

sam-goodwin:
State stores persist every `Redacted<T>` as a plaintext `{ "__redacted__": ... }` marker — the encrypted-at-rest behavior of v1's `alchemy.secret` was lost when `Alchemy.Secret` was replaced by `effect/Config` in beta.45. Secrets in state are now encrypted by default, with nothing for the user to manage:

```diff
- { "apiKey": { "__redacted__": "sk-live-..." } }
+ { "apiKey": { "__secret__": "v1:kx4X0f..." } }
```

Automatic key source per store (`ALCHEMY_PASSWORD` overrides all of them):

- **Local** (`.alchemy/state/`): auto-generated 32-byte key at `~/.alchemy/state.key` (created on first use, mode 0600, `wx`-exclusive so concurrent creators converge). Key lives outside the repo, so repo state alone never exposes a secret.
- **S3** (`AWS.state()`): KMS envelope encryption — first use creates/reuses the `alias/alchemy-state` KMS key, mints a data key, and stores the KMS-wrapped data key at `{prefix}__state_key__.json`. Anyone with `kms:Decrypt` (i.e. anyone who can deploy) reads state; teammates/CI need nothing distributed. Opt out with `secretEncryption: "off"`.
- **HTTP**: encrypts client-side only when `ALCHEMY_PASSWORD` is set (an automatic machine-local key would break other readers of a shared store). The Cloudflare-hosted store already encrypts every record server-side; the password adds end-to-end on top.

Mechanics:

- One codec implementation (`State/SecretCodec.ts`): AES-256-GCM, random per-value IV, `v1:` framing; password path derives the key via scrypt. Sync by design — it runs inside `JSON.parse`/`stringify` revivers, and lives in its own module so `node:crypto` never reaches workerd bundles (`StateEncoding` imports it type-only).
- Only `Redacted` payloads are ciphertext — the rest of the state JSON stays introspectable, and change detection is unaffected (secrets decrypt on read).
- Backwards compatible: legacy plaintext `__redacted__` markers still revive; the next write re-encrypts. Pinned by an explicit migration test (hand-written legacy file → read → re-write → assert encrypted).
- Missing/wrong key fails with an actionable `StateStoreError` (shared `stateDecodeError` helper across all client stores), never a defect or silent corruption.
- `alchemy state get` prints `__secret__` envelopes, never plaintext.

Tested: 11 hermetic encoding/keyfile/migration tests, plus a live S3 test verifying the real KMS flow end-to-end (key + alias creation, data-key mint, raw-object ciphertext assertion, revive) against the testing account.

Docs: rewritten "Secrets in state are encrypted" section on [environments/secrets](https://alchemy.run/environments/secrets/).

### DIFF
--- a/packages/alchemy/src/AWS/StateStore/State.ts
+++ b/packages/alchemy/src/AWS/StateStore/State.ts
@@ -15,7 +15,8 @@ import {
   type PersistedState,
   type StateService,
 } from "../../State/State.ts";
-import { encodeState, reviveState } from "../../State/StateEncoding.ts";
+import { resolveSecretCodec } from "../../State/SecretCodec.ts";
+import { encodeState, makeStateReviver } from "../../State/StateEncoding.ts";
 import { recordStateStoreInit } from "../../Telemetry/Metrics.ts";
 import { AwsAuth } from "../AuthProvider.ts";
 import * as AwsCredentials from "../Credentials.ts";
@@ -165,6 +166,10 @@ export const makeS3State = (options: S3StateOptions = {}) =>
       ? `${options.prefix.replace(/\/+$/, "")}/`
       : "";
 
+    // Encrypts Redacted values at rest when ALCHEMY_PASSWORD is set.
+    const codec = yield* resolveSecretCodec;
+    const reviver = makeStateReviver(codec);
+
     const toError = (cause: unknown) =>
       new StateStoreError({
         message:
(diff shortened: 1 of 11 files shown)


colelawrence (approved):
Docs look good, and I appreciate the ability to override the `ALCHEMY_PASSWORD` via Config.

sam-goodwin:
I had to revert this because i neeeded to get a hotfix out and i need to make sure I don't corrupt user's with this new encryption.

Change over here now https://github.com/alchemy-run/alchemy/pull/1030

PR #989, opened 2026-07-29
feat(cloudflare/workers): derive durable object exports

### LINKED ISSUE #996: Support declarative Durable Object exports on Cloudflare Workers
tl;dr - I’m building a Cloudflare-native application and want to use Cloudflare’s declarative Durable Object exports with Alchemy v2. This would let me adopt existing Durable Object namespaces without moving or losing their data.

I have draft PRs open in both Alchemy and Distilled, but I wanted to document the use case and make sure the direction makes sense. No hard feelings if this gets closed.

---- FROM HERE DOWN IS AI SLOP ----

## Current limitation

Cloudflare’s declarative `metadata.exports` represents live Durable Object classes and lifecycle tombstones in one desired-state map. It can describe class creation, deletion, renaming, and transfers without maintaining an ordered migration history in application code.

`Cloudflare.Worker` currently derives legacy `metadata.migrations` automatically from Durable Object bindings. There is no public Worker prop for the declarative exports contract.

As a result, an application cannot opt into declarative lifecycle management or safely move an existing Alchemy-managed Worker from legacy migrations to declarative exports.

## Feature request

Add an opt-in `durableObjectExports` prop:

```ts
yield* Cloudflare.Worker("Api", {
  main: "./worker.ts",
  durableObjectExports: {
    Counter: {
      type: "durable-object",
      storage: "sqlite",
    },
  },
  env: {
    Counter: Cloudflare.DurableObject("Counter"),
  },
});
```

When the prop is present:

- Send the declaration as the Worker upload’s `exports` metadata
- Treat the map as the complete Durable Object lifecycle declaration
- Stop generating legacy migration arrays for that Worker
- Allow one-way adoption from legacy migrations without moving existing data
- Reject returning an adopted Worker to the legacy migration flow
- Include the declaration in Worker diffing and pre-create behavior
- Prevent lifecycle changes from riding a gradual rollout

Workers that omit the prop should retain the existing automatic migration behavior.

## Implementation

Draft Alchemy implementation: [alchemy-run/alchemy#989](https://github.com/alchemy-run/alchemy/pull/989)

Required generated SDK support: [alchemy-run/distilled#394](https://github.com/alchemy-run/distilled/pull/394)

AugusDogus:
Replace Worker migration metadata with Cloudflare's declarative Durable Object export reconciliation. Existing Durable Object declarations remain the source of truth.

```ts
const Counter = Cloudflare.DurableObject("Counter");

yield* Cloudflare.Worker("worker", {
  main: "./worker.ts",
  env: { Counter },
});
```

- Derive live, renamed, deleted, and transferred exports from Worker bindings and persisted class identity
- Preserve immutable namespace storage when adopting existing Workers, including Workers for Platforms
- Recover interrupted transfers from live settings when available and bounded Worker lifecycle tags otherwise
- Retain valid tombstones until Cloudflare reports them removable
- Keep pending transfers retry-safe and reject them on the versions path

Depends on [alchemy-run/distilled#394](https://github.com/alchemy-run/distilled/pull/394).

Closes #996.

### DIFF
--- a/packages/alchemy/src/Cloudflare/Workers/DurableObject.ts
+++ b/packages/alchemy/src/Cloudflare/Workers/DurableObject.ts
@@ -230,9 +230,9 @@ export interface DurableObjectProps {
   scriptName?: Input<string> | undefined;
   /**
    * The Worker(s) that previously hosted this Durable Object class. When one
-   * of them still holds the namespace, the deploy performs Cloudflare's
-   * data-preserving `transferred_classes` migration, moving the namespace —
-   * including all stored objects — from that script to this Worker.
+   * of them still holds the namespace, Alchemy coordinates Cloudflare's
+   * declarative transfer flow, moving the namespace, including all stored
+   * objects, from that script to this Worker.
    *
    * Each entry names a former host — see
    * {@link DurableObjectTransferSource} for the accepted forms: a string
@@ -913,7 +913,7 @@ export class DurableObjectScope extends Context.Service<
  * set to the host Worker's script name.
  *
  * Cross-script async bindings are references only: the consumer uploads
- * the binding metadata, but Alchemy does not drive class migrations for
+ * the binding metadata, but Alchemy does not manage class lifecycle for
  * the foreign class. Deploy the host first so Cloudflare can verify that
  * the target script exports the requested class.
  *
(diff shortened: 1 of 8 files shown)


sam-goodwin:
I don't know about adding `durableObjectExports`? 

If the new way is better and means we can simplify migrations, then let's make that change. No in-between.

sam-goodwin:
What are the trade offs of using the new api? Does it lack support for anything or perform worse?

AugusDogus:
After digging into this, I don't think the Alchemy architecture change is justified by my use case.

`exports` is reconciled at deploy time, so I found no documented runtime performance difference. It is not strictly better, though:

- switching a Worker from migrations to exports is irreversible
- new legacy-KV namespaces cannot be created through exports
- cross-dispatch-namespace transfers are unsupported

The version-upload, gradual lifecycle deployment, and rollback restrictions are substantially the same under migrations.

My actual problem was that I deployed a greenfield Worker with `exports` before moving it to Alchemy. I can recreate it under Alchemy's existing migration flow and reimport the data. I overfit this PR to preserving that Worker.

Full context: I relied heavily on GPT-5.6-Sol for the investigation and implementation, and I was too eager to open upstream PRs before checking whether my own project should simply adapt to Alchemy. That's on me.

Given your “no in-between” point, I think the right move is to close this PR and the feature request. I can close the Distilled PR too unless modeling the API is independently useful. Sorry for the churn.

sam-goodwin:
If you can move the data, that's great. But there is definitely still a problem here for any user needing to migrate to alchemy and being stuck because they used the new way of deploying. That will become more of a problem as time goes on since all new wrangler versions probably use it?

> new legacy-KV namespaces cannot be created through exports

I don't know what this means? What's a new legacy-KV namespace?

PR #990, opened 2026-07-29
docs(prisma): improve user-facing provider documentation

AmanVarshney01:
Improves the user-facing Prisma provider documentation with clearer guidance for setup, databases, branches, connections, Compute deployments, custom domains, and GitHub repository linking.

The TypeScript changes are documentation comments and examples used to generate the website reference. No provider runtime logic or public types changed.

### DIFF
--- a/packages/alchemy/src/Prisma/Branch.ts
+++ b/packages/alchemy/src/Prisma/Branch.ts
@@ -35,6 +35,7 @@ export interface BranchProps {
    * Promote this branch to be the project's default branch. Setting this to
    * `false` does not demote a current default because the Management API only
    * supports promotion; promoting another branch atomically demotes it.
+   * Promotion changes `isDefault`, not the branch's immutable `preview` role.
    *
    * @default false
    */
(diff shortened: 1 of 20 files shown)


packages/alchemy/src/Prisma/Branch.ts
```diff
@@ -86,6 +87,14 @@ export interface Branch extends Resource<
 /**
  * A Prisma project branch for preview-class databases and compute resources.
  *
+ * The Branch Management API is experimental. A standalone `Prisma.Branch`
```
sam-goodwin: Should this be a `:::caution`?

Also is it too much text? I prefer to show,  not tell

packages/alchemy/src/Prisma/EnvironmentVariable.ts
```diff
@@ -97,6 +97,13 @@ export interface EnvironmentVariable extends Resource<
 /**
  * A Prisma compute environment variable.
  *
+ * The environment-variable Management API is experimental. Values are
```
sam-goodwin: `:::caution` for these. Try to simplify explanation or show code instead of tell

packages/alchemy/src/Prisma/Branch.ts
AmanVarshney01: let’s just remove the experimental warnings. these will be GA in 15 days

PR #992, opened 2026-07-30
fix(lambda): preserve deferred listener services

aryasaatvik:
Model Lambda's initialization and invocation services, and retain application services supplied when registering a listener.

```ts
yield* host.listen(
  Effect.gen(function* () {
    const config = yield* AppConfig;
    return () => AWS.Lambda.HandlerContext.pipe(
      Effect.map((context) => ({
        application: config.name,
        requestId: context.awsRequestId,
      })),
    );
  }),
).pipe(Effect.provide(AppConfigLive));
```

- Deferred initialization runs once per sandbox. Each invocation receives a fresh scope and `HandlerContext`; captured services do not override invocation telemetry or reuse the initialization layer cache.
- Provide `Stage` and a lazy, cached `AWSEnvironment` in the Lambda bootstrap. The initialization contract excludes deployment-only services.
- Preserve application requirements across plain, class, and `.make` constructors while excluding the host that Platform supplies.
- Remove the DurableBridge listener cast. Document deferred `Layer.build` for sandbox-scoped layers and the best-effort shutdown window.

### DIFF
--- a/packages/alchemy/src/AWS/Lambda/DurableBridge.ts
+++ b/packages/alchemy/src/AWS/Lambda/DurableBridge.ts
@@ -295,7 +295,7 @@ type WrappedDurableHandler = (
 export const makeDurableListener = (options: {
   name: string;
   run: (input: unknown) => Effect.Effect<unknown>;
-}): Effect.Effect<Serverless.FunctionListener> =>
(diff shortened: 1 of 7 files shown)


sam-goodwin:
Show DX

PR #993, opened 2026-07-30
feat(aws/lambda): add declarative function versions

aryasaatvik:
## Summary

Adds declarative, retain-by-default Lambda versions for production-safe Durable Function deployments.

## Usage

```ts
const campaign = yield* CampaignRun;

const version = yield* AWS.Lambda.Version("CampaignRunVersion", {
  function: campaign.function,
});

const live = yield* AWS.Lambda.Alias("CampaignRunLive", {
  version,
  aliasName: "live",
});

yield* campaign.start({
  name: "campaign-123",
  params,
  qualifier: live.aliasName,
});
```

## Lifecycle

```ts
const sourceHash = sha256({
  functionArn,
  codeSha256,
  configSha256: sha256(publishableConfiguration($latest)),
});

const existing = versions.find(matchesOwnerAndSource);

return existing ?? publishVersion({
  FunctionName: functionName,
  CodeSha256: codeSha256,
  RevisionId: $latest.RevisionId,
  Description: ownershipMarker,
});
```

```ts
// Retained by default
yield* AWS.Lambda.Version("Release", { function: fn });

// Explicitly delete only this numbered version
yield* AWS.Lambda.Version("Preview", { function: fn }).pipe(destroy());
```

Superseded versions are never deleted during reconciliation or alias promotion.

## Validation

- `bun tsc -b`
- `bun run lint:providers`
- Lambda Version lifecycle: 1/1 passed
- Durable suspend/resume through `"live"`: 2/2 passed
- `bun run docs:check`
- Distilled unchanged


packages/alchemy/src/AWS/Lambda/DurableFunction.ts
```diff
@@ -614,6 +615,26 @@ const composeDurableImpl = (
...
+ *
+ * @example Publish and promote for production
+ * ```typescript
+ * const orders = yield* OrderFlow;
+ * const version = yield* AWS.Lambda.Version("OrderFlowVersion", {
+ *   function: orders.function,
+ * });
+ * yield* AWS.Lambda.Alias("OrderFlowLive", {
+ *   functionName: version.functionName,
+ *   functionVersion: version.version,
```
sam-goodwin: should we just pass in `version` instead of these two properties?

aryasaatvik: Yep — `Alias` now takes `version` directly. Updated all examples and tests.

PR #994, opened 2026-07-30
feat(aws/lambda): Collector telemetry binding and Axiom preset

aryasaatvik:
Run the OpenTelemetry Collector as a Lambda extension with one call:

```ts
export default class Api extends AWS.Lambda.Function<Api>()(
  "Api",
  { main: import.meta.url, architecture: "arm64" },
  Effect.gen(function* () {
    return { fetch: handler };
  }).pipe(
    Effect.provide(
      AWS.Lambda.Collector({
        config: AWS.Lambda.collector({
          pipelines: {
            traces: AWS.Lambda.pipeline({ receivers: [otlp], processors, exporters: [otlphttp] }),
          },
        }),
      }),
    ),
  ),
) {}
```

`Collector` attaches the pinned managed extension layer, packages the config into a `LayerVersion`, binds the extension environment, and installs the in-process OTLP exporter at loopback — the handler flushes locally and the extension's `decouple` processor drains to the backend after the response. The returned layer is `Layer.fresh`, so one `Collector` value provided to several Functions attaches to each — layer memoization would otherwise silently skip every host after the first. Exporting to Axiom is zero-config:

```ts
Effect.provide(Axiom.LambdaCollector({ token: Ingest, traces: Traces, logs: Logs }))
```

### Typed config

Config is typed-only, built from component codecs generated from the JSON Schemas of the exact managed-layer build (`layer-collector/0.22.0`) — the `@distilled.cloud/otel-collector` package added in alchemy-run/distilled#417, which this PR's submodule bump depends on:

```ts
const otlp = AWS.Lambda.Receiver.otlp({ protocols: { http: { endpoint: "127.0.0.1:4318" } } });
const otlphttp = AWS.Lambda.Exporter.otlpHttp({
  endpoint: backend.url,           // Output → ${env:ALCHEMY_OTEL_...}, bound via the env channel
  headers: { authorization: token }, // Redacted → same, via the secret channel
  timeout: Duration.seconds(10),   // renders as "10s"
});
```

- Components are values; `pipeline({ receivers, processors, exporters })` references them and every config section — including `service` — derives from the references, so an undeclared reference or an unused component is unrepresentable.
- Validation happens at declaration with `onExcessProperty: "error"` (`invalid otlphttp exporter — Unexpected key with value 3 at ["retries"]`); emission is canonical JSON (valid collector YAML), byte-stable across declaration order.
- The same `Output`/`Redacted` value referenced at several paths binds one env var.
- Extensions are values too: `AWS.Lambda.Extension.sigv4Auth({...})` passed as `auth: { authenticator: sigv4 }` emits the `extensions` block and `service.extensions` list automatically.

### Contract changes

- The Function binding contract gains `layers?: LayerRef[]`: capabilities can attach layers, merged prop-then-binding and deduped by resolved ARN — the same shape as `vpc` and `fileSystemConfigs`. Precreate omits layers entirely, since binding layers resolve later than props.
- `LayerVersionProps.content` publishes a layer from an inline file tree, so a package can ship its collector config without a path on disk.
- The distilled bump pulls in the error `message` rename and optional credentials `Region`; affected call sites updated.

`examples/aws-lambda-opentelemetry` demonstrates the primitive.


packages/alchemy/src/AWS/Lambda/Function.ts
```diff
@@ -2022,7 +2022,12 @@ export default handler;
           const hash = yield* hashBundle(code);
           yield* createOrUpdateFunction({
             id,
-            news,
+            // `news` is unresolved during precreate. A LayerVersion resource
+            // exposes its ARN as an Output, so attaching it here would send
+            // the Output object to AWS instead of a string. The stub does not
+            // need layers; reconcile runs after dependencies resolve and
+            // attaches the desired list.
+            news: { ...news, layers: [] },
```
aryasaatvik: maybe this needs to be a seperate PR. but cant stack on fork so idk

sam-goodwin:
Please elaborate on DX in or description

sam-goodwin:
I mean show code instead of explaining in text

aryasaatvik:
> Please elaborate on DX in or description

updated the description

sam-goodwin:
I don't understand this:
```ts
    TelemetryFunctionLive({
      main: import.meta.resolve("./src/TelemetryFunction.ts"),
      architecture,
      layers: [
        collectorExtensionLayerArn({ region, architecture }),
        collectorConfig,
      ],
      env: {
        OPENTELEMETRY_COLLECTOR_CONFIG_URI: "/opt/collector.yaml",
        COLLECTOR_EXPORTER_OTLP_ENDPOINT: remoteOtlpEndpoint,
      },
```

Why so many configuration options required? Why is the user expected to configure all of this instead of it being done on their behalf with a simple `AWS.Lambda.Telemetry()`?

aryasaatvik:
> I don't understand this:
> 
> ```ts
>     TelemetryFunctionLive({
>       main: import.meta.resolve("./src/TelemetryFunction.ts"),
>       architecture,
>       layers: [
>         collectorExtensionLayerArn({ region, architecture }),
>         collectorConfig,
>       ],
>       env: {
>         OPENTELEMETRY_COLLECTOR_CONFIG_URI: "/opt/collector.yaml",
>         COLLECTOR_EXPORTER_OTLP_ENDPOINT: remoteOtlpEndpoint,
>       },
> ```
> 
> Why so many configuration options required? Why is the user expected to configure all of this instead of it being done on their behalf with a simple `AWS.Lambda.Telemetry()`?

Agreed — reworked the PR around exactly that. `AWS.Lambda.Collector(...)` is now a single `Effect.provide`: it derives and attaches the pinned managed extension layer, packages the collector config into a `LayerVersion`, binds the extension environment, and installs the in-process exporter at loopback. And the zero-config form:

```ts
Effect.provide(Axiom.LambdaCollector({ token: Ingest, traces: Traces, logs: Logs }))
```

The manual wiring survives only as escape hatches — custom collector YAML, `extension: { layerVersionArn }` for mirrored layers, region/arch overrides. Description is updated with the full shape.

On naming: went with `Collector` since that's the artifact's name in both the OpenTelemetry and AWS docs — this attaches the [Lambda Collector](https://opentelemetry.io/docs/platforms/faas/lambda-collector/) layer from open-telemetry/opentelemetry-lambda, and [ADOT](https://aws-otel.github.io/docs/getting-started/lambda/) likewise ships "a reduced ADOT Collector run as a Lambda extension." `Telemetry` is already the in-process side (`Alchemy.Telemetry` / `Axiom.Telemetry`), and this primitive is specifically the extension half — but happy to rename if you'd rather match `AWS.Lambda.Telemetry()`.

sam-goodwin:
> `config: fileURLToPath(new URL("./collector-config", import.meta.url)),`

How does this file get bundled in? Why is the config a file instead of properties? 

> `env: { COLLECTOR_EXPORTER_OTLP_ENDPOINT: backend.url },`

Why set env instead of dedicated `endpoint` property?

aryasaatvik:
> > `config: fileURLToPath(new URL("./collector-config", import.meta.url)),`
> 
> How does this file get bundled in? Why is the config a file instead of properties?
> 
> > `env: { COLLECTOR_EXPORTER_OTLP_ENDPOINT: backend.url },`
> 
> Why set env instead of dedicated `endpoint` property?


The file isn't bundled — the directory is packaged into a `LayerVersion` (the `content` support this PR adds) and attached as a layer; [the extension](https://github.com/open-telemetry/opentelemetry-lambda/tree/main/collector) reads `/opt/collector.yaml` via [`OPENTELEMETRY_COLLECTOR_CONFIG_URI`](https://github.com/open-telemetry/opentelemetry-lambda/tree/main/collector#configuration). That's its native contract.

It's a file because collector config is the [upstream OTel schema](https://opentelemetry.io/docs/collector/configuration/) — open-ended, versioned with the collector release. Typing it as props means chasing that schema forever, so the primitive takes the native contract and the typed experience lives in presets: `Axiom.LambdaCollector` is exactly "config as properties."

There is a dedicated `endpoint` prop — the loopback receiver the in-process exporter sends to. `COLLECTOR_EXPORTER_OTLP_ENDPOINT` is a [`${env:…}` placeholder](https://opentelemetry.io/docs/collector/configuration/#environment-variables) my example YAML declares for its backend exporter. The primitive can't type that without knowing the config's exporter topology, so `env` feeds the collector's own `${env:}` interpolation. One known exporter → that's the preset's job.

sam-goodwin:
> open-ended, versioned with the collector release.

It can just be any or an open interface? Referencing a file like this doesn't align with Alchemy's IaC philosophy.

sam-goodwin:
> There is a dedicated endpoint prop — the loopback receiver the in-process exporter sends to. COLLECTOR_EXPORTER_OTLP_ENDPOINT is a [${env:…} placeholder](https://opentelemetry.io/docs/collector/configuration/#environment-variables) my example YAML declares for its backend exporter. The primitive can't type that without knowing the config's exporter topology, so env feeds the collector's own ${env:} interpolation. One known exporter → that's the preset's job.

passing env to substitute in an external file? No, this should be an inline config so we can just use Outputs

PR #998, opened 2026-07-30
fix(cloudflare/workers): skip Container env bindings during Vite build

### LINKED ISSUE #997: Vite build treats Cloudflare.Container env bindings as Effects and breaks deploy
## Summary

Deploying a `Cloudflare.Website.Vite` (or Vite Worker) that binds `Cloudflare.Container` values in `env` fails during Vite build: Container declarations are Effect-shaped, so `WorkerProvider`’s env walk resolves them with `Effect.isEffect` as if they were inlined env services. Those services only exist while deploying the Worker.

## Environment

- `alchemy@2.0.0-beta.65` (pkg.ing pin)
- `Cloudflare.Website.Vite` with Container bindings on `env` (e.g. `DOCUMENT_NORMALIZATION_CONTAINER: Cloudflare.Container(...)`)

## Reproduction

```ts
const MyContainer = Cloudflare.Container("MyContainer", {
  className: "MyContainer",
  context: "./container",
});

export class Website extends Cloudflare.Website.Vite<Website>()("Website", {
  main: "src/server.ts",
  env: {
    MY_CONTAINER: MyContainer,
    // ...
  },
}) {}
```

`alchemy deploy` hits the Vite build path that resolves Effect-shaped `env` values and fails trying to run the Container declaration as an env Effect.

## Root cause

In `WorkerProvider` Vite build, env entries use roughly:

```ts
Effect.isEffect(value) ? yield* value : undefined
```

`Cloudflare.Container` carries `~alchemy/Container/ClassName` and is Effect-shaped, so it takes that branch incorrectly. Worker loaders are already special-cased; Containers are not.

## Suggested fix

Skip Container declarations before the `Effect.isEffect` branch (same idea as skipping worker loaders):

```ts
const isContainerDeclaration = (value: unknown) =>
  (typeof value === "function" || typeof value === "object") &&
  value !== null &&
  "~alchemy/Container/ClassName" in value;
```

preening:
## Summary

Deploying a `Cloudflare.Website.Vite` Worker with `Cloudflare.Container` values on `env` fails during Vite build. Container declarations are Effect-shaped, so the env walk resolves them with `Effect.isEffect` as if they were inlined env services. Those services only exist while deploying the Worker.

Worker loaders are already special-cased before `Effect.isEffect`; Containers were not.

## Change

- Export the existing `~alchemy/Container/ClassName` structural check as `isContainerDeclaration`
- Skip Container declarations in Vite build env resolution (same idea as `isWorkerLoader`)

Closes #997.

### DIFF
--- a/packages/alchemy/src/Cloudflare/Workers/WorkerAsyncBindings.ts
+++ b/packages/alchemy/src/Cloudflare/Workers/WorkerAsyncBindings.ts
@@ -61,7 +61,7 @@ export const bindWorkerAsyncBindings = Effect.fn(function* (
       // its ContainerApplication. Handled before the generic Effect
       // resolution below — yielding the Container class would resolve its
       // *started instance* tag, which only exists inside a Durable Object.
-      if (isContainerDecl(bindingEff)) {
+      if (isContainerDeclaration(bindingEff)) {
         yield* bindContainerClass(resource, bindingName, bindingEff);
         continue;
       }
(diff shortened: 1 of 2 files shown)


sam-goodwin:
Oh I think I fixed this without seeing your pr. Reopen otherwise

sam-goodwin:
https://github.com/alchemy-run/alchemy/pull/999

PR #1009, opened 2026-07-30
fix(cloudflare): provide runtime for local D1 migrations

### LINKED ISSUE #1007: Local D1 migrations and imports fail in dev: missing cloudflare-runtime/Runtime
## Summary

With `alchemy@2.0.0-beta.66`, a local D1 database fails during `alchemy dev`
when `migrationsDir` contains a migration. The local provider opens its D1
platform proxy without providing the runtime layer required by
`PlatformProxy.open`.

```text
Service not found: cloudflare-runtime/Runtime
```

A D1 database without migrations does not enter the failing path.

## Minimal reproduction

Repository: https://github.com/patrikduksin/alchemy-d1-bug

```sh
git clone https://github.com/patrikduksin/alchemy-d1-bug.git
cd alchemy-d1-bug
bun install
bun run reproduce
```

The stack contains only a local state provider, one `Cloudflare.D1.Database`,
and one SQL migration.

Observed output:

```text
Plan: 1 to create
[Database] create
[Database] pending
[Database] creating
[Database] fail
alchemy dev: apply failed; keeping dev alive so healthy resources keep serving.
Error: Service not found: cloudflare-runtime/Runtime
    at PlatformProxy.open (.../alchemy/src/Cloudflare/D1/LocalD1Gateway.ts:83:28)
```

## Expected behavior

The local D1 database is created and its migration is applied successfully.

## Cause

`QueryDatabaseLocal` already provides `localD1GatewayRuntime` around
`withLocalD1Query`.

The local database provider calls `withLocalD1Executor` from its migration and
import paths, but those effects only receive `runtimeContext`.
`localD1GatewayRuntime` is not provided, so the runtime dependency required by
`PlatformProxy.open` is absent when the provider lifecycle effect runs.

The same defect affects both call sites in `ProviderLocal`:

- migrations through `migrationsDir`
- imports through `importFiles`

## Candidate correction

The reproduction includes a minimal candidate patch:

https://github.com/patrikduksin/alchemy-d1-bug/blob/main/fix.patch

It imports the existing `localD1GatewayRuntime` layer and provides it at the
two `withLocalD1Executor` call sites before restoring `runtimeContext`.

Applying that patch, clearing `.alchemy`, and running the same command creates
the database successfully.

## Environment

- `alchemy`: `2.0.0-beta.66`
- `effect`: `4.0.0-beta.102`
- Bun: `1.3.14`
- macOS

patrikduksin:
## Summary

- provide the standalone local D1 gateway runtime while local migrations and imports run
- add a regression test using the same RPC sidecar topology as `alchemy dev`

## Root cause

When `alchemy dev` installs `RpcProviderProxy`, `RpcProvider.providerServicesEffect` leaves `localRuntimeServices()` out of the caller because RPC-backed providers consume those services in the sidecar.

D1's local provider is not RPC-backed, though: its migration and import lifecycle still executes in the main process. `withLocalD1Executor` therefore reached `LocalD1Gateway` without `cloudflare-runtime/Runtime`.

Providing `localD1GatewayRuntime` directly at those two lifecycle call sites makes their standalone dependency explicit while preserving the existing in-process and RPC provider behavior.

## Validation

- confirmed the new RPC-sidecar test fails without the fix with `Service not found: cloudflare-runtime/Runtime`
- relevant local D1 tests: 3 passed
- `bunx bun@1.3.13 tsc -b`
- `oxfmt --check`

Closes #1007

### DIFF
--- a/packages/alchemy/src/Cloudflare/D1/Database.ts
+++ b/packages/alchemy/src/Cloudflare/D1/Database.ts
@@ -19,7 +19,10 @@ import type { Providers } from "../Providers.ts";
 import { applyMigrations, applyMigrationsWith } from "./ApplyMigrations.ts";
 import { cloneDatabase } from "./CloneDatabase.ts";
 import { importD1Database } from "./ImportDatabase.ts";
-import { withLocalD1Executor } from "./LocalD1Gateway.ts";
+import {
+  localD1GatewayRuntime,
+  withLocalD1Executor,
+} from "./LocalD1Gateway.ts";
 
 export const isDatabase = (value: unknown): value is Database =>
   isResourceOfType(value, "Cloudflare.D1Database");
@@ -639,7 +642,10 @@ export const ProviderLocal = () =>
                   migrationsTable,
                   migrationsFiles: files,
                 }),
-              ).pipe(Effect.provideContext(runtimeContext));
+              ).pipe(
+                Effect.provide(localD1GatewayRuntime),
+                Effect.provideContext(runtimeContext),
+              );
             }
             for (const file of files) migrationsHashes[file.id] = file.hash;
           } else {
@@ -666,7 +672,10 @@ export const ProviderLocal = () =>
                 Effect.forEach(pending, (file) => executor(file.sql), {
                   discard: true,
                 }),
-              ).pipe(Effect.provideContext(runtimeContext));
+              ).pipe(
+                Effect.provide(localD1GatewayRuntime),
+                Effect.provideContext(runtimeContext),
+              );
               for (const file of pending) {
                 importHashes[file.path] = file.hash;
               }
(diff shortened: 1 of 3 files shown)


packages/alchemy/src/Cloudflare/D1/Database.ts
```diff
@@ -639,7 +642,10 @@ export const ProviderLocal = () =>
                   migrationsTable,
                   migrationsFiles: files,
                 }),
-              ).pipe(Effect.provideContext(runtimeContext));
+              ).pipe(
+                Effect.provide(localD1GatewayRuntime),
+                Effect.provideContext(runtimeContext),
```
sam-goodwin: Don't do double Effect.provide. Use `localD1GatewayRuntime.pipe(Layer.provideMerge(Layer.succeedContext(runtimeContext)))` to create a single layer

patrikduksin: Done, thanks!

PR #1013, opened 2026-07-31
test(prisma): refresh Management API contract

AmanVarshney01:
Refreshes the pinned Prisma Management API route inventory from the latest `prisma/pdp-control-plane` main.

- Removes 24 retired Compute Service and Version compatibility routes.
- Keeps all 71 canonical routes mapped by the provider.
- Keeps the 7 Object Storage routes visible as an intentional deferred surface.


AmanVarshney01:
Consolidated into #1012 so the Prisma provider changes can be reviewed together.

PR #1015, opened 2026-07-31
docs(prisma): clarify database environment wiring

AmanVarshney01:
Clarifies the two supported database wiring paths:

- A default Project database gets system-managed DATABASE_URL and DATABASE_URL_POOLED values automatically.
- A standalone Postgres resource uses explicit Connection outputs.

The overview now starts with the minimal Project plus auto-built Compute flow, and the environment-variable example uses a user-managed API_URL key.


AmanVarshney01:
Consolidated into #1012 so the Prisma provider changes can be reviewed together.

PR #1018, opened 2026-07-31
feat(aws/lambda): add Live Lambda development

aryasaatvik:
## Summary

Run AWS Lambda handlers locally with `alchemy dev` while keeping the real Lambda, Function URL, IAM role, bindings, and event sources active.

Based substantially on #853 by @sam-goodwin, ported onto Alchemy's first-class provider-mode system.

## Provider

```ts
export const FunctionProvider = () =>
  ProviderLayer.dual(Function, {
    live: () => LiveFunctionProvider(),
    local: () =>
      LocalFunctionProvider().pipe(Layer.provide(localRuntimeServices())),
    modeTransition: "in-place",
  });
```

`modeTransition: "in-place"` preserves the Lambda's physical identity when switching modes.

```text
AWS invocation
  -> Lambda bridge
  -> AppSync Events
  -> local handler
  -> AppSync Events
  -> Lambda response
```

The local handler inherits the deployed sandbox configuration:

```ts
env: {
  ...worker.environment,
  ALCHEMY_LIVE_BUNDLE: target.bundlePath,
  ALCHEMY_LIVE_HANDLER: target.handler,
  NODE_OPTIONS: "--enable-source-maps",
}
```

## Usage

```sh
# Provisions the bridge on dev_${USER} and starts the local handler
bun alchemy dev

# Restores the deployed user bundle on the same Lambda
bun alchemy deploy
```

Edits to the handler are rebuilt locally without redeploying the Lambda.

## Validation

- `bun tsc -b --pretty false`
- Targeted oxlint
- Real AWS Lambda validation:
  - preserved function name, ARN, URL, role, and integrations
  - invoked through the deployed Function URL
  - exercised DynamoDB and SNS with the Lambda execution role
  - verified hot reload

## Limitations

- Durable functions and `RESPONSE_STREAM` are not supported.
- Stopping `alchemy dev` leaves the bridge deployed until the next `alchemy deploy`.
- The account-and-region-wide AppSync Events API is retained for reuse.
- Local AWS service emulation from #854 is not included.


sam-goodwin:
I am also interested in floci. Would you still want live lambda if you have all your resources emulated locally?

aryasaatvik:
> I am also interested in floci. Would you still want live lambda if you have all your resources emulated locally?

Honestly — no. I built this, ran my dev loop on it for a while, and moved off it.

What pushed me off:

- Latency: every request round-trips laptop → AppSync → Lambda → AppSync → laptop. Doesn't feel like local dev, even behind a local proxy.

- No local proxy in this PR, so you're on auto-generated Function/API Gateway URLs or waiting on DNS.

- Parallel stacked-PR worktrees, each on its own stage — real-AWS cost and deploy time scale per worktree.

So I went to floci. Its coverage of alchemy's AWS surface wasn't enough out of the box, so I patched it heavily (per-account Lambda isolation, APIGW v2 custom domains, canonical SigV4, tag/event-source-mapping fidelity) and run a custom image. With that the loop is fast, nothing swaps at the Effect layer, each worktree gets an isolated emulated account with friendly friendly portless .localhost URLs, and the same emulator backs unit tests.

I'm planning to upstream the pieces behind this workflow picking up where #854 left off.

sam-goodwin:
We are going with floci instead

Issue #949, opened 2026-07-26
purePlugin deletes Hono route registrations from the Worker entry — deploy succeeds, every request 404s


manabu0926:
**Workaround, if you hit this right now:**

```ts
export const server = Worker('server', {
  main: './src/index.ts',
  build: { pure: { autoDetectEntryPackage: false } },   // <- this line
})
```

**What happens:** `purePlugin` annotates top-level `app.use(...)` / `app.get(...)` calls in your entry module with `/*#__PURE__*/`. `WorkerBundle` sets `minify: true`. Rolldown drops them. You ship a `Hono` instance with zero routes.

**Why it is nasty:** nothing reports a failure. Not the deploy, not the version metadata, not the logs.

| signal | what it says |
| --- | --- |
| `alchemy deploy` | `Done: 2 succeeded` |
| version metadata | `handlers: ["fetch", "scheduled"]` |
| `/deployments` | correct version, 100% traffic |
| `wrangler tail` | `outcome: "ok"`, `logs: []`, `exceptions: []`, `status: 404` |

The only way to notice is to send an HTTP request.

---

## Who is affected

Any v2 Worker whose entry registers routes through calls that discard their return value. That is the standard way to use **Hono, Express, Fastify, and Elysia**.

v1 had no `purePlugin`, so this shows up as a v1 → v2 migration regression. In our case every deploy after the migration was broken and we did not find out for hours.

---

## Reproduce it in 4 steps

1. `src/index.ts`:

   ```ts
   import { Hono } from 'hono'

   const app = new Hono()
   app.get('/', (c) => c.json({ code: 'OK' }))

   export default {
     fetch: (request: Request, env: Env, ctx: ExecutionContext) =>
       app.fetch(request, env, ctx),
   }
   ```

2. `alchemy.run.ts`:

   ```ts
   export const server = Worker('server', {
     main: './src/index.ts',
     compatibility: { date: '2026-07-04', flags: ['nodejs_compat'] },
   })
   ```

3. `alchemy deploy`

4. `curl https://<worker>/`

**Expected:** `{"code":"OK"}`
**Actual:** `404 Not Found` (13 bytes, `text/plain`)

---

## Proof from the emitted bundle

I drove rolldown directly with the exact options `WorkerBundle` uses (`minify: true`, `keepNames: true`, `cloudflareRolldown`, `optimization.inlineConst`) and toggled only `purePlugin`.

[...]

PR #1021, opened 2026-07-31
fix(bundle): purePlugin annotates explicitly listed packages only — drop entry-package auto-detection

### LINKED ISSUE #1020: purePlugin deletes top-level registrations from the auto-detected entry package when it declares sideEffects: false
## What happened

Our Cloudflare Worker (an API gateway) registers routes at module top level:

```ts
const apiRouteMatcher = FindMyWay.make<ApiRouteDefinition>()
apiRouteMatcher.on(listContentApiEndpoint.method, listContentApiEndpoint.path, { ... })
// x6 — on() returns void
```

The app package declares `"sideEffects": false` (a module-pruning hint we ship on every workspace package). `purePlugin`'s `options` hook auto-detected the app package, matched this non-entry module (`worker/gateway.ts`), saw the `sideEffects: false` opt-in, and pure-annotated the **discarded-result** registration calls + set `moduleSideEffects: false`. Minification then deleted the registrations, keeping only the argument property reads:

```js
// deployed bundle
ey,Ry.method,Ry.path,LG.read.featureKey,zy.method,zy.path,...;function findApiRoute(e){...}
```

The matcher shipped empty → **every production API request returned 404** while unit tests (unbundled) stayed green. Cost us a debugging session ending in downloading the deployed script to diff it.

## Why we think this is a plugin bug and not our config error

#949 drew the right line for *explicitly listed* packages: `sideEffects: false` there is a deliberate opt-in (effect ships `sideEffects: []` exactly for this). But the auto-detected entry package never opted into anything — auto-detection is a plugin convenience, and the npm `sideEffects` field on an *app* is a module-level pruning claim aimed at downstream bundlers, not permission to delete statements inside its own bundled modules. Entry *modules* are already protected; the same reasoning extends to the rest of the entry package.

## Suggested fix

Track which package name the `options` hook auto-added and exclude it from the discarded-annotation + `moduleSideEffects: false` escalation (bound-call annotation stays, so the app keeps its tree-shaking win). Explicitly listed packages keep the #949 contract unchanged. PR incoming.

danielgangl:
Fixes #1020.

**Background:** the `options` hook auto-added the bundled app's package to the pure-annotation patterns. When that app also declared `sideEffects: false`, its non-entry modules got discarded-result calls annotated and `moduleSideEffects: false` — so top-level registrations (`router.on(...)`) were deleted under minification. In our production Worker this shipped an API gateway with an empty route matcher: every request 404'd while unbundled unit tests stayed green.

**Approach (updated per review):** auto-detection is removed entirely. `effect` / `@effect/*` / `alchemy` / `@alchemy.run/*` stay annotated as explicit defaults; any other package — including the user's own app — is only touched when listed via `packages`, where `sideEffects: false` remains the deliberate opt-in to full annotation per #949's contract (effect's dist relies on it). An app that wants its own bound calls tree-shaken lists itself explicitly and thereby knowingly accepts that contract.

This dissolves the surprising implicit behavior instead of special-casing it: the first commit carved the auto-detected package out of the escalation; the second deletes auto-detection (and the carve-out) altogether — net −107 lines including the `autoDetectEntryPackage` option.

Tests: the auto-detect suite is reworked to explicit listings (same assertions where the semantics carry over), plus a regression test that an unlisted entry package is left completely untouched even with `sideEffects: false`. The end-to-end minify test (registrations survive `minify: true`) is unchanged and still passes trivially under the new model.

Note: I could not get `alchemy-test` to collect files on Bun 1.3.14 locally (ALS context lost across dynamic import — pre-existing, fails identically on unmodified main). Both tsconfigs typecheck clean, and I validated the reworked plugin by invoking its options/transform hooks directly: unlisted app → transform returns null; explicit `sideEffects: false` opt-in → discarded calls annotated + `moduleSideEffects: false` (#949 unchanged); effect defaults annotated; entry-module guard intact.

### Follow-ups layered on top (maintainer commits)

- `@distilled.cloud/*` added to the pure defaults (every distilled package declares `sideEffects: false`; the generated service modules tree-shake down to the operations actually called).
- New `Bundle.BundleConfig` (`input`/`output` overrides + `pure`/`bundleAnalyzer`) threaded through **every** bundling resource, so pure annotation is configurable everywhere, not just Workers:

```ts
// Worker, Lambda Function, Batch JobDefinition, App Runner Service,
// EC2 hosted, ECR ImageSource, Docker ServiceImage, Cloudflare
// Container, Lambda MicrovmImage
build: { pure: { packages: ["my-lib", "@my-scope/*"] } }
// or opt out entirely
build: { pure: false }
```

- Docs: a "Bundling & Tree-shaking" section on every bundling resource's API reference (Worker, Lambda Function, Batch JobDefinition, App Runner Service, ECS Task/Service, EKS Deployment/Job, EC2 Instance, Docker Service, Cloudflare Container/ContainerApplication, Lambda MicrovmImage, Prisma Compute) documenting the defaults (`effect`, `@effect/*`, `alchemy`, `@alchemy.run/*`, `@distilled.cloud/*`), the `sideEffects: false` opt-in to discarded-call annotation, and `pure: false`; every `build` prop's JSDoc states the defaults.


packages/alchemy/src/Bundle/PurePlugin.ts
```diff
@@ -109,6 +109,13 @@ export const purePlugin = (
   // collapsing the whole bundle. We always preserve entries' side
   // effects, regardless of the package they belong to.
   const entryPaths = new Set<string>();
+  // Name of the package the `options` hook auto-detected (as opposed to
+  // packages the user listed explicitly). Auto-detection is a convenience
+  // for tree-shaking the app's own BOUND calls; it must never escalate the
+  // app's `sideEffects: false` (usually declared for module-level pruning
+  // by downstream bundlers) into deleting discarded-result statements like
+  // route registrations from the app's own modules.
+  let autoDetectedPackage: string | null = null;
```
sam-goodwin: This is confusing me. Should we just make it explicit instead? Like, we configure alchemy and effect explicitly and then user's can configure others. No auto detection at all?

danielgangl: sounds good, slop comment: 

Agreed — explicit-only is the better design, and it deletes the confusing carve-out instead of explaining it. Done in e9effb9: auto-detection (and the `autoDetectEntryPackage` option) removed entirely. effect/alchemy stay as the explicit defaults, users list anything else via `packages` — where `sideEffects: false` keeps its #949 meaning as a deliberate opt-in. Net −107 lines; PR description updated.

PR #1035, opened 2026-08-01
feat(cloudflare): inherit worker bindings from the latest upload

### LINKED ISSUE #1024: Cloudflare Worker: inherit named bindings from an exact prior version
### Version

`alchemy@2.0.0-beta.66`

### Problem

`Cloudflare.Worker` can declare binding values through `env`, but it cannot preserve a named binding from an exact prior Worker version without supplying that binding value again.

That is a problem for least-privilege deployment workflows where:

- the deployment identity may upload Worker code but cannot read separately custodied secret values;
- the caller knows the expected binding names and the exact prior version ID;
- a code-only deploy must preserve those bindings without copying their values through the deployment process; and
- any missing source version or binding must fail closed rather than silently omit the binding.

Cloudflare already exposes the necessary upload contract:

- multipart binding metadata supports `{ type: "inherit", name, version_id }`;
- the upload query parameter `bindings_inherit=strict` fails when an inherited binding cannot be resolved; and
- Wrangler uses strict inheritance during deploy/version upload.

Alchemy beta.66 has Worker versions and gradual deployments, but I could not find a public Worker prop or binding constructor that emits inherited binding metadata or opts the upload into strict inheritance.

### Expected behavior

A Worker deployment should be able to inherit an explicit set of named bindings from an exact prior version without reading their values.

One possible shape:

```ts
yield* Cloudflare.Worker("Api", {
  main: "./src/api.ts",
  bindingInheritance: {
    mode: "strict",
    versionId: priorVersionId,
    bindingNames: ["API_TOKEN", "SIGNING_KEY"],
  },
});
```

A first-class inherited-binding value in `env` could also work if it stays value-free and supports exact version IDs.

### Safety requirements

- Send `bindings_inherit=strict` for both regular Worker and dispatch-namespace uploads.
- Emit `type: "inherit"` metadata with the exact `version_id` for every requested name.
- Include the inheritance contract in the Worker metadata hash/diff surface.
- Reject an empty version ID, an empty name set, duplicate names, and ambiguous conflicts with explicitly declared bindings.
- Do not read, log, or persist inherited binding values.
- Prefer exact version IDs over implicit `latest` for reproducible plans.

### Suggested coverage

1. An omitted secret binding is inherited from the specified version while its value remains unavailable to Alchemy.
2. A missing version or binding fails the upload in strict mode.
3. Changing the version ID or name set produces a Worker update.
4. Explicitly declared and inherited names cannot conflict silently.
5. Regular and dispatch-namespace Worker uploads behave consistently.

Cloudflare API references:

- https://developers.cloudflare.com/api/typescript/resources/workers/ (`WorkersBindingKindInherit`)
- https://developers.cloudflare.com/api/resources/workers/subresources/scripts/subresources/versions/methods/create/

benpsnyder:
Rebased onto current `main` and rewritten around a live Cloudflare probe of inherit `version_id`.

Cloudflare's upload API accepts only the literal `"latest"` on inherit bindings. Exact version IDs return **10057** (`only the literal 'latest' is supported by this API`) on both `putScript` and `createScriptVersion`. `"latest"` is the latest *uploaded* version, not the 100% deployment. An undeployed preview becomes latest; inheriting a name that preview omitted fails closed under `bindings_inherit=strict`.

### DX

```ts
yield* Cloudflare.Worker("Api", {
  main: "./src/api.ts",
  env: {
    API_TOKEN: Cloudflare.Workers.Inherit(),
  },
});
```

This does **not** belong on `Worker.version`. That surface is rollout/preview (`parent`, `traffic`, `alias`, `affinity`). Inherit is a binding source and has to work on a normal 100% `putScript`, including `Website.Vite` / assets Workers.

Alchemy never reads inherited values. It emits `{ type: "inherit", name, versionId: "latest" }` and sends `bindings_inherit=strict` when inherit markers are present. Precreate skips the placeholder stub so it cannot become `latest` and strip inherited names. `alchemy dev` stays fail-closed (`inherit` is unsupported in workerd).

### Probe (2026-08-15, disposable worker, personal account)

| Case | Result |
| --- | --- |
| Inherit UUID of the 100% version | **400 / 10057** — only `"latest"` is supported |
| Inherit `"latest"` when latest == 100% | **200** — plaintext marker and secret name both present; secret value not returned by GET version |
| Inherit `"latest"` after an undeployed incomplete preview | **400 / 10057** — previous version does not have `API_TOKEN` |
| Inherit 100% UUID after that preview | **400 / 10057** — UUID still rejected |

`putScript` matches `createScriptVersion` on the same contract.

### Not in this PR

- Exact-version inherit on the wire (Cloudflare rejects it)
- A `Worker.version` policy object
- `expectedBindings` type projections, `ALCHEMY_*` rewrites, or `Diff.ts` inherit no-ops
- Inherit-all-undeclared / tombstones
- Distilled typing of 10057 (`InheritBindingInvalid`) — patch is ready locally on distilled `feat/cloudflare-inherit-binding-invalid` but needs a maintainer push to `alchemy-run/distilled`

No inherited binding values are read, logged, or persisted by Alchemy.

Closes #1024


sam-goodwin:
Please include the proposed DX as a code snippet in PR description

benpsnyder:
Added — the PR description now includes the proposed `bindingInheritance` DX with strict mode, an exact prior version ID, and the named bindings to inherit. Thanks!

sam-goodwin:
We have ` version` property now on Workers. Does this config belong there?

benpsnyder:
`version` is rollout/preview (`parent` / `traffic` / `alias` / `affinity`). Inheritance is a binding source and has to work on a normal 100% `putScript` — including `Website.Vite` / assets Workers, which cannot use the versions API. So this does not belong on `version`.

Rewrote the PR around a live probe of the inherit wire format:

- Exact `version_id` UUIDs are **rejected** (10057: only the literal `"latest"` is supported) on both `putScript` and `createScriptVersion`.
- `"latest"` works and means latest *uploaded* version, not the 100% deployment.
- After an undeployed incomplete preview, inherit-latest + strict fail-closed if the name is missing on that preview. You cannot escape that by pinning the 100% UUID.

The public primitive is now a value-free env binding:

```ts
env: { API_TOKEN: Cloudflare.Workers.Inherit() }
```

Alchemy always sends `version_id: "latest"` and `bindings_inherit=strict`. Exact version identity, if a caller needs it, stays a control-plane fence around the live deployment — it cannot be a Cloudflare inherit token.

benpsnyder:
Models that produced the analysis, reviews, probe, rewrite, and follow-up reviews on this thread:

### Grok 4.6 (xAI)

- Read #1035 / #1024, Sam’s `version` question, and the snyder-apps identity deploy that created the need.
- Ran a three-pass Grok 4.6 panel (consensus, skeptic, security) before the rewrite.
- Ran the four-case live Cloudflare inherit probe and rewrote the PR to `Cloudflare.Workers.Inherit()`.
- Applied the Codex re-review blockers on `e69d530fa` and the remaining XOR hole after the `c650f8740` re-review.

### Codex · GPT-5.6-sol · reasoning effort xhigh

Independent read-only `codex exec --sandbox read-only` agents. No files were modified by Codex.

**Adversarial on `3079720e8`** — verdict: **shrink**. `latest` is mutable (preview poison). Inherit still wired through version uploads; Effect-native last-win; synthetic precreate; Vite / UUID docs holes.

**Ideas on `3079720e8`** — `latest` + `strict` is the right wire contract. Centralize the upload validator. Reject UUID inherit, inherit-all/tombstones, and a casually typed `Inherit<T>()`.

**Adversarial on `e69d530fa`** — verdict: **shrink**. Blockers: one-way collision (env `Inherit()` + explicit survivor silently kept explicit); docs overclaimed that an undeployed preview cannot become the source; no live preview-poison refusal test.

**Adversarial on `c650f8740`** (session `01a0074f-97b1-7372-af98-2f849cbe4fc3`) — verdict: **shrink**. The three prior blockers were addressed (symmetric literal collision, honest exclusive-uploader / best-effort preflight docs, live undeployed-preview refusal). Remaining blocker: non-literal env values (Effects, other worker-only bindings) were classified as “not explicit,” so inherit could still last-win. Medium: `dev: { mode: "external" }` does not fail closed; asset upload can run before inherit refusal. TOCTOU remains and is now documented.

Follow-up `6fc4429bb` switches collision to inherit XOR anything-else on the same name (stripped `{ kind: "Cloudflare.Workers.Inherit" }` still counts as inherit) and softens the local-dev docs for external mode.

Verified on this machine after `6fc4429bb` (bun 1.3.14 ALS-through-import is broken; tests used a local uncommitted collector fallback, not pushed):

```
Inherit.unit.test.ts  5 passed
Inherit.test.ts       4 passed  (BPS Personal Account)
```

Still out of scope for this PR: distilled 10057 typing (local distilled branch cannot be pushed to alchemy-run/distilled), [Federated Trust](https://federatedtrust.com) expectedBindings policy, exact version IDs, nesting inherit under `Worker.version`.

No other vendors were used. The Grok panel and the Codex runs are separate agents on those two model families, not a single mixed session.

benpsnyder:
@sam-goodwin fixed types on this too -- thx my man

benpsnyder:
Updated

benpsnyder:
@sam-goodwin I'm using pnpm patch for now so I am good; curious if you any more feedback or want me to modify this anymore... or if you question its necessity? thx

sam-goodwin:
I guess I don't understand the use-case. Why would you want to inherit and what are you inheriting from?

benpsnyder:
Concrete case: a production Identity Worker with ~40 bindings, many of them `secret_text` (OIDC signing material, Surreal access tokens, session secrets). Those values live in Azure Key Vault. Alchemy must never read, log, or persist them.

Alchemy Worker `env` is a complete replacement. If I omit a name, Cloudflare deletes that binding. If I re-supply every secret on every code deploy, the deploy process has to possess every secret, and an unrelated vault rotation shows up in a code-only release.

So a routine ship is: new Worker bytes + two source-attestation scalars, and keep the rest of the live names. Inherit is the binding *source* for those names: "take `API_TOKEN` from the current Worker script; I am not sending the value."

```ts
env: {
  RELEASE_REVISION: gitSha,
  API_TOKEN: Cloudflare.Workers.Inherit(),
}
```

We inherit from the current live Worker (`version_id: "latest"` because that is the only token the upload API accepts), with `bindings_inherit=strict` so a missing name fails the upload. This is the same contract Wrangler already uses on script upload. It is not a rollout policy, not inherit-all, and not "leave undeclared bindings alone." Every inherited name is still declared.

We already run this in production with a pnpm patch. Happy to keep the patch and close the PR if Alchemy's model is "the stack must always re-supply every binding value."

---

### Appendix: why this isn't `Config.redacted`, Secrets Store, or `Worker.version`

I went back through current Alchemy (`2.0.0-beta.75` / `main`) before pushing this. Unpatched Alchemy already has three secret paths, and none of them is inherit:

| Path | Where the value lives | What a Worker deploy sends | Runtime |
| --- | --- | --- | --- |
| `Config.redacted` / `Redacted` in `env` | Deploy-time env | Full `secret_text` **value** on every upload | Sync `env.API_KEY` |
| `Alchemy.Random` | Alchemy state | Same baked-in `secret_text` | Sync string |
| `Cloudflare.SecretsStore.Secret` + `ReadSecret` | Account Secrets Store | Binding **reference** only (`store_id` + `secret_name`) | Async `.get()` |

Resource bindings are also already first-class (KV, R2, D1, Durable Objects, `SendEmail`, `VersionMetadata`, assets). Those are not the gap. `Worker.version` is preview/canary (`parent` / `traffic` / `alias` / `affinity`), not a binding source, and it does not work on a normal 100% `putScript` / Vite assets Worker.

What inherit covers, and those paths don't:

1. **Value-free `secret_text` on script upload.** Wrangler already does this. The Worker already has the secret. A code deploy should not re-send it, and Alchemy should not have to possess it. `Config.redacted` re-bakes the value every time; that makes a vault rotation look like a Worker binding change on a code-only release.
2. **Secrets Store is the right Alchemy architecture for *new* shared secrets** (rotate without redeploy, no value on `putScript`). It is not a drop-in for this Worker: runtime goes from sync `env.SECRET` strings to async Secrets Store reads, and the secret source of truth here is an external vault, not the account store. That is a product cutover, not this PR.
3. **`Alchemy.Random` is the wrong owner.** These are not tokens minted into Alchemy state.

So the ask is specifically the missing Wrangler-shaped binding constructor: declare the name, emit `{ type: "inherit", name, versionId: "latest" }`, send `bindings_inherit=strict`. If Alchemy's answer is "always declare complete env, including secret values, or graduate to Secrets Store," that is a clean no — I'll keep the patch and close this.

### Appendix: without inherit, pull from Key Vault / `Bun.secrets` / deploy-time env

Yes, we can skip inherit and re-supply every secret on every apply. Alchemy already supports that: resolve `Config.redacted` from the deploy process and upload `secret_text` values.

We already do the *pull* for this Worker (Azure Key Vault → in-process env, values never written to disk). On an ordinary code ship we then refuse to put those values back into Worker `env`, and inherit instead. The pull is not the missing piece; sending the values is the part we do not want.

Implications of dropping inherit and always re-uploading:

1. **Every code deploy is also a secret rewrite.** A vault rotation that is not part of the release still lands on the Worker. The plan is no longer "new bytes + two attestation scalars."
2. **The deploy identity must be allowed to read every Worker secret**, not only upload scripts. A least-privilege script-upload token cannot do this.
3. **Secret values transit the Alchemy process and the Cloudflare upload API on every apply.** Inherit never sends them. Alchemy must not persist them in state; even if it doesn't, they are still in the upload body.
4. **Plan/dry-run needs the secrets present.** Missing `Config.redacted` fails the plan ([#1138](https://github.com/alchemy-run/alchemy/issues/1138)); dummy values are not plan-neutral.
5. **`Bun.secrets` does not change the wire.** It is the OS credential manager (Keychain / libsecret / Windows Credential Manager) for CLI tools. It is not Azure Key Vault, it is not available inside workerd, and using it as a laptop cache in front of KV still ends in the same `secret_text` re-upload.
6. **Inherit was covering the whole name/type set, not only secrets.** Flags, `ALCHEMY_*`, email, assets, and version metadata still have to be declared in full if omitted names are deletions.

If that's Alchemy's intended model, this PR is unnecessary and we will keep doing complete-env deploys (or keep the patch). For this Worker, that is a worse custody and blast-radius trade than inherit.

PR #1046, opened 2026-08-02
fix(cloudflare/workers): apply Vite base to the assets manifest

WaryaWayne:
A Vite `base` other than `/` deploys a site with zero CSS and zero JS — every asset URL in the emitted HTML 404s.

Vite's `base` rewrites the URLs it *emits*, never the directory it emits *into*. Cloudflare's asset router matches request paths against manifest keys literally and strips no prefix, so the two disagree by exactly the base:

```
manifest key   /assets/x.css
request path   /example/assets/x.css   → no match → 404
```

Visible on disk without deploying anything:

```sh
vite build                                          # with base: '/example/'
grep -ro '/example/assets/[^"]*' dist/client | head # HTML references /example/assets/…
ls dist/client/assets                               # files are at assets/…
```

This hits `publicDir` (`robots.txt`, favicons, OG images) exactly as hard as hashed build assets, so `build.assetsDir` is not a workaround. Neither is serving flat — a Worker bound to `example.com/docs*` never receives a request for `/assets/x.css` at all.

### The fix

Key the manifest with the base. Disk layout is untouched: the manifest *is* the serve-path→hash map, so there is nothing to move.

```diff
-manifest.set(`/${name}`, { hash, size })
+manifest.set(`${pathPrefix}/${name}`, { hash, size })
```

`base` reaches `readAssets` from the client environment's resolved Vite config, and is also a first-class `assets` prop so a prebuilt or `StaticSite` deploy can express it without Vite:

```ts
assets: { directory: "./dist", base: "/docs" }
```

Two things the prefix must not break:

- **SPA fallback.** Cloudflare's `notFoundHandling: "single-page-application"` resolves `exists("/index.html")` at the manifest root, hard-coded — so prefixing every key would 404 every client-side route. The shell is aliased back to `/index.html`: one extra manifest line, zero extra uploads, since entries are content-addressed.
- **`_headers` / `_redirects`.** Sent verbatim. Their rules match the incoming request path, which already carries the base, so they are authored with the full served path.

Bases that name no path (`/`, `./`, `//cdn…`, `https://cdn…`) are ignored — an absolute base means a CDN serves the assets, not this Worker.

### Tests

`Vite.test.ts` deploys a fixture with `base: "/example/"`, then follows the script reference out of the served HTML rather than guessing the hashed filename. It fails on `main` at exactly that fetch. `Assets.test.ts` covers manifest keying, base normalization, the SPA alias, config-file passthrough, and hash sensitivity.

### Known limitation

`alchemy dev` hands miniflare a bare assets directory, so a non-Vite Worker using `assets.base` serves at the root locally. Vite dev is unaffected — its dev server honours `base` itself.

### Possible follow-up

`routes: ["example.com/docs*"]` with assets keyed at the root is always broken. Comparing each route's path prefix against the manifest keys would catch it at plan time, before the upload session. Left out of this PR as it is a separate, opinionated diagnostic.

### DIFF
--- a/packages/alchemy/src/Bundle/Vite.ts
+++ b/packages/alchemy/src/Bundle/Vite.ts
@@ -17,6 +17,7 @@ import {
 
 export interface ViteBuildOutput {
   readonly clientDirectory: string | undefined;
+  readonly base: string | undefined;
   // This is emitted as an Effect instead of a value so we can process it in parallel with reading the client assets.
   readonly serverBundle: Effect.Effect<BundleOutput | undefined, BundleError>;
   readonly externalWorkspaces: Effect.Effect<Set<string>, PlatformError>;
@@ -39,6 +40,7 @@ type RscManifestId = keyof typeof RSC_MANIFEST;
 interface EnvironmentLike {
   readonly name: string;
   readonly config: {
+    readonly base: string;
     readonly root: string;
     readonly build: { readonly outDir: string };
   };
@@ -56,6 +58,7 @@ export const viteBuildOutputPlugin = Effect.fn(function* ({
   const fs = yield* FileSystem.FileSystem;
   const path = yield* Path.Path;
   let clientDirectory: string | undefined;
+  let base: string | undefined;
   let serverEntry: string | undefined;
   const serverChunks = new Map<
     string,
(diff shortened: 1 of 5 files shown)


WaryaWayne:
sorry will fix and send newer one

PR #1048, opened 2026-08-02
fix(cloudflare/workers): apply Vite base to the assets manifest

sam-goodwin:
Fixes the bug reported in Discord: a `Website.Vite` site served behind a path — a zone route like `example.com/docs*`, or any Vite `base` other than `/` — deploys with every asset 404ing. Design and key insights from @WaryaWayne's [#1046](https://github.com/alchemy-run/alchemy/pull/1046).

Vite's `base` rewrites the URLs the build *emits*, never the directory it emits *into* — and Cloudflare's asset router matches request paths against manifest keys literally, stripping no prefix:

```
manifest key   /assets/x.js
request path   /docs/assets/x.js   → no match → 404
```

The fix keys the manifest with the base; disk layout is untouched. `vite.config.ts` is the source of truth for Vite sites — the resolved client-environment `base` is captured out of the build and handed to `readAssets`, so:

```ts
// vite.config.ts
export default defineConfig({ base: "/docs/" });

// alchemy.run.ts — nothing to configure
const docs = yield* Cloudflare.Website.Vite("Docs", {
  routes: [{ pattern: "example.com/docs*", zoneName: "example.com" }],
});
```

For non-Vite deploys (prebuilt directories, `StaticSite`) the same knob is a first-class prop: `assets: { directory: "./dist", base: "/docs" }`.

Details that matter (all from #1046):

- **SPA fallback** — Cloudflare's `notFoundHandling: "single-page-application"` resolves `/index.html` at the manifest root, hard-coded, so the shell is aliased back to `/index.html`: one extra manifest line, zero extra uploads (entries are content-addressed).
- **Bases that name no path** (`/`, `./`, `//cdn…`, `https://cdn…`) are ignored — an absolute base means a CDN serves the assets, not this Worker.
- **`_headers` / `_redirects`** are sent verbatim, not prefixed: their rules match the incoming request path, which already carries the base.
- With no base, manifest keys and hashes are byte-identical to before — no churn for existing deploys.

Live-verified end-to-end: the regression test sets `base` in the fixture's `vite.config.ts`, deploys behind `alchemy-test-2.us/<prefix>/app*`, and round-trips the shell, the hashed module script (asserting a client-module literal so the SPA fallback can't fake a 200), and a deep client-side route through both the zone route and workers.dev.

Known limitation (unchanged from #1046): `alchemy dev` hands the local runtime a bare assets directory, so a non-Vite Worker using `assets.base` serves at the root locally; Vite dev honours `base` itself.

Possible follow-up: routes with a path prefix + a root-keyed manifest is always broken — a plan-time diagnostic comparing route prefixes against manifest keys would catch it before the upload session.

Co-authored with @WaryaWayne (#1046).

### DIFF
--- a/packages/alchemy/src/Cloudflare/Website/Vite.ts
+++ b/packages/alchemy/src/Cloudflare/Website/Vite.ts
@@ -139,6 +139,21 @@ export interface ViteProps<Bindings extends WorkerBindingProps = {}>
  * });
  * ```
  *
+ * @section Serving on a Zone Route with a Path Prefix
+ * Cloudflare matches static assets against the full request pathname,
+ * so a site attached to a route like `example.com/docs*` only serves
+ * assets nested under `/docs`. Set `assets.basePath` and Alchemy nests
+ * the uploaded asset paths under the prefix and applies it as Vite's
+ * `base`, so the emitted asset URLs resolve under the route.
+ *
+ * @example Vite site behind a path-prefixed route
+ * ```typescript
+ * const docs = yield* Cloudflare.Website.Vite("Docs", {
+ *   assets: { basePath: "/docs" },
+ *   routes: [{ pattern: "example.com/docs*", zoneName: "example.com" }],
+ * });
+ * ```
+ *
  * @section Custom Rebuild Scope
  * By default, every non-gitignored file is hashed to decide whether
  * a rebuild is needed. Use `memo` to narrow the scope when your
--- a/packages/alchemy/src/Cloudflare/Website/Vite.ts
+++ b/packages/alchemy/src/Cloudflare/Website/Vite.ts
@@ -142,14 +142,15 @@ export interface ViteProps<Bindings extends WorkerBindingProps = {}>
  * @section Serving on a Zone Route with a Path Prefix
  * Cloudflare matches static assets against the full request pathname,
  * so a site attached to a route like `example.com/docs*` only serves
- * assets nested under `/docs`. Set `assets.basePath` and Alchemy nests
- * the uploaded asset paths under the prefix and applies it as Vite's
- * `base`, so the emitted asset URLs resolve under the route.
+ * assets whose uploaded paths carry the `/docs` prefix. Set Vite's
+ * `base` in your `vite.config.ts` — the emitted HTML references its
+ * assets under the prefix, and Alchemy keys the uploaded asset manifest
+ * with the same resolved `base` so the two always agree.
  *
  * @example Vite site behind a path-prefixed route
  * ```typescript
+ * // vite.config.ts: export default defineConfig({ base: "/docs/" })
  * const docs = yield* Cloudflare.Website.Vite("Docs", {
- *   assets: { basePath: "/docs" },
  *   routes: [{ pattern: "example.com/docs*", zoneName: "example.com" }],
  * });
  * ```
--- a/packages/alchemy/src/Cloudflare/Website/Vite.ts
+++ b/packages/alchemy/src/Cloudflare/Website/Vite.ts
@@ -147,9 +147,17 @@ export interface ViteProps<Bindings extends WorkerBindingProps = {}>
  * assets under the prefix, and Alchemy keys the uploaded asset manifest
  * with the same resolved `base` so the two always agree.
  *
- * @example Vite site behind a path-prefixed route
+ * @example vite.config.ts
  * ```typescript
- * // vite.config.ts: export default defineConfig({ base: "/docs/" })
+ * import { defineConfig } from "vite";
+ *
+ * export default defineConfig({
+ *   base: "/docs/",
+ * });
+ * ```
+ *
+ * @example alchemy.run.ts
+ * ```typescript
  * const docs = yield* Cloudflare.Website.Vite("Docs", {
  *   routes: [{ pattern: "example.com/docs*", zoneName: "example.com" }],
  * });
--- a/packages/alchemy/src/Bundle/Vite.ts
+++ b/packages/alchemy/src/Bundle/Vite.ts
@@ -17,6 +17,12 @@ import {
 
 export interface ViteBuildOutput {
   readonly clientDirectory: string | undefined;
+  /**
+   * The client environment's resolved Vite `base`. The build rewrites
+   * every emitted asset URL with it, so the uploaded asset manifest must
+   * be keyed with the same prefix to agree with the HTML.
+   */
+  readonly base: string | undefined;
   // This is emitted as an Effect instead of a value so we can process it in parallel with reading the client assets.
   readonly serverBundle: Effect.Effect<BundleOutput | undefined, BundleError>;
   readonly externalWorkspaces: Effect.Effect<Set<string>, PlatformError>;
@@ -39,6 +45,7 @@ type RscManifestId = keyof typeof RSC_MANIFEST;
 interface EnvironmentLike {
   readonly name: string;
   readonly config: {
+    readonly base: string;
     readonly root: string;
     readonly build: { readonly outDir: string };
   };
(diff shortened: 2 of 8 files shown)


WaryaWayne:
please see #1046. closed PR that fixes this same issue as well but had merge conflicts before I closed it.

WaryaWayne:
I hit this same bug from the other end — a path-mounted TanStack Start app on the zone route `example.com/example*` that deployed with zero CSS and zero JS — and had a fix in flight (#1046, closed in favour of this one; branch still at `WaryaWayne/alchemy-effect@fix/vite-base-assets` if the diff is useful). Went through this PR carefully since I've been living in this code path. `basePath` as an explicit prop is the better primary design than what I did — it covers `StaticSite` and plain `Worker`, which deriving from Vite never could. A few cases below that look uncovered, most important first.

---

### 1. `basePath` + `notFoundHandling: "single-page-application"` 404s every client route

Cloudflare's asset worker resolves the SPA fallback against the manifest **root**, hard-coded. From miniflare's vendored copy of the production worker (`miniflare/dist/src/workers/assets/assets.worker.js`, `notFound()`):

```js
notFound = async (pathname, request, configuration, exists) => {
  switch (configuration.not_found_handling) {
    case "single-page-application": {
      let eTag = await exists("/index.html", request);   // ← root, always
```

`readAssets` now prefixes every manifest key, so under `basePath: "/docs"` the manifest holds `/docs/index.html` and nothing at `/index.html` — the fallback finds no asset and returns `null`, so `/docs/dashboard` 404s.

`404-page` is unaffected: it walks *up* from the request path (`${cwd}/404.html`), so `/docs/404.html` resolves naturally under a prefix. Only the SPA case is rooted.

This combination is already a supported pattern here — the Foldkit test added in #983 uses `notFoundHandling: "single-page-application"`; it just never sets `basePath`, so nothing currently catches it.

Aliasing the shell back to the root fixes it for one manifest line and zero extra uploads, since manifest entries are content-addressed:

```ts
const indexHtml = manifest.get(`${prefix}/index.html`);
if (
  indexHtml &&
  config.notFoundHandling === "single-page-application" &&
  !manifest.has("/index.html")
) {
  manifest.set("/index.html", indexHtml);
}
```

`uploadAssets` needs no change — both keys strip back to the same path on disk, so the dedupe map is unaffected.

### 2. A `base` already set in `vite.config.ts` is never read back

This is the structural difference between the two approaches, and I think it's the one that matters.

This PR makes `basePath` the source of truth and pushes it *into* Vite as inline config (`Workers/Vite.ts`, `base: overrides?.base`). It never reads the resolved config back — `src/Bundle/Vite.ts` is untouched, so `this.environment.config.base` is still ignored in `viteBuildOutputPlugin`'s `writeBundle`.

That leaves three configurations, and only one of them works:

| Config | Result |
| --- | --- |
| `base` in `vite.config.ts`, no `assets.basePath` | **Silently broken.** HTML emits `/docs/assets/…`, manifest rooted at `/`. No error. |
| `assets.basePath` only | Works. |
| Both | Works, but one value declared in two places with no cross-check. Drift → silent 404s. |

Row 1 is the shape most people who hit this bug already have, because setting `base` in the Vite config is the documented way to serve a Vite app from a subpath — and it's the shape in the original report.

It's also not a shape framework users can simply abandon. TanStack Start derives the router basepath from the public base (`router.basepath ?? publicBase`), so `base` is what makes client-side links respect the prefix; it isn't only an asset-URL concern. Moving it into `alchemy.run.ts` does keep that working — inline config merges into the resolved config before plugin hooks run, so `tanstackStart` still sees `config.base` — but it means a plain `vite build` or `vite dev` outside alchemy no longer reproduces the deployed layout. Anything not routed through alchemy (local dev server, CI build assertions, the `vite build && grep -ro '/docs/assets/'` sanity check) loses the prefix.

Suggestion, keeping `basePath` primary:

- when `basePath` is unset, fall back to the client environment's resolved `base` (read it in `viteBuildOutputPlugin`'s `writeBundle`, where `this.environment.config.base` is already in scope, and normalize it the same way);
- when both are set and disagree, fail loudly rather than letting inline config silently win.

That makes row 1 work, keeps rows 2 and 3 as they are, and keeps `basePath` meaningful for `StaticSite` / plain `Worker` where there is no Vite config to read.

### 3. `basePath` may not invalidate the precomputed-hash skip path

Flagging rather than asserting — I haven't executed this one.

On the `Command.Build` / `StaticSite` path, `normalizePrebuiltAssets` returns `skip: hash === output?.hash?.assets`, and `skip` sets `keepAssets = true`, which preserves the manifest already on Cloudflare. Changing only `basePath` doesn't change the build output, so a caller-supplied hash stays identical, so `skip` stays `true` and the old root-keyed manifest survives the deploy.

The metadata hash *does* change (`workerAssetConfigForHash` strips only `directory` and `hash`, so `basePath` is still in it), so a PUT happens — but `keepAssets` means it carries the stale manifest forward. Worth a test that flips `basePath` on a worker with a precomputed hash and asserts the manifest actually re-keys.

### 4. `_headers` / `_redirects` need to be authored *with* the prefix

Both are matched against the incoming request path, which carries the prefix. Under `basePath: "/docs"` a rule has to read `/docs/old /docs/new 301`, not `/old /new 301`.

Passing them through unrewritten is the right call — rewriting would break anyone who already wrote absolute paths — but the current JSDoc ("read from the asset directory root and applied as site-wide config") reads like they're prefix-relative. One clarifying sentence would save someone real debugging time.

### 5. `normalizeAssetsBasePath` mangles non-path values

```ts
normalizeAssetsBasePath("https://cdn.example.com/")  // → "/https:/cdn.example.com"
normalizeAssetsBasePath("//cdn.example.com/")        // → "/cdn.example.com"
```

It strips leading slashes from anything, so an absolute or protocol-relative value becomes a nonsense path prefix. Only reachable if someone assumes it takes a Vite-style `base` — which Vite *does* accept as an absolute URL, and which point 2 makes a natural assumption. Rejecting (or ignoring) anything that isn't root-relative is a two-line guard.

---

Happy to open a PR for any of these on top of this one once it lands. The SPA alias in particular is ~8 lines and I already have a regression test for it, plus unit coverage for base normalization, the SPA alias firing only in SPA mode, `_headers`/`_redirects` passthrough, and hash sensitivity to the prefix.

One unrelated note while I'm here, since it cost me an hour: the test suite is currently broken on **bun 1.3.14** — every `layer()`-based file (28 of them) fails collection with `alchemy-test: describe/test/hook called outside of a test file collection`, because `AsyncLocalStorage` context is lost across the collector's dynamic import. `test/Bundle/RawPlugin.test.ts` on a clean `main` is 13/13 under the pinned 1.3.13 and 0/13 under 1.3.14. The `devEngines` warning is the only signal, and it scrolls past during install. Happy to file that separately if it's useful.

sam-goodwin:
Thanks — this review landed just after 8aa62f630 reworked the PR to your #1046 design, so the SPA alias, the resolved-base capture, the base normalization, and the `_headers`/`_redirects` wording were already in (largely from your branch — you're co-authored on the commits).

Point 3 was real: fixed in 01c3fbf4f by salting the precomputed hash with the prefix, with a live regression in `AssetsBase.test.ts` that flips `base` under a constant hash.

A follow-up PR for the plan-time route-prefix diagnostic would be welcome.

On bun: 1.3.14 is a known break — stay on the pinned 1.3.13 for now; hopefully bun 1.4 sorts it.

packages/alchemy/src/Cloudflare/Website/Vite.ts
```diff
@@ -139,6 +139,22 @@ export interface ViteProps<Bindings extends WorkerBindingProps = {}>
  * });
  * ```
  *
+ * @section Serving on a Zone Route with a Path Prefix
+ * Cloudflare matches static assets against the full request pathname,
+ * so a site attached to a route like `example.com/docs*` only serves
+ * assets whose uploaded paths carry the `/docs` prefix. Set Vite's
+ * `base` in your `vite.config.ts` — the emitted HTML references its
+ * assets under the prefix, and Alchemy keys the uploaded asset manifest
+ * with the same resolved `base` so the two always agree.
```
sam-goodwin, line 148: Should we show an example vite.config.ts

sam-goodwin: Done in d19027bfd — the section now shows the `vite.config.ts` and `alchemy.run.ts` as separate examples.

PR #1053, opened 2026-08-02
fix(cloudflare/website)!: stop StaticSite duplicating env resources into its namespace

### LINKED ISSUE #1052: Cloudflare.Website.StaticSite duplicates every `env` resource into its own namespace
`Cloudflare.Website.StaticSite` wraps its entire body in `Namespace.push(id)`, so every resource its `env` bindings declare is re-declared as a second copy under `<id>/`. `Cloudflare.Website.Vite` does not do this, which is why the same `env` pattern works there.

### Repro

```ts
const Cache = Cloudflare.KV.Namespace("Cache", {});

Effect.gen(function* () {
  yield* Cache;
  yield* Cloudflare.Website.StaticSite("Site", {
    command: "echo build",
    outdir: "dist",
    main: "./worker.ts",
    env: { CACHE: Cache },
  });
});
```

Registered FQNs:

```
StaticSite:  [ "Cache", "Site/Build", "Site/Cache", "Site/Worker" ]
Vite:        [ "Cache", "Site" ]
```

`Site/Cache` is a duplicate KV namespace — a second physical resource for the same logical one. With a Worker in `env` this means two deployed Workers for one declaration, `created_on` a second apart within a single deploy.

### Cause

[`packages/alchemy/src/Cloudflare/Website/StaticSite.ts#L254-L306`](https://github.com/alchemy-run/alchemy/blob/main/packages/alchemy/src/Cloudflare/Website/StaticSite.ts#L254-L306)

```ts
const makeStaticSite = (id, propsEff) =>
  Effect.gen(function* () {
    const props = yield* asEffect(propsEff);
    // ...
    return yield* Worker("Worker", { ...props, /* ... */ });
  }).pipe(Namespace.push(id)); // <- wraps the whole body
```

`props.env` values are Effects, and they are yielded *inside* the pushed namespace — by `serializeEnv` for the build subprocess, and by `bindWorkerAsyncBindings` for the Worker binding — so the resources they declare register at `<id>/…` instead of the caller's scope.

`Vite` passes `id` straight to `Worker(id, …)` and never pushes a namespace, so its `env` resolves in the caller's scope. Only `Dev` and `Build` actually need the namespace; they carry constant logical ids.

### Reproduced against

`main` @ 42625a4 (`docs: clarify multiple providers per stack (#1051)`), with a stack-compile FQN assertion — no cloud calls:

```
error: expect(received).toEqual(expected)

@@ -2,5 +2,6 @@
    "Cache",
-   "Site",
    "Site/Build",
+   "Site/Cache",
+   "Site/Worker",
  ]

(fail) StaticSite declares env resources in the caller's namespace [14.00ms]
(pass) Vite declares env resources in the caller's namespace
```

### Note on the fix

Making `StaticSite` match `Vite` — namespace only `Build`/`Dev`, pass `id` to `Worker` — moves the Worker's FQN from `<id>/Worker` to `<id>`. That is a breaking state change for existing users: the old row orphans and the site is recreated under a new physical name. `aliases` is resource-*type* only, so there is no FQN-level migration hook today.

apostolos-geyer:
Fixes #1052.

`makeStaticSite` wrapped its whole body in `Namespace.push(id)`, so the Effects in `props.env` resolved inside the pushed namespace and the resources they declare registered at `<id>/…` — a second physical copy of every shared resource. Only `Dev`/`Build` need it; they carry constant logical ids. `Vite` already passes `id` straight to `Worker`.

```diff
-        : yield* Command.Build("Build", { … });
+        : yield* Command.Build("Build", { … }).pipe(Namespace.push(id));

-    return yield* Worker<Bindings, WorkerAssetsConfig, Req>("Worker", {
+    return yield* Worker<Bindings, WorkerAssetsConfig, Req>(id, {
       …
     });
-  }).pipe(Namespace.push(id));
+  });
```

Registered FQNs for a site with `env: { CACHE: Cache }` where `Cache` is declared by the caller:

```diff
- [ "Cache", "Site/Build", "Site/Cache", "Site/Worker" ]
+ [ "Cache", "Site", "Site/Build" ]
```

### Breaking

The Worker's FQN moves from `<id>/Worker` to `<id>`, so its physical name changes. On upgrade the engine creates the site at the new FQN and orphans the old row — the Worker is recreated and gets a new `workers.dev` subdomain. `aliases` is resource-*type* only, so there is no FQN-level migration hook; if one is wanted before this lands, say so and I'll rework it.

### DIFF
--- a/packages/alchemy/src/Cloudflare/Website/StaticSite.ts
+++ b/packages/alchemy/src/Cloudflare/Website/StaticSite.ts
@@ -255,6 +255,14 @@ const makeStaticSite = <
     const ctx = yield* AlchemyContext;
     const props = yield* asEffect(propsEff);
 
+    // `Dev` and `Build` carry constant logical ids, so they are namespaced
+    // under `id` to keep two sites on one stack from colliding. Nothing
+    // else is: the site's own `props` — and the resources its `env`
+    // bindings declare — must resolve in the CALLER's namespace. Pushing
+    // the namespace around the whole body instead re-declares a shared
+    // resource referenced from `env` as a second copy under `<id>/`.
+    // `Vite` already passes `id` straight through to `Worker`.
+
     // In dev mode with a dev.command, declare a DevCommand resource so
     // the sidecar owns the process lifecycle (survives user-code HMR),
     // skip the build, and tell Worker not to start a local instance.
@@ -267,6 +275,7 @@ const makeStaticSite = <
               (typeof props.cwd === "string" ? props.cwd : undefined),
             env: yield* serializeEnv(props.dev.env ?? props.env),
           }).pipe(
+            Namespace.push(id),
             Effect.map((d) =>
               Output.map(d.url, (url) => ({
                 url: url ?? props.dev?.url,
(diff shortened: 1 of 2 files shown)


sam-goodwin:
Good catch. This is a severe drawback of allowing Effects to be yielded and referenced multiple times and also allowing Namespace.push to scope things.

Issue #874, opened 2026-07-18
Effect-valued Worker `env` entries never converge: plans re-update forever


johnrjj:
**Version**: alchemy 2.0.0-beta.63 (also reproduced conceptually against beta.61 sources)

**Summary**: Binding a tagged Worker identity (Effect class) directly in another async Worker's `env` — the pattern the [circular bindings guide](https://v2.alchemy.run/infrastructure-as-effects/circular-bindings/) suggests — produces a **correct live service binding** but a **permanently non-convergent plan**: the Worker re-plans as `update` on every subsequent plan/deploy, forever.

**Reproduction shape** (two async Workers, tagged identities, `isExternal: true`):

```ts
class ExecutionEngineWorker extends Cloudflare.Worker("ExecutionEngineWorker") {}
class AutomationEngineWorker extends Cloudflare.Worker("AutomationEngineWorker") {}

// automation binds execution via the tagged identity in env:
AutomationEngineWorker.make(
  { ...props, env: { EXECUTION_ENGINE: ExecutionEngineWorker } },
  Effect.succeed({}),
)
```

Deploy succeeds and the live binding is correct (verified byte-identical via the Workers settings API). But every subsequent `alchemy plan` shows the Worker as `update`. We verified across **two consecutive applies** — the diff never settles.

**Root cause** (from installed beta.63 sources):

1. `WorkerAsyncBindings` resolves the Worker tag from `env` and produces the service binding from `binding.workerName` — but the original Effect tag remains in desired `props.env` (`src/Cloudflare/Workers/WorkerAsyncBindings.ts:39`).
2. The Worker provider's diff exits immediately for unresolved props: `if (!isResolved(news)) return undefined;` (`src/Cloudflare/Workers/WorkerProvider.ts:2208`), so planning falls back to `havePropsChanged(oldProps, news)` (`src/Plan.ts:863`).
3. On apply, `stripUnresolved` removes the Effect from **stored** props — `Diff.ts:72` maps Effects to `undefined` (`src/Apply.ts:406`).
4. Every later plan therefore compares stored `{ env: {} }` against desired `{ env: { BINDING: <EffectTag> } }` → props changed → update, forever.

The metadata-hash sanitizer's Effect handling never runs because the provider returns at step 2.

[...]

PR #1055, opened 2026-08-02
fix(platform): name the missing layer when a tagged resource is yielded without its impl

### LINKED ISSUE #1054: Yielding a tagged resource without its `.make(...)` layer reaches the provider with undefined props (opaque `TypeError` in Worker pre-create)
A tagged Effect-native resource declared as a bare tag gets both its props and its implementation from `.make(props, impl)`. Yielding the class without that Layer in scope hands the provider `news === undefined`, so the failure surfaces as an opaque `TypeError` deep inside the provider instead of an error naming the class and the missing Layer.

For a Cloudflare Worker:

```ts
export class ApiWorker extends Cloudflare.Worker<ApiWorker, {}>()("ApiWorker") {}
export const ApiWorkerLive = ApiWorker.make({ main: import.meta.url }, Effect.succeed({}));

export default Alchemy.Stack(
  "WorkerMissingImplStack",
  { providers: Cloudflare.providers(), state: Cloudflare.state() },
  Effect.gen(function* () {
    const api = yield* ApiWorker;      // ApiWorkerLive never provided
    return { url: api.url.as<string>() };
  }),
);
```

```
[ApiWorker] pre-creating
[ApiWorker] fail
Cause([Die(TypeError: undefined is not an object (evaluating 'news.name'))])
```

That line is `const name = yield* createWorkerName(id, news.name)` in `WorkerProvider.ts`'s `precreate`. Nothing points at the real mistake, which is the missing `Effect.provide([ApiWorkerLive])`.

### Mechanism

`packages/alchemy/src/Platform.ts` — the tagged branch, when the class's `Self` tag is absent from the context:

```ts
onNone: () =>
  resource(
    id,
    props === undefined
      ? applyTransformProps(id, props)   // <- stays `undefined`
      : externalProps(),
  ),
```

A tagged class WITH props and no impl is the external (non-Effect-native) form, and `externalProps()` spreads to at least `{ isExternal: true }`. A bare tag has neither props nor impl — both live on the un-provided `.make(...)` Layer — so `applyTransformProps(id, undefined)` returns `undefined` verbatim (no `transformProps` hook on most platforms) and the resource is registered with no props at all. `cls.make` normalizes the same value with `transformed ?? {}`; this path does not.

This is not Worker-specific: every `Platform(...)` consumer (Workers, Durable Objects, Containers, Workflows, `AWS.Lambda.Function`, `AWS.ECS.Service`/`Task`, `Kubernetes.Deployment`/`Job`, `Prisma.Compute`, …) reaches its provider with `news === undefined` in this situation. A provider tolerant of missing props deploys a silently misconfigured resource rather than crashing.

### Reproduction

Hermetic, no cloud credentials, plain `bun test` against repo source. A synthetic `Platform` resource whose provider records what it receives:

```ts
const Widget: any = Platform<Widget>("Test.PlatformWidget", {
  createRuntimeContext: () => ({}) as any,
});

const providers = Provider.succeed(Widget, {
  reconcile: Effect.fn(function* ({ id, news }: any) {
    observed = { ran: true, news };
    return { name: news?.name ?? id };
  }),
  // list/diff/delete elided
});

class BareWidget extends Widget()("BareWidget") {}   // no props, no impl, no layer

const Stack = Alchemy.Stack("PlatformMissingImplStack", { providers, state },
  Effect.gen(function* () {
    const widget = yield* BareWidget;
    return { name: widget.name };
  }),
);
```

```
provider ran: true | news: undefined
deploy SUCCEEDED
```

The Cloudflare Worker repro above is equally credential-free — the crash happens before any API call.

### Expected

Fail fast at the yield site with an error naming the class and telling the user to provide its `.make(...)` Layer, e.g.

```
MissingImplementationError: Cloudflare.Worker<ApiWorker> was yielded without its implementation.

`ApiWorker` is declared as a bare tag — no props, no inline implementation — so both come from its `.make(...)` Layer:

  export class ApiWorker extends Cloudflare.Worker<ApiWorker>()("ApiWorker") {}
  export const ApiWorkerLive = ApiWorker.make({ /* props */ }, Effect.gen(function* () { /* ... */ }));

That Layer is not in scope where `ApiWorker` is yielded. Provide it:

  Effect.gen(function* () {
    const instance = yield* ApiWorker;
  }).pipe(Effect.provide([ApiWorkerLive]))
```

The correct user action is documented by `examples/cloudflare-worker/alchemy.run.ts`, which pipes `Effect.provide([WorkerTagLive, SecondaryApiLive, SandboxLive])`. Only the diagnostics are wrong.

Guarding `news?.name` in `WorkerProvider` is the wrong fix — `undefined` props are legitimate elsewhere (the assets-only Worker has neither `main` nor `script`), and it would only move the crash to the next prop read in each of the other platforms.

Reproduced on `42625a4` (upstream `main`). Related but distinct: #1049 / #1050 cover a different pre-create crash (an Output-valued `main` in `isPythonMain`).

apostolos-geyer:
Fixes #1054.

A bare tag — no props, no inline impl — gets both from its `.make(props, impl)` Layer. Yielding it without that Layer in scope registered the resource with `undefined` props, so the failure surfaced as a bare `TypeError` deep in whichever provider first read a prop (`news.name` in `WorkerProvider`'s pre-create).

The yield site can't simply die: a bare-tag yield is also the legitimate forward-reference pattern from #874 — a worker tag bound in another worker's `env` resolves during that worker's async-binding pass, possibly *before* the tag's own `.make` Layer builds and repairs the registration's `Props`. The yield site cannot distinguish a Layer that is merely late from one that is absent.

The fact is static — a property of the resource *type*, declared once where `Platform()` constructs its resource class. Every legitimate platform path (a `.make(props, impl)` Layer build, a tag declared with props, a plain `Worker("id", props)` call) produces defined `Props`, so a platform-typed registration reaching the planner with `Props === undefined` can only be a bare-tag forward reference whose Layer never built:

```ts
// Platform.ts — declared once at type construction; the yield site is untouched
const resource = Resource(type, {
  aliases: hooks.aliases,
  requiresImplementation: true,
});

// Plan.make — the .make Layer's build repairs Props (existing behavior),
// so a satisfied forward reference is inert; a surviving one dies here,
// before any provider runs
if (resource.RequiresImplementation && resource.Props === undefined) {
  yield* Effect.die(missingImplementation(resource.Type, resource.LogicalId));
}
```

Scoping the check to platform types keeps plain no-props reference yields (`yield* Queue("MyQueue")`) planning as noops.

```
MissingImplementationError: Cloudflare.Worker<ApiWorker> was yielded without its implementation.

`ApiWorker` is declared as a bare tag — no props, no inline implementation — so both come from its `.make(...)` Layer:

  export class ApiWorker extends Cloudflare.Worker<ApiWorker>()("ApiWorker") {}
  export const ApiWorkerLive = ApiWorker.make({ /* props */ }, Effect.gen(function* () { /* ... */ }));

That Layer is not in scope where `ApiWorker` is yielded. Provide it to the Stack's program:

  Alchemy.Stack(
    "my-stack",
    { providers, state },
    Effect.gen(function* () {
      const instance = yield* ApiWorker;
    }).pipe(Effect.provide([ApiWorkerLive])),
  )
```

Guarding `news?.name` in `WorkerProvider` was rejected: `undefined` props are legitimate elsewhere (assets-only Workers have neither `main` nor `script`), and the same latent crash exists in every other `Platform(...)` consumer. The three neighbouring branches — tag + Layer provided, tag + props (external), plain non-tagged — are unchanged and covered by `test/platform-missing-impl.test.ts`; the #874 env-tag patterns (including the circular case) are covered by the full `test/Cloudflare/Workers` suite.


packages/alchemy/src/Platform.ts
```diff
@@ -43,6 +44,50 @@ export interface PlatformProps {
...
+      "",
+      `  export class ${id} extends ${type}<${id}>()("${id}") {}`,
+      `  export const ${id}Live = ${id}.make({ /* props */ }, Effect.gen(function* () { /* ... */ }));`,
+      "",
+      `That Layer is not in scope where \`${id}\` is yielded. Provide it:`,
+      "",
+      "  Effect.gen(function* () {",
+      `    const instance = yield* ${id};`,
+      `  }).pipe(Effect.provide([${id}Live]))`,
+      "",
```
sam-goodwin: It needs to be provided to the Stack by the way. Might want to clarify.

Issue #1047, opened 2026-08-02
Cloudflare Worker with directory assets never converges to a no-op plan


magoz:
### Version

`alchemy@2.0.0-beta.66`

The same behavior appears to be present on `main`.

### What happened?

A Cloudflare Worker using directory assets without a caller-supplied hash always appears as changed in `alchemy plan`, even when the directory contents are unchanged.

For example:

```ts
yield* Cloudflare.Worker("Worker", {
  main: "./src/worker.ts",
  assets: {
    directory: "./dist",
  },
});
```

After deploying once, running `alchemy plan` again reports a Worker update.

`readAssets` calculates and stores a deterministic asset hash while applying the update, but the Worker diff returns changed when `assets.hash` is absent:

- [`readAssets` calculates the asset hash](https://github.com/alchemy-run/alchemy/blob/v2.0.0-beta.66/packages/alchemy/src/Cloudflare/Workers/Assets.ts#L156-L253)
- [The Worker diff conservatively returns changed](https://github.com/alchemy-run/alchemy/blob/v2.0.0-beta.66/packages/alchemy/src/Cloudflare/Workers/WorkerProvider.ts#L3863-L3888)

### Expected behavior

When the Worker bundle, configuration, and asset directory contents are unchanged, `alchemy plan` should report a no-op without requiring the caller to implement separate directory hashing.

### Current workaround

Supplying a deterministic `assets.hash` makes successive plans converge, but this requires duplicating asset hashing in application code.

### Context

I encountered this while adding an Alchemy deployment path for Executor’s Cloudflare host. The host builds `dist` before each plan; even when two builds produce identical contents, the Worker still plans an update unless we supply `assets.hash`.

The current workaround computes a content hash for `dist` and passes it to the Worker:

https://github.com/UsefulSoftwareCo/executor/pull/1510#discussion_r3698886427

The review discussion concluded that Alchemy should ideally handle this and suggested opening this issue.

magoz:
It looks like #1045 addresses this exact behavior and was opened shortly before this issue. I’ll leave this open to track the fix through merge and release unless you’d prefer it closed.

PR #1057, opened 2026-08-03
fix(cloudflare/website): pass a string assets hash from StaticSite

### LINKED ISSUE #1056: Cloudflare.Website.StaticSite passes an object as the assets hash, so every deploy re-uploads the assets
### Version

`main` @ 42625a4d8611d2cbd636f1828c6933e163602f7d

### What happened?

`Cloudflare.Website.StaticSite` hands the Worker an **object** where the assets contract declares a `string`, so every `alchemy deploy` re-uploads the whole asset directory even when nothing changed.

`Command.Build`'s `hash` attribute is a pair, not a string ([`Command/Build.ts#L51-L62`](https://github.com/alchemy-run/alchemy/blob/42625a4d8611d2cbd636f1828c6933e163602f7d/packages/alchemy/src/Command/Build.ts#L51-L62)):

```ts
hash: {
  input: string | undefined;
  output: string | undefined;
};
```

`AssetsWithHash.hash` is a single `string` ([`Workers/Worker.ts#L214-L221`](https://github.com/alchemy-run/alchemy/blob/42625a4d8611d2cbd636f1828c6933e163602f7d/packages/alchemy/src/Cloudflare/Workers/Worker.ts#L214-L221)).

StaticSite passes the pair straight through; the `cast(...)` swallows the mismatch ([`Website/StaticSite.ts#L293-L299`](https://github.com/alchemy-run/alchemy/blob/42625a4d8611d2cbd636f1828c6933e163602f7d/packages/alchemy/src/Cloudflare/Website/StaticSite.ts#L293-L299)):

```ts
assets: build
  ? cast({
      directory: build.outdir,
      hash: build.hash,        // <- { input, output }, not a string
      ...props.assets,
    })
  : undefined,
```

Both consumers of that hash compare by **reference**:

```ts
// WorkerProvider.assetsChanged (diff / plan)
if (Predicate.hasProperty(assets, "hash") && assets.hash !== undefined) {
  return assets.hash !== output.hash?.assets;
}

// WorkerProvider.normalizePrebuiltAssets (apply)
skip: effectiveHash === output?.hash?.assets,
```

State round-trips through JSON, so a freshly-built object is never `===` the deserialized one. Result: the plan reports the Worker as changed forever, and the apply never takes the `skip` path — the entire asset directory is re-read and re-uploaded on every deploy.

There is a second-order effect too: when `base` is set, `normalizePrebuiltAssets` builds `` `${hash}#base=${pathPrefix}` `` — with an object that stringifies to `[object Object]#base=/…`.

### Why this is not covered by #1047 / #1045

[#1047](https://github.com/alchemy-run/alchemy/issues/1047) and its fix [#1045](https://github.com/alchemy-run/alchemy/pull/1045) cover the **hash-less** shape — `assets: { directory }` with no precomputed hash — by reading and hashing the directory during diff. #1045 deliberately keeps the precomputed-hash branch ahead of that new code path:

```ts
if (Predicate.hasProperty(assets, "hash") && assets.hash !== undefined) {
  return assets.hash !== output.hash?.assets;   // <- StaticSite still lands here
}
const read = yield* prepareAssets(assets)...    // <- #1045's new directory hashing
```

StaticSite always supplies a hash, so it takes the first branch and never reaches #1045's fix. Verified against `main` with #1045 merged (42625a4d).

Amusingly, #1047's stated workaround is "supply a deterministic `assets.hash`" — StaticSite already does, it just supplies the wrong shape.

### Reproduction

Offline, no cloud credentials: the Cloudflare Worker provider is replaced by a fake that records the resolved `assets` props and mirrors the real provider's precomputed-hash diff branch, and the state store round-trips writes through the same `encodeState` / `reviveState` JSON pipeline `localState()` uses (a plain in-memory store keeps object identity alive and hides the bug).

```ts
const Site = Alchemy.Stack(
  "StaticSiteHashTest",
  { providers: providers(), state: persistingState() },
  Effect.gen(function* () {
    yield* Cloudflare.Website.StaticSite("Site", {
      command: "bash build.sh",
      cwd,
      outdir: "dist",
    });
    return {};
  }),
);

yield* deploy(Site);
expect(typeof seenAssets[0].hash).toBe("string");
yield* deploy(Site);          // nothing changed on disk
expect(reconciles).toBe(1);
```

On `main` (42625a4d), `bun test`:

```
Plan: 2 to create
[Build] create
[Worker] create
...
Done: 4 succeeded
HASH TYPE: object
Plan: 1 to update, 1 to noop
[Build] noop
[Worker] update

[Worker] pending
[Worker] updating
[Worker] updated

Done: 2 succeeded
163 |     expect(reconciles).toBe(1);
                             ^
error: expect(received).toBe(expected)

Expected: 1
Received: 2

(fail) StaticSite passes a string assets hash that converges to a noop
```

The second deploy plans `[Worker] update` with zero source changes.

### Expected behavior

A `StaticSite` whose sources have not changed converges to a no-op plan, and the Worker does not re-upload the asset directory.

`Command.Build.hash.output` is the content hash of `outdir` — exactly the bytes being uploaded — so it is the value the Worker's assets diff wants. It is `undefined` only under `memo: false`, where an absent hash correctly makes the Worker hash the directory itself (now that #1045 landed that path).

apostolos-geyer:
`Command.Build`'s `hash` attribute is a `{ input, output }` pair, but `AssetsWithHash.hash` is a single `string` — and the `cast` hid the mismatch:

```diff
+    const assetsHash: Output.Output<string | undefined> | undefined = build
+      ? Output.map(build.hash, (hash) => hash.output)
+      : undefined;
+
     return yield* Worker<Bindings, WorkerAssetsConfig, Req>("Worker", {
       ...props,
       assets: build
         ? cast({
             directory: build.outdir,
-            hash: build.hash,
+            hash: assetsHash,
             ...props.assets,
           })
         : undefined,
```

The Worker compares that hash by reference (`assets.hash !== output.hash?.assets` in diff, `effectiveHash === output?.hash?.assets` in apply) against the value read back from state. State round-trips through JSON, so an object is never equal to its deserialized self: every plan reported the Worker as changed and every apply re-uploaded the whole asset directory.

`hash.output` is the content hash of the build's `outdir` — exactly the bytes being uploaded. Under `memo: false` no hashes are computed at all, so it is `undefined` and the Worker hashes the directory itself (the path #1045 added).

The local annotation is load-bearing: `cast` erases the object literal's shape, so it is the only thing type-checking the mapping.

`normalizePrebuiltAssets` now treats an explicitly-undefined `hash` as the hash-less shape, matching what `assetsChanged` already does:

```diff
-        if (!Predicate.hasProperty(assets, "hash")) return undefined;
+        if (
+          !Predicate.hasProperty(assets, "hash") ||
+          assets.hash === undefined
+        ) {
+          return undefined;
+        }
```

Without it a greenfield create (`output === undefined`) compares `undefined === undefined`, sets `skip`, and uploads a Worker with no assets.

Not covered by #1047/#1045: those fix the hash-*less* `assets: { directory }` shape. #1045 deliberately keeps the precomputed-hash branch ahead of the new directory hashing, and StaticSite always supplies a hash, so it never reaches it.

Closes #1056

### DIFF
--- a/packages/alchemy/src/Cloudflare/Website/StaticSite.ts
+++ b/packages/alchemy/src/Cloudflare/Website/StaticSite.ts
@@ -14,7 +14,6 @@ import {
 import { asEffect } from "../../Util/types.ts";
 import type { Providers } from "../Providers.ts";
 import type { AssetsConfig } from "../Workers/Assets.ts";
-import { isContainerDecl } from "../Workers/WorkerAsyncBindings.ts";
 import {
   Worker,
   type NormalizedBindings,
@@ -22,6 +21,7 @@ import {
   type WorkerBindingProps,
   type WorkerProps,
 } from "../Workers/Worker.ts";
+import { isContainerDecl } from "../Workers/WorkerAsyncBindings.ts";
 
 export interface StaticSiteProps<Bindings extends WorkerBindingProps = {}>
   extends
@@ -285,23 +285,6 @@ const makeStaticSite = <
           env: yield* serializeEnv(props.env),
         });
 
-    // `AssetsWithHash.hash` is a single `string` that the Worker's assets
-    // diff compares with `!==` against the hash persisted in state, while
-    // `Command.Build` exposes a `{ input, output }` pair. Passing the pair
-    // through made every plan dirty — state round-trips through JSON, so a
-    // freshly built object is never `===` the deserialized one — and the
-    // Worker re-uploaded the whole asset directory on every deploy.
-    //
-    // `output` is the hash of the build's `outdir` contents, i.e. exactly
-    // the bytes being uploaded. It is `undefined` only under `memo: false`,
-    // where an absent hash makes the Worker hash the directory itself.
-    //
-    // The annotation is deliberate: the `cast` below erases the shape of the
-    // whole object literal, so this is what keeps the mapping type-checked.
-    const assetsHash: Output.Output<string | undefined> | undefined = build
-      ? Output.map(build.hash, (hash) => hash.output)
-      : undefined;
-
     // Pure-static sites (neither `main` nor `script`) deploy as
     // assets-only Workers: no script is uploaded and Cloudflare's asset
     // layer serves every request itself.
@@ -310,7 +293,7 @@ const makeStaticSite = <
       assets: build
         ? cast({
             directory: build.outdir,
-            hash: assetsHash,
+            hash: build.hash.output,
             ...props.assets,
           })
         : undefined,
--- a/packages/alchemy/src/Cloudflare/Website/StaticSite.ts
+++ b/packages/alchemy/src/Cloudflare/Website/StaticSite.ts
@@ -285,6 +285,23 @@ const makeStaticSite = <
           env: yield* serializeEnv(props.env),
         });
 
+    // `AssetsWithHash.hash` is a single `string` that the Worker's assets
+    // diff compares with `!==` against the hash persisted in state, while
+    // `Command.Build` exposes a `{ input, output }` pair. Passing the pair
+    // through made every plan dirty — state round-trips through JSON, so a
+    // freshly built object is never `===` the deserialized one — and the
+    // Worker re-uploaded the whole asset directory on every deploy.
+    //
+    // `output` is the hash of the build's `outdir` contents, i.e. exactly
+    // the bytes being uploaded. It is `undefined` only under `memo: false`,
+    // where an absent hash makes the Worker hash the directory itself.
+    //
+    // The annotation is deliberate: the `cast` below erases the shape of the
+    // whole object literal, so this is what keeps the mapping type-checked.
+    const assetsHash: Output.Output<string | undefined> | undefined = build
+      ? Output.map(build.hash, (hash) => hash.output)
+      : undefined;
+
     // Pure-static sites (neither `main` nor `script`) deploy as
     // assets-only Workers: no script is uploaded and Cloudflare's asset
     // layer serves every request itself.
@@ -293,7 +310,7 @@ const makeStaticSite = <
       assets: build
         ? cast({
             directory: build.outdir,
-            hash: build.hash,
+            hash: assetsHash,
             ...props.assets,
           })
         : undefined,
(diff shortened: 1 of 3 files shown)


sam-goodwin:
Is there a regression test we can add that this change fixes?

apostolos-geyer:
Added two `test.provider` cases to `StaticSite.test.ts`.

The first is the defect itself — deploy, then re-plan with zero source changes:

```ts
const site = yield* stack.deploy(program());
yield* roundTripState;
const settled = yield* stack.plan(program());
expect(actionOf(settled, "NoopSite/Build")).toBe("noop");
expect(actionOf(settled, "NoopSite/Worker")).toBe("noop");
expect(typeof site.hash?.assets).toBe("string");
```

On `main` (src reverted, test unchanged — the fix is the only variable):

```
expect(actionOf(settled, "NoopSite/Worker")).toBe("noop");
Expected: "noop"
Received: "update"
(fail) StaticSite: redeploying an unchanged site is a noop (#1056)
```

With the fix: `1 pass`.

### The round-trip is load-bearing

`test.provider`'s scratch store keeps rows **by reference**, so an object-valued
hash compares `===` against itself and the bug is invisible — a version of this
test without `roundTripState` passes on unpatched `main` and proves nothing.
`roundTripState` pushes every row through `encodeState` → JSON → `reviveState`,
which is what `localState` (and R2, and the HTTP store) actually do.

### Second case

`memo: false` computes no hashes, so the site hands the Worker `hash: undefined`.
That is the hash-*less* shape; reading it as a supplied hash makes a greenfield
create compare `undefined === undefined`, set `skip`, and upload a Worker with no
assets. This one guards the `normalizePrebuiltAssets` change in this PR rather
than the original bug, so there is no meaningful "before" to show for it — it
pins a regression the fix could otherwise introduce.

Both pass against this branch; `bun tsc -b` is clean. Verified locally by running
a byte-identical copy under `bun test` (`@/Test/Bun` + `bun:test`) — the
`alchemy-test` CLI doesn't run in my environment: every file, including untouched
ones on `main`, fails collection with local bun 1.3.14 against the pinned 1.3.13.

PR #1061, opened 2026-08-03
feat(prisma): object storage (resources + bindings), locked Postgres state backend, Deployment.redeployOn

wmadden:
This PR completes three gaps in the Prisma provider that show up the moment a real framework embeds it. After it, this works:

```ts
// api.ts — the app entrypoint. `main: import.meta.filename` bundles this
// file, so the Compute is its default export.
import * as Prisma from "alchemy/Prisma";

export const project = Prisma.Project("app", {});

// Object storage: the last deferred Management API surface —
// resources *and* bindings.
export const media = Prisma.Bucket("media", { project });

export default Prisma.Compute(
  "Api",
  { project, appName: "api", main: import.meta.filename, port: 8080 },
  Effect.gen(function* () {
    // Binding a bucket auto-creates a scoped BucketAccessKey and wires
    // endpoint + credentials into this host's environment.
    const bucket = yield* Prisma.ReadWriteBucket(media);
    return {
        const logo = yield* bucket.get("logo.svg");
        // …
      }),
    };
  }).pipe(Effect.provide(Prisma.ReadWriteBucketBinding)),
);
```

```ts
// stack.ts
import * as Prisma from "alchemy/Prisma";
import { postgresState } from "alchemy/State/PostgresState";
import * as Config from "effect/Config";
import Api from "./api.ts";

export default Alchemy.Stack(
  "Media",
  {
    providers: Prisma.providers(),
    // Durable, *locked* state — what Compute's own docs ask for.
    state: postgresState({ url: Config.redacted("STATE_DATABASE_URL") }),
  },
  Effect.gen(function* () {
    const api = yield* Api;
    return { url: api.url };
  }),
);
```

The three ship together because they're the set a production embedder needs at once — we (Prisma Composer, the framework layer over Prisma Cloud) are deleting our own resource implementations in favor of this provider, and these were the things we couldn't do with it. Say the word and we'll split any of them out.

### 1. Object storage: resources + bindings

Covers the 7 previously deferred `/v1/buckets` routes (`deferredRoutes` is now empty in the contract fixture; coverage 71 → 78), with both halves of the equation:

**Resources.** `Bucket` (read-then-create reconcile, replace on identity change, identity-verified delete, real `list` for nuke) and `BucketAccessKey`. The key secret is **reveal-once** — the API never returns it after create, so persisted state is authoritative (the `Connection` pattern). Keys are created under a deterministic `instanceId`-derived physical name and looked up before create, so a crash between the create call and the state write cannot mint a second, unenumerable credential — the orphan from the lost response is found by name and revoked before a fresh key is created.

**Bindings.** `ReadBucket` / `WriteBucket` / `ReadWriteBucket`, following the R2 capability shape: binding a bucket auto-creates a `BucketAccessKey` for that (bucket, capability) — read-scoped for `ReadBucket`; the API has no write-only role, so write bindings carry `read_write` (documented). Per review, credential delivery is the accessor pattern: the binding yields the key's outputs, the host's `RuntimeContext` carries them (Redacted markers included), and no env-var names, host-type dispatch, or `secret_text` mapping exist anywhere in the binding. One provider-side fix fell out: Prisma Compute's runtime context now uppercases its storage keys, because the provider validates env keys to the uppercase POSIX shape — the live suite caught mixed-case keys being rejected. The runtime client reuses the distilled S3 machinery with an endpoint override (Prisma object storage is S3-compatible SigV4); get/head/put/delete/list plus presigned get/put, with presigned GET `contentType` signed as `response-content-type`, matching the S3 bindings.

Docs: `prisma/data/buckets` page (resources, bindings, reference links) + sidebar entry; the three bindings carry `@binding` reference docs.

### 2. Dual-registered dev mode

Per review: `providers()` takes no options; every resource registers its live and dev implementations via `ProviderLayer.dual` (#963), with the dev stubs moved in-module (`Internal/DevStub.ts`). The engine picks by run mode, `Alchemy.remote()` works per-resource, and an embedder that wants different behavior constructs its own layer from the per-resource exports. Defaults unchanged.

### 3. A locked state backend: `State/PostgresState`

`Compute`'s recovery docs tell users to "use a durable, locked state backend" — and no in-tree backend has locking. This adds one on the dependency the repo already carries (`pg`):

- per-`(stack, stage)` **session advisory lock** (`pg_try_advisory_lock(hashtextextended(key, 0))`) held on a reserved connection, with the holder re-verified against `pg_locks` from a *different* pool connection — so a silently dropped lock connection is detected, not trusted;
- a TTL-amortized lease check wrapping every operation; stage-less `deleteStack` locks each stage before touching it;
- schema migration under a transaction-scoped advisory lock, because concurrent `create table if not exists` genuinely fails on Postgres (duplicate `pg_type` errors — reproduced on PG 15).

Deliberately **not** re-exported from the `State` barrel (the barrel is bundled for workers; `pg` must stay off that graph — a comment says so). Deep import: `alchemy/State/PostgresState`. Tests are hermetic stubs per the existing backend convention; the two real-Postgres behaviors verified live during development are documented in comments.

### 4. `Deployment.triggers`

Prisma Cloud snapshots env vars into a deployment at create, and the low-level `Deployment` reuses the running deployment unless the artifact moves — so an env-value-only change (a rotated secret) updates the platform row but never reaches the running app. `triggers` takes the same name and shape as `AWS.ApiGateway.Deployment.triggers`: an opaque key/value map whose changed values plan a replacement. The Prisma-specific detail is storage — values here may be secrets, so they are folded into a salted fingerprint persisted `Redacted` (no plaintext in state) instead of compared as plaintext props. Old rows lack the field and never replace on upgrade.

### 5. One-line core fix: `Aliases` typing

`ResourceClass.Aliases` is `readonly string[] | undefined`, but `ResourceClassLike.Aliases?: readonly string[]` — under a consumer tsconfig with `exactOptionalPropertyTypes: true`, every `Provider.effect(cls, …)` call fails to typecheck (we carry a pnpm patch for this today). The fix widens the optional to `| undefined`.

### Verification

- `bun run format:check` clean; `bun tsc -b` (monorepo) clean.
- The bucket binding suite deploys a real stack per review: `fixtures/` holds three Compute apps binding one shared bucket (read / write / read-write), the tests drive every client operation over `HttpClient` through the deployed apps — including presigned PUT/GET fetched by the test with no credentials of its own — with `beforeAll(deploy)` / `afterAll(destroy)`. A cold full cycle (create → exercise → destroy, 8 resources) runs green against Prisma Cloud. Live runs are guarded by `ALCHEMY_RUN_LIVE_PRISMA_TESTS` + credentials, matching `Compute.live.test.ts`; without credentials the suite skips cleanly and the pure derivation tests still run.
- The live suite keeps catching what the stubbed suite structurally cannot: first the `Redacted` wrapper not surviving env delivery (now handled by the `RuntimeContext` marker round-trip), then — after the accessor rework — mixed-case runtime-context keys failing the provider's uppercase env-key validation. Both fixes were verified by re-running the deployed-host suite to green.
- `PostgresState` now has a live counterpart too (`PostgresState.live.test.ts`, guarded by `ALCHEMY_RUN_LIVE_POSTGRES_STATE_TESTS` + `POSTGRES_STATE_TEST_URL`): the advisory-lock exclusivity, lock release on scope close, and the concurrent transactional migration all run green against real Postgres 15.
- `bun alchemy-test --fast test/Prisma test/State`: 0 failed; core engine suites (exercising the `Resource.ts` change) green.
- `generate-api-reference`: Prisma gains Bucket/BucketAccessKey and the three binding pages; no new category; `docs:check` builds clean.

### Alternatives considered

- **Splitting into separate PRs.** Offered above — kept together because they're one consumer's complete need and the review context overlaps (BucketAccessKey's crash-window design references `Connection`'s; the bindings provision through the resources).
- **Per-operation bindings (the AWS S3 shape).** R2's capability shape fits better: Prisma has one transport and coarse key roles, so ~30 per-op bindings would grant identical credentials with more surface. The capability clients still expose typed per-operation methods.
- **Exporting `postgresState` from the `State` barrel.** Poisons worker bundles with `pg`; deep import + comment instead.
- **A migration-free schema bootstrap** (plain `create table if not exists`). Fails under concurrency on real Postgres; see above.


sam-goodwin:
Can you share sample of DX in PR description

wmadden:
@sam-goodwin I rewrote the PR description - does the example there suffice?

packages/alchemy/src/Prisma/Providers.ts
```diff
@@ -147,6 +149,43 @@ const stackManagementApiLayer = () =>
...
+export interface ProvidersOptions {
+  /**
+   * Replacement provider layer used in `alchemy dev` mode instead of the
+   * built-in dev stubs. This lets an embedding framework supply its own
+   * emulator implementations (e.g. a local Postgres or object-storage
+   * emulator) while keeping the live Management API providers untouched.
+   * The layer must register a provider for every Prisma resource. Live
+   * (non-dev) deployments are unaffected.
+   */
+  dev?: PrismaLocalProviders;
```
sam-goodwin: This isn't something we typically do. Take a look at `Provider.dual` that we set up in https://github.com/alchemy-run/alchemy/pull/963

The user shouldn't provide them. If they want to override them, they can construct a new layer.

wmadden: OK, on it

wmadden: Done in 5c2a64bb. 

All resources now register via `ProviderLayer.dual` with their local stubs moved into the module. Shared helpers are in `Internal/DevStub.ts`. 

Removed `providers({dev})`, `PrismaLocalProviders`, and `liveProviders()`. `providers()` is back to just the collection.

wmadden:
Hey @sam-goodwin, I added one more fix for an issue already present: changing an environment variable's value never reached a running deployment. Prisma Cloud snapshots env vars into the deployment at create, and `Deployment` only replaces when its own props (artifact/port/app) change, so a value-only change updated the platform row but the app kept serving the old value. 

I fixed this by adding an optional `redeployOn` prop to `Deployment`: pass it whatever inputs should force a redeploy (e.g. env values), and a change replaces the deployment. 

The resolved values are salted-hashed and stored `Redacted`, so no plaintext lands in state.

packages/alchemy/src/State/PostgresState.ts
```diff
@@ -0,0 +1,634 @@
...
+ * @example Connection string
+ * ```typescript
+ * import * as Alchemy from "alchemy";
+ * import { postgresState } from "alchemy/State/PostgresState";
+ *
+ * const Stack = Alchemy.Stack(
+ *   "my-stack",
+ *   {
+ *     providers: myProviders(),
+ *     state: postgresState({ dsn: process.env.STATE_DATABASE_URL! }),
```
sam-goodwin: Don't use `process.env`, instead of `Config.string("STATE_DATABASE_URL")` or `Config.redacted("STATE_DATABASE_URL")` if it's sensitive. `process.env` couples everything to node. All code should be written against `effect/Config` instead. Users can do whatever they want, but don't encourage the wrong way of accessing the environment.

packages/alchemy/src/State/PostgresState.ts
```diff
@@ -0,0 +1,634 @@
...
+    const migrate = (client: PostgresStateClient) =>
+      Effect.gen(function* () {
+        const conn = yield* attempt(() => client.connect());
+        const releaseConn = Effect.sync(() => {
+          try {
+            conn.release();
+          } catch {
+            // The pool may already be closed; nothing left to release.
+          }
+        });
```
sam-goodwin: Please use effect SQL instead of the async api

packages/alchemy/src/Prisma/BucketKey.ts
```diff
@@ -0,0 +1,304 @@
...
+  const effectiveSuffix =
+    instanceToken.length >= 12
+      ? instanceToken.slice(0, 12)
+      : fnv1a64(instanceId).slice(0, 12);
+  const maxPrefixLength = 65 - effectiveSuffix.length - 1;
+  return `${name.trim().slice(0, maxPrefixLength)}-${effectiveSuffix}`;
+};
+
+class AmbiguousPrismaBucketKeyError extends Error {
+  readonly _tag = "AmbiguousPrismaBucketKeyError";
```
sam-goodwin: Use `Data.TaggedError`?

packages/alchemy/src/Prisma/BucketKey.ts
```diff
@@ -0,0 +1,304 @@
...
+ *   env: {
+ *     S3_ENDPOINT: key.endpoint,
+ *     S3_BUCKET: key.bucketName,
+ *     S3_ACCESS_KEY_ID: key.accessKeyId,
+ *     S3_SECRET_ACCESS_KEY: key.secretAccessKey,
+ *   },
+ * });
+ * ```
+ */
+export const BucketKey = Resource<BucketKey>("Prisma.BucketKey");
```
sam-goodwin: Interesting. We'll also need to add bindings like we have in src/AWS/S3

Each API gets a binding and will automatically create the BucketKey resource and bind that to the environment.

For reference:
1. look in src/AWS/S3 and see how each operation (e.g. GetObject.ts) has a Binding.Service and a corresponding layer `GetObjectHttp.ts`.
2. look in src/Cloudflare/R2/ReadBucketHttp.ts - see how it creates an Account Api Token and binds granular policies to it.

Right now, your PR only considers the non-effectful DX. Every resource needs effectful bindings for interacting with it.

wmadden: I've added the bindings: ReadBucket / WriteBucket / ReadWriteBucket, following the R2 shape — binding a bucket auto-creates the BucketKey (read-scoped for reads; the API has no write-only role, so write bindings carry read_write), env round-trip follows Connect's naming contract, and the runtime client reuses the S3 machinery with an endpoint override. Get/put presign included.

Overall ~1.8k lines including tests. I debated extracting object storage to a separate PR — the bindings are the latest two commits, but I have a clean split prepared if you'd prefer it: [#1061](https://github.com/alchemy-run/alchemy/pull/1061) would then deliver the state backend, the Aliases fix, dual registration and redeployOn, with buckets + bindings as the follow-up.

Also merged main — the pg optional-peer change (#1069) is honored: PostgresState now lazy-loads the driver with the same install-hint the other drivers use.

packages/alchemy/test/Prisma/BucketBinding.test.ts
sam-goodwin, line 1: Don't stub. Deploy a real stack (in fixtures/) that deploys an actual app and then have the tests execute request with `HttpClient`

sam-goodwin: see how we use beforeAll and afterAll in other tests

packages/alchemy/src/State/PostgresState.ts
wmadden: addressed

packages/alchemy/test/Prisma/BucketBinding.test.ts
wmadden: done

wmadden:
you were right about the stubs. When I rewrote the tests against the real endpoints I found some errors that would only have surfaced when users ran it in production

I also realized I haven't been following the instructions in the repo or established conventions. I educated myself, gave my agent a kick in the pants and corrected my implementation and tests. there should be no novel decisions in my PR any more

BucketBinding.test.ts now follows the R2 Binding.test.ts shape: a fixtures/ stack deploying three Compute apps binding one shared bucket, beforeAll(deploy) / afterAll(destroy), every client method incl. presigned URLs driven over HttpClient. live runs are opt-in like Compute.live.test.ts; a cold create/exercise/destroy cycle runs green against real Prisma Cloud

and the description now shows the two-file export default shape from Compute's JSDoc

packages/alchemy/src/Prisma/BucketKey.ts
wmadden: done

packages/alchemy/src/Prisma/BucketAccessKey.ts
```diff
@@ -0,0 +1,309 @@
...
+   * identity so an interrupted create can be recovered by name instead of
+   * minting a second, unenumerable key.
+   *
+   * @default the resource's logical ID
+   */
+  name?: string;
+  /**
+   * Access role for the key: `"read"` or `"read_write"`.
+   */
+  role: BucketKeyRole;
```
sam-goodwin, line 47: Does Prisma's BucketKey accept an IAM Policy?

packages/alchemy/src/Prisma/BucketAccessKey.ts
```diff
@@ -0,0 +1,309 @@
...
+    bucketKeyId: string;
+    /**
+     * Bucket ID the key belongs to, persisted so deletion can address both
+     * path parameters.
+     */
+    bucketId: string;
+    /**
+     * S3 access key ID.
+     */
+    accessKeyId: string;
```
sam-goodwin, line 66: Should this also be redacted? Or is it public?

website/src/content/docs/prisma/data/buckets.mdx
```diff
@@ -0,0 +1,75 @@
...
+
+```typescript
+import * as Prisma from "alchemy/Prisma";
+
+const bucket = yield* Prisma.Bucket("uploads", {
+  project,
+  name: "uploads",
+});
+
+const key = yield* Prisma.BucketKey("uploads-key", {
```
sam-goodwin: BucketKey as a name confuses me because it makes me think of the S3 Key.

Why not call it `Prisma.AccessKey`? Or is it because it's specific to Bucket? Maybe `Prisma.BucketAccessKey`?

packages/alchemy/src/Prisma/ReadWriteBucket.ts
```diff
@@ -0,0 +1,101 @@
...
+} from "./WriteBucket.ts";
+
+/**
+ * Full-access client for a Prisma Object Store bucket: every operation of
+ * {@link ReadBucketClient} and {@link WriteBucketClient}.
+ */
+export interface ReadWriteBucketClient
+  extends ReadBucketClient, WriteBucketClient {}
+
+/**
```
sam-goodwin, line 27: nit: comment should go on export `const ReadWriteBucket = Binding.Service<ReadWriteBucket>(`

packages/alchemy/src/Prisma/BucketBinding.ts
```diff
@@ -0,0 +1,245 @@
...
+            bindings: bucketWorkerBindings(keys, key),
+          });
+        } else {
+          return yield* Effect.die(
+            new Error(
+              `Prisma.${options.capability}Bucket supports Prisma.Compute, AWS.Lambda.Function, and Cloudflare.Worker runtimes, got '${host?.Type ?? "no host"}'`,
+            ),
+          );
+        }
+      }
```
sam-goodwin: This looks wrong. You don't need to map to environment variables or bindings yourself.

See s3 example https://github.com/alchemy-run/alchemy/blob/main/packages/alchemy/src/AWS/S3/BindingHttp.ts#L55

```ts
const BucketName = yield* bucket.bucketName;
```

You just need to `yield* ` the outputs you need bound to the platform and then create a client by passing the Accessor (an Effect that yields the value) to the client construction. Don't mess around with plain_text vs secret_text. and each individual platform.

sam-goodwin (changes requested):

packages/alchemy/src/Prisma/BucketAccessKey.ts
sam-goodwin: I still think it should be named BucketAccessKey

packages/alchemy/src/Prisma/BucketBinding.ts
wmadden: Done — `makeBucketBinding` now yields the key's outputs and passes the accessors to `makeClient` (the `makeHttpBucketBinding` shape); the env-var naming, `plain_text`/`secret_text` mapping, and per-platform dispatch are all deleted. One provider fix fell out: Prisma Compute's runtime context now uppercases its storage keys, since the provider validates env keys to the uppercase POSIX shape — the deployed-host suite caught mixed-case keys being rejected and runs green after the fix.

packages/alchemy/src/Prisma/BucketAccessKey.ts
wmadden: Renamed to `BucketAccessKey` — resource type, provider, error, tests, docs. On the IAM question: no — a Prisma bucket key takes a single coarse role, `read` or `read_write`; there's no policy document and no write-only role, which is why `WriteBucket` mints a `read_write` credential and enforces write-only client-side (documented on the binding).

packages/alchemy/src/Prisma/BucketAccessKey.ts
wmadden: It's public in the AWS sense: the access key ID travels in cleartext in the `Authorization` header of every signed request, so it identifies but doesn't authenticate. Only `secretAccessKey` stays `Redacted` — the same split as an R2 token's `accountId` vs `value`.

website/src/content/docs/prisma/data/buckets.mdx
wmadden: Renamed across the docs — the buckets page now uses `Prisma.BucketAccessKey` and documents the three bindings with reference links.

packages/alchemy/src/Prisma/ReadWriteBucket.ts
wmadden: Moved onto the `export const` in all three binding modules.

wmadden:
All of the latest round is addressed, plus a sweep of the rest of the PR for the same problem classes.

Your comments:

- **BucketKey → BucketAccessKey.** Renamed everywhere — resource type, provider, error, tests, docs (faa1e8ef).
- **Does the key accept an IAM policy?** No. A Prisma bucket key takes a single coarse role, `read` or `read_write` — no policy document, and no write-only role. That's why `WriteBucket` mints a `read_write` credential and the write-only contract is enforced client-side (documented on the binding).
- **Should `accessKeyId` be redacted?** It's public in the AWS sense: the access key ID travels in cleartext in the `Authorization` header of every signed request, so it identifies but doesn't authenticate. Only `secretAccessKey` stays `Redacted` — the same split as an R2 token's `accountId` vs `value`.
- **The env mapping in `BucketBinding`.** You were right — all of it is gone. `makeBucketBinding` now yields the key's outputs and passes the accessors to `makeClient`, the same shape as `makeHttpBucketBinding`'s token flow; no env-var names, no `plain_text`/`secret_text`, no per-platform dispatch. One provider fix fell out: Prisma Compute's runtime context now uppercases its storage keys, because the provider validates env keys to the uppercase POSIX shape — the deployed-host suite caught mixed-case keys being rejected, and runs green after the fix.
- **Doc placement nit.** The `@binding` blocks now sit on the `export const` in all three binding modules.

Since the theme of this round was "invented instead of followed," I swept the PR for more of it:

- `presignGet` signed `contentType` as a request header — wrong for GET (browsers don't send `Content-Type`), and diverging from `PresignGetObjectHttp`, which maps it to `response-content-type`. Fixed, with per-direction option types.
- `Deployment.redeployOn` is now `Deployment.triggers`, taking the established name and shape from `AWS.ApiGateway.Deployment.triggers`. The Prisma-specific part stays: values here may be secrets, so they're compared as a salted fingerprint persisted `Redacted` instead of as plaintext props.
- Error tags now equal their class names, `ListResult` is a truncated-carries-cursor union like R2's `Objects`, the binding scaffolding is no longer exported from the barrel (the `BindingHttp` rule), a physical-name helper duplicated from `Connection` is shared, and batch deletes are bounded.
- `PostgresState` gained a live suite under the usual env guard — advisory-lock exclusivity, release on scope close, and the concurrent schema migration verified green against real Postgres 15, since the hermetic fake structurally can't prove that SQL.
- Docs: the buckets page now documents the three bindings with reference links, and the state-store page documents `postgresState`.

Verification on the tip: monorepo `tsc -b` clean, all suites green, the deployed-host bucket suite green against Prisma Cloud, `docs:check` clean.

packages/alchemy/src/Prisma/Client.ts
sam-goodwin, line 1: It just occured to me that we're not using distilled for prisma??

wmadden: I don't know what that means, was it directed at me?

sam-goodwin: Not something we should do as part of this PR, but you're supposed to generate a distilled SDK (https://github.com/alchemy-run/distilled) and use that instead of hand-rolling one here in alchemy.

PR #1062, opened 2026-08-03
fix(planetscale): split same-line statement-breakpoint markers

beeirl:
Drizzle-kit appends the `--> statement-breakpoint` marker on the same line for single-line statements:

```sql
CREATE TABLE `users` (
	`id` int NOT NULL
);
--> statement-breakpoint
CREATE INDEX `users_email` ON `users` (`email`);--> statement-breakpoint
CREATE INDEX `users_name` ON `users` (`name`);
```

`STATEMENT_BREAKPOINT` requires a newline before `-->`, so same-line markers survive the split and the merged statement reaches Vitess, which rejects the `-->` token with `syntax error at position 2`. This is the exact failure the split exists to prevent.

```diff
-const STATEMENT_BREAKPOINT = /\r?\n\s*-->\s*statement-breakpoint\s*\r?\n?/g;
+const STATEMENT_BREAKPOINT = /\s*-->\s*statement-breakpoint\s*/g;
```

The `\s*` on both sides still absorbs surrounding newlines and the trailing `;` stays with its statement.

### DIFF
--- a/packages/alchemy/test/Planetscale/MySQL/MySQLMigrations.test.ts
+++ b/packages/alchemy/test/Planetscale/MySQL/MySQLMigrations.test.ts
@@ -1,0 +1,50 @@
+import { splitMySQLStatements } from "@/Planetscale/MySQL/MySQLMigrations";
+import { describe, expect, test } from "alchemy-test";
+
+describe("MySQLMigrations", () => {
+  describe("splitMySQLStatements", () => {
+    test("splits statements separated by breakpoints on their own line", () => {
+      const sql = [
+        "CREATE TABLE `users` (\n\t`id` int NOT NULL\n);",
+        "--> statement-breakpoint",
+        "CREATE TABLE `sessions` (\n\t`id` int NOT NULL\n);",
+      ].join("\n");
+      expect(splitMySQLStatements(sql)).toEqual([
+        "CREATE TABLE `users` (\n\t`id` int NOT NULL\n);",
+        "CREATE TABLE `sessions` (\n\t`id` int NOT NULL\n);",
+      ]);
+    });
+
+    // drizzle-kit puts the marker on the same line after single-line
+    // statements (`...;--> statement-breakpoint`). A newline-anchored split
+    // misses those, and the leftover `-->` makes Vitess bail with
+    // "syntax error at position 2".
+    test("splits breakpoints appended on the same line as the statement", () => {
+      const sql = [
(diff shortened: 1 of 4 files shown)


sam-goodwin:
Please include a regression test

beeirl:
@sam-goodwin sure thing!

packages/alchemy/test/Planetscale/MySQL/MySQLMigrations.test.ts
```diff
@@ -0,0 +1,50 @@
+import { splitMySQLStatements } from "@/Planetscale/MySQL/MySQLMigrations";
+import { describe, expect, test } from "alchemy-test";
+
+describe("MySQLMigrations", () => {
```
sam-goodwin: Does it only apply to mysql or should we patch Postgres and other providers?

Should we maybe centralize it in src/Drizzle or src/SQL and reuse?

beeirl: Only mysql. The split exists because Vitess requires whitespace after `--`, so the `-->` token is a syntax error there. Postgres and sqlite treat `--` as a comment regardless of the next character, so the marker is inert and the Postgres/Neon paths already send the file whole ([PostgresMigrations.ts#L98](https://github.com/alchemy-run/alchemy/blob/74a47648b7cfa90fb2759f22164f3c909f1cb289/packages/alchemy/src/Planetscale/Postgres/PostgresMigrations.ts#L98), [Neon/Migrations.ts#L109](https://github.com/alchemy-run/alchemy/blob/74a47648b7cfa90fb2759f22164f3c909f1cb289/packages/alchemy/src/Neon/Migrations.ts#L109)).

Happy to move `splitMySQLStatements` into `src/SQL/SqlFile.ts` next to `listSqlFiles`/`readSqlFile` so it's reusable. [Drizzle/Schema.ts#L443](https://github.com/alchemy-run/alchemy/blob/74a47648b7cfa90fb2759f22164f3c909f1cb289/packages/alchemy/src/Drizzle/Schema.ts#L443) also writes the marker when joining generated statements, so the marker constant could be shared by both sides. Lmk which you'd prefer and I'll push it.

sam-goodwin: Yeah let's move it to src/SQL so it's easier to discover for other mysql databases we will support in the future

PR #1076, opened 2026-08-04
feat(cloudflare): run vite dev servers in child processes

### LINKED ISSUE #1014: `alchemy dev`: with multiple Website.Vite resources, every inner Vite dev server resolves to port 5173 — the inline `server: { port: 0 }` never yields a random port
## Summary

Under `alchemy dev`, every `Cloudflare.Website.Vite` resource starts its inner Vite dev server with an inline `server: { port: 0 }` (`runVite` in `src/Cloudflare/Workers/LocalWorkerProvider.ts`). Vite treats a falsy port as "not configured" and falls back to its default 5173 (`!configPort` in `startServer`), so with multiple `Website.Vite` resources every inner dev server targets 5173 — and the inline config outranks the app's `vite.config.ts`, so it can't be worked around from the app side.

Result: with default config both servers bind 5173 across IPv4/IPv6 and the proxy serves one app's HTML from the other's URL (silent); with `host: '127.0.0.1'` + `strictPort: true` the loser dies with `Error: Port 5173 is already in use`.

## Reproduction

https://github.com/filipfalcon/mre-alchemy-website-vite-dev-inner-servers-collide-port-5173-already-in-use

1. `bun install`
2. `bun alchemy dev`
3. The second website fails with `Error: Port 5173 is already in use`.

## Expected

Each inner Vite dev server gets its own port; each public URL serves its own app.

## Environment

alchemy 2.0.0-beta.66 · vite 8.0.7 · effect 4.0.0-beta.102 · Bun 1.4.0 · macOS 26.5 (arm64)

### LINKED ISSUE #1019: `alchemy dev`: multiple Website.Vite apps with a `main` worker serve one app's index.html on every URL — modules still route per-app
Under `alchemy dev`, a stack with two `Cloudflare.Website.Vite` resources that declare a custom `main` worker (a plain assets pass-through, `env.ASSETS.fetch(request)`) serves ONE app's `index.html` on BOTH apps' dev URLs, while module requests keep routing to each app's own Vite server. The result is a hybrid page — one app's HTML shell (title, banner markup, inline scripts) hydrated by the other app's JavaScript. Which app wins appears to depend on startup order and can flip between `alchemy dev` restarts.

Two assets-only `Website.Vite` resources (no `main`) do **not** reproduce — the crossing happens in the ASSETS-binding path of the shared local runtime.

**Minimal reproduction:** https://github.com/filipfalcon/mre-alchemy-dev-website-vite-cross-serves-index-html

1. `bun install`
2. `bun alchemy dev`
3. `curl http://localhost:6180/` (AppA's URL) → returns `<title>APP B</title>`
4. `curl http://localhost:6181/` (AppB's URL) → also `<title>APP B</title>`
5. `curl http://localhost:6180/src/only-a.ts` → returns AppA's own module (`ONLY-IN-APP-A`) — the module layer routes per-app while the HTML layer is crossed

**Expected:** each URL serves its own app's `index.html`.

**Actual:** both URLs serve the same app's HTML (see above; winner flips between restarts).

The repro pins each app's inner Vite dev server to a distinct port via a plugin `config` hook to isolate this from #1014 (the `server: { port: 0 }` → 5173 race).

**Environment:** alchemy 2.0.0-beta.66, effect 4.0.0-beta.102, vite 8.2.0, Bun 1.4.0, macOS (arm64). State backend `Cloudflare.state()`.

BlankParticle:
run vite server in child process to make sure everything that relies on `process.cwd()` gets correct cwd, also capture logs from processes and add `[worker] |` prefix to logs
Screenshot 2026-08-04 at 15-24-58

fixes #1014
fixes #1019

vite side bug fix for `server.port` https://github.com/vitejs/vite/pull/23158

### DIFF
--- a/packages/alchemy/src/Cloudflare/Workers/ViteChildRunner.ts
+++ b/packages/alchemy/src/Cloudflare/Workers/ViteChildRunner.ts
@@ -1,0 +1,103 @@
+import { layerRuntime } from "@distilled.cloud/cloudflare-runtime";
+import * as Effect from "effect/Effect";
+import * as Layer from "effect/Layer";
+import * as Stdio from "effect/Stdio";
+import * as Stream from "effect/Stream";
+import * as FetchHttpClient from "effect/unstable/http/FetchHttpClient";
+import * as NodeV8 from "node:v8";
+import { CloudflareAuth } from "../Auth/AuthProvider.ts";
+import * as Credentials from "../Credentials.ts";
+import * as RpcServerEnvironment from "../../Local/RpcServerEnvironment.ts";
+import { PlatformServices, runMain } from "../../Util/PlatformServices.ts";
+import { materializeRuntimeBindings } from "./RuntimeBindings.ts";
+import * as Vite from "./Vite.ts";
+import {
+  DEFAULT_DEV_PORT,
+  type ViteChildConfig,
+  VITE_CHILD_READY_PREFIX,
+  VITE_CHILD_READY_SUFFIX,
+} from "./ViteChild.shared.ts";
+
+const readConfig = Effect.gen(function* () {
+  const stdio = yield* Stdio.Stdio;
+  const chunks = yield* Stream.runCollect(stdio.stdin);
+  return NodeV8.deserialize(Buffer.concat(chunks)) as ViteChildConfig;
+});
+
+const program = Effect.scoped(
+  Effect.gen(function* () {
+    const config = yield* readConfig;
+    const credentials = Credentials.fromAuthProvider().pipe(
+      Layer.provide(CloudflareAuth),
+    );
+    const runtimeContext = yield* layerRuntime({
+      api: { accountId: config.accountId },
+      storage: { directory: config.storageDirectory },
+    }).pipe(
+      Layer.provide(Layer.mergeAll(credentials, FetchHttpClient.layer)),
+      Layer.build,
+    );
+    // The non-runtime fields are consumed here; everything else in
+    // `config.worker` is the runtime worker shape `viteDev` expects, so new
+    // runtime fields flow through without being re-listed.
+    const {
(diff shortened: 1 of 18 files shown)


packages/alchemy/src/Cloudflare/Workers/ViteChildRunner.ts
```diff
@@ -0,0 +1,103 @@
...
+      );
+    });
+    return yield* Effect.never;
+  }),
+);
+
+runMain(
+  program.pipe(
+    Effect.provide(RpcServerEnvironment.fromEnv()),
+    Effect.provide(PlatformServices),
```
sam-goodwin: Never use two susbsequent Effect.provide.

Always `Effect.provide(RpcServerEnvironment.fromEnv().pipe(Layer.provideMerge(PlatformService)))`

If RpcServerEnvironment doe not depend on PlatformService, then just do `Effect.provide([RpcServerEnvironment.fromEnv(), PlatformService])`

packages/alchemy/src/Cloudflare/Workers/ViteChildRunner.ts
```diff
@@ -0,0 +1,103 @@
...
+    );
+    const url = devServer.resolvedUrls?.local[0];
+    if (!url) {
+      return yield* Effect.die("Vite child started without a local URL");
+    }
+    yield* Effect.sync(() => {
+      process.stdout.write(
+        `${VITE_CHILD_READY_PREFIX}${url}${VITE_CHILD_READY_SUFFIX}\n`,
+      );
+    });
```
sam-goodwin: Could this be:
```ts
yield* Console.log(`${VITE_CHILD_READY_PREFIX}${url}${VITE_CHILD_READY_SUFFIX}`)
```

PR #1077, opened 2026-08-04
feat(aws-lambda): add first-class container image functions

### LINKED ISSUE #657: feat(AWS Lambda): support container image functions
## Summary

Add support for deploying AWS Lambda functions from container images, alongside the existing ZIP-based packaging flow.

AWS Lambda supports OCI-compatible container images, which are useful when a function needs native system dependencies, larger artifacts, custom runtimes, or an existing Docker-based build pipeline.

## Proposed API

Allow an AWS Lambda function to select an image package and configure either:

- an existing ECR image URI (ideally including digest support), or
- a local Docker build context/Dockerfile that Alchemy builds, publishes to ECR, and deploys.

For example:

```ts
class Worker extends AWS.Lambda.Function<Worker>()("Worker", {
  image: {
    uri: "123456789012.dkr.ecr.us-east-1.amazonaws.com/worker@sha256:...",
  },
}) {}
```

A local-build API could look roughly like:

```ts
class Worker extends AWS.Lambda.Function<Worker>()("Worker", {
  image: {
    context: ".",
    dockerfile: "Dockerfile",
  },
}) {}
```

The exact API shape is open for discussion.

## Expected behavior

- Configure the Lambda package type as `Image`.
- Accept an ECR image URI, including tags and immutable digests.
- Support Lambda image settings such as command, entry point, and working directory where applicable.
- Treat ZIP-specific options such as handler and runtime as incompatible with image functions and report a clear configuration error.
- Detect image changes and update the function on deployment.
- Preserve the current ZIP-based behavior as the default.
- If local builds are supported, manage or integrate with ECR authentication, image publishing, architecture selection, and deployment dependencies.

## Motivation

Container-image support would make it possible to use Alchemy for Lambda workloads that depend on native binaries, custom runtimes, or artifacts that do not fit comfortably into the ZIP packaging workflow. It would also make migration easier for teams that already package Lambda functions as Docker images.

## Acceptance criteria

- A Lambda function can be deployed from an existing ECR image URI.
- Updating the image reference updates the deployed function.
- Image configuration supports Lambda's image-specific command/entry-point settings.
- Invalid combinations of image and ZIP/runtime options fail with an actionable error.
- The feature is covered by tests and documented with an example.

I’d be happy to help refine the API or test an implementation.

westerlund:
Closes #657.

`AWS.Lambda.Function` can now deploy container images, alongside zip functions.

Build straight from your Dockerfile — Alchemy builds the image, pushes it to a managed private ECR repository, and wires up the function:

```typescript
const api = yield* AWS.Lambda.Function("ContainerApi", {
  image: {
    context: "./lambda",
    dockerfile: "Dockerfile",
  },
  architecture: "arm64",
  functionUrl: true,
});
```

Or deploy an image that already lives in ECR, by tag or immutable digest:

```typescript
const worker = yield* AWS.Lambda.Function("Worker", {
  image: {
    uri: "123456789012.dkr.ecr.eu-west-1.amazonaws.com/worker:production",
  },
  architecture: "x86_64",
});
```

Override the image's entrypoint without rebuilding it:

```typescript
const processor = yield* AWS.Lambda.Function("Processor", {
  image: {
    uri: "123456789012.dkr.ecr.eu-west-1.amazonaws.com/processor@sha256:...",
    command: ["app.handler"],
    entryPoint: ["/lambda-entrypoint.sh"],
  },
  architecture: "x86_64",
});
```

Everything else about `Function` works the same for image functions — bindings, env, Function URLs, VPC, event sources. Props are typed as zip XOR image, so mixing `main` with `image` is a compile error rather than a deploy failure.

### Behavior worth knowing

- Deploys are content-addressed: the context, Dockerfile, `.dockerignore`, build args, and architecture identify the image, so an unchanged app deploys without rebuilding. Repointing an ECR tag out-of-band is detected by digest on the next plan and rolled forward.
- Switching a function between zip and image replaces it (`PackageType` is create-time-only on Lambda).
- Context builds push to a function-owned immutable ECR repository that is deleted with the function; external repositories referenced by `image.uri` are never modified or deleted.

------------------------------------------------------------------

For transparency, GPT-5.6 Sol, with Kit Langton's Effect skill, wrote all of the code. I've reviewed the code along the way and before creating the PR. I have made several real tests to my own AWS account including deploying, destroying, changing architecture, updates, etc. I'm currently running Lambdas in a stage environment deployed using this feature.


sam-goodwin:
I just found myself needing this. Will take a look today and try and merge before next release

PR #1078, opened 2026-08-04
feat(cloudflare/workers): add Worker.lookup data source for external Workers

filipfalcon:
Reference a Worker deployed outside the stack (Wrangler, the dashboard, another pipeline) and bind it as a `service` binding, without provisioning or taking ownership of it.

    env: { AUTH: Cloudflare.Worker.lookup({ name: "auth-service" }) }

Follows the VpcService.lookup shape: an Output that reads the live account during plan/deploy and resolves to the target's attributes branded with the resource Type, so binding classification treats it like the managed resource. Existence is probed via getScriptScriptAndVersionSetting, so a name typo fails the deploy rather than producing a broken binding.

Exported from Workers/index.ts as types only — that module is re-exported flat from Cloudflare/index.ts, so a bare `lookup` const would surface as `Cloudflare.lookup`.

### DIFF
--- a/packages/alchemy/src/Cloudflare/Workers/InferEnv.ts
+++ b/packages/alchemy/src/Cloudflare/Workers/InferEnv.ts
@@ -44,6 +44,7 @@ import type { VersionMetadataBinding } from "./VersionMetadataBinding.ts";
 import type { Worker } from "./Worker.ts";
 import type { WorkerEntrypointBinding } from "./WorkerEntrypoint.ts";
 import type { WorkerLoader as WorkerLoaderResource } from "./WorkerLoader.ts";
+import type { WorkerLookup } from "./WorkerLookup.ts";
 
 export type InferEnv<W> =
   W extends Effect.Effect<infer A, infer _E, infer _R>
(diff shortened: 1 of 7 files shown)


sam-goodwin:
My only push back here is that I don't want to implement these over and over for each resource manually and instead it should be something we implement in the `Provider`. A new lifecycle method `lookup` so that all resources can define how they are looked up and implemented.

Similar to `Resource.ref`, we want `Resource.lookup`. But the problem is that `Resource.ref` has a stable interface, but lookup is resource-specific (e.g. bucketArn vs vpcId vs ...). So we'd need the `Resource` type to declare its lookup interface.

filipfalcon:
> My only push back here is that I don't want to implement these over and over for each resource manually and instead it should be something we implement in the `Provider`. A new lifecycle method `lookup` so that all resources can define how they are looked up and implemented.
> 
> 
> 
> Similar to `Resource.ref`, we want `Resource.lookup`. But the problem is that `Resource.ref` has a stable interface, but lookup is resource-specific (e.g. bucketArn vs vpcId vs ...). So we'd need the `Resource` type to declare its lookup interface.

I think the lookup interface declaration is a good way to go. Do you want me to spin up the clanker on this or would you rather think this through more?

PR #1091, opened 2026-08-04
fix(bundle): externalize the workerd virtual module

aryasaatvik:
Keep workerd's virtual module external when bundling Lambda functions whose dependency graph reaches Cloudflare runtime helpers outside workerd.

```ts
if (moduleId === "cloudflare:workers") return true;
```

The Lambda bundler applies this by default, and the regression exercises the Lambda plan without a caller-provided workaround.

The packaged dev executable does not need the workerd virtual module. It now imports the leaf `alchemy/Cli/exec` entry instead of traversing the full CLI barrel and Cloudflare graph, so its tsdown configuration needs no special externalization.

The isolated-install resolution architecture is handled by #1324; this PR retains only the independent runtime-module externalization.

### DIFF
--- a/packages/alchemy/tsdown.config.ts
+++ b/packages/alchemy/tsdown.config.ts
@@ -23,6 +23,10 @@ export default [
   // the cli bundle below).
   defineConfig({
     entry: ["bin/exec.ts"],
+    // Supplied by workerd at runtime; it has no filesystem existence, and
+    // rolldown resolves before tree-shaking, so the guarded import trips
+    // [UNRESOLVED_IMPORT] even though it is dead code outside workerd.
+    external: ["cloudflare:workers"],
     format: ["esm"],
     clean: false,
     shims: true,
--- a/packages/alchemy/tsdown.config.ts
+++ b/packages/alchemy/tsdown.config.ts
@@ -23,10 +23,6 @@ export default [
   // the cli bundle below).
   defineConfig({
     entry: ["bin/exec.ts"],
-    // Supplied by workerd at runtime; it has no filesystem existence, and
-    // rolldown resolves before tree-shaking, so the guarded import trips
-    // [UNRESOLVED_IMPORT] even though it is dead code outside workerd.
-    external: ["cloudflare:workers"],
     format: ["esm"],
     clean: false,
     shims: true,
--- a/packages/alchemy/bin/exec.ts
+++ b/packages/alchemy/bin/exec.ts
@@ -1,4 +1,4 @@
-import { exec } from "alchemy/Cli";
+import { exec } from "alchemy/Cli/exec";
 import { runMain } from "alchemy/Util/PlatformServices";
 
 exec().pipe(runMain);
--- a/packages/alchemy/src/AWS/Lambda/FunctionBundle.ts
+++ b/packages/alchemy/src/AWS/Lambda/FunctionBundle.ts
@@ -168,6 +168,11 @@ export const makeFunctionBundler = Effect.gen(function* () {
       parentId: string | undefined,
       isResolved: boolean,
     ): boolean => {
+      // Alchemy's cross-provider runtime helpers contain a guarded dynamic
+      // import of workerd's virtual module. A Lambda can never resolve it;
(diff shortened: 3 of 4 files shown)


packages/alchemy/src/AWS/Lambda/Function.ts
```diff
@@ -1180,6 +1180,11 @@ export const FunctionProvider = () =>
           parentId: string | undefined,
           isResolved: boolean,
         ): boolean => {
+          // Alchemy's cross-provider runtime helpers contain a guarded
+          // dynamic import of workerd's virtual module. A Lambda can never
+          // resolve it; keep the import external so the runtime fallback
+          // handles it instead of tripping [UNRESOLVED_IMPORT] before DCE.
+          if (moduleId === "cloudflare:workers") return true;
```
sam-goodwin, line 1187: This makes sense

packages/alchemy/src/Bundle/Bundle.ts
```diff
@@ -325,6 +326,35 @@ export const watch = (
 const ENTRY_PREFIX = "\0virtual:alchemy-entry:";
 // oxlint-disable-next-line no-control-regex
 const ENTRY_REGEX = /^\0virtual:alchemy-entry:/;
+const ALCHEMY_PRIVATE_DEPENDENCY_REGEX = /^@distilled\.cloud\//;
```
sam-goodwin: These are not all private. All the distilled SDKs are in there. What motivated this?

aryasaatvik: Fair — "private" is the wrong word; they're published. The intent is provenance, not privacy: resolve `@distilled.cloud/*` from alchemy's own install rather than the consumer's package graph. 

What motivated it: I consume alchemy in another monorepo from a packed tarball built off a worktree carrying my open PRs plus distilled patches that aren't on npm yet. (`bun link` isn't workable there — linking pulls alchemy's whole source tree into the dev server's watch graph, >10k files, and the process hits the macOS open-file limit.) The patched SDK builds ride inside alchemy's install, so bundled user code has to resolve `@distilled.cloud/*` to those copies — the consumer graph either doesn't have them or hoists the unpatched registry versions.

It's not specific to that workflow though: imports reached from the virtual entry have no importer path, so without the plugin resolution falls back to the consumer graph and only works when the package manager happens to hoist — and the versions it hoists aren't necessarily the ones alchemy's install was built against.

packages/alchemy/src/Bundle/Bundle.ts
```diff
@@ -325,6 +326,37 @@ export const watch = (
...
+    filter: { id: ALCHEMY_SDK_DEPENDENCY_REGEX },
+    handler(source) {
+      try {
+        return fileURLToPath(import.meta.resolve(source));
+      } catch {
+        return null;
+      }
+    },
+  },
+});
```
sam-goodwin, line 359: What is the purpose behind this?

packages/alchemy/tsdown.config.ts
```diff
@@ -23,6 +23,10 @@ export default [
   // the cli bundle below).
   defineConfig({
     entry: ["bin/exec.ts"],
+    // Supplied by workerd at runtime; it has no filesystem existence, and
+    // rolldown resolves before tree-shaking, so the guarded import trips
+    // [UNRESOLVED_IMPORT] even though it is dead code outside workerd.
+    external: ["cloudflare:workers"],
```
sam-goodwin: Why does the alchemy cli need this?

aryasaatvik: You’re right—the CLI doesn’t need this. The warning was [UNRESOLVED_IMPORT] Could not resolve "cloudflare:workers" because bin/exec.ts imported the broad alchemy/Cli barrel, causing Rolldown to traverse unrelated Cloudflare modules before tree-shaking. The bundle still completed and treated the import as external. Importing alchemy/Cli/exec directly removes the warning without this configuration. After rechecking the remaining Lambda change against #1324, I don’t think this PR retains a justified behavior, so I’m going to close it as superseded.

PR #1092, opened 2026-08-04
fix(cloudflare/workers): resolve zone inference with exact name lookups

aryasaatvik:
Resolve Worker custom-domain and route zones with account-scoped exact-name lookups instead of scanning only the first page of an account's zones.

```text
api.staging.example.com
  -> zones?account.id=...&name=api.staging.example.com
  -> zones?account.id=...&name=staging.example.com
  -> zones?account.id=...&name=example.com                 found
```

- Walk parent labels until an exact zone is found.
- Share cached in-flight inference lookups within a reconcile pass.
- Keep explicitly pinned zone resolution separate so a hostname can safely target different pinned zones.


packages/alchemy/src/Cloudflare/Workers/WorkerProvider.ts
```diff
@@ -1024,6 +1023,41 @@ const resolveWorkerMetadataHash = ({
...
+      Effect.catch(() =>
+        Effect.die(
+          `Could not infer Cloudflare Zone for hostname "${hostname}". ` +
+            "Ensure the parent zone exists in this account.",
+        ),
+      ),
+    );
+    zoneCache.set(hostname, zoneId);
+    return zoneId;
+  });
```
sam-goodwin: Does something like this exist already in Zone.ts? I also think it probably belongs there instead of WorkerProvider?

aryasaatvik: Yes — the walk itself is `resolveZoneId` in `Zone/lookup.ts`, and this PR routes through it; what was left here was just the per-run cache + env resolution + the die message. Moved that wrapper into `Zone/lookup.ts` next to `resolveZoneId` and moved the tests with it.

packages/alchemy/src/Cloudflare/Zone/lookup.ts
```diff
@@ -46,6 +47,45 @@ export const resolveZoneId = ({
...
+      hostname,
+    }).pipe(
+      Effect.catch(() =>
+        Effect.die(
+          `Could not infer Cloudflare Zone for hostname "${hostname}". ` +
+            "Ensure the parent zone exists in this account.",
+        ),
+      ),
+    );
+    zoneCache.set(hostname, zoneId);
```
sam-goodwin: This is not atomic. Should probably store an Effect in the Map and set it to `zoneCache.set(hostname, yield* Effect.cached(resolve...)`

That way it's only ever computed once.

aryasaatvik: Done. the map now stores `Effect.cached` lookups built synchronously inside `Effect.suspend` (the `Action.ts` shape) — get-and-set is atomic and the lookup runs once per hostname. Added a concurrency test pinning exactly one fetch for N concurrent resolutions.

packages/alchemy/src/Cloudflare/Zone/lookup.ts
```diff
@@ -46,6 +47,67 @@ export const resolveZoneId = ({
...
+                Effect.die(
+                  `Could not infer Cloudflare Zone for hostname "${hostname}". ` +
+                    "Ensure the parent zone exists in this account.",
+                ),
+              ),
+            );
+          }),
+        ),
+      );
+      zoneCache.set(hostname, lookup);
```
sam-goodwin, line 100: Like this:

```ts
if (!zoneCache.has(hostname)) {
  const eff = yield* Effect.cached(..)
  zoneCache.set(eff)
}
return yield* zoneCache.get(eff)
```

aryasaatvik: simplified

aryasaatvik:
Closing as superseded by #1241, which moved Worker domain zone inference onto `resolveZoneId`'s exact `?name=` lookups and added zone pinning on `Worker.domain`.

I'll open a follow-up with:

- `resolveZoneId` failing with a typed `ZoneNotFound` instead of `new Error(...)`
- one memo for pinned and inferred lookups via the existing `Util/cached-function.ts` (atomic, keeps the typed error channel)

PR #798, opened 2026-07-10
docs(cloudflare): cloudflare foldkit example

filipfalcon:
Adds a minimal [Foldkit](https://foldkit.dev) example: an Elm-style counter SPA (single `src/main.ts`) deployed with `Cloudflare.Website.Vite`.

`effect` is pinned to `4.0.0-beta.88` — Foldkit's exact peer — instead of the catalog; noted in the README.

PR #1120, opened 2026-08-06
feat(cloudflare/website): Foldkit resource

filipfalcon:
Foldkit apps are client-only Vite projects with no SSR or server runtime, so `Website.Foldkit` is `Website.Vite` with Foldkit's defaults applied rather than a new source provider.

- `assets.notFoundHandling` defaults to `single-page-application`, so deep links reach the client router with no config; an explicit `assets` merges over the default instead of replacing it
- `viteEnvironments` dropped (Foldkit has no RSC split); `script`/`bundle`/`source` rejected at the type level
- `main` kept, for deployments that carry a Worker entry in front of the assets (API routes, error reporting, Durable Objects)


filipfalcon:
Follow up on https://github.com/alchemy-run/alchemy/pull/798 and https://github.com/alchemy-run/alchemy/pull/983. First-party support would be neat.

packages/alchemy/src/Cloudflare/Website/Foldkit.ts
```diff
@@ -0,0 +1,298 @@
...
+ * //   fetch(request: Request, env: { TICKER: KVNamespace; ASSETS: Fetcher }) {
+ * //     const url = new URL(request.url);
+ * //     if (url.pathname === "/api/ticker") {
+ * //       return env.TICKER.get("ticker:clubs").then((body) =>
+ * //         new Response(body, { headers: { "content-type": "application/json" } }),
+ * //       );
+ * //     }
+ * //     return env.ASSETS.fetch(request);
+ * //   },
+ * // };
```
sam-goodwin: WHy are these comments?

sam-goodwin: Why is this commented out

filipfalcon: I made Claude to take an exact template from `Cloudflare.Website.Waku`. It should now be consistent. About custom entrypoint I personally rely on it for wiring a Foldkit app Worker with Sentry.

PR #1124, opened 2026-08-06
fix(deps)!: upgrade to effect 4.0.0-beta.104

Mkassabov:
The peer range `>=4.0.0-beta.102 || >=4.0.0` admitted effect beta.104, which renamed `Schema.TaggedErrorClass` to `Schema.TaggedError`. A fresh install could therefore resolve to a version where the CLI died at import:

```
TypeError: Schema.TaggedErrorClass is not a function
    at alchemy/src/Auth/AuthProvider.ts:43:39
```

This migrates to beta.104 and raises the floor so the range can no longer resolve to an incompatible version.

### API changes

beta.104 removed four things we used:

```diff
-export class AuthError extends Schema.TaggedErrorClass<AuthError>()(
+export class AuthError extends Schema.TaggedError<AuthError>()(

-  Command.withHidden,
+  Command.unlisted,
```

`HttpPlatform` gained two required members, so the worker stubs now declare them (matching effect's own web layer, and what workerd's `CompressionStream` supports):

```diff
 const HttpPlatformStub = Layer.succeed(HttpPlatform.HttpPlatform, {
+  platform: "web",
+  compression: HttpPlatform.makeCompressionWeb({
+    algorithms: ["gzip", "deflate"],
+    transform: (algorithm) => HttpPlatform.compressionTransformWeb(algorithm),
+  }),
   fileResponse: () => Effect.die("HttpPlatform.fileResponse not supported"),
```

And `Sse.decode` now surfaces `SseError`, which was escaping the declared error channel:

```diff
 Stream.catchTag("Retry", (retry) => Stream.die(retry)),
+Stream.catchTag("SseError", (cause) =>
+  Stream.fail(toAiError(cause, "streamText")),
+),
```

### Also

`@effect/sql-mysql2` and `@effect/sql-sqlite-do` are added to `minimumReleaseAgeExcludes` — `effect` and the other `@effect/sql-*` packages were already listed, so their absence was an oversight.

Submodules are pinned to the merged migrations: distilled alchemy-run/distilled#431 (v1.0.0-rc.3), cloudflare-tools alchemy-run/cloudflare-tools#111.

### Breaking

`effect >=4.0.0-beta.104` is now the minimum peer. Consumers on beta.102/103 must upgrade in lockstep.


sam-goodwin (approved):

sam-goodwin:
Closing in favor of #1132 — we accidentally reproduced this exact migration independently (down to the same `SseError` mapping), but on top of the cloudflare-tools submodule-to-monorepo move (#1122), so the runtime/frameworks fixes now land in `packages/cloudflare-runtime` / `packages/cloudflare-frameworks` directly rather than via a submodule pin, which this branch predates.

#1132 also goes a bit further: effect `beta.105` instead of `beta.104`, distilled pinned one commit later (picks up alchemy-run/distilled#430), and a root `overrides` forcing a single `effect` instance — foldkit exact-pins older betas as peers, which otherwise splits the install into two incompatible effect copies on a fresh resolve.

The `Command.unlisted` catch and the `minimumReleaseAgeExcludes` additions from this PR were carried over verbatim — thank you @Mkassabov!

PR #1127, opened 2026-08-06
feat(celld, rivet): Workers and Durable Objects for CF/Rivet/Celld

sam-goodwin:
Add provider-owned Celld and Rivet Workers and Durable Objects, with Effect-native implementations, typed callers, and AWS ECS hosting. Celld v0.5.0 deployment uses object storage and private management APIs—not Wrangler or `celld deploy`. The generated Celld SDK is in [Distilled #621](https://github.com/alchemy-run/distilled/pull/621).

Each provider owns its public runtime contracts. Shared internals handle construction, request scopes, context propagation, telemetry, and dispatch; Cloudflare retains its native APIs. The examples below enumerate Celld first, then Rivet. Handler fragments run inside a Worker request or Durable Object method; binding setup runs during initialization.

The main merge carries durable callbacks, schema-based `RpcDurableObject`, RPC over hibernatable WebSockets, validated application attachments, and Durable Object SQL migrations into all three providers. Callback dispatch, RPC transport and recovery metadata, browser clients, and migration execution share implementations; adapters retain native persistence and lifecycle behavior. Workflows remain available on Cloudflare and Celld, not Rivet. RivetKit 2.3.10 blocks automatic idle sleep while a raw WebSocket remains open; the adapter supports recovery after explicit native sleep without forcing retirement. Corresponding website guides follow the same structure while documenting those differences.

## Celld: fleets, Workers, and Applications

A Fleet owns infrastructure. A Worker stages an immutable bundle. An Application publishes and activates the complete Worker graph. Resources and Workers select their fleet through `Celld.Fleet.layer`, rather than a fleet argument on every declaration.

```ts
// api.ts
import * as Celld from "alchemy/Celld";

export class Api extends Celld.Worker<Api>()("Api") {}
```

```ts
// api.live.ts
import * as HttpServerResponse from "effect/unstable/http/HttpServerResponse";
import { Api } from "./api.ts";

export default Api.make(
  { main: import.meta.url, expose: "public" },
  Effect.succeed({
    fetch: Effect.succeed(HttpServerResponse.text("Hello from Celld")),
  }),
);
```

```ts
// alchemy.run.ts
import * as AWS from "alchemy/AWS";
import * as Celld from "alchemy/Celld";
import * as Layer from "effect/Layer";
import { Api } from "./api.ts";
import ApiLive from "./api.live.ts";

export class Cells extends Celld.Fleet<Cells>()("Cells", {
  instances: { min: 2, max: 4, targetCpu: 60 },
  cpu: 512,
  memory: 1024,
}) {}

export const App = Celld.Application("App", { entrypoint: Api }).pipe(
  Effect.provide(
    ApiLive.pipe(Layer.provideMerge(Celld.Fleet.layer(Cells))),
  ),
);

export default Alchemy.Stack(
  "celld-app",
  {
    providers: Layer.mergeAll(
      AWS.providers(),
      Celld.providers(),
      Celld.EcsFleet(),
    ),
    state: AWS.state(),
  },
  Effect.gen(function* () {
    const app = yield* App;
    return { url: app.url, revision: app.revision };
  }),
);
```

`Celld.EcsFleet()` defaults to Fargate. The host provisions backing storage, networking, compute, and private management. Celld embeds V8; this is not a workerd host or a complete Cloudflare API implementation.

## Celld: public ingress, custom domains, and network placement

Workers are private by default. Only the Application entrypoint may request public ingress. `domain` implies public exposure and uses the selected DNS implementation for certificate validation and records.

```ts

export default Api.make(
  { main: import.meta.url, domain: "api.example.com" },
  Effect.succeed({
    fetch: Effect.succeed(HttpServerResponse.text("Hello over HTTPS")),
  }).pipe(Effect.provide(Cloudflare.CloudflareDns())),
);
```

Use `AWS.Route53Dns()` instead when Route 53 manages DNS, and include the selected DNS provider's resource providers in the stack. Supply `vpc` to use an existing network; otherwise the host composes one.

```ts
export class Cells extends Celld.Fleet<Cells>()("Cells", {
  instances: 2,
  cpuArchitecture: "X86_64",
  vpc: {
    vpcId: "vpc-0123",
    subnetIds: ["subnet-a", "subnet-b"],
    securityGroupIds: ["sg-callers"],
  },
}) {}
```

Public ingress requires public subnets in at least two Availability Zones. Administration remains private through an IAM-invoked VPC Lambda; public Worker ingress does not expose the management API.

## Celld: Worker configuration and native binding interop

Effect capabilities register bindings automatically. Worker props also expose explicit environment values, native binding declarations, compatibility settings, and bundler overrides for integrations that need them.

```ts
const workerProps = {
  main: import.meta.url,
  env: { APP_NAME: "reports" },
  celldVersion: "0.5.0",
  compatibilityDate: "2026-09-01",
  compatibilityFlags: ["nodejs_compat"],
  bindings: [{ type: "worker_loader", name: "CUSTOM_LOADER" }],
} satisfies Celld.CelldWorkerProps;
```

Binding names must be unique across native capabilities, Durable Objects, assets, and environment variables. Version checks target the pinned Celld runtime; compatibility flags do not imply support for every Cloudflare feature.

## Celld: Durable Objects and typed RPC

Separate the declaration from its implementation. Yielding the declaration in Worker initialization registers the namespace and returns typed stubs.

```ts
import type { RuntimeContext } from "alchemy";

export interface CounterShape {
  increment(): Effect.Effect<number, never, RuntimeContext>;
}

export class Counter extends Celld.DurableObject<Counter, CounterShape>()(
  "Counter",
) {}

export const CounterLive = Counter.make(
  Effect.gen(function* () {
    const state = yield* Celld.DurableObjectState;
    return Effect.succeed({
      increment: () => state.storage.transaction((tx) =>
        Effect.gen(function* () {
          const next = ((yield* tx.get<number>("count")) ?? 0) + 1;
          yield* tx.put("count", next);
          return next;
        }),
      ),
    });
  }),
);
```

```ts
export default Api.make(
  { main: import.meta.url, expose: "public" },
  Effect.gen(function* () {
    const counters = yield* Counter;
    return {
        const count = yield* counters.getByName("lobby").increment();
        return yield* HttpServerResponse.json({ count });
      }),
    };
  }).pipe(Effect.provide(CounterLive)),
);
```

The implementation returns an Effect that builds each instance. Runtime methods declare `RuntimeContext`; requests receive their own scope. Declare tagged RPC failures with the Durable Object's `{ errors: [MyError] }` option to reconstruct them on callers.

## Celld: Durable Object storage, SQLite, alarms, and WebSockets

`Celld.DurableObjectState` exposes identity, KV, transactions, SQLite, alarms, `waitUntil`, concurrency blocking, and native WebSocket handling.

```ts
// Inside a Durable Object method; state was captured at initialization.
yield* state.storage.put("profile", { name: "Sam" });
const profile = yield* state.storage.get<{ name: string }>("profile");
const entries = yield* state.storage.list({ prefix: "profile" });

yield* state.storage.sql.exec("CREATE TABLE IF NOT EXISTS notes (body TEXT)");
yield* state.storage.sql.exec("INSERT INTO notes VALUES (?)", "hello");
const cursor = yield* state.storage.sql.exec<{ body: string }>(
  "SELECT body FROM notes",
);
const notes = yield* cursor.toArray();

const now = yield* Effect.sync(() => Date.now());
yield* state.storage.setAlarm(now + 60_000);
const alarmAt = yield* state.storage.getAlarm();
yield* state.storage.deleteAlarm();
```

Add lifecycle handlers to the object returned by the implementation:

```ts
return Effect.succeed({
  alarm: () => state.storage.put("alarm-fired", true),
    const [response, socket] = yield* Celld.upgrade();
    yield* Effect.sync(() => socket.serializeAttachment({ room: "lobby" }));
    return response;
  }),
  webSocketMessage: (socket: Celld.WebSocket, message: string | ArrayBuffer) =>
    socket.send(typeof message === "string" ? message : new Uint8Array(message)),
});
```

Forward a Worker's request to `rooms.getByName("lobby").fetch(request)` to reach the object's HTTP/WebSocket handler. Connected sockets and tags are available through `state.getWebSockets()` and `state.getTags()`. Celld's storage surface excludes Cloudflare point-in-time recovery bookmarks.

## Celld: native storage declarations and binding layers

Declare storage once and resolve it inside the selected Fleet scope. Declarations have no per-resource fleet prop.

```ts
export const Cache = Celld.KV.Namespace("Cache");
export const Database = Celld.D1.Database("Database", {
  migrations: "./migrations",
});
export const Files = Celld.R2.Bucket("Files");
export const JobsQueue = Celld.Queues.Queue("JobsQueue");
export const FailedQueue = Celld.Queues.Queue("FailedQueue");
```

Provide capability implementations on the Worker initialization effect. Compose dependencies into one layer and use one `Effect.provide` per pipe.

```ts
const StorageBindings = Layer.mergeAll(
  Celld.KV.ReadWriteNamespaceBinding,
  Celld.D1.QueryDatabaseBinding,
  Celld.R2.ReadWriteBucketBinding,
);

const storageWorker = Effect.gen(function* () {
  const cache = yield* Celld.KV.ReadWriteNamespace(yield* Cache);
  const db = yield* Celld.D1.QueryDatabase(yield* Database);
  const files = yield* Celld.R2.ReadWriteBucket(yield* Files);
  return {
      const cached = yield* cache.get("greeting");
      const tables = yield* db.prepare(
        "SELECT name FROM sqlite_master WHERE type = ?",
      ).bind("table").all();
      const objects = yield* files.list();
      return yield* HttpServerResponse.json({ cached, tables, objects });
    }).pipe(Effect.orDie),
  };
}).pipe(Effect.provide(StorageBindings));
```

## Celld: KV reads, writes, metadata, and expiration

Choose `ReadNamespace`, `WriteNamespace`, or `ReadWriteNamespace`, with the corresponding `*Binding` layer.

```ts
// Worker initialization:
const cache = yield* Celld.KV.ReadWriteNamespace(yield* Cache);

// Inside a handler:
yield* cache.put("profile", '{"name":"Sam"}', {
  metadata: { version: 1 },
  expirationTtl: 120,
});
const profile = yield* cache.get<{ name: string }>("profile", "json");
const withMetadata = yield* cache.getWithMetadata("profile", "json");
const page = yield* cache.list({ prefix: "profile", limit: 100 });
yield* cache.delete("profile");
```

These access levels restrict the TypeScript client surface; they are not separate native authorization grants.

## Celld: D1 queries, batches, and deploy-time migrations

`prepare` and `bind` build statements synchronously. `all`, `first`, `run`, and `raw` execute as Effects. `batch` runs statements transactionally, and `raw` on the database client exposes the native binding for library interop.

```ts
// Worker initialization:
const db = yield* Celld.D1.QueryDatabase(yield* Database);

// Inside a handler:
const inserted = yield* db.prepare(
  "INSERT INTO users (id, name) VALUES (?, ?)",
).bind("user-1", "Sam").run();

const user = yield* db.prepare(
  "SELECT id, name FROM users WHERE id = ?",
).bind("user-1").first<{ id: string; name: string }>();

const batch = yield* db.batch([
  db.prepare("UPDATE users SET name = ? WHERE id = ?").bind("Samuel", "user-1"),
  db.prepare("SELECT id, name FROM users WHERE id = ?").bind("user-1"),
]);
```

```sql
-- migrations/0001_users.sql
CREATE TABLE users (id TEXT PRIMARY KEY, name TEXT NOT NULL);
```

`migrations` accepts the shared SQL migration input. Deployment applies migrations through the private native operator with a transaction, lock, and hash ledger. Changed applied migration content is rejected. The host bootstraps operators before application publication, and the root application keeps them available even without corresponding Worker bindings.

## Celld: R2 objects, streaming, conditions, and multipart uploads

Choose `ReadBucket`, `WriteBucket`, or `ReadWriteBucket`, with the matching binding layer. A bucket is an isolated keyspace in the Fleet's object store, not another AWS S3 bucket.

```ts
// Worker initialization:
const files = yield* Celld.R2.ReadWriteBucket(yield* Files);

// Inside a handler:
yield* files.put("hello.txt", "Hello", {
  httpMetadata: { contentType: "text/plain" },
  customMetadata: { author: "Sam" },
});
const object = yield* files.get("hello.txt");
const text = object ? yield* object.text() : null;
const metadata = yield* files.head("hello.txt");
const page = yield* files.list({ prefix: "uploads/", limit: 100 });

if (metadata) {
  const replaced = yield* files.put("hello.txt", "Updated", {
    onlyIf: { etagMatches: metadata.etag },
  });
}
yield* files.delete("hello.txt");
```

Writes also accept Effect byte streams. Multipart clients expose creation, part upload, completion, abort, and resume. Resume is node-local; restarting or switching nodes can lose the upload. Native part ETags are empty, and completion uses part numbers. Conditional writes never silently discard their preconditions. Customer-key encryption and R2 event notifications are outside this surface.

## Celld: queue producers, push consumers, retries, and dead letters

```ts
// Inside producer Worker initialization:
const producer = yield* Celld.Queues.WriteQueue(yield* JobsQueue);

// Inside its request handler:
yield* producer.send({ id: "job-1" });
yield* producer.sendBatch([
  { body: { id: "job-2" } },
  { body: { id: "job-3" } },
]);
```

Provide `Celld.Queues.WriteQueueBinding` on the producer. Register consumption on a background Worker with `Celld.Queues.EventSourceLive`:

```ts
import * as Stream from "effect/Stream";

export class Jobs extends Celld.Worker<Jobs>()("Jobs") {}

export const JobsLive = Jobs.make(
  { main: import.meta.url },
  Effect.gen(function* () {
    yield* Celld.Queues.consumeQueueMessages<{ id: string }>(
      yield* JobsQueue,
      {
        batchSize: 10,
        maxWaitTime: "1 second",
        maxRetries: 3,
        retryDelay: "5 seconds",
        deadLetterQueue: yield* FailedQueue,
      },
      (messages) => Stream.runForEach(messages, (message) =>
        Effect.log({ id: message.body.id, attempt: message.attempts }),
      ),
    );
    return {};
  }).pipe(Effect.orDie, Effect.provide(Celld.Queues.EventSourceLive)),
);

export default JobsLive;
```

Successful handlers acknowledge unsettled messages; failures retry them. Explicit `message.ack()` and `message.retry({ delaySeconds })` remain available via `Effect.sync`. Delivery is at least once; handlers must complete settlement before returning. There is no pull-consumer API.

Include the consumer in the published graph:

```ts
const app = yield* Celld.Application("App", {
  entrypoint: Api,
  workers: [Jobs],
}).pipe(
  Effect.provide(
    Layer.mergeAll(ApiLive, JobsLive).pipe(
      Layer.provideMerge(Celld.Fleet.layer(Cells)),
    ),
  ),
);
```

## Celld: durable Workflows and instance controls

Define a Workflow during Worker initialization. Its initialization Effect captures dependencies and returns the durable run function.

```ts
const reports = yield* Celld.Workflow(
  "Reports",
  Effect.succeed((input: { title: string }) =>
    Effect.gen(function* () {
      const title = yield* Celld.Workflows.task(
        "prepare",
        Effect.succeed(input.title.toUpperCase()),
        { retries: { limit: 3, delay: "1 second", backoff: "linear" } },
      );
      yield* Celld.Workflows.sleep("cooldown", "10 seconds");
      const approval = yield* Celld.Workflows.waitForEvent<{ approved: boolean }>(
        "approval",
        { type: "approval", timeout: "1 day" },
      );
      return { title, approved: approval.payload.approved };
    }),
  ),
);
```

```ts
// Inside request handlers:
const instance = yield* reports.create({
  id: "daily-report",
  params: { title: "Daily" },
});
yield* instance.sendEvent({ type: "approval", payload: { approved: true } });
const status = yield* instance.status();

// Alternative lifecycle operations:
const existing = yield* reports.get("daily-report");
yield* existing.pause();
yield* existing.resume();
yield* existing.restart();
yield* existing.terminate();
yield* existing.delete();
```

The API also provides reusable `class Reports extends Celld.Workflow<Reports>()(...)`, `createBatch`, `deleteBatch`, `sleepUntil`, and `WorkflowStepContext` for attempt information. Names are scoped to the hosting script. Native Workflow schedules, rollback, and sensitive outputs are not supported.

## Celld: cron handlers

```ts
const implementation = Effect.gen(function* () {
  yield* Celld.cron("0 * * * *", (event) =>
    Effect.log({ cron: event.cron, scheduledTime: event.scheduledTime }),
  );
  return {};
}).pipe(Effect.provide(Celld.CronEventSourceLive));
```

Cron schedules belong to the Application's entrypoint; schedules on secondary Workers are rejected. Publication activation confirms graph adoption, not a future cron invocation. Celld v0.5.0 cannot safely retire a persisted cron cell when its root loses all schedules or changes script identity. Alchemy rejects those transitions before publication instead of stranding native wakeups. Same-script updates between nonempty schedule lists remain supported.

## Celld: static assets and SPA routing

```ts
import { HttpServerRequest } from "effect/unstable/http/HttpServerRequest";

export default Api.make(
  {
    main: import.meta.url,
    expose: "public",
    assets: {
      directory: "public",
      binding: "ASSETS",
      notFoundHandling: "single-page-application",
      runWorkerFirst: true,
    },
  },
  Effect.gen(function* () {
    const assets = yield* Celld.Assets();
    return {
        const request = yield* HttpServerRequest;
        if (request.url.startsWith("/api/health")) {
          return yield* HttpServerResponse.json({ ok: true });
        }
        return yield* assets.fetch(request);
      }),
    };
  }).pipe(Effect.provide(Celld.AssetsBinding)),
);
```

The directory is relative to the Worker entry module. Asset publication includes `_headers` and `_redirects`; `htmlHandling`, missing-file handling, and Worker-first route selection are configurable.

## Celld: same-fleet service bindings

```ts
// Api initialization; Service is a separately declared Celld.Worker.
const implementation = Effect.gen(function* () {
  const service = yield* Celld.Fetch(yield* Service, { bindingName: "SERVICE" });
  return {
      const request = yield* HttpServerRequest;
      return yield* service(request);
    }),
  };
}).pipe(Effect.provide(Celld.FetchBinding));
```

Include `Service` in `Application.workers`. The native binding is fetch-only and accepts Effect HTTP client requests or forwarded server requests. It preserves the original request origin; it is distinct from the cross-host RPC gateway.

## Celld: calling an activated Application from Lambda

`Celld.bindWorker(App, Api)` requires an activated Application and its entrypoint Worker, rather than a merely staged Worker. It attaches supported callers to the private fleet network and supplies the gateway credential.

```ts
export default class Backend extends AWS.Lambda.Function<Backend>()(
  "Backend",
  { main: import.meta.url },
  Effect.gen(function* () {
    const api = yield* Celld.bindWorker(App, Api);
    const counters = api.durableObject<CounterShape>("Counter");
    return {
        const count = yield* counters.getByName("lobby").increment();
        return yield* HttpServerResponse.json({ count });
      }),
    };
  }),
) {}
```

The client also exposes authenticated `api.fetch(HttpClientRequest.get("/hello"))` and typed Worker RPC methods. Failed RPC calls are not automatically replayed: a failed response can follow a successful mutation.

## Celld: dynamic Worker loading and entrypoints

Register `Celld.WorkerLoader()` at initialization, then load code inside a request. Anonymous Workers are disposed with that request's scope.

```ts
import * as HttpClientRequest from "effect/unstable/http/HttpClientRequest";

// Worker initialization:
const loader = yield* Celld.WorkerLoader("LOADER");

// Inside a handler:
const loaded = yield* loader.load({
  compatibilityDate: "2026-09-01",
  mainModule: "main.js",
  modules: {
    "main.js": `
      import { WorkerEntrypoint } from "cloudflare:workers";
      export class Calculator extends WorkerEntrypoint {
        add(a, b) { return a + b + this.ctx.props.offset; }
      }
      export default { fetch() { return new Response("loaded"); } };
    `,
  },
  globalOutbound: null,
});
const entrypoint = yield* loaded.getEntrypoint();
const response = yield* entrypoint.fetch(HttpClientRequest.get("https://loaded.test/"));
const calculator = yield* loaded.getEntrypoint<{
  add(a: number, b: number): number;
}>("Calculator", { props: { offset: 3 } });
const sum = yield* calculator.add(2, 4);
```

`loader.get(name, getCode)` memoizes named code; the callback is lazy. Select an entrypoint before fetching—the loaded handle is not itself a Fetcher. JavaScript and WebAssembly modules are supported. RPC is direct, single-method invocation, without pipelining or introspection. Handles remain request-local; native disposal is not an eviction-completion acknowledgement.

## Celld: Durable Object facets and loopback capabilities

Loaded Durable Object classes can back named child facets. `state` is the current parent Durable Object's `Celld.DurableObjectState`; `loader` is initialized on its hosting Worker/object.

```ts
// Inside a parent Durable Object method:
const loaded = yield* loader.load({
  compatibilityDate: "2026-09-01",
  mainModule: "facet.js",
  modules: {
    "facet.js": `
      import { DurableObject } from "cloudflare:workers";
      export class Child extends DurableObject {
        fetch() { return new Response("child"); }
      }
      export default { fetch() { return new Response("root"); } };
    `,
  },
  globalOutbound: null,
});
const childClass = yield* loaded.getDurableObjectClass("Child", {
  props: { tenant: "acme" },
});
const child = yield* state.facets.get("child", () => ({ class: childClass }));
const response = yield* child.fetch(HttpClientRequest.get("https://facet.test/"));
yield* state.facets.abort("child");
```

`state.facets.delete(name)` schedules storage deletion; neither abort nor delete is a completion acknowledgement. Nesting is limited to four levels including the root. `state.exports` exposes declared native loopback capabilities, and `state.props` exposes structured-clone properties. This is Celld's supported subset, not the complete workerd exports interface.

## Celld: EC2, Containers, and sandbox execution

Containers require dedicated Docker hosts. Select EC2 capacity and install runsc through the host layer:

```ts
providers: Layer.mergeAll(
  AWS.providers(),
  Celld.providers(),
  Celld.EcsFleet({
    capacity: { type: "ec2", instanceType: "m7i.large" },
    containerRuntime: "runsc",
  }),
)
```

Use `cpuArchitecture: "X86_64"` on the Fleet for `m7i.large`. The host configures Docker, gVisor/runsc, the network fence, scratch storage, and ECS registration. Fargate rejects Container/Sandbox requirements.

A Container declaration attaches an image to a same-script SQLite Durable Object. It is not an independently provisioned cloud container or a separate `Celld.Sandbox` API.

```ts
class Tool extends Celld.Container<Tool>()("Tool", {
  image: "node:22-alpine",
  instanceType: "dev",
  maxInstances: 4,
  ociRuntime: "runsc",
}) {}

const startup: Celld.Containers.ContainerStartupOptions = {
  entrypoint: [
    "node", "-e",
    "require('http').createServer((req,res)=>res.end('ready')).listen(3000,'0.0.0.0')",
  ],
  enableInternet: false,
  env: { MODE: "sandbox" },
};

// In the owning Durable Object's outer initialization effect:
const init = Effect.gen(function* () {
  const tool = yield* Tool;
  return Effect.succeed({
    run: () => Effect.gen(function* () {
      yield* tool.getTcpPort(3000);
      const process = yield* tool.exec(["node", "-e", "console.log('hello')"]);
      return yield* process.output();
    }),
  });
}).pipe(Effect.provide(Celld.Containers.layer(Tool, startup)));
```

`getTcpPort` starts lazily and waits for HTTP readiness. Await readiness before the first `exec`; native startup acknowledgement is earlier than Docker readiness, and exec is never replayed automatically. `output()` returns stdout/stderr bytes and an exit code; nonzero exit codes must be handled explicitly. Process handles and TCP sockets belong to the calling request's scope.

The client also exposes `running`, `start`, `monitor`, `signal`, `destroy`, `setInactivityTimeout`, and port `fetch`/plain-TCP `connect`. Startup `entrypoint` replaces the image CMD, preserving its ENTRYPOINT. `image` accepts a registry image or Dockerfile path; publication requires platform-aware Docker inspect/save and exports only the Fleet architecture.

## Celld: generated Bun/Node containers and typed RPC

Language runtime and OCI isolation are separate choices. Keep the declaration outside the process implementation module.

```ts
// tool.ts
import type { RuntimeContext } from "alchemy";

export class Tool extends Celld.Container<Tool, {
  add(a: number, b: number): Effect.Effect<number, never, RuntimeContext>;
}>()("Tool", { runtime: "bun", ociRuntime: "runsc" }) {}
```

```ts
// tool.runtime.ts
import { Tool } from "./tool.ts";

export default Tool.make(
  { main: import.meta.url },
  Effect.succeed({
    add: (a, b) => Effect.succeed(a + b),
    fetch: Effect.succeed(HttpServerResponse.text("ready")),
  }),
);
```

Provide this module's default-exported layer on the Stack, and `Celld.Containers.layer(Tool)` inside the owning Durable Object. `yield* Tool` returns both native container controls and typed RPC methods:

```ts
// Capture at object initialization:
const tool = yield* Tool;

// Call inside an object method:
const sum = yield* tool.add(2, 4);
```

`runtime: "node"` selects Node instead of Bun. Generated programs listen on port 3000 and initialize inside the container process, not at deployment. Native control names such as `exec` are reserved; application RPC methods use distinct names.

## Celld: persistence, publication, and runtime boundaries

- Removing storage or Application declarations retains persistent data and ownership records. Removal is not permission to erase data or transfer ownership; destructive cleanup requires a separate explicit action.
- Application publication uses ordered object writes, conditional ownership checks, a nonexpiring lock, and a recovery journal. It is not a multi-object atomic transaction. Activation checks the complete selected graph on live nodes before returning its revision.
- An Application lists every secondary service and consumer Worker explicitly. Only its entrypoint is served through the root listener.
- Container writable disks are ephemeral across destroy, node restart, and object migration. Image/runtime changes require explicit retirement of cached per-cell container specifications; stopping an isolate alone is insufficient.
- Container limits are approximate, disk quotas are not enforced, and snapshots, hard execution timeouts, and outbound interception are unavailable. Portless images have no native startup-readiness signal. The default OCI runtime is not a VM isolation boundary; use the configured `runsc` runtime where required.
- Binding clients may be initialized once, but operations, streams, sockets, and process handles run within their invoking request. Native access-level views and compatibility flags do not promise full Cloudflare behavior.

## Rivet: clusters and private ECS hosting

A Cluster hosts Rivet Engine. A Worker is an outbound-connected actor runner, not an HTTP Worker server. `Rivet.EcsCluster()` runs the engine and runners on ECS Fargate.

```ts
// cluster.ts
import * as Rivet from "alchemy/Rivet";

export class Actors extends Rivet.Cluster<Actors>()("Actors", {
  instances: 1,
  cpu: 1024,
  memory: 2048,
}) {}

export class ActorWorker extends Rivet.Worker<ActorWorker>()("ActorWorker") {}
```

```ts
// alchemy.run.ts
import * as AWS from "alchemy/AWS";
import * as Rivet from "alchemy/Rivet";
import * as Layer from "effect/Layer";
import { ActorWorker } from "./cluster.ts";
import ActorWorkerLive from "./worker.ts";

export default Alchemy.Stack(
  "rivet-app",
  {
    providers: Layer.mergeAll(
      AWS.providers(),
      Rivet.providers(),
      Rivet.EcsCluster(),
    ),
    state: AWS.state(),
  },
  Effect.gen(function* () {
    const worker = yield* ActorWorker;
    return { endpoint: worker.endpoint };
  }).pipe(Effect.provide(ActorWorkerLive)),
);
```

The engine and actor gateway stay on the private network. The runner has no inbound ports. Supplying Worker `expose` or `domain` is rejected, including `expose: "private"`; leave those props unset.

## Rivet: storage backends, scaling, and existing networks

The default single-node engine uses RocksDB on ephemeral task disk. Supply PostgreSQL for persistence across engine replacement and multi-node hosting.

```ts
export class Actors extends Rivet.Cluster<Actors>()("Actors", {
  postgresUrl: process.env.RIVET_POSTGRES_URL!,
  instances: 2,
  cpuArchitecture: "X86_64",
  vpc: {
    vpcId: "vpc-0123",
    subnetIds: ["subnet-a", "subnet-b"],
    securityGroupIds: ["sg-engine"],
  },
}) {}
```

`instances > 1` without PostgreSQL is rejected. `rivetVersion`, `image`, CPU, memory, architecture, tags, and `publicUrl` configure the engine; `publicUrl` is an advertised origin, not a request to publish the actor gateway. Database provisioning and retention remain the chosen database provider's responsibility.

## Rivet: provider-owned Durable Objects and implementation layers

Rivet has its own actor identity, KV, SQLite, alarms, and socket contracts. The declaration/implementation pattern matches Celld without pretending the runtime semantics are identical.

```ts
// document.ts
import * as Rivet from "alchemy/Rivet";
import type { RuntimeContext } from "alchemy";

export interface DocumentShape {
  put(text: string): Effect.Effect<void, never, RuntimeContext>;
  get(): Effect.Effect<string | undefined, never, RuntimeContext>;
}

export class Document extends Rivet.DurableObject<Document, DocumentShape>()(
  "Document",
) {}

export const DocumentLive = Document.make(
  Effect.gen(function* () {
    const state = yield* Rivet.DurableObjectState;
    return Effect.succeed({
      put: (text) => state.storage.put("text", text),
      get: () => state.storage.get<string>("text"),
    });
  }),
);
```

```ts
// worker.ts
import { Actors, ActorWorker } from "./cluster.ts";
import { Document, DocumentLive } from "./document.ts";

export default ActorWorker.make(
  {
    cluster: Actors,
    main: import.meta.url,
    desiredCount: 2,
    cpu: 512,
    memory: 1024,
  },
  Effect.gen(function* () {
    yield* Document;
    return {};
  }).pipe(Effect.provide(DocumentLive)),
);
```

Yield each actor declaration to register it; merge multiple implementation layers with `Layer.mergeAll`. Runner `image`, `rivetkitVersion`, architecture, and build options are configurable. Deployments advance the runner version so Rivet can drain older actor versions.

## Rivet: typed actor calls from Lambda

`Rivet.bindWorker` connects supported callers to the private cluster network. Use a public application endpoint, such as Lambda, for application authentication and authorization before invoking actors.

```ts
import * as Duration from "effect/Duration";
import * as HttpServerResponse from "effect/unstable/http/HttpServerResponse";
import { ActorWorker } from "./cluster.ts";
import type { DocumentShape } from "./document.ts";

export default class Backend extends AWS.Lambda.Function<Backend>()(
  "Backend",
  { main: import.meta.url, timeout: Duration.seconds(30) },
  Effect.gen(function* () {
    const actors = yield* Rivet.bindWorker(ActorWorker);
    const documents = actors.durableObject<DocumentShape>("Document");
    return {
        const document = documents.getByName("welcome");
        yield* document.put("Hello from Rivet");
        return yield* HttpServerResponse.json({ text: yield* document.get() });
      }),
    };
  }),
) {}
```

Include the Lambda in the stack and provide `ActorWorkerLive` so its binding resolves the Worker. Actor naming is stable through `getByName`. The engine admin token authenticates management/runners, not application actor calls; private network reachability is the gateway boundary.

## Rivet: actor identity, KV, SQLite, and concurrency

```ts
// Capture during actor initialization:
const state = yield* Rivet.DurableObjectState;
const identity = { id: state.actorId, key: state.key, name: state.name };

// Inside an actor method:
yield* state.storage.put({ title: "Draft", version: 1 });
const title = yield* state.storage.get<string>("title");
const entries = yield* state.storage.list({ prefix: "title", limit: 10 });
yield* state.storage.delete("version");

yield* state.storage.sql.exec("CREATE TABLE IF NOT EXISTS notes (body TEXT)");
yield* state.storage.sql.exec("INSERT INTO notes VALUES (?)", "hello");
const notes = yield* state.storage.sql.exec<{ body: string }>(
  "SELECT body FROM notes",
);
```

Rivet SQL returns materialized row arrays, not Celld's SQL cursor. KV follows Rivet's native actor-state persistence policy. Concurrent actions do not acquire Cloudflare input/output gates: a multi-step read/modify/write operation needs application-level coordination. Cloudflare storage transactions, bookmarks, synchronous KV, and `blockConcurrencyWhile` are not offered by this contract.

## Rivet: alarms and invocation-scoped background work

```ts
// In the object returned by Document.make's inner Effect:
return Effect.succeed({
  arm: (delayMs: number) => Effect.gen(function* () {
    const now = yield* Effect.sync(() => Date.now());
    yield* state.storage.setAlarm(now + delayMs);
  }),
  cancel: () => state.storage.deleteAlarm(),
  alarm: () => state.storage.put("alarm-fired", true),
  notify: () => state.waitUntil(Effect.log("background work")),
});
```

Add callable methods to the declaration's shape when exposing them over RPC. `getAlarm()` returns the current timestamp. Setting an alarm replaces/cancels the previous native event; generation checks reject stale deliveries. `waitUntil` uses the invoking action's context, and `state.raw` resolves that context as a runtime Effect rather than retaining an old invocation.

## Rivet: WebSocket lifecycle and native interop

Rivet delivers already-accepted sockets through its native hibernation lifecycle. Implement lifecycle handlers on the actor; no Celld-style upgrade or Cloudflare `WebSocketPair` is needed. Direct gateway clients negotiate `["rivet", "rivet_encoding.bare"]`; browser access needs an authenticated, WebSocket-capable application ingress.

```ts
return Effect.succeed({
  webSocketMessage: (socket: Rivet.WebSocket, message: string | ArrayBuffer) =>
    Effect.gen(function* () {
      yield* state.setWebSocketTags(socket, ["room:lobby"]);
      yield* socket.send(message);
    }),
  webSocketClose: (
    _socket: Rivet.WebSocket,
    code: number,
    reason: string,
    wasClean: boolean,
  ) => Effect.log({ code, reason, wasClean }),
  webSocketError: (_socket: Rivet.WebSocket, error: unknown) => Effect.log(error),
});
```

`state.getWebSockets(tag?)` enumerates sockets registered in the current activation. Rivet restores retained sockets lazily through `webSocketOpen`, not during the object constructor. Tags and schema-validated application attachments persist in native connection state. Sockets expose `send`, `close`, `readyState`, `setAttachment(schema, value)`, `getAttachment(schema)`, and their native `ws`. Automatic Effect RPC uses separate protected recovery metadata: idle sockets can survive native sleep, while unfinished requests are reset rather than replayed. RivetKit 2.3.10 keeps actors with open raw sockets active, so automatic idle hibernation remains unavailable on that version; explicit native sleep and connection recovery are distinct from that policy.

## Rivet: RPC failures, streams, and capability boundaries

```ts
import * as Data from "effect/Data";
import * as Stream from "effect/Stream";

class Rejected extends Data.TaggedError("Rejected")<{ reason: string }> {}

interface JobShape {
  fail(): Effect.Effect<never, Rejected, Alchemy.RuntimeContext>;
  values(): Stream.Stream<number, never, Alchemy.RuntimeContext>;
}

class Job extends Rivet.DurableObject<Job, JobShape>()("Job", {
  errors: [Rejected],
}) {}

const JobLive = Job.make(Effect.succeed(Effect.succeed({
  fail: () => Effect.fail(new Rejected({ reason: "invalid job" })),
  values: () => Stream.range(1, 3),
})));
```

Tagged failures are reconstructed by typed callers. Ordinary actor actions collect finite Stream results before returning them; these actions do not provide incremental streaming, and infinite streams cannot return a response. `Rivet.RpcDurableObject` uses the shared Effect RPC WebSocket protocol for incremental streams, interruption, and scoped connection ownership. `Alchemy.makeCallback` uses native persistent jobs and wakeups; actor KV, SQL transactions, and scheduling remain separate persistence boundaries.

Rivet Workers host actors rather than `fetch`, queue, cron, or Workflow handlers. This PR does not add Celld KV/D1/R2, WorkerLoader/facets, or Container APIs to Rivet. Persistence, concurrency, sockets, and transport retain Rivet-specific behavior rather than claiming Cloudflare compatibility.


K-Mistele:
any updates on this @sam-goodwin or anything I can do to help?

sam-goodwin:
> any updates on this @sam-goodwin or anything I can do to help? 

Been keeping it updated with main. Need to just find right time to commit to it. It impacts cloudflare workers and durable objects because i refactored the runtime out into code shared by rivet and celld. 

I'll try and get it done this week.

jperezrealini:
would it make sense to create abstract `Alchemy.DurableObject`, `Alchemy.Worker`, `Alchemy.Queue`, etc. and provide the underlying provider via layers?

sam-goodwin:
> would it make sense to create abstract `Alchemy.DurableObject`, `Alchemy.Worker`, `Alchemy.Queue`, etc. and provide the underlying provider via layers?

I did have that in this PR earlier but backed out since it's already a heavy lift. I think my priority right now is laying the physical groundwork per cloud which I can then package into Layers to implement `Alchemy.DurableObject` - but after building a dedicated interface per provider (Cloudflare vs Celld vs RIvet) so that we uncover their unique differences. Try to abstract their implementation to avoid duplicating code and learn from that. Once it's working, derive an abstract `Alchemy.*` model that is directly portable to any cloud.

jperezrealini:
Makes perfect sense. No premature abstraction.

PR #1140, opened 2026-08-08
feat(aws/website): web frameworks on AWS

sam-goodwin:
AWS hosting for all six web frameworks, mirroring the Cloudflare architecture: dual local/live providers, each framework's own HMR dev server under `alchemy dev`, and S3 + CloudFront + streaming-Lambda deployments. Every framework is live-verified end to end (SSR direct from the Lambda URL and through CloudFront, API routes, static assets and prerendered/SSG pages from S3, clean destroy) and dev-verified through the RPC sidecar.

All composites share the same shape: on deploy, the framework server runs on a Lambda Function URL with response streaming, assets and prerendered pages live in a private S3 bucket, and a CloudFront Function routes each request at the edge via a KeyValueStore file manifest. Under `alchemy dev` the framework's own dev server runs instead (native HMR, zero cloud resources); `Alchemy.remote()` opts back into the live deployment.

### Next.js

The full OpenNext serverless topology: streaming server Lambda, arm64 image-optimization Lambda behind `/_next/image` (via a new `metadata.image` edge-routing seam), SQS FIFO revalidation queue + consumer Lambda + event source mapping, DynamoDB tag cache, and a dedicated ISR cache bucket, with IAM wired through the binding channel.

```ts
const site = yield* AWS.Website.Nextjs("Web", {
  domain: { name: "app.example.com", hostedZoneId: zone.hostedZoneId },
});
```

### Nuxt

nitro's `aws-lambda` preset with response streaming — your `nuxt.config.ts` loads natively, prerendered route rules land in S3.

```ts
const site = yield* AWS.Website.Nuxt("Web", {
  server: {
    memorySize: 2048,
    environment: { NUXT_PUBLIC_API_BASE: api.url },
  },
});
```

### Astro

The adapter is provided by the integration (nothing to add to `astro.config.mjs`). `output: "static"` deploys assets-only with no Lambda.

```ts
const site = yield* AWS.Website.Astro("Web");

const docs = yield* AWS.Website.Astro("Docs", {
  rootDir: "./docs",
  astro: { output: "static" },
});
```

### SvelteKit

In-memory adapter — no `svelte.config.js` adapter to install. Prerendered routes serve from S3 via the edge router's `.html` fallback; server routes stream from the Lambda.

```ts
const site = yield* AWS.Website.SvelteKit("Web");
```

### Waku

RSC server on the streaming Lambda (fork of upstream's aws-lambda adapter: Hono + rscMiddleware); SSG pages serve from S3.

```ts
const site = yield* AWS.Website.Waku("Web");
```

### Octane

A zero-dependency marker adapter over octane's native node build — the emitted fetch handler gets a `toLambdaHandler` entry written beside it.

```ts
// octane.config.ts
import { aws } from "@alchemy.run/web-frameworks/octane/aws-adapter";
export default defineConfig({ adapter: aws() });
```

```ts
const site = yield* AWS.Website.Octane("Web");
```

### Dual `AWS.Website.Server` (the Cloudflare-architecture mirror)

`ProviderLayer.dual`: live runs the framework build with the `./aws` deploy target; local (`LocalProvider.make`, sidecar-owned, config-hash restarts) runs the framework kit's `dev()`. Composites branch on `AlchemyContext.dev` + `ProviderModePolicy` — dev declares only the local Server, deploy/`remote()` declares the cloud fan-out. `server.environment` reaches both modes: Lambda env vars on deploy, the dev server's process env under `alchemy dev` (a changed value restarts the dev server) — asserted end-to-end in the Nuxt local + live tests.

### `@alchemy.run/web-frameworks` (renamed from `cloudflare-frameworks`)

Platform-neutral integrations with per-platform subpath targets: `./{framework}/aws` deploy targets and `./aws-lambda` (Function URL payload-v2 ↔ web `Request`/`Response`, streaming via `awslambda.streamifyResponse` + buffered). BREAKING: projects depend on `@alchemy.run/web-frameworks` now; `cloudflare-runtime`/`cloudflare-test-tools` keep their names. Astro and Waku gained the same `make`/`layer` split the other frameworks had (their unit suites stay green).

### `Lambda.Function` `bundle: false`

```ts
yield* Lambda.Function("Server", {
  main: ".output/server/index.mjs",
  handler: "handler",
  bundle: false, // ship the directory as-is (entry + chunks + node_modules)
});
```

Framework outputs are already self-contained deployment units; re-bundling orphaned CJS exports-map subpaths (vue's `require("entities/decode")` 500'd every SSR request). The Handler string derives from `main`'s basename.

### `AWS.Website.StaticSite` correctness (was broken standalone on main)

- OAC + S3 bucket policy bound in standalone and Router modes (every request 403'd without it)
- Real default origin (S3 + OAC static, server host for SSR) so CloudFront's custom-error-page fetches resolve
- `spa: true` (miss → index.html 200) vs `errorPage` (real 404), mutually exclusive
- KV file manifest derived from `AssetDeployment`'s uploaded file list at deploy time (was a plan-time `readdirSync` walk, empty for built sites)
- Router: `distributionArn` output, bucket policies for inline bucket routes, and SSR-safe caching — the shared behavior now uses a defaultTTL-0 cache policy (+ `AllViewerExceptHostHeader`) instead of managed CachingOptimized, which cached header-less SSR responses for a day. A gated live test runs Nuxt through a shared Router: SSR via the router's distribution (first live proof of the placeholder→Lambda edge origin switch), distinct API responses, static + prerendered from S3.

### Docs

AWS frontend guides for all six frameworks plus StaticSite spa/errorPage semantics under `website/src/content/docs/aws/frontend/`.

### Follow-ups (documented, not blocking)

- **ISR revalidation e2e test** — the ISR plumbing (queue, consumer, tag table) deploys and is verified to exist, but no test yet proves a `revalidate: N` page actually re-renders after expiry. Deploy + wait + poll, so it would run as a gated slow test.
- **Tag-cache seeding** — OpenNext emits initial DynamoDB tag-cache data that upstream inserts by invoking a provided function once at deploy. We create the table but skip the seed; fresh apps work, apps calling `revalidateTag()` on prerendered pages need it. Requires an invoke-once-per-deploy action.
- **Symlink-safe `bundle: false`** — the zip walker flattens symlinks, which breaks bun/pnpm store layouts inside shipped `node_modules` (the `@swc/helpers` incident). Our integrations avoid it; a user pointing `bundle: false` at their own symlinked tree could still hit it.


packages/alchemy/src/AWS/Website/Astro.ts
```diff
@@ -0,0 +1,308 @@
...
+     */
+    architecture?: "x86_64" | "arm64";
+  };
+  /**
+   * Static asset upload configuration.
+   */
+  assets?: StaticSiteAssetsProps & {
+    fileOptions?: AssetFileOption[];
+  };
+  /**
```
sam-goodwin: Declare an interface instead of inline intersection

packages/alchemy/src/AWS/Website/Nextjs.ts
```diff
@@ -0,0 +1,474 @@
...
+ *     memorySize: 2048,
+ *     environment: {
+ *       API_BASE: api.url,
+ *     },
+ *   },
+ * });
+ * ```
+ */
+export const Nextjs = (id: string, props: NextjsProps = {}) =>
+  Effect.gen(function* () {
```
sam-goodwin: Use Effect.fn instead of (..) => Effect.gen . Make sure you do that everywhere in this PR too

packages/alchemy/src/AWS/Website/Nextjs.ts
```diff
@@ -0,0 +1,474 @@
...
+
+    // Image optimization: OpenNext installs sharp's linux-arm64 binaries
+    // into the bundle, so the function architecture is always arm64. Its
+    // s3 image loader reads originals from the asset bucket.
+    const imageFunction = yield* LambdaFunction("ImageOptimization", {
+      main: fromDist("image-optimization-function/index.mjs"),
+      handler: "handler",
+      isExternal: true,
+      bundle: false,
+      runtime: "nodejs22.x",
```
sam-goodwin: Should we go to node 24? Better to follow latest unless it's not supported

packages/alchemy/src/AWS/Website/Nuxt.ts
```diff
@@ -0,0 +1,235 @@
...
+ *     memorySize: 2048,
+ *     environment: {
+ *       NUXT_PUBLIC_API_BASE: api.url,
+ *     },
+ *   },
+ * });
+ * ```
+ */
+export const Nuxt = (id: string, props: NuxtProps = {}) =>
+  Effect.gen(function* () {
```
sam-goodwin: Effect.fn

packages/alchemy/src/AWS/Website/StaticSite.ts
```diff
@@ -203,6 +275,16 @@ export const StaticSite = (id: string, props: StaticSiteProps) =>
...
+    if (props.spa && props.errorPage) {
+      return yield* Effect.die(
+        `Cannot provide both "spa" and "errorPage". A SPA answers misses with the index page (200); "errorPage" answers them with a real 404.`,
+      );
+    }
+    if (server && (props.spa || props.errorPage)) {
+      return yield* Effect.die(
+        `A site with a server origin routes misses to the server; "spa" and "errorPage" do not apply.`,
+      );
+    }
```
sam-goodwin: Where did the design of props.spa and props.errorPage come from? Is that inspired by SST or something?

packages/alchemy/src/AWS/Website/SvelteKit.ts
```diff
@@ -0,0 +1,241 @@
...
+
+    const server = yield* LambdaFunction("Server", {
+      main: build.serverEntry as unknown as string,
+      handler: "handler",
+      isExternal: true,
+      // The AWS deploy target's finishing pass writes `dist/server` as a
+      // complete Node deployment unit (entry + chunks) — ship it as-is.
+      bundle: false,
+      runtime: "nodejs22.x",
+      architecture: props.server?.architecture,
```
sam-goodwin: This looks mostly like a copy/paste for each web framework. Is that intentional? Can we share most of the logic across all websites? What would that look like? NextJS was possibly the weirdest, but the rest look remarkably similar.

sam-goodwin (commented):
Looks like a bit too much copy/paste acorss them. Look for ways to share.

PR #1145, opened 2026-08-09
chore: use new pr-package flow for faster package previews

BlankParticle:
Replace wave-based PR package publication with one content-addressed job for the complete Alchemy package set.

- Store each backing tarball by `(package name, SHA-256)`; byte size is validation and metadata, not identity
- Probe and upload through hash-only package URLs
- Use `Alchemy-Tags`, `Alchemy-Tarball-Hash`, and `Alchemy-TTL` protocol headers
- Point commit dependencies before exposing short, branch, and PR aliases
- Rewrite internal dependencies to full-commit package URLs, with no graph-specific tags
- Replace incompatible tag indexes rather than carrying a compatibility layer
- Validate uploaded bytes against the URL SHA-256 through R2
- Schedule TTL expiration with Alchemy named Durable Object events
- On expiry, remove KV tags still pointing at the tarball, delete the R2 blob, and clear its Durable Object state
- Retry failed expiration cleanup after one minute
- Run publishing on one `ubuntu-22-large` runner
- Keep checkout, setup, Turbo cache, workspace builds, package contents, and commenting explicit in the caller
- Copy the root README only into `packages/alchemy` before building; the Action performs no README injection
- Resolve Distilled packages from their explicitly marked submodule commit
- Pin the merged implementation from alchemy-run/actions#10
- Restore the temporarily disabled pull-request workflows

Production verification passed end to end. Distilled published successfully, then Alchemy reused all six shared Distilled tarballs byte-for-byte. Alchemy publishing, dependency rewriting, the Actions summary, and the grouped copyable PR comment all completed successfully.

Depends on alchemy-run/distilled#436.

### DIFF
--- a/packages/pr-package/README.md
+++ b/packages/pr-package/README.md
@@ -91,7 +91,7 @@ The stack output gives you the worker URL and the auto-generated bearer token. S
 | Option          | Type                                | Default          | Notes                                                                |
 | --------------- | ----------------------------------- | ---------------- | -------------------------------------------------------------------- |
 | `parseAliasUrl` | `(url: URL) => AliasMatch \| null`  | `() => null`     | Maps any non-`/projects/...` GET to `{ pkgName, tag }` for a 301.    |
-| `defaultTtl`    | `string` (Effect Duration)          | `"3 weeks"`      | TTL applied when an upload doesn't pass `X-TTL`.                     |
+| `defaultTtl`    | `string` (Effect Duration)          | `"3 weeks"`      | TTL applied when a tag request doesn't pass `Alchemy-TTL`.           |
 
 `AliasMatch` is `{ pkgName: string; tag: string }`. Returning `null` falls through to the regular `/projects/:pkgName/...` matcher.
 
@@ -114,9 +114,9 @@ Assigns tags to an existing backing tarball without uploading its bytes again.
 Headers:
 
 - `Authorization: Bearer <token>` (required)
-- `X-Tarball-Hash: <sha256>` (required)
-- `X-Tags: <json-array>` (required) — e.g. `["main","abc1234","abc1234abc1234..."]`
-- `X-TTL: <duration>` (optional) — e.g. `"7 hours"`, `"3 weeks"`. Effect `Duration` syntax.
+- `Alchemy-Tarball-Hash: <sha256>` (required)
+- `Alchemy-Tags: <json-array>` (required) — e.g. `["main","abc1234","abc1234abc1234..."]`
+- `Alchemy-TTL: <duration>` (optional) — e.g. `"7 hours"`, `"3 weeks"`. Effect `Duration` syntax.
 
 If a tag already points elsewhere, it moves to the new tarball. A tarball is deleted after its final tag is removed.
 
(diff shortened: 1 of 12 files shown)


packages/pr-package/README.md
```diff
@@ -99,38 +99,45 @@ The stack output gives you the worker URL and the auto-generated bearer token. S
...
+Assigns tags to an existing backing tarball without uploading its bytes again.
 
 Headers:
+
 - `Authorization: Bearer <token>` (required)
-- `Content-Type: application/gzip`
+- `X-Tarball-Hash: <sha256>` (required)
+- `X-Tarball-Size: <bytes>` (required)
 - `X-Tags: <json-array>` (required) — e.g. `["main","abc1234","abc1234abc1234..."]`
 - `X-TTL: <duration>` (optional) — e.g. `"7 hours"`, `"3 weeks"`. Effect `Duration` syntax.
```
Mkassabov: Apparently we're not supposed to use `X-` anymore can we prefix with `Alchemy-` or `PKG-` or something?

packages/pr-package/README.md
```diff
@@ -99,38 +99,45 @@ The stack output gives you the worker URL and the auto-generated bearer token. S
...
 All routes are scoped by `:pkgName`, which can be scoped (`@scope/name`) or unscoped (`name`) — matches npm package naming.
 
-### `PUT /projects/:pkgName/packages` — upload
+### `HEAD /projects/:pkgName/packages/:sha256/:size` — probe
+
+Checks whether the backing tarball identified by `(package name, SHA-256, byte size)` already exists. Authentication is required. Returns 200 when present and 404 otherwise.
+
+### `PUT /projects/:pkgName/packages/:sha256/:size` — upload
+
+Uploads the raw `.tgz` stream when the content-addressed backing tarball is absent. Authentication, `Content-Type: application/gzip`, and a matching `Content-Length` are required. Repeating the request is idempotent and does not overwrite existing content.
```
Mkassabov: TTL should be on the bundle not on the tag, we want to expire the actual object in R2 so we don't build up long term cost. Expiring the tags doesn't provide a benefit

packages/pr-package/README.md
```diff
@@ -99,38 +99,45 @@ The stack output gives you the worker URL and the auto-generated bearer token. S
...
-Returns the `.tgz` with `cache-control: public, max-age=31536000, immutable`. No auth required (resourceIds are unguessable UUIDs).
+Returns the `.tgz` with `cache-control: public, max-age=31536000, immutable`. No auth required; the URL itself is content-addressed.
 
 ### `DELETE /projects/:pkgName/tags/:tag` — remove tag
 
-Auth required. If the tag was the resource's last one, the blob and metadata are also deleted.
+Auth required. If the tag was the tarball's last one, the backing blob is also deleted.
 
-### `GET /projects/:pkgName/packages/:resourceId/stats` — download stats
+### `GET /projects/:pkgName/packages/:sha256/:size/stats` — download stats
```
Mkassabov: Why do we need size here?

Mkassabov (changes requested):

PR #1147, opened 2026-08-09
chore: add license and third-party attribution metadata

BlankParticle:
## Summary

- Add Apache-2.0 licensing files and third-party attribution documentation.
- Add package license metadata and include notices in published packages.
- Mark vendored and adapted source files with licensing references.
- Update build and release workflows to copy package metadata automatically.


packages/cloudflare-test-tools/LICENSE
sam-goodwin, line 1: Can we have this LICENSE once in the root instead of duplicated per package?

BlankParticle: ah thats left over, i intend to have a root license only already 
it's copied at build for publishing

PR #1148, opened 2026-08-09
feat(cloudflare): add Vite framework Worker contributions

aryasaatvik:
Frameworks that generate a Worker entry often expect the official Cloudflare Vite plugin to own workerd and deployment. Running that plugin beside Alchemy would create two infrastructure authorities, so this adds a framework-neutral handoff into Alchemy's existing runtime and Worker provider.

```ts
// vite.config.ts
export default defineConfig({
  plugins: [
    framework(),
    Cloudflare.cloudflareViteFramework(frameworkWorkerConfig()),
  ],
});
```

```text
framework Vite plugin
  -> generated/virtual Worker entry + flags + hosted DO declarations
  -> cloudflareViteFramework() normalizes the contribution
  -> Alchemy's single Cloudflare runtime plugin
  -> local workerd namespaces / production Worker metadata
  -> WorkerProvider append-only DO migrations
```

### Ownership

- `cloudflareViteFramework()` contributes configuration only; it does not start workerd or install a second Cloudflare runtime plugin.
- The generated entry must preserve and re-export the application Worker surface. Competing entries are rejected instead of silently dropping either owner.
- Application bindings, routes, exports, and Durable Objects remain owned by the Worker. Exact DO duplicates stay application-owned; aliases reuse the hosted class; conflicting declarations fail before upload.
- Alchemy merges framework compatibility flags and hosted DO bindings into production metadata, then assigns stable `vite:<binding>` identities for append-only migrations.
- Immutable `version.parent` Workers reject `vite` configuration because that API cannot carry the framework build seam.

### Fixture

The Flue-shaped fixture generates a virtual Worker entry that re-exports the application fetch handler and application-owned Durable Object, then adds one framework-owned Durable Object. It proves both local namespaces, normalized contribution metadata, compatibility-flag deduplication, collision handling, stable migration identity, and exactly one Alchemy Cloudflare runtime plugin. No Wrangler configuration or official Cloudflare Vite plugin is added.


sam-goodwin:
I don't understand what this does. Can you explain the use case a bit more?

aryasaatvik:
The use case is a framework that generates the Worker entry and hosted Durable Object declarations while the application already deploys that Worker through Alchemy. Adding the official Cloudflare Vite plugin beside Alchemy would give both plugins ownership of workerd, bindings, deployment metadata, and migrations. This seam lets the framework contribute only its entry, compatibility flags, and hosted classes; Alchemy's existing runtime/plugin/provider remains the single infrastructure owner. I've expanded the PR body and fixture to show that flow, including preservation of the application Worker and its Durable Object without adding a second Cloudflare plugin.

aryasaatvik:
A concrete example is a project using Flue together with Alchemy. Flue generates part of the Worker: a virtual entry point, a FlueAgent Durable Object, its binding, and compatibility flags such as nodejs_compat.
That project may already use Alchemy to define and deploy the rest of the Worker—its request handler, application-owned Durable Objects, bindings, local runtime, and production configuration.
The problem is that Flue would normally install Cloudflare’s Vite plugin to manage its generated pieces. Running that plugin alongside Alchemy would give two systems responsibility for the same Worker.
This PR provides a cleaner handoff: Flue contributes the Worker code and declarations it generates, while Alchemy combines them with the application’s code and remains responsible for the complete Worker, including local development, production deployment, bindings, and Durable Object migrations.
Flue is the example, but the API is framework-neutral. Any Vite framework that generates Worker code or hosted Durable Objects could use the same handoff without introducing a second Cloudflare plugin or a separate Wrangler configuration.

sam-goodwin:
I don't think this is the right way to integrate flue. Change is too large and not something I can merge right now.

PR #1150, opened 2026-08-10
fix(cloudflare): preserve unresolved container attachments

danieljvdm:
Preserve a container’s Durable Object attachment when its Worker’s namespace output is temporarily unresolved.

- Prefer the observed live attachment over stale cached state.
- Recreate a missing application only with its exact recorded namespace; reject missing or malformed recovery data.
- Distinguish unresolved bindings from intentional attachment removal in planning and reconciliation.

Replace mocked provider tests with deployed Worker/container fixtures in `Container.test.ts`, including HTTP round trips and state restoration.


sam-goodwin:
Please add a regression test that reproduces this error and tests (with test.provider) that we recover from it in test/Cloudflare/Workers/Worker.test.ts

danieljvdm:
Added the regression test (`test.provider`, in `test/Cloudflare/Workers/Worker.test.ts`): **"container reconcile with an unresolved DO attachment preserves the live application"**. It deploys a Worker with a Container in `env`, doctors the scratch state to the incident's shape (worker's `durableObjectNamespaces` blanked, container's image hash staled), and asserts the redeploy keeps the live application (same id, attachment intact, verified out-of-band); a second leg deletes the application out-of-band and asserts the next reconcile recreates it with the namespace resolved by class name. It is red on `main` with the exact incident error (`BadRequest VALIDATE_INPUT` with `durable_objects.*` required) — and it also caught that this branch as originally submitted still took the destructive path: the stale attachment arrives as `{ namespaceId: undefined }` (truthy container, unresolved inner value), slipping past both the guard and the by-class fallback. Amended `getDurableObjects` to treat a namespaceId-less attachment as unresolved (54ab33c); with that, both legs are green.

danieljvdm:
anything more I need to do?

sam-goodwin:
Sorry, fell off my radar. Looking

danieljvdm:
@sam-goodwin bump on this

PR #1154, opened 2026-08-10
feat(aws): local dev with floci

sam-goodwin:
Local dev for AWS. `alchemy dev` now runs your AWS stack on your machine — no AWS account, no credentials, no config — using [floci](https://floci.io) (MIT LocalStack-style emulator, run from [our fork](https://github.com/alchemy-run/floci)). `alchemy deploy` is unchanged: same stack file, real cloud. Stacked on #1140.

### Zero-setup local dev

```sh
bun alchemy dev
# no AWS credentials — using the local emulator
```

No stored profile, no env vars needed. The emulator starts automatically in Docker (one container, `alchemy-floci`, shared across projects and sessions) and your stack deploys into it in seconds:

```ts
const files = yield* AWS.S3.Bucket("Files");
const jobs  = yield* AWS.SQS.Queue("Jobs");
const users = yield* AWS.DynamoDB.Table("Users", {
  partitionKey: "id",
  attributes: { id: "S" },
});
const api = yield* AWS.Lambda.Function("Api", { main: "./src/api.ts" });
```

```
Api.functionUrl  http://b71ff5c0….lambda-url.us-east-1.localhost:4566/
Jobs.queueUrl    http://localhost:4566/000000000000/jobs-dev-x7k2
Users.tableArn   arn:aws:dynamodb:us-east-1:000000000000:table/users-dev-p3m9
```

Everything runs for real, locally: the Lambda executes in a runtime container behind a working Function URL, SQS→Lambda event source mappings pump messages, S3 notifications fire, EventBridge routes. SDK calls inside your function need zero config — `S3.putObject(...)` just lands in the local bucket.

### Hot reload

Save `./src/api.ts` and the running function serves the new code in **~100–500ms**. No redeploy, no restart — the dev session watches your code, rebuilds, and the emulator swaps it under the live Function URL.

### Mix in the real cloud with `Alchemy.remote()`

```ts
// everything else is local; this one hits real AWS
const secret = yield* AWS.SecretsManager.Secret("ProdSecret", { ... })
  .pipe(Alchemy.remote());
```

If a `remote()` resource needs credentials you don't have, `alchemy dev` tells you before touching anything:

```
CredentialsRequired: 1 resource runs against the real cloud via Alchemy.remote():
  - ProdSecret (AWS.SecretsManager.Secret)
Run `alchemy login --profile <name>`, or set CI=1 to use environment credentials.
```

Switching a resource between local and live (or dev → deploy) plans a **replacement** — dev state never silently becomes cloud state.

### Supported in this PR

S3, SQS, DynamoDB, SNS, EventBridge, Secrets Manager, SSM, IAM roles, Lambda (with hot reload + Function URLs + event source mappings). Resources without local emulation run against the real cloud as before — mixed stacks just work.

### Emulator management

The container outlives dev sessions (Docker is the supervisor); nothing to babysit. Overrides when you want them:

```sh
ALCHEMY_FLOCI_IMAGE=floci:dev bun alchemy dev   # your own build
bun floci:dev                                    # contributors: live-reload emulator from .vendor/floci
```

The default image is our patched release (`ghcr.io/alchemy-run/floci`) — emulator gaps we hit get fixed in the fork and released, most recently SNS data-protection ops, event-source-mapping tags, and streaming Function URLs.

> [!WARNING]
> BREAKING: `@alchemy.run/web-frameworks` is renamed **`@alchemy.run/frontend-frameworks`** — the home for frontend-framework integrations across all clouds. Update your devDependency.


packages/alchemy/src/AWS/DynamoDB/Table.ts
```diff
@@ -549,6 +549,11 @@ export const TableProvider = () =>
           Effect.catchTag("PolicyNotFoundException", () =>
             Effect.succeed(undefined),
           ),
+          // Local emulators (floci/LocalStack) don't implement resource
+          // policies — treat "operation not supported" as "no policy".
+          Effect.catchTag("UnknownOperationException", () =>
+            Effect.succeed(undefined),
+          ),
```
sam-goodwin, line 556: Should we gate this on whether alchemy is in dev mode or not?

Issue #1149, opened 2026-08-10
pnpm peers check fails on a fresh alchemy@2.0.0-beta.70 install (Cloudflare and Effect)


shuuheyhey:
## Summary

A fresh pnpm install of `alchemy@2.0.0-beta.70` produces an inconsistent peer graph. `pnpm peers check` fails for both the Cloudflare packages and the Effect packages.

## Reproduction

```sh
mkdir alchemy-peer-repro
cd alchemy-peer-repro
pnpm init
pnpm add --ignore-scripts \
  alchemy@2.0.0-beta.70 \
  effect@4.0.0-beta.102 \
  @effect/platform-node@4.0.0-beta.102 \
  @effect/platform-bun@4.0.0-beta.102
pnpm peers check
```

Tested with pnpm 11.21.0.

## Actual result

```text
✕ unmet peer effect
  Installed: 4.0.0-beta.102
  Wanted:
    ^4.0.0-beta.106:
      @effect/platform-node-shared@4.0.0-beta.106
      @effect/sql-d1@4.0.0-beta.106
      @effect/sql-sqlite-do@4.0.0-beta.106
      @effect/vitest@4.0.0-beta.106

✕ unmet peer @distilled.cloud/cloudflare
  Installed: 1.0.0-rc.2
  Wanted:
    ^0.30.3:
      @distilled.cloud/cloudflare-runtime@0.16.1
      @distilled.cloud/cloudflare-vite-plugin@0.16.1
```

The published Alchemy package currently installs:

- `@distilled.cloud/cloudflare@1.0.0-rc.2`
- `@distilled.cloud/cloudflare-runtime@0.16.1`
- `@distilled.cloud/cloudflare-vite-plugin@0.16.1`

## Expected result

The officially published Alchemy version should provide a peer-consistent dependency set so that a fresh install passes `pnpm peers check` without consumer-side peer overrides.

PR #1146 appears to consolidate the Cloudflare runtime packages on main, but no npm release includes that change yet. Could you publish a follow-up beta from current main, or document an official compatible version set for beta.70?

This currently blocks a Cloudflare resource adoption because we do not want to deploy with a guessed peer override.

sam-goodwin:
@BlankParticle can you take a look at this?

shuuheyhey:
I traced this against the published manifests and the current `main` preview.

[...]

PR #1156, opened 2026-08-10
test(release): check pnpm peer graph

shuuheyhey:
Prevent #1149 from recurring by installing the just-published `alchemy` PR package with pnpm 11.21.0 and failing when its peer graph is inconsistent.

The check follows the same package URLs emitted by the PR-package workflow, disables pnpm's exotic-subdependency block only for those trusted preview URLs, and ignores install scripts.

This covers the published-manifest boundary that workspace installs and the Bun lockfile cannot expose. The current `main` preview resolves a coherent Effect beta.106 and Distilled rc.3 graph.

### DIFF
--- a/.github/workflows/pr-package.yml
+++ b/.github/workflows/pr-package.yml
@@ -69,6 +69,34 @@ jobs:
             ]
           pr-package-token: ${{ secrets.PR_PACKAGE_TOKEN }}
 
+      - name: Check pnpm peer graph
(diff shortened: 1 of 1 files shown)


BlankParticle:
we already maintain peer deps properly, new alchemy releases are just blocked due to a drizzle issue, this change is not needed in CI
we will release new alchemy version soon once the drizzle issue is resolved

PR #1168, opened 2026-08-11
perf(cli): lazily load command graph

patrikduksin:
Reduce CLI startup work by loading only the selected command and delaying command-only services until a handler actually runs.

```text
                              before     after
CLI help mean                1,213 ms    140 ms
dev to first Plan mean       1,713 ms    488 ms
CLI main module import         727 ms     88 ms
```

CPU and differential-import profiling showed that help eagerly loaded every command, all built-in auth providers, generated AWS/Cloudflare clients, telemetry exporters, HTTP clients, and UI dependencies. The HTTP exporter/client path alone accounted for roughly 334 ms of sampled self-time.

- Load only the selected command; retain lightweight placeholders for root help and load all commands for completion generation
- Skip telemetry, credentials, HTTP, artifact, and UI layers when no command handler runs
- Load built-in auth providers only when login/profile needs them
- Load the version-check HTTP client only when its cache is stale
- Add a repeatable startup benchmark and focused command-loader coverage

The representative Cloudflare consumer stack import remains unchanged at roughly 0.85 seconds, so the reduction comes from removing unrelated CLI initialization rather than optimizing the benchmark fixture.


BlankParticle:
Ok so did some cleanup to bring this into a state we can merge it
removed the benchmark since we don't really need to benchmark startup time too much, and general use of alchemy always need to wait for network and I/O
this version is probably bit slower than what you did, but this is atleast doing the lazy loading of commands and providers

sam-goodwin:
I care about a benchmark. I want to have a measure of it. Maybe even commented on GitHub like how effect comments the bundle size of their packages to track tree shaking.

BlankParticle:
do you want to have a benchmark as part of this pr, or later when we also add bundle size benchmarks too?

sam-goodwin:
Thanks for taking the initiative to fix this but since we are re-writing the CLI right now, I am going to close this in favor of that re-write. We'll heavily optimize and benchmark the new one.

patrikduksin:
@sam-goodwin sure no problem, excited to try new cli :)

PR #1170, opened 2026-08-11
feat(cloudflare-runtime): opt-in workerd inspector via WORKERD_INSPECTOR_WORKER

Butch78:
workerd's `--inspector-addr` exposes the Chrome DevTools protocol for the isolate — headless CPU profiling of the JS and wasm running in a dev worker, no browser attached. `Workerd.serve` already counts an `--inspector-addr` listener in its control-socket handshake, but nothing passes the flag, so a dev-session worker cannot be profiled today.

This adds an env-gated debug facility in the spirit of `WORKERD_DUMP_CONFIG`:

```sh
WORKERD_INSPECTOR_WORKER=my-worker alchemy dev
# then: curl http://127.0.0.1:9229/json  →  attach CDP, Profiler.start/stop
```

`WORKERD_INSPECTOR_ADDR` overrides the default `127.0.0.1:9229`. Scoped to one named worker because every worker runs in its own workerd process — an unconditional fixed address would collide across a multi-worker session. Unset, the serve args are unchanged.

A typed per-worker option (alongside `logging`) would be the fuller shape; the env gate works from any consumer without an API change.

### DIFF
--- a/packages/cloudflare-runtime/src/core/Runtime.ts
+++ b/packages/cloudflare-runtime/src/core/Runtime.ts
@@ -209,6 +209,16 @@ export const RuntimeLive = Layer.effect(
             concurrency: "unbounded",
           },
         );
+        // Debug facility: expose workerd's V8 inspector (the Chrome DevTools
+        // protocol) for one worker. `WORKERD_INSPECTOR_WORKER=<worker name>`
+        // selects it; `WORKERD_INSPECTOR_ADDR` overrides the default listen
+        // address. Scoped to a single named worker because every worker runs
+        // in its own workerd process, so an unconditional fixed address
+        // would collide across them.
+        const inspectorAddr =
+          process.env.WORKERD_INSPECTOR_WORKER === worker.name
+            ? (process.env.WORKERD_INSPECTOR_ADDR ?? "127.0.0.1:9229")
+            : undefined;
         const ports = yield* workerd.serve(
           {
             sockets: [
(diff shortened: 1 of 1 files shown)


sam-goodwin:
What's the best way to test this?

Butch78:
Manual, two steps — the env var names the worker, then any CDP client attaches:

```sh
WORKERD_INSPECTOR_WORKER=my-worker alchemy dev
curl http://127.0.0.1:9229/json   # → [{"id":"user-worker","webSocketDebuggerUrl":"ws://127.0.0.1:9229/user-worker",…}]
```

`chrome://inspect` → "Open dedicated DevTools for Node" works against that address, or headless:

```ts
const ws = new WebSocket("ws://127.0.0.1:9229/user-worker");
let id = 1;
ws.onopen = () => ["Profiler.enable", "Profiler.start"].forEach((method) => ws.send(JSON.stringify({ id: id++, method })));
// drive some requests, then Profiler.stop returns a .cpuprofile
```

We validated it on our stack (multi-worker `alchemy dev`, Rust-wasm worker + Durable Object): the named worker exposes the target, unnamed workers are untouched, and a 25s `Profiler.start/stop` under load returned a 63k-sample profile with the wasm module visible in `Debugger.scriptParsed`. One thing worth knowing for docs: the value has to be the runtime `worker.name` (for us an internal generated name printed at startup), not the resource id the user wrote.

I didn't add a vitest case because the package's suite currently fails at import on an unmodified checkout (`Schema.TaggedErrorClass is not a function`); happy to add one alongside a fix for that if useful.

sam-goodwin:
How do we support multiple workers?

PR #1175, opened 2026-08-11
refactor: make process cleanup explicit and move node-utils into the workspace

BlankParticle:
## Summary

- remove import-time process signal and exit-hook side effects from Alchemy
- make spawned-process, RPC sidecar, and local registry cleanup explicit and scoped
- replace the auth lock with an Effect-scoped cross-process filesystem lease
- move `@alchemy.run/node-utils` into the monorepo as an `ignore`-only package
- port the applicable upstream `ignore` and `proper-lockfile` test coverage
- add regression coverage for the remaining workerd orphan case after abrupt CLI termination

## Process cleanup

Alchemy no longer installs process-wide signal or exit listeners merely by being imported. The CLI owns foreground-child behavior directly, including signal forwarding, IPC bridging, stderr filtering, and propagation of the child's exit result.

RPC sidecars retain graceful scoped cleanup and terminate when their parent connection disappears. Cloudflare local registry cleanup is awaited and ordered so stale registry files are removed before shutdown completes.

Hard termination such as `SIGKILL` cannot run in-process finalizers. A real `alchemy dev` regression test records the remaining gap: when the CLI is SIGKILLed, a descendant workerd process can still become orphaned. The test remains an expected failure until process-independent supervision is implemented.

## Effect auth lock

`Auth/Lock.ts` now uses an in-process semaphore plus an atomic lock directory for cross-process exclusion. The lock is scoped to the protected Effect, refreshes its mtime with a heartbeat, reaps stale locks, and verifies ownership before releasing.

Tests adapted from `node-proper-lockfile` cover:

- same-key serialization and different-key concurrency
- live cross-process exclusion and retry behavior
- release after success, failure, and interruption
- heartbeat refresh and stale-lock recovery
- ownership changes while a lock is held
- a real child process being SIGKILLed followed by stale recovery

Upstream cases for manual `check`, `unlock`, synchronous APIs, custom filesystems, and path realpath behavior do not apply to the scoped Effect API.

## Workspace node-utils package

`@alchemy.run/node-utils` now lives at `packages/node-utils` and intentionally exports only `./ignore`. Alchemy and cloudflare-runtime consume it through a workspace dependency. Package versions, project references, release workflows, package-file copying, and third-party license attribution are wired into the monorepo.

The matcher is a typed port of `ignore@7.0.5`. Its applicable upstream suite is ported, including:

- the complete behavior-case matrix for `filter`, `createFilter`, `ignores`, and `checkIgnore`
- options and public-method regressions
- native `git check-ignore` parity tests
- hidden ignore fixtures
- default and named import smoke coverage

The upstream cross-copy-instance regression restored symbol-marker detection rather than `instanceof`, preserving compatibility between instances originating from different module copies.

## Runtime compatibility

TypeScript fixture launching checks the child Node runtime's `process.features.typescript` capability. Transform flags are supplied only when supported, including safe behavior on older Node versions and Node 26's strip-only implementation.

## Verification

- node-utils: 414 tests pass
- auth lock focused suite: 14 pass; the slow heartbeat case is skipped only under `FAST=1`
- focused RPC cleanup, Cloudflare registry, runtime compatibility, and workerd regression suites are included in the branch


packages/alchemy/src/Auth/Lock.ts
```diff
@@ -18,79 +35,152 @@ import { rootDir } from "./Profile.ts";
...
+      info.mtime,
+      (mtime) => mtime.getTime() < now - Duration.toMillis(STALE),
+    );
+    if (isStale) {
+      yield* fs.remove(lockPath, { recursive: true, force: true });
+    }
+  }).pipe(
+    Effect.catchIf(
+      (error) => error.reason._tag === "NotFound",
+      () => Effect.void,
```
sam-goodwin: `Effect.catchTag("NotFound"`

BlankParticle: the error is of type `PlatformError`, tag is on the `.reason`, I also thought I would use `catchTag`, but looks like this is easier

packages/alchemy/src/Auth/Lock.ts
```diff
@@ -1,9 +1,26 @@
...
 import { rootDir } from "./Profile.ts";
 
+const semaphores = new Map<string, Semaphore.Semaphore>();
+
+// Keep STALE well above REFRESH — a saturated machine can starve the
+// heartbeat fiber for seconds.
+const STALE = Duration.seconds(30);
+const REFRESH = Duration.seconds(10);
+const RETRY_INTERVAL = Duration.millis(50);
+const DEFAULT_TIMEOUT: Duration.Input = "2 minutes";
```
sam-goodwin: nit: may as well be consistent with how you create durations

`Duration.minutes(2)` or switch to string form

packages/alchemy/src/Auth/Lock.ts
```diff
@@ -18,79 +35,152 @@ import { rootDir } from "./Profile.ts";
...
+    Effect.catchIf(
+      (error) => error.reason._tag === "AlreadyExists",
+      () => reapStale.pipe(Effect.andThen(Effect.fail(new LockHeld()))),
+    ),
+    Effect.retry({
+      while: (error) => error._tag === "LockHeld",
+      schedule: Schedule.spaced(RETRY_INTERVAL),
+      times: Math.ceil(
+        Duration.toMillis(timeout) / Duration.toMillis(RETRY_INTERVAL),
+      ),
```
sam-goodwin: Hmm, why compute times? Probably better to just set a timeout on the Schedule.

packages/alchemy/src/Auth/Lock.ts
```diff
@@ -18,79 +35,152 @@ import { rootDir } from "./Profile.ts";
...
+    }),
+    Effect.catchTag("LockHeld", () =>
+      Effect.die(
+        new Error(
+          `Timed out waiting for the alchemy auth lock '${lockPath}' — another alchemy ` +
+            `process has held it for over ${Duration.toSeconds(timeout)}s. If no other ` +
+            `alchemy process is running, delete the lock directory and retry.`,
+        ),
```
sam-goodwin: Add a named and TaggedError since this looks like an intentional failure

packages/alchemy/src/Auth/Lock.ts
```diff
@@ -18,79 +35,152 @@ import { rootDir } from "./Profile.ts";
...
+      Effect.andThen(fs.remove(lockPath, { recursive: true, force: true })),
+      Effect.ignore,
+    ),
+  );
+
+  yield* fs.readFileString(ownerPath).pipe(
+    Effect.filterOrFail((current) => current === owner),
+    Effect.andThen(Clock.currentTimeMillis),
+    // NB: utimes interprets a bare number as *seconds* since epoch.
+    Effect.flatMap((now) => fs.utimes(lockPath, new Date(now), new Date(now))),
```
sam-goodwin:
```ts
Effect.andThen(Clock.currentTimeMillis.pipe(Effect.map(now => new Date(now))),
Effect.flatMap((now) => fs.utimes(lockPath, now, now)),
```

packages/alchemy/src/Auth/Lock.ts
```diff
@@ -18,79 +35,152 @@ import { rootDir } from "./Profile.ts";
...
+    Effect.repeat(Schedule.spaced(REFRESH)),
+    Effect.catch(() =>
+      Effect.logWarning(
+        `auth lock compromised (continuing): '${lockPath}' is no longer owned by this process`,
+      ),
+    ),
+    Effect.forkScoped,
```
sam-goodwin, line 130: Why fork?

packages/alchemy/src/Auth/Lock.ts
```diff
@@ -18,79 +35,152 @@ import { rootDir } from "./Profile.ts";
...
+  options?: { readonly timeout?: Duration.Input },
+) => {
+  const safeKey = sanitizeLockKey(key);
+  let semaphore = semaphores.get(safeKey);
+  if (semaphore === undefined) {
+    semaphore = Semaphore.makeUnsafe(1);
+    semaphores.set(safeKey, semaphore);
+  }
+  return semaphore.withPermit(
+    Effect.scoped(
```
sam-goodwin: nit: i prefer Effect.scoped at the end of the .pipe

packages/alchemy/src/Auth/Lock.ts
```diff
@@ -18,79 +35,152 @@ import { rootDir } from "./Profile.ts";
...
+        // sees it uninitialised.
+        const lockPath = path.join(rootDir, "lock", `${safeKey}.lock`);
+        yield* acquireFileLock(
+          lockPath,
+          options?.timeout ?? DEFAULT_TIMEOUT,
+        ).pipe(
+          Effect.catchIf(isUnlockable, (error) =>
+            Effect.logWarning(
+              `auth lock unavailable (${errnoCode(error) ?? error.reason._tag} at '${lockPath}') — continuing without cross-process locking`,
+            ),
```
sam-goodwin: Why are we continuing instead of dying?

packages/alchemy/src/Local/RpcSpawner.ts
```diff
@@ -138,12 +136,7 @@ export const make = Effect.fn(function* ({
       Effect.ignore,
       Effect.forkScoped,
     );
-    const unregister = exitHook(() => {
-      killProcessGroup(handle.pid, "SIGKILL");
-    });
-    const kill = handle
-      .kill({ forceKillAfter: "500 millis" })
-      .pipe(Effect.tap(() => Effect.sync(unregister)));
+    const kill = handle.kill({ forceKillAfter: "500 millis" });
```
sam-goodwin, line 142: What is replacing this? What now does best-effort kill?

packages/alchemy/test/AWS/AccessAnalyzer/TestLease.ts
```diff
@@ -1,73 +1,38 @@
...
-              held = false;
-              return accessAnalyzerTestLock.release(1).pipe(Effect.asVoid);
-            }),
-          ),
-        );
-      }),
-    ),
+    acquire: Effect.gen(function* () {
+      const acquired = yield* Deferred.make<void, unknown>();
+      scope = yield* Scope.make();
```
sam-goodwin: Is this the scope that the finalizer in withLock is attached to?

Why are we extracting it to the `makeAccessAnalyzerTestLease` closure and sharing with `release`?

packages/cloudflare-runtime/src/core/bindings/assets/Assets.ts
```diff
@@ -101,8 +101,20 @@ export const AssetsLive = Layer.effect(
             }),
         ),
       );
-      const { assetsIgnoreFunction } = yield* Effect.promise(() =>
-        createAssetsIgnoreFunction(dir),
+      const { assetsIgnoreFunction } = yield* createAssetsIgnoreFunction(
+        dir,
+      ).pipe(
+        Effect.provideService(FileSystem.FileSystem, fs),
+        Effect.provideService(Path.Path, path),
```
sam-goodwin: Never use two subsequent `Effect.provide*`, always one `Effect.provide` + chained `Layer.*` or, in this case, just do: `Effect.provide([Layer.succeed(FileSystem.FileSystem, fs), Layer.succeed(Path.Path, path)])`

But also, why do you need to provide the file system and path to this Effect? Surely it can use the surrounding Context's?

packages/cloudflare-runtime/src/core/registry/Registry.ts
```diff
@@ -168,22 +166,12 @@ export const RegistryLive = Layer.effect(
...
             // would unregister the live replacement.
-            // If scope finalizers fail to run, a synchronous exit hook ensures the entry is removed.
-            const unregisterSync = () => {
-              try {
-                if (NodeFs.readFileSync(entryPath, "utf8") === serialized) {
-                  NodeFs.unlinkSync(entryPath);
-                }
-              } catch {}
-            };
-            const unregister = exitHook(unregisterSync);
```
sam-goodwin, line 179: What is this replaced with?

I remember @john-royal had strong reasons for using sync here.

packages/cloudflare-runtime/src/core/workerd/Workerd.ts
```diff
@@ -111,12 +110,9 @@ const make = (
           ],
           Buffer.from(serializeConfig(config)),
         );
-        // Scope finalizers may not run if the parent exits, so we use an exit hook to ensure we always kill the process.
-        const unregister = exitHook(() => handle.kill());
         yield* Effect.addFinalizer(() =>
           Effect.sync(() => {
             handle.kill();
-            unregister();
```
sam-goodwin, line 119: I hope this direction works, but I must say i am also concerned. What is our new story for best effort process killing?

packages/alchemy/src/Auth/Lock.ts
john-royal: Ever tried `Effect.catchReason("PlatformError", "NotFound", () => Effect.void)`?

packages/alchemy/src/Local/RpcSpawner.ts
john-royal: Since the immediate child process isn't detached, you can probably do without the exit hook there.

Where I'm less sure is on children of that child process (e.g. the workerd process spawned by the Worker provider). I *think* Effect's `forceKillAfter` will kill the child process group, but you'll want to verify that.

packages/cloudflare-runtime/src/core/registry/Registry.ts
john-royal: I did, but the problem was that the Effect finalizer wasn't running reliably. If it does with `exitHook` gone, you should be alright.

The one thing I'd be concerned about is even if Alchemy isn't calling `process.exit` anymore, we can't guarantee that the user's code - or a package they're using - doesn't still call it. If it does, you're back to square one. That makes me afraid to remove the synchronous cleanup entirely, though I obviously see why you're doing it.

Ultimately my reason for slapping synchronous exit hooks everywhere was that leftover stuff kept popping up otherwise. It definitely feels like a bandaid. If you're confident you've addressed the root cause, rip off the bandaids.

packages/cloudflare-runtime/src/core/workerd/Workerd.ts
john-royal: Unless this is a detached process, this might not be necessary because it can't outlive the patent anyway. Not sure though.

john-royal (commented):
Left some comments but I'm all for ripping off bandaids and the logic seems sound.

I feel like there could be more regression tests here? But I'm also on my phone right now so maybe I just didn't see, or maybe existing tests are sufficient. Just wanted to flag that since behind every synchronous exit hook is a zombie process that either I or a user found.

packages/alchemy/src/Local/RpcSpawner.ts
BlankParticle: `workerd` is a grandchild which should be cleaned up in most cases, the only case we can't handle is `SIGKILL` on parent, which is what foreground child was doing, its a full node process which checks if the parent dies and kill all child/grandchild processes, but not worth it to run another node process to just do that
I have added a skipped test with TODO to address this, I have ideas on how to not have orphan workerd processes, but that will be in a later pr, worst case we need to add this back if my plan doesnt work out later

packages/cloudflare-runtime/src/core/registry/Registry.ts
BlankParticle: since I have move vite and other things into child processes, the likely hood of random things calling process.exit is very low
if we run into that anyways, we can hijack process.exit to let effect runtime shutdown properly first

BlankParticle:
I fixed most of the things suggested, about the orphan workerd issue, there isn't much we can't do sadly it looks like, even miniflare doesn't handle it

I thought I would vendor in `foreground-child` watchdog back again, but looks like it didn't really work anyways before
Screenshot 2026-08-12 at 17-11-04

packages/alchemy/src/Auth/Lock.ts
sam-goodwin, line 1: Let's port over the tests https://github.com/moxystudio/node-proper-lockfile/tree/master/test and adapt them to effect solution

packages/cloudflare-runtime/src/core/test/Workerd.test.ts
```diff
@@ -376,6 +375,7 @@ layer(services)((it) => {
     { timeout: 30_000 },
   );
 
+  it.skip("TODO: workerd shuts down after an uncatchable parent SIGKILL", () => {});
```
sam-goodwin, line 378: these testes are in-process, can we do one that tests via `alchemy` CLI

packages/alchemy/test/AWS/AccessAnalyzer/TestLease.ts
```diff
@@ -1,73 +1,40 @@
...
+      yield* withLock(
```
sam-goodwin: Since this is only a test, i think we can just do this in process with a semaphore

packages/alchemy/bin/cli.js
```diff
@@ -66,27 +131,6 @@ if (runtime === "bun" && isDev) {
 // .ts only runs under bun.
 args.push(runtime === "bun" ? tsEntry : jsEntry, ...process.argv.slice(2));
 
-process.on("uncaughtException", (error) => {
-  // STATUS_CONTROL_C_EXIT on Windows — the watchdog inherits Ctrl-C from the
-  // console and exits with this code instead of a SIGINT signal.
-  const WIN_CTRL_C = 3221225786;
-  if (
```
sam-goodwin, line 73: Did we lose something here? Looks like a specific error

sam-goodwin (approved):

PR #1176, opened 2026-08-11
feat(cloudflare): bind Pipelines streams and legacy pipelines to Workers

spinx:
`toBinding()` had no arm for `Cloudflare.Pipelines.Stream` or `LegacyPipeline`, so a stream in a Worker's `env` fell through the classifier chain to the `json` fallback and deployed as a JSON binding holding the resource's attributes. Every
  other piece already shipped: the wire type (`PutScriptBindingPipelines`) is in distilled and already inside `WorkerBinding`, `RuntimeBindings.ts` handles `case "pipelines"`, and both resource guards exist.

  ```ts
  // WorkerAsyncBindings.ts — toBinding()
  } else if (isPipelinesStream(binding)) {
    return { type: "pipelines", name: bindingName, pipeline: binding.streamId };
  } else if (isLegacyPipeline(binding)) {
    return { type: "pipelines", name: bindingName, pipeline: binding.name };
  }
  ```

  The two generations address the binding differently, hence the two arms:

  - A **stream** is bound by id — wrangler's `stream = "<STREAM_ID>"`, [renamed from `pipeline`](https://developers.cloudflare.com/changelog/product/pipelines/) but lowering to the same API field.
  - A **legacy pipeline** is bound by name — the API documents the field as "Name of the Pipeline to bind to", matching `LegacyPipelineAttributes.name` ("unique per account; the API identifier").

  `WorkerBindingResource` gains both resources, and `InferEnv` maps them to `Pipeline` so `env.EVENTS.send(...)` types natively:

  ```ts
  // `Pipeline` is not an ambient global — only declared inside the module.
  import type { Pipeline } from "cloudflare:pipelines";
  ```

### DIFF
--- a/packages/alchemy/test/Cloudflare/Pipelines/StreamBinding.test.ts
+++ b/packages/alchemy/test/Cloudflare/Pipelines/StreamBinding.test.ts
@@ -60,7 +60,15 @@ const sendEvent = (url: string, nonce: string) =>
     }),
   );
 
-const stack = beforeAll(deploy(Stack));
+const stack = beforeAll(
+  deploy(Stack).pipe(
+    Effect.retry({
+      while: (e) => String(e).includes("Unable to authenticate request"),
+      schedule: Schedule.exponential("1 second"),
+      times: 6,
+    }),
+  ),
+);
 afterAll.skipIf(!!process.env.NO_DESTROY)(destroy(Stack));
 
 test(
--- a/packages/alchemy/test/Cloudflare/Pipelines/StreamBinding.test.ts
+++ b/packages/alchemy/test/Cloudflare/Pipelines/StreamBinding.test.ts
@@ -1,0 +1,119 @@
+import * as Cloudflare from "@/Cloudflare";
+import { CloudflareEnvironment } from "@/Cloudflare/CloudflareEnvironment";
+import * as Test from "@/Test/Alchemy";
+import * as workers from "@distilled.cloud/cloudflare/workers";
+import { expect } from "alchemy-test";
+import * as Data from "effect/Data";
+import * as Effect from "effect/Effect";
+import { MinimumLogLevel } from "effect/References";
+import * as Schedule from "effect/Schedule";
+import * as HttpClient from "effect/unstable/http/HttpClient";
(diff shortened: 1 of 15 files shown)


sam-goodwin:
Please add test fixtures exercising these

packages/alchemy/test/Cloudflare/Pipelines/fixtures/stack.ts
```diff
@@ -0,0 +1,33 @@
...
+import * as pathe from "pathe";
+
+export const EventsStream = Cloudflare.Pipelines.Stream("BindingStream", {});
+
+export const AsyncWorker = Cloudflare.Worker("PipelinesAsyncWorker", {
+  main: pathe.resolve(import.meta.dirname, "async-worker.ts"),
+  env: {
+    EVENTS: EventsStream,
+  },
+});
```
sam-goodwin, line 13: You also need to exercise the Effectufl workers, or were they already tested and we were just missing async stuff?

packages/alchemy/test/Cloudflare/Pipelines/StreamBinding.test.ts
```diff
@@ -0,0 +1,127 @@
...
+      ]),
+    }),
+  );
+
+const stack = beforeAll(
+  deploy(Stack).pipe(
+    Effect.retry({
+      while: (e) => String(e).includes("Unable to authenticate request"),
+      schedule: Schedule.exponential("1 second"),
+      times: 6,
```
sam-goodwin: You shouldn't retry this.

packages/alchemy/test/Cloudflare/Pipelines/fixtures/stack.ts
spinx: Didn't know about that feature. Added.

sam-goodwin: Not quite. I pushed what it should have been here: https://github.com/alchemy-run/alchemy/pull/1176/changes/1e88712a096bca6c8c12c2747e801402dac4b2c7

PR #1187, opened 2026-08-12
feat(git): git hosting on Workers, Durable Objects, and R2

sam-goodwin:
Git hosting on Cloudflare Workers + Durable Objects + R2, built with Alchemy Effect-native Workers, shipped inside `alchemy` as the `./Git` export. Full design with measurements in [`packages/alchemy/src/Git/DESIGN.md`](https://github.com/alchemy-run/alchemy/blob/claude/git-service-alchemy-workers-7ba779/packages/alchemy/src/Git/DESIGN.md). Battle-tested by hosting **the alchemy monorepo itself** (full history, 44k objects) on a live deployment, cloned back byte-identically under `git fsck --strict`.

Distributed as **building blocks**, not an appliance. Nothing ships a Worker: every component is an Effect `Context.Service` with swappable `Layer` implementations, assembled into a Worker you own.

```ts
import * as Git from "alchemy/Git";

class Api extends Git.Api.middleware(Authenticated) {} // yours: who may call what

const GitLive = Git.Server.layer(Api).pipe(
  Layer.provide(Git.Handlers),              // the engine's implementation of every route
  Layer.provide(AuthenticatedLive),         // your middleware
  Layer.provide(Git.ReposDurableObject),    // refs + objects + pulls
  Layer.provide(Git.RegistryDurableObject), // owner/name -> repoId
  Layer.provide(Git.BlobStoreR2(MyBucket)), // packs & clone bundles
  Layer.provide(Git.HasherInline),          // push verification
);

export default class GitHost extends Cloudflare.Worker<GitHost>()(
  "git",
  { main: import.meta.url, ...Git.GIT_WORKER_OPTIONS },
  Effect.gen(function* () {
    const git = yield* Git.Server;
    return { fetch: git.fetch };
  }).pipe(Effect.provide(GitLive)),
) {}
```

### Every plane is one `HttpApi` of route classes

`alchemy/Http` (new) makes a route a class: an `HttpApiEndpoint` that also names the tag of its implementation, so one route can have many implementations, each a Layer, and Effect's client, OpenAPI, groups, and middleware are unchanged.

```ts
export class GetRepo extends Http.get<GetRepo>()("get", "/repos/:owner/:repo", {
  params: RepoPath,
  success: Repo,
  error: [RepoNotFound],
}) {}

export const GetRepoLive = GetRepo.make(
  Effect.gen(function* () {
    const repos = yield* Git.RepoStore;               // build time
    return Effect.fn(function* ({ params }) {        // request time
      return yield* repos.get(params.owner, params.repo);
    });
  }),
);
```

`Git.Api` is six groups of such routes: `Repos`, `Refs`, `Objects` (incl. the raw blob and file reads), `Pulls`, `Protocol` (`info/refs`, `git-upload-pack`, `git-receive-pack`), and `GitHub`. Streaming routes declare no payload and answer with the response they build. `Git.Handlers` is the default implementation of every route, one `*Live` each. `Git.Server.layer(api)` serves any API derived from `Git.Api` and requires an implementation of every route plus the API's middleware; a Layer provided nearer than `Handlers` overrides a route. Under it, `Http.handlers(api)` mounts every route group under the API's middleware, and `Http.Platform` is the platform layer for Workers and Lambda.

```ts
export class AppApi extends Git.Api.add(AppRoutes).middleware(Authenticated) {}

Git.Server.layer(AppApi).pipe(
  Layer.provide([MeLive, GitHubUserLive]),   // yours, and one of the engine's replaced
  Layer.provide(Git.Handlers),               // the engine's for the rest
  Layer.provide(AuthenticatedLive),          // the middleware AppApi declares
)
```

Leave a route unimplemented and the compiler names it. `Git.ServerLive` is `Server.layer(Git.Api)` with `Handlers`: every route, nothing in front.

### No auth inside the engine

The engine holds no users, no credentials, and no policy. No route declares middleware or an auth error; who may call which route is the `HttpApi` middleware the host puts on the API that mounts them (`Git.Api.middleware(Authenticated)`), before the engine sees a request. The REST plane, the git wire (a `git` client can only send HTTP Basic), the raw reads, and the GitHub facade all sit behind it, so it is written once, in user land, with whatever knows the host's users.

```ts
export class Authenticated extends HttpApiMiddleware.Service<
  Authenticated,
  { provides: Session; requires: Alchemy.RuntimeContext }
>()("app/Authenticated", { error: Unauthorized }) {}

export const AuthenticatedLive = Layer.effect(
  Authenticated,
  Effect.gen(function* () {
    const auth = yield* Auth;                       // Better Auth
    const registry = yield* Git.RegistryStore;
    return (httpEffect, { endpoint }) =>
      Effect.gen(function* () {
        const user = yield* resolve;                // API key (git, gh) or session cookie
        const { owner } = yield* HttpRouter.params;
        if (user !== undefined && (owner === undefined || owner === user.id)) {
          return yield* Effect.provideService(httpEffect, Session, { user });
        }
        const request = yield* HttpServerRequest;
        if (Git.isRead(endpoint, request) && (yield* isPublic)) {
          return yield* Effect.provideService(httpEffect, Session, { user: user ?? null });
        }
        return unauthorized;                        // 401 + WWW-Authenticate, so git prompts
      });
  }),
);
```

`Git.isRead(endpoint, request)` is the one thing the engine says about its routes: which ones only read (REST and raw reads, the facade's GETs, the fetch advertisement, `git-upload-pack`). The middleware asks `Git.RegistryStore` whether a repository is `public`, a flag the engine stores and reports and attaches no meaning to.

The one decision a middleware cannot make, which refs a push may move, is git's pre-receive hook, and **`Git.Hooks`** is that hook as an optional service. It runs in the Worker, inside the request (so whatever the middleware provided is readable with `Effect.serviceOption`), with the parsed ref updates, before `git push` moves anything, before the REST ref writes, and before a merge. The wire reports the reason per ref; REST answers a typed `403`, `HookRejected`.

```ts
export const ProtectedMain = Layer.succeed(Git.Hooks, {
  preReceive: ({ repo, updates }) =>
    Effect.gen(function* () {
      const session = yield* Effect.serviceOption(Session);
      const user = Option.isSome(session) ? session.value.user : null;
      return updates.flatMap((u) =>
        u.ref === "refs/heads/main" && user?.id !== repo.owner
          ? [{ ref: u.ref, reason: "only the owner moves main" }]
          : [],
      );
    }),
});
```

The engine's own hash route is `Git.InternalApi`, mounted by `Server.layer` beside the host's API and outside its middleware, authenticated with a deploy-time `Alchemy.Random` secret no user holds. There is no tokens table, no admin key, and no principal.

`examples/cloudflare-git-service` is the taught shape: Better Auth as a service, one middleware for sessions and API keys with owners-write / anyone-reads-public, a `Me` route of its own, a React UI that signs in and mints keys. `bun test` deploys it and proves sign-up → create → key → push → anonymous clone → a stranger's push refused → a private repo hidden. The docs' Getting Started is the same host behind one shared secret in twenty lines of middleware.

### The blocks

- **`Git.Server`** — the composed handler for all three planes behind one `fetch`; `Git.Api` + `Git.Handlers` are the pieces it is made of.
- **`Git.BlobStore`** — bulk bytes (packs, clone bundles, oversize objects, push spill) behind a contract of ranged reads, multipart writes, and prefix listing. `Git.BlobStoreR2(bucket)` and `Git.BlobStoreS3()` ship; the S3 store declares its `AWS.S3.Bucket` (or takes yours), is built over the S3 bindings, and the Worker gets a least-privilege identity minted for it — the S3 HTTP bindings gained the Worker-host path the Lambda ones had. One provided layer serves both runtime contexts.
- **`Git.RepoStore`** — the repository block as a service (`getByName(repoId)` → the Repo DO's typed stub), provided by **`Git.ReposDurableObject`**; the seam for hooking `commitPush`/`updateRef`/`mergePull` with a decorating Layer, or for implementing `GitRepoShape` over another store. **`Git.RegistryStore`** (via `Git.RegistryDurableObject` or `Git.RegistryD1`) is its sibling.
- **`Git.Hasher`** — where push verification runs: `HasherInline`, `HasherWorkerLoader`, `HasherLambda`, all under `alchemy/Git/Hasher`.
- **`Git.Hooks`** — above.

### Architecture

- **`GitRepo` DO** (one per repo, ULID-keyed) — refs, object index, loose object bytes, commit graph in SQLite; runs the v0 wire protocol so ref CAS is one `transactionSync`.
- **`GitRegistry` DO** (singleton) — `owner/name → repoId`, uniqueness, listing from denormalized columns (no per-row DO wakes).
- **R2/S3** — compacted packs, clone bundles, oversize objects, spilled push bodies. Objects stored pre-deflated, so serving a pack is concatenation + SHA-1.
- Pure format/protocol layer under `src/Git/Protocol/` (pkt-line, sideband, object codecs, delta apply, pack parse/write), covered by fixture tests against real-git packs; route classes under `Api/`, storage under `Store/`, alarm jobs under `Jobs/`, hashers under `Hasher/`.

### Protocol scope

Smart HTTP v0: clone, incremental fetch (`multi_ack_detailed` + `no-done`), shallow `--depth`, thin-pack push with ofs/ref deltas, full connectivity check, `report-status(-v2)`, atomic pushes, annotated tags with peeled advertisement. Conformance suite drives the real `git` CLI (push, rebase, merge, cherry-pick, force-push, tags) against a deployed stack.

### Pull requests and diffs

Server-side merge-base via a paint walk over the commit graph, tree diffs with identical-subtree pruning, and PR create/list/merge — fast-forward plus trivial three-way merges, with `expectedHeadOid` as a race guard. The GitHub facade maps these onto GitHub's shapes (a merged PR is `state: "closed"` with `merged: true`), so `gh api` and Octokit work against the host unmodified; `/user` is one route away from answering with your own user record (provide a Layer for `Git.GitHubUser` nearer than `Handlers`).

### Public repos (GitHub model)

Repos carry a `public` flag the engine stores and reports; what it grants is the middleware's decision. The tutorial's and the example's middleware serve anonymous reads of public repos (REST reads, raw files, advertisement, `upload-pack`, so `git clone <url>` needs no credentials) and answer `401 + WWW-Authenticate` otherwise. `ListRepos` lists everything the Registry holds; `public: true` narrows it.

### Pushes: the pack never enters the Durable Object

The push body streams through the stateless Worker and the Repo DO only receives *rows* over RPC — `beginPush` / `stagePush` / `readPushBase` / `commitPush`. The Worker spills the body to blob storage in uniform 8 MiB parts as it arrives, verifies it (inflate + SHA-1 + delta resolution) as parts land, and stages promoted rows that point into the spilled object; the DO's part of a push is SQL. Peak DO memory, CPU and egress are independent of push size.

### Scaling the read path, measured

Refs stay in the DO (the only thing needing serialized CAS); immutable content-addressed bytes move to blob storage (compaction into real packfiles + precomputed clone bundles keyed by ref snapshot); the **Worker** streams bundle bytes so no pack byte transits the DO.

**Head snapshot — reads never wake the DO.** The DO republishes a small `{repoId}/head` object after every ref move (push, REST ref write, merge); the Worker answers the advertisement and bundle-covered clones straight from it (`x-git-served-by`) for every caller, since access was decided in front of the route.

| anonymous reads, concurrency 32 | before | after |
| --- | --- | --- |
| raw full clones | 69.4/s · 17.4 MiB/s | **101.0/s · 25.3 MiB/s** |

**Geometric pack merging.** Compaction restores a factor-2 geometric invariant by pure concatenation; superseded packs grace-delete so in-flight ranged reads finish.

| dynamic fetch, repo compacted in 6 increments | before | after |
| --- | --- | --- |
| ms/fetch (2.3 MiB pack) | 730 | **300–334** |

**Clone throughput sweep** (DESIGN.md §22; 36 MiB / 15.6k-object repo, one client):

| path | before | after |
| --- | --- | --- |
| raw bundle, 1 stream | 20.3 MiB/s | **98.4 MiB/s** (0.56 s end to end) |
| sideband bundle — what every `git clone` uses | 15.5 MiB/s | **80.2 MiB/s** (0.69 s) |
| dynamic fetch TTFB (compacted repo) | 27.5 s | **0.7 s** |
| 4 concurrent raw clones, aggregate | 29.7 MiB/s | 93 MiB/s |

**Push-ingest sweep** (DESIGN.md §22.5–22.15; a 15.6k-object, 40 MiB delta-heavy push and a 44 MiB push of whole blobs, real `git push`, GitHub on a fresh repo from the same client for scale):

| layer | parallelism | leaves Cloudflare | loose push | synthetic push | incremental |
| --- | --- | --- | --- | --- | --- |
| GitHub | | | 3.3 s | 4.6 s | 1.2 s |
| start of sweep | | | 63.6 s | — | — |
| `HasherInline` | none | no | 9.9–13.5 s | 8.8 s | 0.24–0.4 s |
| `HasherLambda` | one Lambda per chunk | yes | 6.8–7.9 s | 6.3–6.7 s | 0.27–0.8 s |
| `HasherWorkerLoader` | 4 isolates | no | 7.1–7.8 s | 6.8 s | 0.24 s |
| + delta jobs, resolved pack, lazy flip | | | **5.9 s best, 6.5 s median** | **4.8 s best, 6.4 s median** | 0.24 s |

- On Workers a service-binding subrequest executes on the caller's thread, so in-Worker hashing is single-threaded; `Hasher` is pluggable compute. `HasherLambda(HasherFunction)` binds `AWS.Lambda.InvokeFunction` cross-cloud (the function-scoped Lambda `*Http` bindings provision the assume-role identity for a Worker host, inlined in `BindingHttp.ts`); `HasherWorkerLoader()` hashes on four dynamically loaded Workers, measured to run in parallel with the caller. The Worker bundler gained a `?worker` import for the loader's module.
- Deltas resolve as hasher jobs; a spilled push's resolved blobs become a resolved pack staged as coordinates; a push is live when its `pushes` row commits (`LIVE_OBJECTS`); the commit graph inserts multi-row; connectivity checks only the edges that leave the push.

### Docs

A Git hub under **Source & CI** that reads as one story: Overview and Getting Started, a four-part tutorial (build the server in one file behind a shared-secret middleware, repositories and public reads, your own API with Better Auth and owners, a team in the middleware and `main` protected by a hook), Using your host, Building blocks (one page per block), and Recipes with the scaling model and complete stacks. The Git tab has the official Git mark.

### Also in this PR

- `AWS/Lambda/BindingHttp.ts`: the Worker-host AWS identity is declared inline (the engine de-dupes declarations; the old per-Scope memo store is gone) and distilled operations are called directly per request (`Effect.flatMap` over an `OperationMethod` died at runtime).
- `HttpServerRequest.toWeb` returns the raw source `Request` whenever there is one, so `request.modify({ headers })` was silently dropped when forwarding through a `Fetcher`; the forward now rebuilds only when the headers differ.
- `Cloudflare.R2.Bucket` with `forceDestroy` re-empties and retries on `BucketNotEmpty`, for a writer that lands an object between the sweep and the delete.


ericclemmons:
You mad lad, you’ve done it again! 🙌

werkamsus:
king shit

K-Mistele:
this would be so cool if it were compatible wtih #1127

sam-goodwin:
> this would be so cool if it were compatible wtih #1127 

That's the plan lol.

PR #1191, opened 2026-08-12
fix(alchemy): move @effect/vitest to devDependencies

kristof-siket:
## What this changes

Moves `@effect/vitest` from `dependencies` to `devDependencies` in `packages/alchemy`. It is a test-utility integration, imported only by the `alchemy/Test/Vitest` export, and declaring it as a runtime dependency pulls a test toolchain into every consumer's dependency resolution.

## The problem it fixes

Because `@effect/vitest` sits in `dependencies`, npm resolves its hard peer `vitest` for every consumer of `alchemy`, and `vitest` brings `vite@8`, whether or not the consumer uses the test utilities. For most consumers this only widens resolution. It becomes a real failure for consumers whose repository already pins `esbuild` outside vite 8's peer range (`^0.27 || ^0.28`), for example an app that builds with `esbuild@^0.24`:

```
npm error ERESOLVE could not resolve
npm error Found: esbuild@0.24.2   (the consumer's own build tool)
npm error Could not resolve dependency:
npm error peerOptional esbuild@"^0.27.0 || ^0.28.0" from vite@8.2.1
```

The chain:

```
alchemy (dependencies)
  -> @effect/vitest   (hard peer: vitest, auto-installed by npm)
     -> vitest        (hard dep: vite@8)
        -> vite@8     (peer: esbuild ^0.27 || ^0.28)
           -> conflicts with the consumer's own esbuild pin
```

Whether the conflict actually fails an install depends on npm version and resolver ordering. Measured on one affected repository with identical inputs: npm 10 with an existing lockfile installed fine at one hour and failed at the next; npm 11 with a lockfile present fails deterministically. Consumers with no conflicting pin install fine, which is why this stays invisible until someone's build tool overlaps with vite's range.

## Why the move is safe

`@effect/vitest` is imported in exactly one published file: `packages/alchemy/src/Test/Vitest.ts`, the vitest-specific test-utility module. Anyone importing `alchemy/Test/Vitest` is running vitest tests and already has the test toolchain installed. `vitest` itself is already declared in `devDependencies` here, and that same file imports it; this move makes `@effect/vitest` follow the convention its sibling import already lives under. The dedicated `packages/alchemy-test` package points the same direction: test-framework integration is its own layer.

## Context

Found through the Prisma Composer integration, which brings `alchemy` into end-user repositories (see PR #1061 by @wmadden for the ongoing collaboration); end users pin all kinds of esbuild versions for their own builds, which is how the overlap surfaced. Happy to adjust the fix to however you prefer it done.


packages/alchemy/package.json
```diff
@@ -401,6 +400,7 @@
   },
   "devDependencies": {
     "@aws/durable-execution-sdk-js": "^2.1.0",
+    "@effect/vitest": "catalog:",
```
sam-goodwin, line 403: it needs to be an optional peer (alchemy/Test/Vitest uses it)

kristof-siket: @sam-goodwin thanks a lot! i've added it to the peer deps list as optional!

PR #1195, opened 2026-08-12
feat(cloudflare): named state stores and monotonic version guard

### LINKED ISSUE #912: Cloudflare.state({ workerName }) is documented but unsupported, and bootstrap workerName is only partially propagated
## Summary

The state-store documentation says that a separate Cloudflare state store can
be selected with:

```ts
state: Cloudflare.state({
  workerName: "alchemy-state-store-team-a",
}),
```

However, `Cloudflare.state()` currently accepts no options and always uses
`STATE_STORE_SCRIPT_NAME` (`"alchemy-state-store"`).

The lower-level exported `bootstrap()` function does accept a `workerName`,
but that value is only propagated to parts of the bootstrap lifecycle. The
physical Worker definition, existing-store detection, login URL, credential
cache, and secret lookup continue to use the default state-store identity.

As a result, the documented per-team/dedicated-store configuration does not
appear to be usable end to end.

## Versions checked

- `alchemy@2.0.0-beta.63`
- Upstream `main` at `7c926aa628c70d15610bb469634bda5bf70c769b`

## Documentation

The current documentation says:

> Pass `workerName` to use a separate state store — for example, when you want
> a dedicated store per Cloudflare account or per team.

https://github.com/alchemy-run/alchemy/blob/7c926aa628c70d15610bb469634bda5bf70c769b/website/src/content/docs/state-store/index.mdx#L111-L119

## Compile-time reproduction

```ts

export default Alchemy.Stack(
  "Example",
  {
    state: Cloudflare.state({
      workerName: "alchemy-state-store-team-a",
    }),
  },
  Effect.void,
);
```

The call does not type-check because `state` has a zero-argument signature:

```ts
export const state = () => ...
```

Source:

https://github.com/alchemy-run/alchemy/blob/7c926aa628c70d15610bb469634bda5bf70c769b/packages/alchemy/src/Cloudflare/StateStore/State.ts#L56-L63

## Bootstrap propagation

`bootstrap()` does expose `workerName` and resolves it into `scriptName`:

https://github.com/alchemy-run/alchemy/blob/7c926aa628c70d15610bb469634bda5bf70c769b/packages/alchemy/src/Cloudflare/StateStore/State.ts#L245-L260

That name is used for the bootstrap stage and related messages, but several
subsequent operations still use the default identity:

1. The physical Worker resource is explicitly named
   `STATE_STORE_SCRIPT_NAME`:

   https://github.com/alchemy-run/alchemy/blob/7c926aa628c70d15610bb469634bda5bf70c769b/packages/alchemy/src/Cloudflare/StateStore/Api.ts#L101-L105

2. `isStateStoreServing()` always probes the default Worker URL:

   https://github.com/alchemy-run/alchemy/blob/7c926aa628c70d15610bb469634bda5bf70c769b/packages/alchemy/src/Cloudflare/StateStore/State.ts#L842-L859

3. `loginWithCloudflare()` reads the default auth-token secret, derives the
   default Worker URL, and writes the shared default credentials file:

   https://github.com/alchemy-run/alchemy/blob/7c926aa628c70d15610bb469634bda5bf70c769b/packages/alchemy/src/Cloudflare/StateStore/State.ts#L717-L787

This suggests that calling:

```ts
yield* Cloudflare.bootstrap({
  workerName: "alchemy-state-store-team-a",
});
```

can change the bootstrap stack/stage identity without consistently changing
the state-store service to which later operations attach.

I have not run this reproduction against a disposable Cloudflare account
because bootstrap can create or update account-level state infrastructure.
The compile-time mismatch and incomplete propagation above are visible
directly in the published API and current source.

## Expected behavior

One of the following contracts should be made explicit:

### Option A: named state stores are supported

- `Cloudflare.state({ workerName })` accepts the documented option.
- The same name is used consistently for:
  - bootstrap state
  - physical Worker identity
  - availability and version checks
  - Worker URL derivation
  - login and credential-cache identity
- The intended treatment of the authentication and encryption secrets is
  documented and tested: either per-store identities or explicitly shared
  account-level identities.
- A test proves that the default store and a named store can coexist without
  attaching to or refreshing the wrong backend.

### Option B: one Cloudflare state store per account is supported

- Remove the `Cloudflare.state({ workerName })` example from the documentation.
- Remove or clarify `BootstrapOptions.workerName` so it cannot imply a separate
  usable state service.
- Document the account-wide Worker, credential, authentication, encryption,
  and upgrade failure domain.

## Actual behavior

The documentation advertises Option A, while the public `state()` API and
several runtime paths currently behave like Option B. `bootstrap.workerName`
sits between those two contracts and appears incomplete.

### LINKED ISSUE #1194: Cloudflare state-store version mismatch forces lockstep upgrades and can let older clients downgrade the account singleton
## Motivation

The default Cloudflare state store is an account-wide singleton, but its
contract version is compiled into each Alchemy CLI release. A version mismatch
can redeploy that singleton to the invoking client's version.

That makes a normal application dependency upgrade an account-wide lockstep
migration across otherwise unrelated stacks.

```text
Cloudflare account
└── alchemy-state-store (one Worker + one DO + one credential set)
    ├── Stack A / repository A / alchemy beta.40 -> expects store v5
    └── Stack B / repository B / alchemy beta.71 -> expects store v7

beta.71 invocation: v5 != v7 -> upgrade singleton to v7
beta.40 invocation: v7 != v5 -> redeploy singleton back to v5
```

We have exactly this topology today: one existing beta.40 stack and one new
beta.71 stack in the same Cloudflare account. The live store is currently v5.
We cannot safely introduce the beta.71 stack until every supported beta.40
remote-state entrypoint is removed or upgraded.

## Why this is a blocker

The two stacks have independent codebases, release schedules, infrastructure,
and rollback needs. Nevertheless, the singleton/version behavior means:

1. **They cannot upgrade independently.** The first beta.71 plan/deploy may
   change the state backend used by the beta.40 stack.
2. **An older client can undo the upgrade.** beta.40 treats a different version
   as stale and redeploys the store it was built with.
3. **Canarying the new application is not isolated.** Even a new stage still
   attaches to and may upgrade the same account-level state service.
4. **Rollback is no longer code-only.** After upgrading the singleton, reverting
   one app to beta.40 can mutate shared control-plane infrastructure.
5. **Credential authority becomes cross-stack.** The one state bearer decrypts
   state for every stack using the singleton, so an app-specific operator or CI
   principal necessarily receives account-wide state authority.

This is distinct from #912. That issue tracks the documented-but-incomplete
`Cloudflare.state({ workerName })` isolation escape hatch. This issue is about
the safety and version-negotiation behavior of the default shared singleton,
including accounts that intentionally want one shared store.

### beta.40 expects v5 and redeploys mismatches

- [`STATE_STORE_VERSION = 5`](https://github.com/alchemy-run/alchemy/blob/v2.0.0-beta.40/packages/alchemy/src/Cloudflare/StateStore/Api.ts#L21-L32)
- [`redeployIfStale` is called before trusting existing credentials](https://github.com/alchemy-run/alchemy/blob/v2.0.0-beta.40/packages/alchemy/src/Cloudflare/StateStore/State.ts#L261-L274)
- [A mismatch is logged as stale and redeployed](https://github.com/alchemy-run/alchemy/blob/v2.0.0-beta.40/packages/alchemy/src/Cloudflare/StateStore/State.ts#L643-L677)

### beta.71 expects v7 and upgrades mismatches in place

- [`STATE_STORE_VERSION = 7`](https://github.com/alchemy-run/alchemy/blob/v2.0.0-beta.71/packages/alchemy/src/Cloudflare/StateStore/Api.ts#L23-L34)
- [A mismatch enters the in-place upgrade path](https://github.com/alchemy-run/alchemy/blob/v2.0.0-beta.71/packages/alchemy/src/Cloudflare/StateStore/State.ts#L113-L153)
- [`--yes` accepts the upgrade automatically](https://github.com/alchemy-run/alchemy/blob/v2.0.0-beta.71/packages/alchemy/src/Cloudflare/StateStore/State.ts#L132-L143)

I have not intentionally exercised the downgrade against a disposable account:
the source paths establish the mutation behavior, and doing so would mutate a
state service and its Durable Object. Our live observation is limited to the
public `/version` response reporting v5.

## Workarounds we are considering

| Workaround | Why it is insufficient as the long-term contract |
| --- | --- |
| Upgrade every Alchemy consumer in lockstep, then upgrade v5 -> v7 once | Couples unrelated repositories and turns one beta update into a coordinated account migration. This is our likely immediate workaround. |
| Add an application-side preflight that refuses unless the live store version exactly matches | Prevents accidental mutation, but every consumer must implement and retain the same out-of-band guard. |
| Temporarily force the older stack to local state | Removes the downgrade path, but loses shared locking/state and is unsuitable for team/CI operations. |
| Use `Cloudflare.state({ workerName })` | The documented option is not implemented end to end in beta.71; tracked by #912. |
| Move one stack to a custom/AWS/Postgres state backend | Adds another provider, credential system, and operational failure domain solely to isolate Alchemy itself. |
| Give CI the account-wide state bearer or `Secrets Store Write` | Enables automation but broadens a repository-specific principal into cross-stack state authority. |

## Desired behavior

At minimum, an older client must never automatically replace a newer deployed
state-store contract.

Any one of these designs could resolve the blocker:

1. **Monotonic store upgrades:** clients may upgrade an older compatible store,
   but seeing a newer contract always fails closed before mutation—even with
   `--yes`.
2. **Version negotiation:** the store reports a protocol range/capabilities and
   clients attach when compatible rather than requiring exact equality.
3. **Explicit migrations:** changing the store contract requires a dedicated
   operator command; ordinary `plan`, `dev`, and `deploy` never change shared
   state infrastructure implicitly.
4. **Working named-store isolation:** #912 is completed so independent stacks
   can opt out of the singleton failure domain.

The strongest result would combine monotonic/version-aware behavior for the
default singleton with working named-store isolation.

## Acceptance criteria

- A test runs a newer store with an older client and proves the older client
  fails before uploading or redeploying the state-store Worker.
- `--yes` cannot authorize a downgrade implicitly.
- The CLI clearly distinguishes “older compatible,” “migration required,” and
  “client too old for deployed store.”
- Documentation calls out the account-wide upgrade/rollback boundary.
- If named stores are the recommended isolation mechanism, #912 is fixed in the
  same or an explicitly referenced release.

sam-goodwin:
Closes #912, closes #1194.

### Named state stores (#912)

`Cloudflare.state({ workerName })` now works end to end:

```ts
state: Cloudflare.state({ workerName: "alchemy-state-store-team-a" }),
```

The worker name flows through a `Context.Reference` (`StateStoreWorkerName`, default `alchemy-state-store`) into the state-store stack, so `Api.ts` can stay a static `export default Worker(...)` — its props are now an Effect that resolves the name from context:

```ts
export default Worker(
  "Api",
  Effect.gen(function* () {
    return { name: yield* StateStoreWorkerName, main: import.meta.url, ... };
  }),
  ...
);
```

`deployStateStore` provides the reference on the stack body; inside the deployed bundle the reference falls back to its default, where the name is irrelevant.

Everything the issue listed as only partially propagated is now keyed by the selected store:

- physical Worker identity, serving probe, `/version` check, and workers.dev URL derivation
- login (`readSecretViaEdge` targets the named worker and its suffixed secret)
- credentials cache: `cloudflare-state-store-<name>.json` per store
- secrets: `AlchemyStateStoreToken_<name>` / `AlchemyStateStoreEncryptionKey_<name>` (default store keeps the historical un-suffixed names; logical ids stay fixed since the worker binds by logical id) — so one store's bearer token can't authenticate against, nor its key decrypt, another store's state
- `bootstrap({ workerName })` and `teardownStateStore({ workerName })` only touch their own store's worker, secrets, and credential cache

### Monotonic version guard (#1194)

`checkStateStoreVersion` results are classified as `match | missing | store-older | store-newer`:

- `store-older` keeps the existing upgrade flow (prompt, `--yes`, CI error).
- `store-newer` fails closed with `StateStoreClientTooOldError` **before** any mutation — `--yes` cannot authorize a downgrade. The only downgrade path is the explicit operator command `alchemy cloudflare bootstrap --force`.

### Tests

- `Names.test.ts` — pure coverage of the version classification and per-store identity derivation.
- `NamedStore.test.ts` — live: bootstraps a named store on the testing account, proves state roundtrip, proves the default and named stores can't see each other's state and don't redeploy each other, then tears the named store down and verifies the default store's worker and secrets survive. All green against the real API; the default-store `State.test.ts` suite (18 tests) also passes unchanged.

### DIFF
--- a/packages/alchemy/src/Cloudflare/StateStore/Api.ts
+++ b/packages/alchemy/src/Cloudflare/StateStore/Api.ts
@@ -17,10 +17,11 @@ import {
 import { ReadSecret } from "../SecretsStore/ReadSecret.ts";
 import { ReadSecretBinding } from "../SecretsStore/ReadSecretBinding.ts";
 import { Worker } from "../Workers/Worker.ts";
+import { StateStoreWorkerName } from "./Names.ts";
 import Store from "./Store.ts";
 import { AuthToken } from "./Token.ts";
 
-export const STATE_STORE_SCRIPT_NAME = "alchemy-state-store" as const;
+export { STATE_STORE_SCRIPT_NAME } from "./Names.ts";
 
 /**
  * Version of the deployed Cloudflare State Store worker contract.
@@ -100,15 +101,22 @@ export const STATE_STORE_VERSION = 7 as const;
  */
 export default Worker(
   "Api",
-  {
-    name: STATE_STORE_SCRIPT_NAME,
-    main: import.meta.url,
-    workersDev: true,
-    compatibility: {
-      flags: ["nodejs_compat"],
-      date: "2026-03-17",
-    },
-  },
+  // Effect-valued props: the worker's physical name resolves from the
+  // `StateStoreWorkerName` reference so `Cloudflare.state({ workerName })`
+  // and `bootstrap({ workerName })` can deploy independent named stores.
+  // Un-provided (including inside the deployed bundle, where the name is
+  // irrelevant) it defaults to `alchemy-state-store`.
+  Effect.gen(function* () {
+    return {
+      name: yield* StateStoreWorkerName,
+      main: import.meta.url,
+      workersDev: true,
+      compatibility: {
+        flags: ["nodejs_compat"] as string[],
+        date: "2026-03-17",
+      },
+    };
+  }),
   Effect.gen(function* () {
     const remoteSecret = yield* ReadSecret(AuthToken);
     const store = yield* Store;
--- a/packages/alchemy/src/Cloudflare/StateStore/CredentialsFile.ts
+++ b/packages/alchemy/src/Cloudflare/StateStore/CredentialsFile.ts
@@ -13,6 +13,17 @@ import type { HttpStateStoreCredentials } from "../../State/HttpStateStore.ts";
 export const CREDENTIALS_FILE = "cloudflare-state-store";
 
 /**
+ * Per-store credentials cache file. The default store keeps the
+ * historical {@link CREDENTIALS_FILE} name; named stores (see
+ * `Cloudflare.state({ workerName })`) each get their own file so two
+ * stores never share a cached `{ url, authToken }` pair.
+ */
+export const stateStoreCredentialsFile = (workerName: string) =>
+  workerName === "alchemy-state-store"
+    ? CREDENTIALS_FILE
+    : `${CREDENTIALS_FILE}-${workerName}`;
+
+/**
  * On-disk shape of the cached Cloudflare state-store credentials.
  *
  * Extends the generic {@link HttpStateStoreCredentials} (`{ url, authToken }`)
--- a/packages/alchemy/src/Cloudflare/StateStore/Names.ts
+++ b/packages/alchemy/src/Cloudflare/StateStore/Names.ts
@@ -1,0 +1,57 @@
+import * as Context from "effect/Context";
+
+/**
+ * Default name of the state-store Worker deployed by
+ * `Cloudflare.state()` / `alchemy bootstrap cloudflare`.
+ */
+export const STATE_STORE_SCRIPT_NAME = "alchemy-state-store" as const;
+
+/**
+ * The physical name of the state-store Worker for the current
+ * deploy, as a `Context.Reference` so it can flow from
+ * `Cloudflare.state({ workerName })` (or `bootstrap({ workerName })`)
+ * into the state-store stack — the Worker resource in `Api.ts` and
+ * the Secrets Store secrets in `Token.ts` — without those modules
+ * taking constructor parameters (their default exports must stay
+ * statically analyzable for the worker bundler).
+ *
+ * The default keeps every un-parameterized consumer (including the
+ * re-evaluated module inside the deployed worker bundle, where the
+ * name is irrelevant) on the historical `alchemy-state-store`
+ * identity.
+ */
+export const StateStoreWorkerName = Context.Reference<string>(
+  "Alchemy/Cloudflare/StateStoreWorkerName",
+  { defaultValue: () => STATE_STORE_SCRIPT_NAME },
+);
+
+/**
+ * Cloudflare Secrets Store secret names only allow `[A-Za-z0-9_]`.
+ * Worker names allow dashes, so fold anything else to `_` when a
+ * worker name is embedded in a secret name.
+ */
+const sanitizeSecretName = (name: string) =>
+  name.replace(/[^A-Za-z0-9_]/g, "_");
+
+/**
+ * Per-store name of the bearer-token secret in the account Secrets
+ * Store. The default store keeps the historical un-suffixed name so
+ * existing deployments keep resolving their secret; named stores get
+ * a suffixed name so each store has its own credential authority —
+ * one store's bearer token must never authenticate against another.
+ */
+export const authTokenSecretName = (workerName: string) =>
+  workerName === STATE_STORE_SCRIPT_NAME
+    ? "AlchemyStateStoreToken"
+    : `AlchemyStateStoreToken_${sanitizeSecretName(workerName)}`;
+
+/**
+ * Per-store name of the state-encryption-key secret. Same suffixing
+ * rule as {@link authTokenSecretName}: each store encrypts its state
+ * with its own key, so credentials for one store cannot decrypt
+ * another store's state.
+ */
+export const encryptionKeySecretName = (workerName: string) =>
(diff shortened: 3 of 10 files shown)


luxass:
Hey Sam, the docs sent me here while I was setting up a named state store. The pkg.ing artifact has expired though. 

If you pick this up again, could you refresh the pkg.ing artifact? 😊

samcxps:
Bumping this PR/issue - would love for this to be merged in :)

PR #1197, opened 2026-08-12
fix: better platform runtime detection and error messages

BlankParticle:


packages/alchemy/src/AWS/AppRunner/Service.ts
```diff
@@ -998,7 +998,7 @@ export const ServiceProvider = () =>
               realMain,
               virtualEntryPlugin(
                 (importPath) => `
-import { BunServices } from "@effect/platform-bun";
+import { BunServices } from "alchemy/Util/PlatformServices";
```
sam-goodwin: This doesn't need to change, this is the generated entrypoint

BlankParticle:
ok the first pass was too much, now its only fixing the places that matters

sam-goodwin (approved):

PR #1205, opened 2026-08-12
feat(RDS): support parameters on DBParameterGroup

skusez:
## Problem

`AWS.RDS.DBParameterGroup` can create a group but not set anything in it — `DBParameterGroupProps` exposes only `dbParameterGroupName`, `family`, `description` and `tags`. Engine settings have to be applied out of band with the CLI or console, so they are invisible to the plan and drift silently.

## Change

Adds a `parameters?: Record<string, string>` prop, reconciled against the values RDS reports as user-modified:

- entries added or changed are written with `modifyDBParameterGroup`
- entries dropped from props are reset to the engine default with `resetDBParameterGroup`
- the diff baseline is live cloud state, not prior props, so the group converges after a console edit and on adoption

`parameters` is also surfaced as an attribute, hydrated from the API in `read` (unlike `tags`, `describeDBParameters` does return values). `list` returns `{}` for the same reason it returns `{}` for tags — a describe per group would make enumeration O(groups) round trips.

## Consistency with Neptune

`AWS/Neptune/DBParameterGroup.ts` already implements this feature. This is deliberately a transcription of it rather than a new design, so the two read the same:

- the `readParameters` / `readUserParameters` split, and `toUserParameterRecord`
- 20-parameters-per-call chunking on both modify and reset
- `retryWhileParameterGroupBusy` for `InvalidDBParameterGroupStateFault`, which back-to-back calls hit while the group has pending changes
- static parameters applied with `pending-reboot`, dynamic with `immediate`
- `ResetAllParameters: false`, resetting only the names this resource dropped

If a shared helper for the two is wanted, happy to follow up — I kept them independent here to avoid widening the diff.

## Notes

- No change to `diff`: a `parameters` change falls through to `reconcile` rather than forcing a replacement. Name/family/description still replace.
- Deploy credentials now need `rds:ModifyDBParameterGroup`, `rds:DescribeDBParameters` and `rds:ResetDBParameterGroup`.
- `DBClusterParameterGroup` has the same gap and the same API shape. Left out of this PR to keep it reviewable — happy to mirror it if you'd like it in the same change.

## Testing

- `bun tsc -b` — no new errors. Note `main` currently has a pre-existing failure in `packages/alchemy/src/AWS/DynamoDB/Table.ts:779` (`"UnsupportedOperation"` missing from the generated error-tag union); identical with and without this change.
- `bun run format` clean.
- Adds a test to `packages/alchemy/test/AWS/RDS/DBParameterGroup.test.ts` covering write → update → reset, asserting against live `Source=user` state. **I have not been able to run it** — it needs a live AWS account, so it is unexercised and would benefit from a maintainer run.

### DIFF
--- a/packages/alchemy/src/AWS/RDS/DBParameterGroup.ts
+++ b/packages/alchemy/src/AWS/RDS/DBParameterGroup.ts
@@ -2,7 +2,7 @@ import * as rds from "@distilled.cloud/aws/rds";
 import * as Effect from "effect/Effect";
 import * as Schedule from "effect/Schedule";
 import * as Stream from "effect/Stream";
-import { isResolved } from "../../Diff.ts";
+import { deepEqual, isResolved } from "../../Diff.ts";
 import { createPhysicalName } from "../../PhysicalName.ts";
 import * as Provider from "../../Provider.ts";
 import { Resource } from "../../Resource.ts";
@@ -24,7 +24,17 @@ export interface DBParameterGroupProps {
   description?: string;
   /**
    * Instance parameter overrides, e.g. `{ time_zone: "Australia/Sydney" }`.
-   * Parameters removed from this map are reset to their engine defaults.
+   *
+   * When set, this map is the group's entire user-owned surface: entries are
+   * written, and any parameter RDS reports as user-set but absent here is
+   * reset to its engine default. `{}` therefore resets every override, while
+   * OMITTING the prop leaves parameters alone entirely — which is what makes
+   * it safe to adopt a group that was tuned elsewhere.
+   *
+   * Values must be in the form RDS reports back (it canonicalises some — a
+   * boolean set as `ON` reads back as `1`), or the two never compare equal
+   * and every deploy re-issues the modify.
+   *
    * Static parameters are applied with `pending-reboot`, dynamic parameters
    * with `immediate`.
    */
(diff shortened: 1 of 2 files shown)


skusez:
Pushed two fixes after a review pass, plus one issue I'm flagging rather than fixing here.

**Omitted `parameters` was destructive.** `news.parameters ?? {}` collapsed "not managed here" into "reset everything". Since `reconcile` also runs on the adopt path, adopting a group tuned elsewhere would have wiped it on first deploy — and worse, every existing user of this resource is in that position by construction, because there was no way to set parameters through it and they must have set them out of band. An omitted map now skips the sync entirely; `{}` still means reset everything. Documented on the prop.

**Out-of-band drift never converged.** The sync diffs live state, but it only ever ran when *props* changed — the engine schedules `reconcile` from a props comparison, so a console edit to a managed parameter was never remediated. `diff` now compares the desired map against `output.parameters`.

Also: restored the `@example` for `parameters` (lost in a rebase on my end), documented that values must be in RDS's canonical form or the comparison never settles, and tightened the `byName` key type.

---

**Flagging, not fixing:** `diff` returns `{ action: "replace" }` for name/family/description without `deleteFirst: true`. For a user-supplied `dbParameterGroupName` that's create-first onto a name that already exists — `DBParameterGroupAlreadyExistsFault` gets swallowed at the re-read, adopting the *old* group, and GC then deletes the replacement by the same name. This predates the PR and applies equally to `Neptune/DBParameterGroup.ts`, so I've left it alone rather than widen the diff — but parameters make the failure messier, since the new family's parameters get applied to the old-family group. Happy to fix here or separately, whichever you prefer.

`bun tsc -b` still clean (the `DynamoDB/Table.ts:779` failure is pre-existing on `main`).

PR #1206, opened 2026-08-13
feat!: Effectful web frameworks

sam-goodwin:
Website constructs on both clouds accept the Worker-style third Effect argument: one Worker / one Lambda serves the frontend framework **and** an effect-native backend, with typed capability bindings collected at plan time — replacing the two-worker service-binding pattern.

```ts
// src/backend.ts — the only alchemy file a user writes
import * as KV from "alchemy/Cloudflare/KV";
import * as Website from "alchemy/Cloudflare/Website";

export const Visits = KV.Namespace("Visits");

export default class Backend extends Website.Vite<Backend>()(
  "Site",
  { main: import.meta.url },
  Effect.gen(function* () {
    const kv = yield* KV.ReadWriteNamespace(yield* Visits); // binding collected at plan
    return {
      visits: () => kv.get("count").pipe(Effect.map((v) => Number(v ?? 0))),
      bump: Effect.fn(function* () { /* increment */ }),
    };
  }).pipe(Effect.provide(KV.ReadWriteNamespaceBinding)),
) {}
```

**Zero framework-file edits on every framework, both clouds** — delivery is auto-injected (OpenNext wrapper override, nitro `handlers` injection, vite virtual entry, adapter artifact wraps), with explicit `alchemy/Serve` mounts remaining as documented escape hatches that auto-injection detects and stands down for.

### The trust model: schema-less RPC is for trusted callers only

A backend's methods are **not** a public API. There is no reserved public RPC path, no anonymous wire, and no browser RPC client. The only ways in from the outside are surfaces you explicitly create:

- **routes you claim** with a `fetch` handler (`server.routes`),
- **a schema you mount** — an effect `HttpApi` on the fetch handler, validated on both ends,
- **a framework transport you write** — a server function, server action, form action, or API route.

Inside those trusted server surfaces, calls reach the backend through the **value form**: `createClient(Backend)` dispatches methods **in-process** (no HTTP — the server code runs in the same Worker/Lambda as the backend class). Internal service-to-service stays native: Cloudflare Worker→Worker over **service-binding JS-RPC** (`env.BACKEND.bump(1)` — never public HTTP), AWS Lambda→Lambda over **invoke bindings**.

```ts
// SSR / server code only — in-process, typed, cookie-aware
const backend = createClient(Backend, { headers: getRequestHeaders });
const n = await backend.bump();
```

`createClient` snapshots headers thunks synchronously per call (concurrent SSR requests never cross identities) and, on AWS classes, builds the Lambda/Node runtime through the class's serve shell (`ServeShell.runtime`). The `"browser"` condition of `alchemy/Client` throws with guidance — browser code brings its own schema. A method-only backend (no `fetch`) compiles **zero** public routing rules.

### Architecture

- **Collect-only mode** (both clouds): the impl's init Effect runs at plan time in the engine — `host.bind`, Config interception, env merge — while the framework-built bundle ships byte-for-byte (`runtimeDelivery: "external" | "wrapper"` stamp). AWS: `impl + bundle:false` on `Lambda.Function` feeds `attachBindings` (env + IAM).
- **`alchemy/Serve` runtime bridge** (+ top-level adapters `alchemy/Next`, `alchemy/Nitro`, `alchemy/Astro`, `alchemy/SvelteKit`, and the CF entry helper `alchemy/Serve/Worker`): lazy WeakMap-memoized isolate-scope layer build, per-request Scope, env-resolution ladder, prerender/SSG guard. World-selected platform: workerd worlds never touch platform-node (foreign bundlers externalize `node:*` — a static import 500'd every SSR dispatch); Node worlds load it dynamically. Engine-graph leaf tags (`StackTag`, `CloudflareEnvironmentTag`) keep `node:os`/`node:net`/OAuth `node:http` out of framework server bundles. Users import none of this in the zero-setup flow — only `alchemy/Client` in server code.
- **Routing — strict route ownership**: `server.routes` decides who serves each path when the backend has a `fetch`. Inside the routes the effect fetch is authoritative (its 404s are real 404s); outside, the framework serves. Exclusion globs (`["/api/*", "!/api/auth/*"]`) carve paths out. Compiles to `assets.runWorkerFirst` (Cloudflare) and the CloudFront KV-router `serverRoutes` (AWS). Generated CloudFront Function code is compacted under the service's 10KB cap, with a worst-case size pin in tests.
- **`AWS.Website.Vite` — full parity with `Cloudflare.Website.Vite`**: the SPA arm is a convention layer over StaticSite (`vite build`/`vite dev` resolved from the project's OWN `node_modules/.bin`, `dist/`, spa fallback); `ssr: true` rides the FrameworkSite topology through a new generic `@alchemy.run/frontend-frameworks/vite` integration — the project's own `vite.config.*` builds programmatically (Environment API; client + ssr environments), the SSR entry's fetch export wraps into a streaming Lambda entry, and the effectful backend composes ahead of it (dev = vite's own dev server with a routes-scoped effect middleware). This serves the whole vite-plugin framework camp — TanStack Start (shipped, `examples/aws-tanstack`), SolidStart, Octane, Waku — with one integration; the orchestrator frameworks (Nuxt/nitro, Astro adapters, Next/turbopack) keep their per-framework seams by necessity.
- **Non-fetch handlers** (`queue`/`scheduled`/DO exports): v1 on all Cloudflare paths via the generated entries; on Next/Nuxt AWS they deploy as a sibling effect Lambda from the same site module.
- **Wiring handshake**: external-delivery deploys scan the built bundle for the serve sentinel — impl wired but entry missing is a loud typed deploy error with a per-framework fix-it; auto-injection stands down when an explicit mount is detected (or `server: { takeover: false }`), on deploy AND in dev.
- **Dev**: effect code rides each framework's own dev graph (vite module runner in workerd with HMR; kit/nitro/next dev in the sidecar) against the same `dev:` local simulators; AWS dev lowers the collected env into the sidecar process.

### Per-framework DX

All examples run **one identical demo** — a server-rendered visits counter, a bump button, and an async queue leg (enqueue → consumer on the same backend class → UI polls until it catches up). KV+Queues on Cloudflare, DynamoDB+SQS on AWS. UIs are real `.tsx`/SFC components (Tailwind + shadcn-patterned `button`/`card`/`input`, no `<script>` tags anywhere). The browser never imports the backend; each framework's own transport is the public API.

| Framework | Public API (the wire) | Server → backend | Client UI |
|---|---|---|---|
| TanStack Start (CF + AWS) | `createServerFn` server functions | value form in server fns | React + shadcn |
| Next.js (CF + AWS) | `"use server"` actions; RSC reads direct | value form in actions/RSC | React + shadcn |
| Astro (CF + AWS) | Astro Actions (zod-validated) | value form in actions + frontmatter | React island + shadcn |
| SvelteKit (CF + AWS) | form actions + a JSON GET route | value form in `+page.server.ts` | Svelte + shadcn-svelte pattern |
| Nuxt (CF + AWS) | nitro `server/api/*` routes | value form in nitro routes | Vue + shadcn-vue pattern |
| Vite SPA (`aws-vite`, `AWS.Website.Vite`) | **effect `HttpApi` schema** | `HttpApiBuilder` on the backend fetch | React + shadcn, `HttpApiClient` from the schema import |
| Waku (CF + AWS) | plain framework showcase (effectful arm deferred) | — | React + shadcn |

<details>
<summary><b>TanStack Start — server functions</b> — <code>examples/cloudflare-tanstack</code></summary>

```ts
// src/server/visits.ts — the wire IS TanStack's server-fn transport
import { createServerFn } from "@tanstack/react-start";
import { backend } from "../lib/backend"; // value-form client (server-only module)

export const getVisits = createServerFn().handler(() => backend.visits());
export const bumpVisits = createServerFn({ method: "POST" }).handler(() => backend.bump());
export const enqueueJob = createServerFn({ method: "POST" })
  .inputValidator((message: string) => message)
  .handler(({ data }) => backend.enqueue(data));
```

```tsx
// src/components/visits-card.tsx — client component, zero backend imports
import { Button } from "./ui/button";
import { bumpVisits } from "../server/visits";

export function VisitsCard({ initial }: { initial: number }) {
  const [visits, setVisits] = useState(initial);
  return <Button onClick={async () => setVisits(await bumpVisits())}>Bump visits</Button>;
}
```
</details>

<details>
<summary><b>Next.js — server components + actions</b> — <code>examples/{cloudflare,aws}-website-nextjs</code></summary>

```ts
// app/actions.ts
"use server";
import { createClient } from "alchemy/Client";
import Backend from "./backend.ts";

const backend = createClient(Backend);
export async function bumpVisits() { return backend.bump(); }
export async function getProcessed() { return backend.processed(); }
```

```tsx
// app/page.tsx — server component: direct in-process read, no HTTP at all
export default async function Home() {
  const backend = createClient(Backend);
  return <VisitsCard initial={await backend.visits()} processed={await backend.processed()} />;
}
```
</details>

<details>
<summary><b>Astro — Actions + React island</b> — <code>examples/{cloudflare,aws}-website-astro</code></summary>

```ts
// src/actions/index.ts — typed, zod-validated action transport
import { defineAction } from "astro:actions";
import { z } from "astro:schema";
const backend = createClient(Backend);

export const server = {
  bump: defineAction({ handler: () => backend.bump() }),
  enqueue: defineAction({ input: z.object({ message: z.string() }), handler: ({ message }) => backend.enqueue(message) }),
};
```

```astro
---
// src/pages/index.astro — SSR read in frontmatter; the island replaces script tags
const visits = await createClient(Backend).visits();
---
<VisitsCard client:load initial={visits} />
```
</details>

<details>
<summary><b>SvelteKit — form actions</b> — <code>examples/{cloudflare,aws}-website-sveltekit</code></summary>

```ts
// src/routes/+page.server.ts
const backend = createClient(Backend);

export const load = async () => ({ visits: await backend.visits(), processed: await backend.processed() });
export const actions = {
  bump: async () => ({ visits: await backend.bump() }),
  enqueue: async ({ request }) => backend.enqueue(String((await request.formData()).get("message"))),
};
```

`+page.svelte` posts through `use:enhance` forms (CSRF-protected, works without JS); the queue section polls a `GET /api/processed` `+server.ts` route.
</details>

<details>
<summary><b>Nuxt — nitro routes</b> — <code>examples/{cloudflare,aws}-website-nuxt</code></summary>

```ts
// server/api/visits.get.ts (siblings: visits.post.ts, jobs.get.ts, jobs.post.ts)
import { createClient } from "alchemy/Client";
import Backend from "../backend.ts";

const backend = createClient(Backend);
export default defineEventHandler(() => backend.visits());
```

The value form lives only in `server/` (nitro's server graph), so the CF vue-server-graph limitation never applies; the CF backend drops its fetch claim entirely — nitro routes own the API surface. `app/pages/index.vue` uses `useFetch`/`$fetch` with shadcn-vue-patterned components.
</details>

<details>
<summary><b>Vite SPA — the HttpApi flagship</b> — <code>examples/aws-vite</code></summary>

```ts
// src/api.ts — THE CONTRACT. No backend imports; browser-safe by construction.
export class SiteApi extends HttpApi.make("SiteApi").add(
  HttpApiGroup.make("visits")
    .add(HttpApiEndpoint.get("get", "/api/visits").addSuccess(Schema.Number))
    .add(HttpApiEndpoint.post("bump", "/api/visits").addSuccess(Schema.Number))
    .add(HttpApiEndpoint.post("enqueue", "/api/jobs")
      .setPayload(Schema.Struct({ message: Schema.String }))
      .addSuccess(Schema.Struct({ queued: Schema.Boolean }))),
) {}
```

```tsx
// browser — typed client built FROM THE SCHEMA IMPORT alone
const client = await HttpApiClient.make(SiteApi);
await client.visits.bump(); // validated on both ends by the same schema
```

The effectful `AWS.Website.StaticSite` backend mounts `HttpApiBuilder.group` handlers over DynamoDB + SQS bindings on its `fetch`; bad payloads 400 at the schema boundary before any effect code runs.
</details>

### Tests

Live + `.local.test.ts` coverage per framework per cloud: plan suites pin every arm on both clouds (collect-only stamping, serve-shell regression across all seven AWS class arms, delivery-tier stamps, stand-down on explicit mounts and `takeover: false`, the CloudFront Function 10KB size pin); dev roundtrips run against real local simulators; live deploys assert SSR, framework-transport wires (Start server-fn ids, Next action ids scraped from client chunks, Astro `/_actions`, form actions, nitro routes, HttpApi paths — plus a pin that `POST /api/__rpc/*` is dead), queue consumption, and capability writes verified out-of-band via distilled. Unit suites cover the serve bridge, entry wrappers, JS-RPC proxy (positional args, typed-failure envelopes, defect separation), `createClient` (concurrent header-thunk identity, the AWS shell seam, browser-condition guidance throw), the nitro adapter's three body sources, and the Next dev dispatch. Example integ/dev tests drive the canonical demo end to end on both clouds.

### Known limitations

- **CF Nuxt value form in the vue server graph**: nuxt's vite-builder resolves the alchemy graph node-flavored into the vue SSR build, which breaks on workerd. Moot for the shipped pattern (nitro routes keep the value form in `server/`), but a hand-rolled value-form call inside a `.vue` SSR branch still hits it.
- **CF Waku effectful arm**: typed but not yet delivered (no effect descriptor threading; JSDoc defers it).

### Deferred (per design)

- **Cross-service call matrix**: CF→CF service-binding JS-RPC ships now; AWS→AWS typed method calls over Lambda Invoke (an invoke-RPC envelope — IAM-private, no HTTP); cross-cloud (Lambda→Worker, Worker→Lambda) as authenticated HTTP with capability-minted credentials bound to both sides (shared secret v1; CF Access service tokens / mTLS above it; SigV4 to an `AWS_IAM` Function URL in reverse). Never anonymous, never public-by-default.
- Waku + Octane effectful arms, `Serve.exports` for user-authored custom entries.

Design doc, per-framework research, and the amendment history are in the session scratchpad (`DESIGN.md`, `RPC-PIVOT.md`).


examples/aws-vite/src/backend.ts
```diff
@@ -0,0 +1,205 @@
...
+ * `spa: true`.
+ *
+ * The program's non-`fetch` methods are the trusted-caller RPC surface
+ * (in-process `createClient(Site)` from server code, AWS invoke-style
+ * bindings from sibling functions). The static frontend is untrusted — it
+ * talks ONLY through the schema-validated HttpApi mounted on `fetch`.
+ */
+export default class Site extends Vite<Site>()(
+  "Site",
+  {
```
sam-goodwin: I see aws-vite backend.ts uses process.env. Never do this. It should be using effect/Config. Vite's props can be an Effect, so resolve the config in there.

bad:
```ts
const domainName = process.env.WEBSITE_DOMAIN;
const hostedZoneId = process.env.WEBSITE_ZONE_ID;
const domainAliases = process.env.WEBSITE_ALIASES?.split(",")
  .map((part) => part.trim())
  .filter(Boolean);
```

good:
```ts
export default class Site extends Vite<Site>()(
  "Site",
  Effect.gen(function*() {
    const domainName = yield* Config.string("WEBSITE_DOMAIN") // etc.
```

examples/aws-vite/src/backend.ts
```diff
@@ -0,0 +1,205 @@
...
+      ),
+      // Trusted-caller RPC methods (in-process value form, invoke bindings).
+      visits,
+      bump,
+      enqueue,
+      processed,
+    };
+  }).pipe(
+    Effect.provide([S3.GetObjectHttp, S3.PutObjectHttp, SQS.SendMessageHttp]),
+    Effect.provide(QueueEventSource),
```
sam-goodwin: Never do subsequent Effect.provide. Build a single layer with Layer.* and provide it once

examples/aws-website-astro/src/backend.ts
```diff
@@ -0,0 +1,151 @@
...
+          return { count: Number(count ?? "0"), last: last ?? null };
+        }),
+    };
+  }).pipe(
+    Effect.provide([
+      DynamoDB.GetItemHttp,
+      DynamoDB.PutItemHttp,
+      SQS.SendMessageHttp,
+    ]),
+    Effect.provide(QueueEventSource),
```
sam-goodwin: No double Effect.provide

examples/aws-website-nuxt/server/api/hello.ts
sam-goodwin, line 1: Can we delete this file?

examples/aws-website-nuxt/server/api/jobs.get.ts
```diff
@@ -0,0 +1,8 @@
+// GET /api/jobs — the queue consumer's async state, read back from KV.
+import { createClient } from "alchemy/Client";
+import Backend from "../backend.ts";
+
+export default defineEventHandler(async (event) => {
+  const backend = createClient(Backend, { headers: event.headers });
```
sam-goodwin: Why would we ned headers? This is just for calling in-process RPC methods which don't require headers.

If they are going to call the fetch, they can do that with `backend.fetch(..)`.

Also, this should be created once in the module scope, not created per request.

examples/aws-website-nuxt/server/backend.ts
```diff
@@ -0,0 +1,139 @@
...
+          return { count: Number(count ?? "0"), last: last ?? null };
+        }),
+    };
+  }).pipe(
+    Effect.provide([
+      DynamoDB.GetItemHttp,
+      DynamoDB.PutItemHttp,
+      SQS.SendMessageHttp,
+    ]),
+    Effect.provide(QueueEventSource),
```
sam-goodwin, line 213: No double Effect.provide!

examples/aws-website-waku/README.md
sam-goodwin, line 1: why does examples/aws-website-waku not have a backend.ts ? I don't see an effectful website in this

examples/cloudflare-tanstack/src/lib/backend.ts
```diff
@@ -0,0 +1,14 @@
...
+// code reaches the backend exclusively through the TanStack server
+// functions in src/server/visits.ts. `getRequestHeaders` is a thunk
+// resolved fresh per call, so this module-scope client stays
+// per-request-correct (methods can read the caller's cookies via
+// `HttpServerRequest`).
+import { getRequestHeaders } from "@tanstack/react-start/server";
+import { createClient, type RpcClient } from "alchemy/Client";
+import Backend from "../backend.ts";
+
+export const backend: RpcClient<typeof Backend> = createClient(Backend, {
```
sam-goodwin: Why does this need an explicit type? we can infer it

examples/cloudflare-tanstack/src/server.ts
```diff
@@ -0,0 +1,53 @@
...
+    ) {
+      return new Response("forbidden", { status: 403 });
+    }
+
+    // Effect API first (undefined = "not mine"), then the framework
+    // serves pages, assets, and TanStack server functions.
+    return (
+      (await site.fetch(request, env, ctx)) ??
+      framework.fetch(request, env, ctx)
+    );
```
sam-goodwin: How is "not mine" detected? we don't support returning undefined from `fetch`, do we?

examples/cloudflare-website-astro/src/actions/index.ts
```diff
@@ -0,0 +1,24 @@
...
+import { createClient } from "alchemy/Client";
+import { z } from "astro/zod";
+import Backend from "../backend.ts";
+
+const backend = (ctx: ActionAPIContext) =>
+  createClient(Backend, { headers: ctx.request.headers });
+
+export const server = {
+  bump: defineAction({
+    handler: (_input, ctx) => backend(ctx).bump(),
```
sam-goodwin: Create it once per module (when imported). Headers aren't needed for rpc. They can be passed explicitly if the user is integrating with an effectful `fetch`

examples/cloudflare-website-nextjs/app/api/hello/route.ts
sam-goodwin, line 1: Can we get rid of this? We should only have api endpoints connecting to the effectful backend

examples/cloudflare-website-nuxt/server/api/hello.ts
sam-goodwin, line 1: get rid of hello world apis. replace with ones that call the effectful backends

examples/cloudflare-website-astro/src/backend.ts
```diff
@@ -0,0 +1,119 @@
...
+      processed: Effect.fn(function* () {
+          const count = yield* visits.get("processed-count");
+          const last = yield* visits.get("processed-last");
+          return { count: Number(count ?? "0"), last: last ?? null };
+        }, Effect.orDie),
+    };
+  }).pipe(
+    Effect.provide(KV.ReadWriteNamespaceBinding),
+    Effect.provide(Queues.WriteQueueBinding),
+    Effect.provide(Queues.EventSourceLive),
```
sam-goodwin: No triple Effect.provide. Use Layer.* to build one layer

examples/monorepo-frontends/alchemy.run.ts
```diff
@@ -0,0 +1,35 @@
...
+    state: AWS.state(),
+  },
+  Effect.gen(function* () {
+    // Each Website class (declared in its package's src/backend.ts with
+    // its Effect program) is itself the construct — yielding it deploys
+    // the site from its nested package root.
+    const nextjs = yield* NextjsSite;
+    const nuxt = yield* NuxtSite;
+    const astro = yield* AstroSite;
+    const sveltekit = yield* SvelteKitSite;
```
sam-goodwin, line 28: This does not look complete. We need to test all frameworks

examples/monorepo-frontends/alchemy.run.ts
```diff
@@ -0,0 +1,35 @@
+// ONE stack deploying an effectful AWS Website per framework, each built
```
sam-goodwin, line 1: Should we be testing cloudflare too?

packages/alchemy/src/AWS/Lambda/Function.ts
```diff
@@ -171,11 +173,92 @@ export interface FunctionUrlConfig {
...
+  verify?: boolean;
+  /**
+   * Whether alchemy may take over the framework's server entry to deliver
+   * the runtime half automatically (the auto-inject tier). Set `false` to
+   * force the explicit tier: the framework artifact deploys byte-for-byte
+   * and you mount the program yourself via `alchemy/Serve`.
+   * @default the framework integration's tier (auto where supported)
+   */
+  takeover?: boolean;
+}
```
sam-goodwin, line 193: Do we still need this if we're now forcing the user to integrate expliitly with the effectful backend instead of implicitly? We don't need to know the routes. They can encode that in their entrypoint fetch override

PR #1207, opened 2026-08-13
feat(kubernetes): direct image strategy for Deployment (skip the managed-registry mirror)

microagi-andy:
## What

Adds `imageStrategy?: "mirror" | "direct"` to `Kubernetes.Deployment`'s `ImageDeploymentProps`:

```ts
yield* Kubernetes.Deployment("api", {
  cluster,
  image: "europe-west4-docker.pkg.dev/my-project/apps/api@sha256:…",
  imageStrategy: "direct",
});
```

- `"mirror"` (**default**, unchanged behaviour) — the image is copied into the cluster adapter's managed registry (ECR on EKS) and the pod pulls the mirrored copy.
- `"direct"` — the supplied reference is used verbatim; the registry adapter is never invoked.

## Why

`Kubernetes.Deployment` currently routes *every* image source through the platform's managed registry whenever the adapter has one, including a pre-built `image` reference. Today the pass-through path only exists for registry-*less* clusters (`resolveWorkloadImage`'s `adapter.registry === undefined` branch).

That means an image that already lives in a registry the cluster can pull from — a repo-wide Artifact Registry, an internal Harbor, a public base image, an image built by a CI pipeline that already pushed it — gets mirrored purely to satisfy the API. The cost is a managed repository per deployment, a duplicate copy of every layer, and a pull/push of the whole image on each deploy. For large ML/CUDA images that is tens of gigabytes of pointless traffic and storage.

`"direct"` lets the caller say "the cluster can already pull this", and the deployment just references it.

## How it stays backwards-compatible

- The prop is optional and defaults to `"mirror"`, so the resolved image, hash, and persisted registry state of every existing deployment are byte-identical.
- It is only accepted on `ImageDeploymentProps`. `main` / `context` sources are unchanged and still require a managed registry — a `@ts-expect-error` test pins that.
- The registry-less-cluster pass-through path is untouched.

### Retiring the old repository on a mirror → direct switch

Flipping an existing deployment to `"direct"` orphans the managed repository it used to own. `resolveWorkloadImage` therefore returns the previous registry state as `cleanupState`, and the reconciler deletes it **after** `waitForDeploymentRollout` confirms the new pods are up — so the mirrored image stays pullable until nothing references it. `diff` returns `action: "update"` with `registry` removed from `stables` for that transition, since registry ownership legitimately changes.

`isDeploymentRolloutComplete` implements Kubernetes' completed-rollout conditions (`observedGeneration >= generation` and `replicas == updatedReplicas == availableReplicas == spec.replicas`) and is exported so it can be unit-tested without a cluster.

## Tests

`packages/alchemy/test/Kubernetes/Deployment.test.ts`, using a fake `ClusterAdapterService` whose registry records every call:

- **mirrors pre-built images by default** — no `imageStrategy` ⇒ `registry.resolve` is called and its `imageUri` wins; `cleanupState` is `undefined`.
- **uses a pre-built image directly without invoking the registry** — `"direct"` ⇒ zero registry calls, `imageUri` is the supplied reference, `state` is `undefined`, and the previously persisted registry state comes back as `cleanupState`.
- **hashes direct images without invoking the managed registry** — the plan-time `workloadImageHash` path also bypasses `registry.hash`.
- **recognizes only a fully completed Kubernetes rollout** — `isDeploymentRolloutComplete` against complete, under-replicated, and stale-generation statuses.
- A compile-time surface check asserting `imageStrategy` is accepted on `ImageDeploymentProps` and rejected on `BundledDeploymentProps`.

```
$ cd packages/alchemy && bun alchemy-test test/Kubernetes/
Tests: 0 failed | 17 passed | 2 skipped (5 files)
```
(the 2 skipped are the pre-existing live-AWS E2E cases)

```
$ bun tsc -b packages/alchemy/tsconfig.test.json
```
No new diagnostics — the single remaining error (`AWS/DynamoDB/Table.ts:779`) reproduces identically on `main` at `bee6451a2`.

`bunx oxfmt --check packages/alchemy/{src,test}/Kubernetes` is clean.

## Follow-up (not in this PR)

`Kubernetes.Job` shares `resolveWorkloadImage` / `workloadImageHash` and has the same `ImageJobProps` shape, so the same option is a natural extension there. It is left out here because a Job has no rollout to gate the old-repository cleanup on, and that deserves its own decision. The plumbing in `internal/workload.ts` is platform-neutral and ready for it.

### DIFF
--- a/packages/alchemy/src/Kubernetes/Deployment.ts
+++ b/packages/alchemy/src/Kubernetes/Deployment.ts
@@ -45,6 +45,7 @@ import {
   connectionIdentity,
   connectionOfOutput,
   deepMerge,
+  imageSourceKind,
   imagePlatformOf,
   makeServerBootstrap,
   resolveWorkloadImage,
@@ -200,6 +201,13 @@ export interface ImageDeploymentProps extends DeploymentPropsBase {
    * reference is used verbatim.
    */
   image: string;
+  /**
+   * How the image is delivered on clusters with a managed registry.
+   * `"mirror"` copies it into that registry; `"direct"` uses the supplied
+   * reference verbatim and requires the cluster to be able to pull it.
+   * @default "mirror"
+   */
+  imageStrategy?: "mirror" | "direct";
 }
 
 export type DeploymentProps =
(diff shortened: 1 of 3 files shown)


sam-goodwin:
Duplicate of https://github.com/alchemy-run/alchemy/pull/1207 ?

microagi-andy:
this is 1207 ;-)

microagi-andy:
duplicate of https://github.com/alchemy-run/alchemy/pull/1098

My clanker ran into it after I made it update to latest alchemy beta and made a new PR.

PR #1221, opened 2026-08-14
fix(cloudflare/workers): paginate zone inference for domains and routes

spinx:
Worker custom domains and routes infer their zone from `GET /zones`, which Cloudflare pages at 20 by default. On an account with more zones, every zone past the first page is invisible to inference and the deploy dies:

```
Could not infer Cloudflare Zone for hostname "app.example.com".
Ensure the parent zone exists in this account.
```

The listing was also unscoped, so a token with access to several accounts could match a zone the Worker's account doesn't own.

Replaced with the exact-name walk already backing the explicit `zone` prop:

```diff
-const zoneList = yield* zones.listZones({})
-const match = zoneList.find((z) => z.name === candidate)
+const hit =
+  zoneCache.get(candidate) ??
+  (yield* findZoneByName({ accountId, name: candidate }))
```

Request count now tracks the hostname's label count instead of the account's zone count, and the cache memoizes the matched zone name, so sibling hostnames on one zone (`api.x.com`, `app.x.com`) cost a single lookup.

`ZonePaging.test.ts` covers it, gated on `CLOUDFLARE_TEST_MANY_ZONES_PROFILE` naming a profile whose account has more than one page of zones. It picks the target by diffing the exhaustive paged listing against the default single-page response, so it stays valid as the account's zone set changes, and no-ops on accounts that can't reproduce.

### DIFF
--- a/packages/alchemy/src/Cloudflare/Workers/WorkerProvider.ts
+++ b/packages/alchemy/src/Cloudflare/Workers/WorkerProvider.ts
@@ -2,7 +2,6 @@ import * as durableObjectsApi from "@distilled.cloud/cloudflare/durable-objects"
 import * as rulesets from "@distilled.cloud/cloudflare/rulesets";
 import * as workers from "@distilled.cloud/cloudflare/workers";
 import * as wfp from "@distilled.cloud/cloudflare/workers-for-platforms";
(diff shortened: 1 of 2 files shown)


sam-goodwin:
Did you close this because it was already fixed?

PR #1222, opened 2026-08-14
fix(planetscale): env overrides for OAuth client + redirect URI

### LINKED ISSUE #1166: PlanetScale OAuth login fails: invalid redirect_uri for https://alchemy.run/auth/callback
### Environment
- `alchemy` 2.0.0-beta.70 (from the npm dist tag; identical behavior on the checkout this CLI ships)
- macOS arm64, bun 1.3.9
- `alchemy login` → PlanetScale provider

### Steps to reproduce
1. Run `alchemy login` and select the PlanetScale provider. The CLI prints the authorization URL and starts the local callback listener on `http://localhost:9976/auth/callback` (confirmed alive via `GET /auth/ping` → 200).
2. Open the printed URL. The PlanetScale consent screen loads, but the authorization request is rejected *before* any code is issued, with an error page reading just `invalid redirect uri`.

The request that fails is exactly what `Planetscale/OAuthClient.ts` builds:

```
GET https://app.planetscale.com/oauth/authorize
  ?client_id=pscale_app_aa12e3938baebb788aac443f66e422da
  &redirect_uri=https%3A%2F%2Falchemy.run%2Fauth%2Fcallback
  &response_type=code
  &state=<state>
```

- `client_id`: `pscale_app_aa12e3938baebb788aac443f66e422da`
- `redirect_uri`: `https://alchemy.run/auth/callback` (i.e. `OAUTH_REDIRECT_URI` — interpolated as a plain string, so I also see the literal URI in the address bar)

### Expected
The consent screen approves and redirects to `https://alchemy.run/auth/callback?code=...&state=...` (or to `http://localhost:9976/auth/callback` via the relay), and the CLI stores PlanetScale credentials in the profile.

### Actual
PlanetScale returns `invalid redirect uri` before consent. No `code` is ever issued, so the CLI's local callback / paste-back path never gets invoked and the login times out.

### Notes / likely cause
- `OAUTH_REDIRECT_URI` is `https://alchemy.run/auth/callback` — a live callback that would bounce to the user's local host (the CLI handles `OAUTH_LOCAL_CALLBACK_URI` = `http://localhost:9976/auth/callback`).
- An `invalid redirect_uri` at the *authorize* step means the PlanetScale OAuth application registered under that `client_id` does not list `https://alchemy.run/auth/callback` as an allowed callback — i.e. the app's registered redirect URIs on PlanetScale don't match the URI the CLI sends, or the app/client record was rotated/recreated.
- All other flows are unaffected: Cloudflare OAuth login works (the `default` profile authenticates Cloudflare fine).

### Request
- Re-register / verify the redirect URIs for the `pscale_app_aa12e3938baebb788aac443f66e422da` application so `https://alchemy.run/auth/callback` (and the `localhost:9976` loopback) are allowed, or
- ship a configurable `redirect_uri` / auth endpoint override so users can point the CLI at a correction in the meantime.

Happy to provide any more detail or run a debug build.

Mkassabov:
PlanetScale OAuth login fails with `invalid redirect uri` at the authorize step (#1166).

Root cause (confirmed against the app's dashboard settings): [#962](https://github.com/alchemy-run/alchemy/pull/962) switched the CLI's `redirect_uri` from `http://localhost:9976/auth/callback` to the hosted relay `https://alchemy.run/auth/callback`, but the PlanetScale OAuth app (`pscale_app_aa12e393...`) still had only the loopback URI registered. PlanetScale exact-matches `redirect_uri` before consent, so no code was ever issued. Cloudflare was unaffected because its app registration was updated.

The fix is registering exactly `https://alchemy.run/auth/callback` (no trailing slash) as the app's Redirect URI — done dashboard-side, no release required. This PR captures what the code can:

- Document the exact-match requirement, the #1166 drift, and the trailing-slash foot-gun (the live relay page 307s to `/auth/callback/`; the registration must match what the CLI sends, not the canonical URL) on the `OAUTH_REDIRECT_URI` constant.
- Make the OAuth app values overridable for self-registered PlanetScale OAuth applications:

```ts
export const OAUTH_CLIENT_ID = env(
  "ALCHEMY_PLANETSCALE_OAUTH_CLIENT_ID",
  "pscale_app_aa12e3938baebb788aac443f66e422da",
);
// likewise ALCHEMY_PLANETSCALE_OAUTH_CLIENT_SECRET and
// ALCHEMY_PLANETSCALE_OAUTH_REDIRECT_URI
```

- Relay page error copy said "Cloudflare did not authorize the request." for all providers — now provider-neutral.

Closes #1166 (fixed by the redirect URI re-registration; verified alongside this PR).

### DIFF
--- a/packages/alchemy/src/Planetscale/OAuthClient.ts
+++ b/packages/alchemy/src/Planetscale/OAuthClient.ts
@@ -41,12 +41,45 @@ export interface Authorization {
  * {@link OAUTH_REDIRECT_URI}. Scopes are configured on the application
  * itself, not requested per-authorization. Rotate by registering a new
  * secret and cutting a release.
+ *
+ * All three values (client id, client secret, redirect URI) can be
+ * overridden via `ALCHEMY_PLANETSCALE_OAUTH_CLIENT_ID` /
+ * `ALCHEMY_PLANETSCALE_OAUTH_CLIENT_SECRET` /
+ * `ALCHEMY_PLANETSCALE_OAUTH_REDIRECT_URI` to point the CLI at a
+ * self-registered PlanetScale OAuth application.
  */
-export const OAUTH_CLIENT_ID = "pscale_app_aa12e3938baebb788aac443f66e422da";
-export const OAUTH_CLIENT_SECRET =
-  "pscale_app_secret_yyZ3Q8oe99GP9_yA5wrA5er6RuN6Lz9dC66Bj1OJzpg";
+const env = (key: string, fallback: string): string => {
+  const value = process.env[key];
+  return value !== undefined && value !== "" ? value : fallback;
+};
 
-export const OAUTH_REDIRECT_URI = "https://alchemy.run/auth/callback";
+export const OAUTH_CLIENT_ID = env(
+  "ALCHEMY_PLANETSCALE_OAUTH_CLIENT_ID",
+  "pscale_app_aa12e3938baebb788aac443f66e422da",
+);
+export const OAUTH_CLIENT_SECRET = env(
+  "ALCHEMY_PLANETSCALE_OAUTH_CLIENT_SECRET",
+  "pscale_app_secret_yyZ3Q8oe99GP9_yA5wrA5er6RuN6Lz9dC66Bj1OJzpg",
+);
+
+/**
+ * PlanetScale validates `redirect_uri` against the OAuth application's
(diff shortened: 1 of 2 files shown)


Mkassabov:
Root cause was the PlanetScale OAuth app's registered Redirect URI (still the pre-#962 loopback URI); fixed by re-registering https://alchemy.run/auth/callback in the dashboard — no code change needed.

PR #1227, opened 2026-08-15
feat(cloudflare/access): protect Workers by identity

### LINKED ISSUE #1225: Access.Application: destinations type does not allow Worker destinations
**Package:** `alchemy@2.0.0-beta.72`
**Files:** `src/Cloudflare/Access/Application.ts` and the generated `lib/Cloudflare/Access/Application.d.ts`

## Summary

`ApplicationDestination` only allows three shapes — `public`, `private`, and `via_mcp_server_portal`. Cloudflare Access also lets an application target a Worker directly, but there is no way to express that in Alchemy without a type error.

The variants that are missing:

| `type` | What it protects |
| --- | --- |
| `worker` | Every ingress surface of one Worker |
| `preview_worker` | Only that Worker's preview URLs |
| `all_workers` | Every Worker on the account, including future ones |
| `all_preview_workers` | The same, but previews only |

## Reproduction

```ts
import { Access, Worker } from "alchemy/cloudflare";

const admin = await Worker("admin", { entrypoint: "./src/index.ts" });

await Access.Application("admin-access", {
  type: "self_hosted",
  destinations: [{ type: "worker", workerId: admin.workerId }],
});
```

## Actual behavior

TypeScript rejects the call:

```
Type '{ type: "worker"; workerId: string; }' is not assignable to type 'ApplicationDestination'.
```

## Expected behavior

The call type-checks, the same way `{ type: "private", ... }` does.

## Notes

This looks like a **types-only** gap. `Application.ts` forwards `destinations` to the Cloudflare API verbatim — `body.destinations = news.destinations` around line 989 — and reads the response back with a cast at line 793. Nothing in the resource inspects or rewrites the `type` field, so once the union is widened the existing code path already handles these destinations.

Two smaller things in the same block:

1. The doc comment above the type says *"Cloudflare supports three destination flavours"* but already lists four bullets, so the count should be dropped or updated.
2. `ApplicationDestination` is a closed union used for **both** the request and the response (`destinations` on the resource output). If Cloudflare adds another destination kind, reading back an existing application produces a value that does not match the declared type. It may be worth allowing an unknown fallback member on the response side.

## Suggested fix

```diff
--- a/src/Cloudflare/Access/Application.ts
+++ b/src/Cloudflare/Access/Application.ts
@@
  * - `via_mcp_server_portal` — routes via a managed MCP server portal.
+ * - `worker` / `preview_worker` — every ingress surface for one Worker, or
+ *   only that Worker's previews.
+ * - `all_workers` / `all_preview_workers` — account-wide protection for all
+ *   current and future Workers, or only their previews.
  */
 export type ApplicationDestination =
   | { type: "public"; uri: string }
   | {
       type: "private";
       hostname?: string;
       cidr?: string;
       l4Protocol?: "tcp" | "udp";
       portRange?: string;
       vnetId?: string;
     }
-  | { type: "via_mcp_server_portal"; mcpServerId: string };
+  | { type: "via_mcp_server_portal"; mcpServerId: string }
+  | { type: "worker"; workerId: string }
+  | { type: "preview_worker"; workerId: string }
+  | { type: "all_workers" }
+  | { type: "all_preview_workers" };
```

Worth confirming the exact field name Cloudflare expects for the Worker id (`workerId` is what we send) and the four `type` strings against the current Access API before merging.

## Workaround in use

A local `patches/alchemy@2.0.0-beta.72.patch` containing exactly the diff above. It works, but the patch has to be regenerated on every release, so having this upstream would help.

mschwarzmueller:
Closes #1225. Protect a Worker with Cloudflare Access by identity — one Worker, its previews, or every Worker / preview in the account — and read the authenticated identity at runtime.

This is the feature from the [Access for Workers docs](https://developers.cloudflare.com/workers/configuration/cloudflare-access/) and the [launch post](https://blog.cloudflare.com/workers-protected-by-access/). #1225 asked for the destination union; this PR also wires the control plane, the script-tag identity, and `ctx.access`.

```ts
const worker = yield* Cloudflare.Worker("Api", { main: "./src/api.ts" });
const allowOrg = yield* Cloudflare.Access.Policy("AllowOrg", {
  decision: "allow",
  include: [{ cloudflareAccountMember: {} }],
});
yield* Cloudflare.Access.Application("ProtectApi", {
  type: "self_hosted",
  destinations: [{ type: "worker", workerId: worker.workerId }],
  policies: [allowOrg.policyId],
});
```

`worker.workerId` is the immutable script `tag`, not the script name. Cloudflare's `worker_id` field is that tag.

```ts
export type ApplicationDestination =
  | { type: "public"; uri: string }
  | { type: "private"; /* … */ }
  | { type: "via_mcp_server_portal"; mcpServerId: string }
  | { type: "worker"; workerId: string }
  | { type: "preview_worker"; workerId: string }
  | { type: "all_workers" }
  | { type: "all_preview_workers" };
```

```ts
const ctx = yield* Cloudflare.WorkerExecutionContext;
const aud = yield* ctx.access.aud;
const identity = yield* ctx.access.identity;
// or: yield* Cloudflare.Access.getIdentity
```

Depends on [alchemy-run/distilled#458](https://github.com/alchemy-run/distilled/pull/458) for `AccessApplicationNotFound` and `AccessDestinationConflict`. Draft until that merges and this PR retargets the submodule pin to the merge SHA.

A second application for the same worker destination is rejected (`AccessDestinationConflict`) — no steal/adopt. Cloudflare echoes destinations faithfully; we do not invent a `domain`.

Account-wide `all_workers` / `all_preview_workers` live tests stay behind `CLOUDFLARE_TEST_ACCESS_ACCOUNT_WIDE=1`. Local `alchemy dev` simulation of `ctx.access` is a follow-up.

Guide: `website/src/content/docs/cloudflare/security/access-workers.mdx`.

### DIFF
--- a/packages/alchemy-test/src/Registry.ts
+++ b/packages/alchemy-test/src/Registry.ts
@@ -2,13 +2,11 @@
  * Per-file registration state.
  *
  * While a test file's module body is evaluating, `describe`/`test`/hook
- * calls register nodes against that file's collector. The runner imports
- * all test files in PARALLEL; attribution stays correct because the
- * collector is carried by AsyncLocalStorage — the module loader propagates
- * the async context of the `import()` call into the module's top-level
- * evaluation (and into microtasks queued from it), so each file's
- * registrations resolve to its own root no matter how many imports are in
- * flight.
+ * calls register nodes against that file's collector. Collection is
+ * sequential: Bun 1.3.x does not propagate AsyncLocalStorage through
+ * dynamic `import()`, so a module-level fallback attributes registrations
+ * to the file currently importing. Do not raise collect concurrency
+ * without restoring ALS (the fallback is process-global).
  *
  * The storage lives on `globalThis` so that a duplicated module instance
  * (e.g. two resolutions of the package) still shares one registry.
@@ -26,6 +24,11 @@ const key = Symbol.for("alchemy-test/registry");
 const storage: AsyncLocalStorage<FileContext> = ((globalThis as any)[key] ??=
   new AsyncLocalStorage<FileContext>());
 
+// Bun 1.3.x does not propagate AsyncLocalStorage through dynamic
+// `import()`. Collection is sequential (see Runner.ts), so a module
+// fallback attributes `describe`/`test` to the file currently importing.
+let importFallback: FileContext | undefined;
+
 /**
  * Collect one file: run `f` (the file's dynamic import + microtask flush)
  * with a fresh root as the ambient collector, and return the root.
@@ -35,12 +38,19 @@ export const collect = async (
   f: () => Promise<void>,
 ): Promise<FileSuite> => {
   const root = makeFileSuite(file);
-  await storage.run({ current: root }, f);
-  return root;
+  const ctx: FileContext = { current: root };
+  const previous = importFallback;
+  importFallback = ctx;
+  try {
+    await storage.run(ctx, f);
+    return root;
+  } finally {
+    importFallback = previous;
+  }
 };
 
 const currentContext = (): FileContext => {
-  const context = storage.getStore();
+  const context = storage.getStore() ?? importFallback;
   if (context === undefined) {
     throw new Error(
       "alchemy-test: describe/test/hook called outside of a test file collection. " +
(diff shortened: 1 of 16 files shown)


sam-goodwin:
Was this satisfied by https://github.com/alchemy-run/alchemy/pull/1228?

Might want to rebase this PR on top of that if there's still an issue

PR #1229, opened 2026-08-15
fix(cloudflare-runtime): keep module-runner export-types detection alive with ctx.waitUntil

schutzelaars:
When you run a Vite dev server, there's a piece that checks what a Worker file exports so it can wire things up correctly. That check runs in the background  and it starts the check but doesn't wait for it to finish before moving on.

The problem: if the check is still running by the time the runtime considers "that request" done, whatever the check tries to do next has nowhere left to run, and it crashes with `Uncaught Error: Disallowed operation called within global scope`.

The fix: tell the runtime to keep things alive until the check actually finishes (`ctx.waitUntil`), instead of firing it and walking away. Same fix already used elsewhere in this codebase for the exact same situation (`workflows-shared/binding.ts`).


schutzelaars:
I couldn't reproduce this. It turned up while hunting a different bug in our own project that shared the same generic error text.

PR #1237, opened 2026-08-15
fix(cloudflare): deduplicate local worker bindings

### LINKED ISSUE #1232: Local Workflow invocation exhausts macOS ephemeral ports through loopback connection storm
## Summary

One local Workflow invocation opens about 16,350 loopback TCP connections. This exhausts the macOS ephemeral port range and makes the Worker return HTTP 500.

Workerd reports:

```text
kj/async-io-unix.c++:1518: failed: connect(): Can't assign requested address; address = 127.0.0.1:<ephemeral-port>
```

The destination address is valid. macOS returns `EADDRNOTAVAIL` because the connection storm leaves no ephemeral source port available.

## Environment

- macOS 26.6.1 (25G76), arm64
- Node.js 24.19.0
- `alchemy@2.0.0-beta.72`
- `@alchemy.run/cloudflare-runtime@2.0.0-beta.72`
- macOS ephemeral range: 49152–65535 (16,384 ports)
- `net.inet.tcp.msl=15000`

## Stack shape

The stack defines an Effect Workflow and binds it to a Worker in the same script:

```ts
export class GuideWorkflow extends Cloudflare.Workflow<GuideWorkflow>()(
  "GuideWorkflow",
  Effect.succeed(
    Effect.fn("GuideWorkflow.run")(function* (input) {
      return yield* Cloudflare.Workflows.task(
        "run",
        Effect.succeed(input),
      );
    }),
  ),
) {}

export default class AppWorker extends Cloudflare.Worker<AppWorker>()(
  "AppWorker",
  { main: import.meta.url },
  Effect.gen(function* () {
    const workflow = yield* GuideWorkflow;
    return {
        const instance = yield* workflow.create({
          id: "repro",
          params: { value: "test" },
        });
        return HttpServerResponse.json({ id: instance.id });
      }),
    };
  }),
) {}
```

The Alchemy stack creates the local Workflow descriptor and Worker binding:

```ts
const guideWorkflow = Cloudflare.Workflow("GuideWorkflow", {
  className: "GuideWorkflow",
});

const worker = yield* Cloudflare.Worker("AppWorker", {
  main: "./worker.ts",
  env: { GuideWorkflow: guideWorkflow },
});
```

Run the stack with `alchemy dev`, then make one request to the Worker.

## Expected result

The local Workflow instance starts and the Worker returns its instance ID. TCP connection counts remain bounded.

## Actual result

The Worker returns HTTP 500. One request creates enough loopback connections to exhaust the ephemeral port range.

Clean socket counts before starting Alchemy:

```text
ESTABLISHED 16
LISTEN      16
TIME_WAIT    0
```

Immediately before the Workflow request:

```text
ESTABLISHED 34
LISTEN      26
TIME_WAIT    2
```

Immediately after one request:

```text
ESTABLISHED 32721
LISTEN          26
TIME_WAIT        7
```

After terminating Alchemy:

```text
TIME_WAIT          32704
loopback_TIME_WAIT 32699
```

A local TCP connection appears as both endpoints in `netstat`, so approximately 16,350 connections were created. That closely matches the 16,384-port ephemeral range.

I reproduced this after terminating every Alchemy/workerd process and waiting 35 seconds, which exceeds twice the configured 15-second MSL. There were no stale `TIME_WAIT` sockets before the run.

## Suspected area

The connection storm begins when the local Workflow binding is invoked. The relevant forwarding path appears to be `ExternalWorkflow` in `RegistryProxy.worker.ts`:

```ts
env.REGISTRY_DEBUG_PORT
  .connect(service.debugPortAddress)
  .getEntrypoint(service.service, "WorkflowBinding")
```

The behavior is consistent with the local Workflow proxy resolving back through itself instead of reaching the owning Workflow implementation. This is a suspected cause, not a confirmed diagnosis.

## Workaround

An in-process orchestration test works, but it does not test Cloudflare Workflow durability. Increasing the ephemeral-port range would only delay the failure.

LukeTandjung:
Deduplicate local Worker bindings by binding name before generating workerd configuration.

Effect Worker metadata and an explicit stack `env` can both register the same Workflow binding. Passing both descriptors to workerd can route the same-worker Workflow through its own registry proxy, recursively opening loopback connections. Local bindings now follow Cloudflare's last-write-wins behavior.

Closes #1232.

### DIFF
--- a/packages/alchemy/src/Cloudflare/Workers/LocalWorkerProvider.ts
+++ b/packages/alchemy/src/Cloudflare/Workers/LocalWorkerProvider.ts
@@ -313,13 +313,7 @@ export const LocalWorkerProvider = () =>
         const props = news as WorkerProps;
         const name = yield* createWorkerName(id, props.name);
         const compatibility = getCompatibility(props);
-        // Worker metadata and explicit `env` can both describe the same
-        // binding (for example, an Effect Workflow class plus its stack
-        // descriptor). workerd requires binding names to be unique; keeping
-        // both can make a same-worker Workflow resolve through its own
-        // registry proxy indefinitely. Match Cloudflare's last-write-wins
-        // binding semantics while preserving the first insertion order.
-        const bindingDescriptors = new Map<string, WorkerBinding>();
+        const bindingDescriptors: WorkerBinding[] = [];
         const durableObjectNamespaces: Record<
           string,
           RuntimeDurableObject & { uniqueKey: string }
@@ -351,10 +345,7 @@ export const LocalWorkerProvider = () =>
                 uniqueKey: namespaceId,
                 sql: true,
               };
-              bindingDescriptors.set(binding.name, {
-                ...binding,
-                namespaceId,
-              });
+              bindingDescriptors.push({ ...binding, namespaceId });
             } else {
               if (
                 binding.type === "workflow" &&
@@ -368,7 +359,7 @@ export const LocalWorkerProvider = () =>
                   className: binding.className,
                 };
               }
-              bindingDescriptors.set(binding.name, binding);
+              bindingDescriptors.push(binding);
             }
           }
           if (data.devRemote) {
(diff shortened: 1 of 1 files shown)


LukeTandjung:
Closing because the deduplication diagnosis was incorrect. The reproduced workerd config contains one Workflow binding and routes directly to the local engine; this PR does not address the actual recursion.

PR #1239, opened 2026-08-15
test: fix ubuntu workerd shutdown race in CI

BlankParticle:
## Summary

- identify the scoped Workerd instance with a unique response sentinel instead of ephemeral-port occupancy
- avoid false shutdown failures when Linux immediately reuses the port for another concurrently spawned Workerd
- keep the shutdown probe bounded

## Validation

- complete Workerd test file passed 10 consecutive local runs
- Ubuntu CI passed 10 consecutive complete Workerd-file runs
- Ubuntu CI passed the full Cloudflare package suite under normal concurrent load

### DIFF
--- a/packages/cloudflare-runtime/src/core/test/Workerd.test.ts
+++ b/packages/cloudflare-runtime/src/core/test/Workerd.test.ts
@@ -4,10 +4,10 @@ import * as Effect from "effect/Effect";
 import * as Exit from "effect/Exit";
 import * as Layer from "effect/Layer";
 import * as Predicate from "effect/Predicate";
+import * as Random from "effect/Random";
 import * as Schedule from "effect/Schedule";
 import * as NodeNet from "node:net";
 import * as Workerd from "../workerd/Workerd.ts";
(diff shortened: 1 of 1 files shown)


packages/cloudflare-runtime/src/core/test/Workerd.test.ts
```diff
@@ -245,24 +245,28 @@ layer(services)((it) => {
...
+        // Linux may immediately give this ephemeral port to another workerd
+        // spawned by a concurrently-running test. Port occupancy therefore
+        // cannot identify whether *this* process survived scope closure.
+        // Probe the unique response instead: refusal, timeout, or a different
+        // body all prove the original process is no longer serving here.
+        const stopped = yield* Effect.tryPromise(async () => {
+          const response = await fetch(`http://127.0.0.1:${port}/`, {
+            signal: AbortSignal.timeout(1_000),
+          });
+          return (await response.text()) !== sentinel;
```
sam-goodwin, line 260: Why are you not using `HttpClient`

Issue #942, opened 2026-07-24
Cloudflare Worker: omitted domain deletes out-of-band custom domains


benpsnyder:
### What happened

A Cloudflare Worker deployment detached an existing custom domain even though the stack did not declare any domain change. The hostname had been attached outside Alchemy, while the Alchemy `Worker` omitted the optional `domain` property.

This is reproducible with Alchemy `2.0.0-beta.64`, and the same reconciliation logic is present at vendored upstream commit `31edd3c4b2f0f3310fad07f5423aee20cf72be8d`.

### Minimal reproduction

1. Create a Worker with Alchemy and omit `domain`.
2. Attach `app.example.com` to that Worker through Cloudflare's Worker Domains API or dashboard.
3. Change the Worker bundle or an environment binding.
4. Run an Alchemy plan/deploy for the Worker.

The deploy deletes the custom-domain attachment.

### Provider path

In `packages/alchemy/src/Cloudflare/Workers/WorkerProvider.ts`:

1. `normalizeDomains(undefined)` returns `[]`.
2. Worker reconciliation calls `reconcileDomains(name, desiredDomains)` even when `domain` is omitted.
3. `reconcileDomains` lists every live domain for that Worker.
4. `toRemove` contains every live hostname not in the empty desired set.
5. The provider calls `workers.deleteDomain` for each entry.

The top-level plan reports a Worker update; it does not enumerate the nested Worker-domain DELETE as a separate destructive operation.

### Why this was surprising

`domain` is optional and its documentation says it binds one or more hostnames. It does not state that omission means "detach every live domain attached to this Worker." By comparison, the adjacent `crons` documentation explicitly says to pass an empty array to remove all triggers.

Cloudflare is behaving as requested here; the provider explicitly issues the delete.

### Expected behavior

One of these contracts would avoid the surprise:

- `domain: undefined` preserves/unmanages live custom domains, while `domain: []` explicitly detaches all; or
- adopting/updating a Worker with live domains and an omitted `domain` fails with a clear error requiring the caller to declare an explicit intent; or
- at minimum, the plan and documentation clearly identify that omitted `domain` will delete all live Worker-domain attachments.

### Suggested regression coverage

- Attach a custom domain to a Worker out of band.
- Update the Worker with `domain` omitted.
- Assert that omission preserves the attachment or fails before mutation.
- Separately assert that `domain: []` performs the intentional detach-all behavior.

[...]

PR #1241, opened 2026-08-15
fix(cloudflare): pin Worker.domain zoneId and paginate listZones

### LINKED ISSUE #1240: Worker.domain zone inference uses one unpaginated listZones page and cannot pin zoneId
`Worker.domain` (and `routes` without an explicit zone) infer the Cloudflare zone by listing the account's zones once and walking DNS labels. That helper is not the same lookup the rest of the Cloudflare provider already uses.

### What happens

```ts
// WorkerProvider.ts — inferZoneIdForHostname
const zoneList = yield* zones.listZones({})
// first page only, no account.id filter, no name query

const parts = hostname.split(".")
for (let i = 0; i < parts.length - 1; i++) {
  const candidate = parts.slice(i).join(".")
  const match = zoneList.find((z) => z.name === candidate)
  if (match) return match.id
}
```

The same provider already has the correct tools:

```ts
// Zone/lookup.ts — used by Worker.routes when zone/zoneName is set
zones.listZones.pages({ account: { id: accountId } })  // exhaustive
findZoneByName({ accountId, name })                     // GET /zones?account.id=&name=&per_page=1
resolveZoneId({ accountId, zone, hostname })
```

`Worker.routes` can pin `zoneId` / `zone` / `zoneName`. `WorkerDomainConfig` cannot — only `name` / `aliases` / `redirects`.

Cloudflare's zone list is paginated (default page size 20). An account with more zones than one page can fail to infer, or attach to the wrong parent if a suffix match appears on page 1 and the real zone does not.

### Why this is not #942 / #588 / #784

- [#942](https://github.com/alchemy-run/alchemy/issues/942) was "omitted `domain` deletes this Worker's attachments." That is fixed: omit unmanages; `null` detaches.
- Domain reconcile lists attachments for **this script** (`listDomains({ service: scriptName })`). It does not walk the zone's other Workers.
- This issue is attach-time **zone selection**, not omit/delete semantics and not unresolved Effect inputs.

### Proposed contract

1. Replace `inferZoneIdForHostname` with `resolveZoneId` / `findZoneByName` (same as routes that already pass `zone` / `zoneName`).
2. Add optional `zoneId` / `zone` / `zoneName` on `WorkerDomainConfig`, same shape as `WorkerRouteConfig`. Explicit zone wins; inference is the fallback.

```ts
domain: {
  name: "app.example.com",
  zoneId: "...",          // or zone / zoneName
  aliases: ["www.example.com"],
}
```

### Non-goals

- Binding inheritance (#1035 / #1024)
- Changing omit-`domain` preservation (#942)
- Detaching or adopting hostnames owned by a different Worker
- A second Worker-domain resource type

### Acceptance

- A unit test feeds `infer` / `resolve` a hostname whose parent zone is not on the first `listZones` page and proves the named lookup (or explicit `zoneId`) still attaches.
- `Worker.domain` with `zoneId` never calls `listZones`.
- Docs for `WorkerDomainConfig` match `WorkerRouteConfig` on zone pinning.

benpsnyder:
Closes [#1240](https://github.com/alchemy-run/alchemy/issues/1240).

`Worker.routes` can pin a zone (`zoneId` / `zone` / `zoneName`) and, when it does, looks the zone up with `resolveZoneId` (`GET /zones?account.id=&name=`). `Worker.domain` could not pin a zone. Attach inferred the zone by listing the account's zones **once** (`listZones({})`) and walking DNS labels against that first page.

Cloudflare paginates zone lists (default 20). An account with more zones than one page can fail to attach, or attach to the wrong parent if a suffix match is on page 1 and the real zone is not. The rest of the Cloudflare provider already has the right helpers (`findZoneByName`, `listZones.pages`, `resolveZoneId`). Domain attach was the leftover caller.

This is attach-time **zone selection**. It is not [#942](https://github.com/alchemy-run/alchemy/issues/942) (omitted `domain` used to delete this Worker's attachments — already fixed: omit unmanages, `null` detaches). Reconcile still lists domains for **this script** only; it does not walk other Workers on the same zone.

```ts
// before — first page of every zone in the token's view
zones.listZones({})

// after — same path routes already use
resolveZoneId({ accountId, zone, hostname })
```

```ts
domain: {
  name: "app.example.com",
  zoneId: "...", // or zone / zoneName; omit to infer by hostname
  aliases: ["www.example.com"],
}
```

An explicit pin never lists zones. Unpinned hostnames walk parent labels with a named lookup, not an account-wide list.

Not in this PR: binding inheritance, omit-`domain` semantics, a second Worker-domain resource, or adopting a hostname owned by another Worker.

### DIFF
--- a/packages/alchemy/src/Cloudflare/Workers/Worker.ts
+++ b/packages/alchemy/src/Cloudflare/Workers/Worker.ts
@@ -338,7 +338,8 @@ export interface WorkerDomainConfig {
    * The canonical hostname (e.g. `"example.com"`). Attached to the Worker
    * as a Cloudflare custom domain — DNS record and edge certificate are
    * managed automatically. The Cloudflare zone is inferred from the
-   * hostname and must already exist in the account.
+   * hostname unless {@link zoneId}, {@link zone}, or {@link zoneName} is
+   * set. The zone must already exist in the account.
    *
    * When set, `https://<name>` is the Worker's primary `url` output.
    */
@@ -359,6 +360,23 @@ export interface WorkerDomainConfig {
    * `urls`.
    */
   redirects?: string[];
+  /**
+   * Cloudflare zone ID for every hostname in this config. Equivalent to
+   * Wrangler's `zone_id`. Wins over {@link zone} / {@link zoneName}.
+   * When omitted, each hostname's zone is looked up by name
+   * (`GET /zones?name=`), not by listing the account's first page of zones.
+   */
+  zoneId?: string;
+  /**
+   * Cloudflare zone name, e.g. `"example.com"`. Equivalent to Wrangler's
+   * `zone_name`.
+   */
+  zoneName?: string;
+  /**
+   * Zone reference — a zone ID, zone name, or `{ zoneId, name? }` object.
+   * Alternative to {@link zoneId} / {@link zoneName}.
+   */
+  zone?: ZoneReference;
 }
 
 export interface WorkerRouteConfig {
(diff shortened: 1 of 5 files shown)


sam-goodwin:
Looks good. Please fix type errors

benpsnyder:
> Looks good. Please fix type errors

Ready

PR #1249, opened 2026-08-18
feat(cli)!: normalize default stage names

sethcarlton:
**BREAKING:** Default stages now use `dev-<safe-user>` instead of `dev_<user>`. User names that require normalization or truncation get a deterministic hash to avoid collisions.

To target an existing `dev_<user>` deployment, pass the old stage explicitly. Explicit `--stage` and `STAGE` values are used as-is.

### DIFF
--- a/packages/alchemy/src/Cli/StageName.ts
+++ b/packages/alchemy/src/Cli/StageName.ts
@@ -1,0 +1,34 @@
+import { createHash } from "node:crypto";
+import * as Effect from "effect/Effect";
+import * as Schema from "effect/Schema";
+import { base32 } from "../Util/base32.ts";
+
+const MAX_DEFAULT_STAGE_LENGTH = 32;
+const HASH_LENGTH = 6;
+const prefix = "dev-";
+const defaultUserPattern = /^[a-z0-9](?:[a-z0-9-]*[a-z0-9])?$/;
+
+export const StageNameSchema = Schema.String.check(
+  Schema.isPattern(/^[a-z0-9]+([-_a-z0-9]+)*$/i),
+);
+
+export const createDefaultStageName = Effect.fn(function* (user: string) {
+  const maxUserLength = MAX_DEFAULT_STAGE_LENGTH - prefix.length;
+  if (user.length <= maxUserLength && defaultUserPattern.test(user)) {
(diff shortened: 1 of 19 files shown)


sam-goodwin:
Why do we need to change `_` to `-`? normalization sounds fine, but why change default?

sethcarlton:
> Why do we need to change `_` to `-`? normalization sounds fine, but why change default?

@sam-goodwin If we leave the default with the current logic it would get auto-normalized to a `-` with a hash. So it would still end up being a different stage name/breaking change. So if the normalization logic stays, I think it's better that the default is something that doesn't need normalization so it's predictable.

I'm not sure how else to handle it. If the current default stays, it won't work with certain resources (R2, Planetscale branches). Unless it was normalized at evaluation time when used for resource names? but the stage in state would still need the non-normalized version right? 

I'm very open to alternatives here. It's just a bummer that the current defaults can cause errors out of box.

sam-goodwin:
Can we first discuss what requires normalization? If it's resources, then they should be changing how they generate physical names (see PhysicalName.ts). We can't come up with a stage naming scheme that will satsify every resource in the world. They're all different. Perhaps we're solving this in the wrong palce.

sethcarlton:
Yea, resources. 

Ok I somehow missed PhysicalName.ts entirely and looking at it, it seems to do what we want already. Pretty sure I only hit issues because of explicitly passing a name which contained stage `name: deployment.isProductionStage ? database.defaultBranch : dev-${deployment.stage}`

I still don't love that the default stage always contains an underscore which could commonly cause this, but maybe this change is too invasive.

sam-goodwin:
I have also seen dashes cause this. I don't know if `_` is worse than `-` in the general case.

sam-goodwin:
We should have explicit names be validated for each resource.

sethcarlton:
Makes sense. Validate and error before attempting to create the resource?

sam-goodwin:
Correct, we might want to add a `validate` method to `Provider` to serve this purpose. Although, it can be difficult to validate during plan because not all Output values will be known. Otherwise, we may as well validate in `reconcile` but that will cause the failure to occur when the plan is being applied. No physical way around this.

PR #1251, opened 2026-08-18
feat(better-auth): upgrade wrapper to 1.7.1

xofromthemoon:
Upgrade `@alchemy.run/better-auth` to better-auth 1.7.1 and run the 1.7 schema steps at deploy so users do not have to follow the [upgrade guide](https://www.better-auth.com/docs/guides/1-7-upgrade-guide) by hand.

Migrations now fingerprint table indexes (so `account.issuer` retriggers deploy), backfill missing issuers from provider IDs and stored Microsoft `id_token` claims, and surface `UnsafeMigrationError` as `BetterAuthMigrationError`. Deploy fails instead of guessing on identity collisions, Microsoft rows without `oid`, leftover `scimProvider` / `oauthApplication` tables, or duplicate device codes.

```ts
yield* prepareSchemaUpgrades(options, database);
const { getMigrations, UnsafeMigrationError } = yield* Effect.promise(
  () => import("better-auth/db/migration"),
);
```

Drizzle consumers switch to the Relations v2 adapter and spread generated `authRelations`:

```ts
const db = drizzle({
  client: pool,
  relations: { ...relations, ...schema.authRelations },
});
```

`secondaryStorage.increment` now requires a TTL.


packages/better-auth/src/BetterAuth.ts
```diff
@@ -41,11 +41,15 @@ export interface BetterAuthProps extends Omit<
...
    * deployed runtime) and re-runs only when the auth schema (plugins,
-   * additional fields) or the target database changes.
+   * additional fields, indexes) or the target database changes.
    *
    * `false` opts out. `true` on a Database layer without migration support
    * (Memory, Drizzle) fails the deploy with a descriptive error.
    *
+   * Better Auth 1.7 refuses to add a required column with no default to a
+   * populated table (notably `account.issuer`). Backfill first using the
+   * [1.7 upgrade guide](https://www.better-auth.com/docs/guides/1-7-upgrade-guide).
```
sam-goodwin: Is this something we should do on behalf of users?

xofromthemoon: Do you mean the backfilling of the account.issuer?

sam-goodwin: Whatever we expect the user to do manually here to upgrade, can we do that for them?

xofromthemoon: Added some backfillings

sam-goodwin:
I had an agent look at this. You can see my commits. There's some issue wth migrations?

patrikduksin:
@xofromthemoon if you need some help lemme know

xofromthemoon:
@patrikduksin feel free to contribute on this pr since I won't be able to have a look on it until next week

sam-goodwin:
superseded by https://github.com/alchemy-run/alchemy/pull/1732

Issue #1106, opened 2026-08-05
AWS env auth: unset AWS_ACCOUNT_ID silently hangs every deploy — getAccountId's STS call re-enters the in-flight cached AWSEnvironment via the Endpoint service (incomplete fix of #704)


mepuka:
> **Amended 2026-08-05.** The original report blamed the auth file lock (`Auth/Lock.js`) and an unbounded API retry. Both were wrong; verification traced the real mechanism, which is an Effect-level self-await, and found this to be residual of #704. Corrected in full below — apologies for the initial misdirection.

## Summary

On `2.0.0-beta.67`, any `alchemy deploy` of a stack that touches an AWS resource with env-based auth (`method: "env"`) hangs **forever** — zero output, 0% CPU, no HTTP request ever dispatched — when `AWS_ACCOUNT_ID` is unset, even though key/secret/token/region are all present and valid. Setting `AWS_ACCOUNT_ID` fixes it completely.

Same symptom, trigger, and mechanism class as **#704** (closed 2026-06-30). That fix is present in beta.67 but incomplete: it provided `Region` and `Credentials` directly in `getAccountId` and missed `Endpoint`.

## Root cause

1. `resolveCredentials`, `method: "env"` branch (`src/AWS/AuthProvider.ts`) falls back to `getAccountId(...)` when `AWS_ACCOUNT_ID` is absent.
2. `getAccountId` calls `STS.getCallerIdentity({})`, providing `Credentials` and `Region` directly (`Layer.succeed` / `Region.of`) — exactly the #704 fix, with its explanatory comment still in place.
3. The distilled client (`@distilled.cloud/aws`, `client/api.ts`) additionally resolves the **`Endpoint`** service: `Effect.serviceOption(Endpoint.Endpoint)`. Nothing shadows it, so it finds the ambient `Endpoint.fromEnvironment` (`src/AWS/Endpoint.ts`), whose value is `Effect.map(<cached AWSEnvironment effect>, env => env.endpoint)`.
4. That cached `AWSEnvironment` effect (`src/AWS/Environment.ts`) **is the effect currently executing** — `auth.read` is its body. The fiber awaits its own `Effect.cached` deferred and suspends permanently, *before* the STS request is dispatched.

Consistent with the observed evidence: the stall is at 0% CPU (a suspended fiber, not a poll loop), no network activity, and instrumentation shows execution stopping at exactly the `Endpoint` service resolution.

[...]

PR #1252, opened 2026-08-18
fix(aws): Endpoint.none to avoid deadlock

### LINKED ISSUE #1231: alchemy deploy hangs with zero output and no network activity (event loop idle in kevent)
## Summary

`alchemy deploy` on an AWS stack hangs indefinitely with **zero output** — not even at `--log-level debug` — and never opens a network socket. The Node event loop goes idle in `kevent` with nothing pending that can wake it, which looks like an unresolved promise / fiber deadlock rather than a slow or retrying operation.

Reproduces on **2.0.0-beta.63 and 2.0.0-beta.72**, and on **effect 4.0.0-beta.107 and 4.0.0-rc.109** — four combinations, identical behaviour.

## Environment

- alchemy `2.0.0-beta.72` (also reproduced on `2.0.0-beta.63`)
- effect `4.0.0-rc.109` (also reproduced on `4.0.0-beta.107`)
- Node `v24.14.1`, Bun `1.3.14`, macOS on Apple Silicon
- AWS credentials via `aws configure export-credentials --format env` (access key + secret + session token), `AWS_REGION=us-east-1`
- `ALCHEMY_TELEMETRY_DISABLED=1`, `DO_NOT_TRACK=1`

## Reproduction

```ts
// alchemy.aws-probe.ts
export default Alchemy.Stack(
  awsProbeContract.stackName,
  {
    providers: AWS.providers(),
    state: AWS.state({ prefix: "op0/probes" }),
  },
  Effect.gen(function* () { /* one S3 bucket + one SQS queue */ }),
);
```

```
$ eval "$(aws configure export-credentials --format env)"
$ export AWS_REGION=us-east-1 ALCHEMY_TELEMETRY_DISABLED=1
$ alchemy deploy alchemy.aws-probe.ts --stage probe-smoke --dry-run --yes --log-level debug
# ...no output at all. Killed after 10 minutes. stdout is 0 bytes.
```

## The process is idle, not busy

Sampled with `sample <pid>` while hung:

```
2335 Thread_…  DispatchQueue_1: com.apple.main-thread  (serial)
+ 2335 start  (in dyld) + 6992
+   2335 node::Start(int, char**)  (in node) + 604
+     2335 node::NodeMainInstance::Run()  (in node) + 268
+       2335 node::SpinEventLoopInternal(node::Environment*)  (in node) + 248
+         2335 uv_run  (in node) + 408
+           2335 uv__io_poll  (in node) + 784
+             2335 kevent  (in libsystem_kernel.dylib) + 8

2335 Thread_…: V8Worker
+     2335 node::TaskQueue<node::TaskQueueEntry>::Locked::BlockingPop()
+       2335 uv_cond_wait
+         2335 _pthread_cond_wait
```

And no sockets are open:

```
$ lsof -p <pid> -a -i
(no output)
```

So it is not waiting on AWS, not retrying, and not spinning — the loop has nothing to do and nothing will wake it.

## What I ruled out, each by test

| suspect | how it was ruled out |
|---|---|
| credentials | the same exported creds work: `aws sts get-caller-identity` returns the account, `aws s3api list-buckets` lists buckets |
| region | missing region produced its own clean `AuthError: AWS region not found`; fixed and the hang continues |
| telemetry | `ALCHEMY_TELEMETRY_DISABLED=1` + `DO_NOT_TRACK=1` — no change |
| the auth lock | a later run reported `Timed out waiting for the alchemy auth lock … held for over 120s`, but that lock was held **by my own first hung run**. Clearing `~/.alchemy/lock/` does not prevent the original hang |
| alchemy version | beta.63 and beta.72 both |
| effect version | 4.0.0-beta.107 and 4.0.0-rc.109 both |
| `createBucket` | ran the exact call manually — `aws s3api create-bucket --bucket alchemy-state-<acct>-us-east-1-an --bucket-namespace account-regional` succeeded **instantly**. The hang persists with the bucket already present, so `ensureStateBucket`'s create path is not it |
| a retry loop | the post-create wait is bounded — `Schedule.max([Schedule.spaced("1 second"), Schedule.recurs(10)])` ≈ 10s |

The earliest stack trace I captured (before the hang became silent) pointed here, which is why I focused on the state store:

```
at state_store.init (…/alchemy/lib/Telemetry/Metrics.js:96:32)
  at Effect.fn (…/alchemy/lib/Cli/commands/_shared.js:187:77)
  at Effect.fn (definition) (…/alchemy/lib/Cli/commands/deploy.js:47:33)
```

## Two separate asks

1. **The deadlock.** `deploy` should not be able to reach a state where the event loop is idle with no pending I/O and no timer. If there is a wait here that can never be satisfied, it needs a timeout with a message naming what it was waiting for.

2. **The silence, which is arguably worse.** Ten minutes of `--log-level debug` produced **0 bytes**. Whatever the root cause, a user cannot distinguish "hung", "slow", "waiting on a prompt", or "network down". Emitting a line before state-store init — and one before each auth/lock acquisition — would have made this a one-minute diagnosis instead of a multi-hour one. The lock-timeout message that *did* eventually appear was excellent precisely because it named the file and the remedy; the silent path deserves the same.

## What I could not determine

Whether the deadlock is AWS-state-specific or general to `deploy`. I did not test a Cloudflare stack in the same environment, so I cannot say whether `AWS.state()` is implicated or merely where this particular stack happens to stop.

Happy to run any diagnostic patch — the repro is one stack file and takes ten minutes to confirm either way.

sam-goodwin:
Closes #1231.

`alchemy deploy` on an AWS stack could hang forever with zero output and no network activity. It is a self-deadlock, not credentials, S3, or the auth lock.

`getAccountId` runs STS `GetCallerIdentity` from *inside* the construction of `AWSEnvironment`. It already guards `Region` against the resulting cycle, but passed `Layer.empty` for the endpoint:

```diff
             Region.of(region),
-            endpoint ? Endpoint.of(endpoint) : Layer.empty,
+            endpoint ? Endpoint.of(endpoint) : Endpoint.none,
```

`Layer.empty` lets the call fall through to the ambient `Endpoint.fromEnvironment` — provideMerge'd by both `AWS.state()` and `AWS.providers()` — whose service Effect reads the endpoint off `AWSEnvironment`:

```ts
// distilled protocol.ts — every AWS op resolves this
const customEndpoint = yield* yield* Effect.serviceOption(Endpoint.Endpoint)
```

That re-enters the `Effect.cached` `AWSEnvironment` resolution already in flight on the same fiber, so the fiber awaits its own cache deferred forever: no timer, no socket, event loop idle in `kevent`. The cross-process auth lock is held for the whole hang, which is why the *next* run reports `Timed out waiting for the alchemy auth lock … held for over 120s`.

### Trigger

`method: "env"` credentials (what exported `AWS_*` gives you) with no `AWS_ACCOUNT_ID` (so STS is needed) and no `AWS_ENDPOINT_URL` (so `Layer.empty` is taken). The `stored`-without-`accountId` backfill path hits it too. Not state-store specific — `Endpoint.fromEnvironment` is in `AWS.providers()` as well.

### Verification

Probe stack with `AWS.state()`, profile configured `{ method: "env" }`, dummy `AWS_*` credentials:

| run | before | after |
|---|---|---|
| no `AWS_ENDPOINT_URL` | hangs indefinitely, 0 bytes out, 0 inet fds | fails in ~2.7s: `The security token included in the request is invalid` |
| `AWS_ENDPOINT_URL` set | fails in ~4s (took the `Endpoint.of` branch) | unchanged |

### The silence (issue's second ask)

A stalled deploy printed nothing at any log level, and `.alchemy/log/out` was empty too — nothing on the `deploy → state_store.init → auth lock` path logged at all.

- Debug lines for the state-store init and each auth-lock acquire/acquired/release.
- A stall notice every 30s naming the operation and lock path:

```
alchemy: AWS.read (profile 'prod') has been running for 30s while holding the
auth lock '/Users/me/.alchemy/lock/prod-AWS.lock'. Re-run with --log-level debug for detail.
```

  Via `Console.error`, not the logger — with telemetry on, `TelemetryLive` replaces the console logger with the OTLP one, so log records never reach the terminal of the user watching it hang. Suppressed for `login`/`configure`, which legitimately block on the user and have a prompt on screen.

- `--log-level` now reaches the console. Telemetry-on runs sent every record to `.alchemy/log/out` and nothing to stderr, so `--log-level debug` changed nothing visible. A console sink is added when the flag is explicitly passed; with telemetry off the default logger is still in the set, so nothing double-prints.

No hard timeout on the locked effect: a slow-but-real SSO refresh or OAuth round-trip would be killed by one. The notice makes a stall visible without declaring it fatal.

Before / after on the same stalling run, telemetry on, no flags:

```diff
-(nothing, indefinitely)
+alchemy: AWS.read (profile 'probe') has been running for 30s while holding the auth lock '…'.
```


arthtyagi:
Confirming this is still needed against the current published release, and adding an independent verification of the fix.

`2.0.0-beta.74` (published 2026-08-21, current `latest`) still ships the unfixed branch, and so does `main` at `e9fe594`:

```
packages/alchemy/src/AWS/AuthProvider.ts:186
    endpoint ? Endpoint.of(endpoint) : Layer.empty,
```

`Endpoint.none` is absent from the published tarball in both `src/AWS/Endpoint.ts` and `lib/AWS/Endpoint.js`.

One detail from the beta.72 to beta.74 window is worth naming, because it makes the line look already handled. beta.72's `getAccountId` had no endpoint provision at all. beta.74's gained the `endpoint` parameter and the `Endpoint.of(endpoint)` branch. But the `Layer.empty` fallback is the branch every ordinary run takes, since a user with plain exported `AWS_*` credentials has no `AWS_ENDPOINT_URL`. So the deadlock is unchanged in practice across all three releases, while the diff reads as if it were addressed.

Independent verification, arrived at separately before finding this PR:

* Applied the semantically identical one-liner, `Layer.succeed(Endpoint.Endpoint, Effect.succeed(undefined))` in place of `Layer.empty`, as a patch over `2.0.0-beta.72`.
* Profile `{ method: "env" }`, valid credentials in the environment, `AWS_ACCOUNT_ID` deliberately unset, no other workaround.
* Before: `deploy` produced zero bytes and never terminated.
* After: `deploy` of a stack using `AWS.state()` plus eight resources reached `Done: 8 succeeded` in about 60 seconds, `destroy` reached `Done: 4 succeeded` in about 71 seconds, and a sweep afterwards found no orphaned instances, queues, functions or roles, with the state bucket empty.

Two notes for anyone working around this before the merge:

1. The patch has to reach `src/`, not only `lib/`. Under bun the package resolves to `src/AWS/AuthProvider.ts`, so a `lib/`-only patch changes nothing and reads as though the fix did not work.
2. The silence is worth keeping in scope rather than splitting off. The stall emits nothing at `--log-level trace` either, and the records that do accumulate under `.alchemy/log/out` are not pointed to from anywhere a user would look while watching a hang. The console sink and the 30s stall notice in the second commit are what turn this from a multi-hour bisect into one line on screen.

Also relevant: #1106 describes the same self-await and proposes the same one-line fix, so this PR should close that issue too. It is currently linked only to #1231.

arthtyagi:
This is still reproducible on `2.0.0-beta.74`, the current `latest` on npm. I checked the published
tarball rather than the branch, in both the source and the build that actually ships:

```
x-2.0.0-beta.74/package/src/AWS/AuthProvider.ts:186
    endpoint ? Endpoint.of(endpoint) : Layer.empty,

x-2.0.0-beta.74/package/lib/AWS/AuthProvider.js:79
    endpoint ? Endpoint.of(endpoint) : Layer.empty)), Effect.flatMap((self) => self.Account
```

When no custom endpoint is configured, which is the default for anyone deploying to real AWS,
`Layer.empty` leaves the ambient `Endpoint.fromEnvironment` to resolve. That provider reads
`env.endpoint` off the `Effect.cached` `AWSEnvironment` effect that this very STS call is in the
middle of constructing, so the fiber waits on itself.

What makes this worth merging rather than working around: the comment two lines above the same call
site already documents this exact hazard, for the sibling service.

```ts
// Provide Region directly from the resolved inputs. Relying on the
// ambient Region provider (Region.fromEnvironment) here would
// deadlock: it derives the region from AWSEnvironment, which is the
// very service still being constructed by this STS call.
Region.of(region),
```

`Region` is pinned for precisely the reason `Endpoint` needs to be. `Endpoint.none` in this PR is
that same pin, which is why the change is small and why the failure is total rather than partial.

The symptom is bad out of proportion to the fix. There is no error, no timeout and no network
activity: `alchemy deploy` hangs with zero output and the event loop parked in `kevent`, which reads
as a credentials or connectivity problem rather than a resolution cycle. That is what #1231 was
opened about.

We have been carrying this exact change as a `patchedDependencies` patch against `beta.72` in
production since it was diagnosed, applied to both `src/AWS/AuthProvider.ts` and
`lib/AWS/AuthProvider.js`, because bun resolves `alchemy/*` to `src/*.ts` while node resolves to
`lib/*.js` and the launcher re-execs whichever runtime invoked it. Patching only one entrypoint
leaves the other still hanging. No regressions since.

We also confirmed the deadlock is not a `beta.72` artifact that later releases fixed by accident:
`beta.73` and `beta.74` both ship `Layer.empty` at this site, so upgrading is not a workaround.

Measured in our repository against the real `AWSEnvironment`, the real `Endpoint.fromEnvironment`
and the real `Endpoint` tag, with no network and no deploy:

```
ambient Endpoint only (unpatched shape):    DEADLOCK, timed out after 3008ms
Endpoint pinned at the call site (this PR):  completed in 1ms
```

Could this be merged? It is a two-line fix for a failure mode that is silent, total, and hits the
default path for every AWS user.

One small note on shape, take it or leave it: `endpoint ? Endpoint.of(endpoint) : Endpoint.none` is
strictly better than what we patched in, which pins unconditionally. Ours is fine on `beta.72`
because that call site has no `endpoint` parameter yet, but on `beta.74` the conditional is what
keeps a LocalStack or floci account lookup from reaching real STS. We will align to yours when this
lands.

PR #1256, opened 2026-08-19
fix(cloudflare): don't let the dev registry freeze when fs.watch fails

sam-goodwin:
Fixes a user report ([repro](https://github.com/LordCoughmann/alchemy-writequeue-binding-repro)) where adding a Queue producer binding to a Worker made a `Website.Vite` Worker's service binding to it fail with `503 Worker "<backend>" not found`, while direct requests to the same Worker kept working. Confirmed fixed by the reporter on both the minimal repro and their real app.

`alchemy dev` runs a Vite website's workerd in a **different process** from a plain Worker's, so the two find each other only through the on-disk dev registry (`~/.local/state/alchemy/registry`). Each process refreshed its in-memory view of that directory from `fs.watch` alone — one snapshot at startup, then watcher events forever after:

```ts
// Registry.ts, before
yield* (
  supportsWatcher
    ? fs.watch(directory).pipe(/* … */, Stream.mapEffect(() => refresh))
    : Stream.fromEffect(refresh).pipe(Stream.repeat(Schedule.spaced(100)))
).pipe(Stream.runDrain, Effect.forkScoped)
```

That single signal is load-bearing and it can stop in two shapes. It can **fail**, and it can **end** — a closed `fs.watch` handle calls `Queue.endUnsafe`, so the stream completes normally, `runDrain` returns, the forked fiber retires, and the registry stops refreshing with nothing logged and no error anywhere. From the outside the two are identical: the website's frozen snapshot never gains the backend, and its service binding answers "not found" for the rest of the session.

The Queue is only the trigger. A producer binding for a queue **nothing consumes** makes the Worker subscribe to the registry for a `queue-consumer` target, which is what gives it a registry-proxy worker to build and boot before it registers itself. That pushes its registration past the website's snapshot. Without the queue it registers earlier, the snapshot happens to contain it, and everything looks fine.

### The fix

- Poll the registry directory (1s) as a safety net alongside the watcher, and drop to the unwatched rate (100ms) if the watcher stops — whether it failed or ended.
- Log which of the two happened, so a frozen registry is diagnosable next time instead of silent.
- Don't let a failed snapshot or a failed target-map push tear down the loops that keep discovery current — each push carries the full map, so the next change recovers whatever one dropped.
- `subscribe` dedupes by value, so polling doesn't push identical target maps to every proxy.

```ts
// Registry.ts, after
fs.watch(directory).pipe(
  /* … */,
  Stream.concat(Stream.suspend(() => watcherStopped("ended"))),
  Stream.catchCause((cause) => watcherStopped("failed", cause)),
  Stream.merge(poll(WATCHED_POLL_INTERVAL)),
)
```

`Registry (watcher fails)` / `Registry (watcher ends)` inject a `FileSystem` whose `watch` does each, then register a worker the way another process does (a file appearing in the directory). Both time out before this change and pass after.

### On the root cause

Worth being straight about what is and isn't established. This change is verified against the failure *mode*, by injecting a stopped watcher — I never reproduced the reporter's environment. Their repro runs clean on macOS at `2.0.0-beta.72`, which is why the Queue looked like the culprit rather than the timing shift it causes.

The reporter has since ruled out both environmental guesses in the original version of this description: their project is on the native WSL filesystem, not `/mnt/c`, and their inotify limits are healthy (`max_user_watches: 524288`, `max_user_instances: 128`, and raising the latter changes nothing). So **why** their watcher stops delivering is still unknown — the "ends" path above is the leading candidate precisely because it needs no error and leaves no trace. The fix stands either way: a registry whose only refresh signal is a best-effort watcher is wrong regardless of which platform quirk trips it.

### For @LordCoughmann, if you're up for one more datapoint

This would tell us whether `fs.watch` delivers events at all on your box. Save as `watch-probe.mjs`, run `node watch-probe.mjs` in WSL:

```js
import fs from "node:fs";
import os from "node:os";
import path from "node:path";

const dir = process.env.XDG_STATE_HOME
  ? path.join(process.env.XDG_STATE_HOME, "alchemy", "registry")
  : path.join(os.homedir(), ".local", "state", "alchemy", "registry");
fs.mkdirSync(dir, { recursive: true });
console.log("watching", dir);

let events = 0;
const w = fs.watch(dir, (event, name) => {
  events++;
  console.log("event:", event, name);
});
w.on("error", (e) => console.log("WATCHER ERROR:", e.message));
w.on("close", () => console.log("WATCHER CLOSED"));

const probe = path.join(dir, "watch-probe.json");
setTimeout(() => fs.writeFileSync(probe, "{}"), 500);
setTimeout(() => fs.writeFileSync(probe, '{"n":2}'), 1500);
setTimeout(() => {
  fs.rmSync(probe, { force: true });
  w.close();
  console.log(events > 0 ? `OK: ${events} events` : "BROKEN: 0 events delivered");
  process.exit(events > 0 ? 0 : 1);
}, 3000);
```

`BROKEN: 0 events` or a `WATCHER CLOSED` before the end confirms it. `OK: 2 events` means the watcher is fine in isolation and something in the dev session is stopping it — also useful to know, and a reason to look elsewhere. Either result is a real answer; no rush, the fix doesn't depend on it.

### DIFF
--- a/packages/alchemy/test/Cloudflare/Queues/ProducerServiceBinding.local.test.ts
+++ b/packages/alchemy/test/Cloudflare/Queues/ProducerServiceBinding.local.test.ts
@@ -1,0 +1,95 @@
+import * as Cloudflare from "@/Cloudflare/index.ts";
+import * as Test from "@/Test/Alchemy";
+import { expect } from "alchemy-test";
+import * as Data from "effect/Data";
+import * as Effect from "effect/Effect";
+import { MinimumLogLevel } from "effect/References";
+import * as Schedule from "effect/Schedule";
+import * as HttpClient from "effect/unstable/http/HttpClient";
+import * as pathe from "pathe";
+
+const { test } = Test.make({
+  providers: Cloudflare.providers(),
+  dev: true,
+});
+
+const logLevel = Effect.provideService(
+  MinimumLogLevel,
+  process.env.DEBUG ? "Debug" : "Info",
+);
+
+class WorkerNotReady extends Data.TaggedError("WorkerNotReady")<{
+  status: number;
+}> {}
+
+const getReady = (url: string) =>
+  Effect.gen(function* () {
+    const client = yield* HttpClient.HttpClient;
+    return yield* client.get(url).pipe(
+      Effect.flatMap((res) =>
+        res.status === 200
+          ? Effect.succeed(res)
+          : Effect.fail(new WorkerNotReady({ status: res.status })),
+      ),
+      Effect.retry({
+        while: (error) => error instanceof WorkerNotReady,
+        schedule: Schedule.exponential("250 millis"),
+        times: 20,
+      }),
+    );
+  }).pipe(Effect.orDie);
+
+/**
+ * A producer binding for a queue nothing consumes makes the worker subscribe
+ * to the dev registry (it has to find whichever instance consumes the queue),
+ * which in turn gives it a registry-proxy service it would not otherwise
+ * have. That must not disturb how OTHER workers reach it: the caller's
+ * service binding resolves the target through the same registry.
+ *
+ * Both workers run in this process here, so this covers the in-process wiring
+ * only — the cross-process path (`alchemy dev` runs a Vite website's workerd
+ * in a child process) is pinned by the registry's own tests.
+ */
+test.provider(
+  "service binding reaches a worker that produces to an unconsumed queue",
+  (stack) =>
+    Effect.gen(function* () {
+      yield* stack.destroy();
+
+      const deployed = yield* stack.deploy(
+        Effect.gen(function* () {
+          const queue = yield* Cloudflare.Queues.Queue("ProducerOnlyQueue");
+          const target = yield* Cloudflare.Worker("producer-target", {
+            main: pathe.resolve(
+              import.meta.dirname,
+              "fixtures/producer-target-worker.ts",
+            ),
+            env: { QUEUE: queue },
+          });
+          const caller = yield* Cloudflare.Worker("producer-caller", {
+            main: pathe.resolve(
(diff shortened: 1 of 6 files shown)


LordCoughmann:
## Probe result

Ran the probe on the WSL box (`XDG_STATE_HOME` unset, native WSL2 filesystem):

```
watching /home/user/.local/state/alchemy/registry
event: rename watch-probe.json
event: change watch-probe.json
event: change watch-probe.json
OK: 3 events
```

`fs.watch` delivers events fine in isolation — the watcher is not broken at the OS/inotify level.

Environment limits, for completeness:

```
max_user_watches: 524288
max_user_instances: 128
```

Raising `max_user_instances` to `512` does **not** fix the issue.

## Additional finding: the Vite child never holds a registry watch

From the live dev session (process-level, while the `503` is happening):

- The **main local runner** holds an inotify watch on the registry directory.
- The **Vite child** (`ViteChildRunner`) holds **no** registry watch at all — its only inotify fd is Vite's own source watcher.

So the child's registry `SubscriptionRef` is driven by a watch that was never established (or was established and immediately dropped), and its proxy target map stays frozen at the boot snapshot.

## Temporary workaround

We're running with the poll fallback forced on until this lands:

```
CLOUDFLARE_RUNTIME_FILE_SYSTEM_SUPPORTS_WATCHER=false
```

This reliably fixes the symptom, which also confirms the poll path is correct.

## What this narrows

The open question moves from "why does the watcher stop delivering mid-session" to:

> Why does the Vite child never establish (or immediately tear down) the registry watch?

The poll fallback still fixes the symptom regardless, so this is a follow-up observation, not a blocker — it just points the remaining root-cause hunt at the child's watch wiring rather than the watcher itself.

PR #1264, opened 2026-08-19
fix(cloudflare/dns): disambiguate DNS record adoption by content/priority

### LINKED ISSUE #1262: Cloudflare DNS adoption is ambiguous for records sharing `(name, type)`
**AI-drafted; human-reviewed and authorized.** On behalf of user by GPT-5.6 Sol via T3 Code. The user reviewed and approved this issue's content and explicitly authorized its publication.

When adopting an existing Cloudflare zone, `Cloudflare.DNS.Record` looks up records by `(zoneId, name, type)` and selects the first match.

Multiple records with the same name and type are valid and common—for example MX fallbacks, multiple TXT records, or multiple A records. This makes `--adopt` ambiguous because several Alchemy resources may select the same physical record.

```ts
Cloudflare.DNS.Record("PrimaryMx", {
  zoneId,
  name: "example.com",
  type: "MX",
  content: "aspmx.example.com",
  priority: 1,
});

Cloudflare.DNS.Record("BackupMx", {
  zoneId,
  name: "example.com",
  type: "MX",
  content: "backup.example.com",
  priority: 10,
});
```

Could adoption support an explicit `recordId` or another unambiguous selector? At minimum, detecting multiple matches and returning an actionable error would prevent adopting the wrong record.

sam-goodwin:
DNS record adoption looked up records by `(zoneId, name, type)` and took the first match. Several records legitimately share that key (MX fallbacks, multiple TXT/A records), so two Alchemy resources could adopt the same physical record and clobber each other.

`findByNameType` now collects **all** matches and, when more than one exists, the declared `content`/`priority` must select exactly one:

```ts
// both adopt their own physical record now — previously both matched the first MX
Cloudflare.DNS.Record("PrimaryMx", { zoneId, name, type: "MX", content: "mx1.example.com", priority: 1 });
Cloudflare.DNS.Record("BackupMx",  { zoneId, name, type: "MX", content: "mx2.example.com", priority: 10 });
```

```
- 1 candidate → returned as-is (preserves the adopt-then-modify flow)
->1 candidates, exactly one exact `content`/`priority` match → that record
->1 candidates, no exact match → treated as missing (a new sibling record is created)
->1 candidates, still ambiguous → fails with an error listing candidate ids/contents/priorities
```
Applies to the `read` adoption scan, the reconcile fallback scan, and the `DnsRecordAlreadyExists` self-heal path.

Fixes alchemy-run/alchemy#1262

### DIFF
--- a/packages/alchemy/src/Cloudflare/DNS/Record.ts
+++ b/packages/alchemy/src/Cloudflare/DNS/Record.ts
@@ -138,6 +138,15 @@ export type Record = Resource<
  * set. This protects hand-edited records (especially the apex `A`/`AAAA`
  * and email DKIM/SPF records that the dashboard often manages) from
  * being clobbered.
+ *
+ * Several records may legitimately share `(name, type)` — MX fallbacks,
+ * multiple TXT records, round-robin A records. When the scan finds more
+ * than one candidate, the declared `content` (and `priority`) must match
+ * exactly one record; a record with no exact match is treated as missing
+ * (a new sibling record is created), and a still-ambiguous match fails
+ * with an error listing the candidates. To adopt one record out of such
+ * a set, declare its current `content`/`priority` verbatim first, then
+ * change them in a follow-up deploy.
  * @resource
  * @product DNS
  * @category Domains & DNS
@@ -246,7 +255,10 @@ export const RecordProvider = () =>
       //    behind the adopt policy before reconcile ever runs.
       let foundByScan = false;
       if (!observed) {
-        const existing = yield* findByNameType(zoneId, news.name, news.type);
+        const existing = yield* findByNameType(zoneId, news.name, news.type, {
+          content,
+          priority: news.priority,
+        });
         if (existing) {
           foundByScan = true;
           observed = existing;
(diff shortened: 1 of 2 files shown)


packages/alchemy/src/Cloudflare/DNS/Record.ts
```diff
@@ -407,27 +428,78 @@ const observeById = (zoneId: string, dnsRecordId: string) =>
...
+            candidates
+              .map(
+                (r) =>
+                  `  - id=${r.id} content=${JSON.stringify(r.content)}` +
+                  (r.priority === undefined ? "" : ` priority=${r.priority}`),
+              )
+              .join("\n"),
+        ),
+      );
+    }),
```
sam-goodwin: Re-write as Effect.flatMap(Effect.fn(..)) and give that a Data.TaggedError

packages/alchemy/src/Cloudflare/DNS/Record.ts
```diff
@@ -382,12 +397,18 @@ export const RecordProvider = () =>
...
+      // resource corresponds to.
       const zoneId = output?.zoneId ?? (olds?.zoneId as string | undefined);
       const name = output?.name ?? olds?.name;
       const type = output?.type ?? olds?.type;
       if (zoneId && name && type) {
-        const observed = yield* findByNameType(zoneId, name, type);
+        const observed = yield* findByNameType(zoneId, name, type, {
+          content: (olds?.content as string | undefined) ?? output?.content,
+          priority: olds?.priority as number | undefined,
+        });
```
sam-goodwin, line 412: Why so much casting? Are the types broken

PR #1235, opened 2026-08-15
feat!: overhaul the Alchemy CLI

BlankParticle:
Top of stack #1236: `feat/clikit` → `feat/profile` → `feat/cli-overhaul`.

This PR completes the CLI overhaul on top of CliKit and profiles: it reorganizes commands, replaces the remaining legacy renderer with `SigilCli`, introduces module-owned control services, and models finite control work as typed scoped operation streams.

```text
alchemy
├─ deploy | plan | destroy
├─ drift [--repair]
├─ dev
├─ logs [--follow]
├─ state list|read|delete
├─ unsafe nuke
└─ provider
   ├─ check-env
   ├─ aws bootstrap|teardown
   └─ cloudflare bootstrap|teardown|token|state logs
```

### Control and operation model

- Top-level control modules own their routes: stack, drift, logs, state, profile, provider, and unsafe nuke.
- Stack, drift, profile, and nuke keep their Schema tagged event unions in their existing module files.
- A small shared `Operation` primitive exposes a scoped event stream with a typed `Succeeded` result variant.
- Expected failures remain in the stream error channel; stopping consumption interrupts the Effect performing the work.
- The former operation registry, snapshots, replay buffer, status polling, string tags, and `unknown` payload envelopes are removed.
- CLI and TUI consumers match domain events directly without operation IDs or payload casts.

### Rendering and commands

- The final root view is `SigilCli`; remaining Ink identifiers and filenames are removed.
- Plan, approval, apply, drift, profile, nuke, help, and state explorer surfaces use CliKit.
- Planning and loading messages display the selected stage beside the message.
- Clack and the legacy `tui/` renderer are deleted.
- Hidden compatibility commands explain migrations without executing removed behavior.
- Exit codes are script-safe: `0` success, `1` failure or decline, and `130` cancellation.

### Command migration

| Old | New |
| --- | --- |
| positional stack file | `--config` / `-c` |
| `tail` | `logs --follow` |
| `sync` | `drift --repair` |
| `destroy --dry-run` | `plan --destroy` |
| top-level `aws` / `cloudflare` | `provider aws` / `provider cloudflare` |
| state-specific nouns | `state list|read|delete <path>` |

Sidecar logs stream into the exec child's logger without tearing live output. Every CLI run also writes a pruned debug log under `~/.alchemy/logs` with invocation context and full failure causes.

Verification includes the package type-check plus focused AlchemyControl, CliKit, planning, stage, import, and interruption suites.

PR #1274, opened 2026-08-20
feat(cli): add a --detailed flag to show property changes in "alchemy plan" and "alchemy deploy"

henningpokriefke:
> [!NOTE]
> This needs to be changed after PR #1235 to add on to the CLI overhaul from Blank. I dont know what/if you guys are planning everything else in the property diff direction that does not comply with my proposal. This is just how I thought it could fit in :)  

### Reason I would like this to be added

We want to migrate to alchemy from using Terraform and we like seeing the planned changes more detailed.

### What changes

This adds an opt-in `--detailed` view to `alchemy plan` and `alchemy deploy`:

```sh
alchemy plan --detailed
alchemy deploy --detailed
```

The default output stays compact:

plan-compact-default

With `--detailed`, the same plan includes the declared property changes:

plan-detailed-update

### Why a flag

Basically everything stays the same, except the user wants to opt in into more detailed output.

### What the detailed view shows

- creates show the inputs declared in the stack
- updates and replacements show additions, removals, and changed values
- deletes stay compact

The diff compares persisted `state.props` with the desired `props` already in the plan. It explains changes to declared inputs, not live cloud drift.

The Logging CLI and Ink UI share one structured `PropertyChange` result. Paths are sorted, arrays are compared by index, and rendering never evaluates a value:

- secrets stay `(redacted)`
- Outputs become `(known after apply)`
- raw Effects and functions become `(computed)`

I tried aligned tables, fixed columns, and always-stacked values along the way. Inline `old → new` rows were the easiest to scan at normal widths, with a stacked before/after layout only when the terminal gets narrow.

### Create previews

Detailed creates show every declared input already present in the plan, including unresolved inputs represented as Outputs.

They cannot list future provider-generated fields such as an ARN, queue URL, or database endpoint. Those values are not in `props` yet, so the preview shows what Alchemy actually knows instead of guessing the rest. 

Supporting a Terraform-style list of future computed fields would need a separate provider preview contract as far as I understood the that part of the repo.

plan-detailed-create

### Try it locally (that test was primarily for local development to test different look and feels; idk if that should be merged tbh)
The test deploys a stack (against not real cloud, just the state will be created) with detailed deploy, then plans an normal update in step 2 and then plans a detailed update in step 3 so you can see these 3 versions.

```sh
cd packages/alchemy
bun test/Cli/MultiCloudPlanPropertyDiff.test.ts --verbose
```

### More examples

Redacted and deferred values:

plan-detailed-security-deferred

At narrower widths, before and after values move onto separate lines:

plan-detailed-update-narrow
plan-detailed-output-narrow

`alchemy deploy --detailed` shows the same preview before the existing approval prompt. The flag changes the preview, not the deploy flow.

deploy-detailed-approval

### DIFF
--- a/packages/alchemy/src/Apply.ts
+++ b/packages/alchemy/src/Apply.ts
@@ -24,6 +24,7 @@ import {
   makeScopedArtifacts,
 } from "./Artifacts.ts";
 import {
+  type PlanDisplayOptions,
   type PlanStatusSession,
   type ScopedPlanStatusSession,
   Cli,
@@ -89,6 +90,10 @@ export type AppliedPlan<P extends Plan> = {
     : Simplify<P["resources"][id]["resource"]["attr"]>;
 };
 
+export interface ApplyOptions {
+  planDisplay?: PlanDisplayOptions;
+}
+
 interface ResourceTracker {
   output: any;
   props: any;
@@ -148,6 +153,7 @@ const instrumentLifecycle =
 
 export const apply = <P extends Plan>(
   plan: P,
+  options: ApplyOptions = {},
 ): Effect.Effect<
   Input.Resolve<P["output"]>,
   | Output.InvalidReferenceError
@@ -176,7 +182,7 @@ export const apply = <P extends Plan>(
     }
 
     const cli = yield* Cli;
-    const session = yield* cli.startApplySession(plan);
+    const session = yield* cli.startApplySession(plan, options.planDisplay);
     const state = yield* yield* State;
     const stack = yield* Stack;
     const stage = yield* Stage;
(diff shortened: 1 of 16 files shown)


henningpokriefke:
This needs to be changed after PR #1235 to add on to the CLI overhaul from Blank. I dont know what/if you guys are planning everything else in the property diff direction that does not comply with my proposal. This is just how I thought it could fit in :)

henningpokriefke:
Closing this since `--detailed` and property-level plan output have landed as part of the new CLI in #1235. The current implementation covers the original intent and integrates it with the new plan UI.

PR #1297, opened 2026-08-21
fix(cloudflare/workers): preserve DO migration state after precreate

### LINKED ISSUE #1296: Cloudflare Worker precreate can emit duplicate new_sqlite_classes migration
### What happened

On a fresh preview deployment using `alchemy@2.0.0-beta.72`, Alchemy precreated a Worker with a Durable Object class, observed its namespace, and then failed during the real upload because the class was already depended on by existing Durable Objects.

The same deployment input succeeded after cleanup and redeployment.

### Why this appears to happen

The normal Worker path precreates a placeholder with `newSqliteClasses`, then polls until `durableObjectNamespaces` proves the namespace exists. During reconcile, however, `WorkerNotFound` / `WorkerHasNoVersions` is converted to `undefined`. The fallback settings read in `putWorker` also catches every error and converts it to `undefined`.

That lets a transient settings observation failure erase the precreate evidence. Migration planning reads its logical-id-to-class map from `oldSettings.tags`, sees no previous class, and emits `new_sqlite_classes` for the same class a second time, which Cloudflare rejects.

The precreate output already persists the same `alchemy:dos:` tags. They are available to reconcile even while the live settings read is racing propagation.

This is distinct from a legitimate greenfield Workers for Platforms user Worker, which skips precreate and can correctly treat a missing script as absent.

### Proposed behavior

When attributes for the script are already held, retry the settings read briefly (bounded backoff) before concluding there are no existing settings. A Worker that really is gone still falls through to the upsert, so out-of-band deletion keeps self-healing.

A focused fix with deterministic regression coverage is in #1297.

sshkeda:
Fixes #1296.

Precreate uploads a stub that registers a Worker's Durable Object classes, then reconcile reads the script settings to learn which classes already exist. Under load that read can 404 "has no versions" even though precreate just observed the class, so migration planning sent `new_sqlite_classes` for the same class again:

```text
Cannot apply new-sqlite-class migration to class 'Counter' that is already depended on by existing Durable Objects
```

When attributes for the script are already held, retry the read briefly before concluding there are no existing settings:

```ts
Effect.retry({
  while: (error) =>
    output !== undefined &&
    (error._tag === "WorkerNotFound" ||
      error._tag === "WorkerHasNoVersions" ||
      error._tag === "DispatchNamespaceScriptNotFound" ||
      error._tag === "DispatchNamespaceNotFound"),
  schedule: Schedule.max([Schedule.exponential(250), Schedule.recurs(6)]),
}),
```

A worker that really is gone still falls through to the upsert after ~16s, so out-of-band deletion self-heals as before.

`DurableObjectPrecreateRace.test.ts` deploys a Durable Object Worker against the real API and answers the two settings reads after precreate with 404 "has no versions".

### DIFF
--- a/packages/alchemy/test/Cloudflare/Workers/WorkerProvider.test.ts
+++ b/packages/alchemy/test/Cloudflare/Workers/WorkerProvider.test.ts
@@ -1,5 +1,6 @@
 import {
   encodeDurableObjectTags,
+  getObservedHostedDurableObjectClassNames,
   getDurableObjectTagMap,
   normalizeStateDomains,
   observeWorkerSettingsForMigration,
@@ -13,7 +14,6 @@ import {
   stateCustomDomains,
   stateWorkerDomain,
 } from "@/Cloudflare/Workers/WorkerProvider";
-import { CloudflareRateLimited } from "@distilled.cloud/cloudflare/Errors";
 import * as workers from "@distilled.cloud/cloudflare/workers";
 import { describe, expect, test } from "alchemy-test";
 import * as Effect from "effect/Effect";
@@ -23,6 +23,18 @@ import * as Schedule from "effect/Schedule";
 
 describe("WorkerProvider", () => {
   describe("observeWorkerSettingsForMigration", () => {
+    test("uses only namespaces for classes hosted by this Worker", () => {
+      expect(
+        getObservedHostedDurableObjectClassNames(
+          {
+            ForeignClass: "foreign-namespace-id",
+            SeatConnector: "hosted-namespace-id",
+          },
+          new Set(["SeatConnector"]),
+        ),
+      ).toEqual(["SeatConnector"]);
+    });
+
     test.effect(
       "retries a missing-settings race after precreate proved the DO exists",
       () =>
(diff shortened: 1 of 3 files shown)


packages/alchemy/test/Cloudflare/Workers/WorkerProvider.test.ts
```diff
@@ -17,6 +19,82 @@ import * as Effect from "effect/Effect";
...
+      const migration = planDurableObjectClassMigration({
+        previousClassName: previousClassNames.ExampleDurableObject,
+        className: "ExampleDurableObject",
+      });
+
+      expect(migration.newSqliteClasses).toEqual([]);
+      expect(previousClassNames).toEqual({
+        ExampleDurableObject: "ExampleDurableObject",
+      });
+    });
```
sam-goodwin: These are not good tests. You should test behavior, not implementation details. Please implement a regression test that reproduces the issue that demanded this fix and get rid of these tests.

sshkeda: Dropped those tests and added a live regression test that reproduces the race. Also simplified the fix to a bounded retry on the settings read.

PR #1298, opened 2026-08-21
feat(stripe): add Stripe resources, auth, and nuke wiring

Mkassabov:
Add a Stripe cloud to Alchemy: `STRIPE_API_KEY` auth, resources with HTTP bindings, live tests, nuke via `Stripe.providers()`, and an effectful Worker example. Rebased onto `main`. Connect and Issuing lifecycles are skipIf-gated (`STRIPE_TEST_CONNECT=1`, `STRIPE_TEST_ISSUING=1`) with ungated entitlement probes.

```ts
import * as Stripe from "alchemy/Stripe";

const product = yield* Stripe.Product("pro-plan", { name: "Pro Plan" });
const retrieve = yield* Stripe.RetrieveProduct(product);
const live = yield* retrieve();
```

Ownership uses Stripe metadata keys `alchemy_stack` / `alchemy_stage` / `alchemy_id` (`:` is not allowed). Objects without a hard delete deactivate (`active: false`). `stacks/nuke.ts` includes `Stripe.providers()`.

Bindings follow the Fly/AWS HTTP pattern: each operation is a `Binding.Service` whose `*Http` layer injects the host token and calls distilled. Provide the `*Http` layer on the Function/Worker Effect with `Effect.provide([...])`. Stripe has no public mint API for restricted keys; `RestrictedApiKey` collects binding permissions and injects a Dashboard RAK or the account secret until that API exists.

Examples (both run end to end against a live test account; see `test/integ.test.ts` in each):

- `examples/stripe-billing` — the SaaS subscription loop on one Worker. Product / Price / Coupon / `BillingPortalConfiguration` declared on the Worker. `POST /checkout` creates a Customer and a subscription-mode Checkout Session and returns the hosted URL. `POST /portal` opens the Billing Portal. `consumeEvents` on `checkout.session.completed` + `customer.subscription.{created,updated,deleted}` + `invoice.payment_failed` keeps an entitlement record in KV that `GET /subscription/:customerId` serves — the thing the app gates paid features on.
- `examples/stripe-connect` — a Connect platform. `POST /merchants` creates an Express account (with `card_payments` / `transfers` capabilities), records it in a D1 `merchants` table, and returns an Account Link with both `return_url` and `refresh_url`. `POST /merchants/:id/onboarding` mints a fresh link for merchants who abandoned the form. `consumeEvents` on `account.updated` copies `details_submitted` / `charges_enabled` / `payouts_enabled` into D1; `GET /merchants/:id` is what the platform checks before letting a merchant sell.

New in this revision: `CreateCheckoutSession` / `CreateBillingPortalSession` bindings, `CustomerSubscription{Created,Updated,Deleted}` and `AccountUpdated` event classes, `checkout_sessions_*` permissions. Docs rewritten as two use-case walkthroughs of those examples (`/stripe/guides/subscriptions`, `/stripe/guides/connect`) plus an events reference; the generic bindings/catalog/sell-from-a-worker pages are folded in.

### consumeEvents

`consumeEvents` provisions the `WebhookEndpoint` itself: it derives the delivery URL from `worker.url`, enables the selected event types, and binds the minted signing secret onto the Worker. No separate endpoint declaration is needed.

```ts
yield* Stripe.consumeEvents("Events", {
  events: [Stripe.CustomerCreated, Stripe.InvoicePaid],
}, Effect.fn(function* (event) {
  yield* Effect.log(event.type);
}));
```

Provide `Stripe.ConsumeEventsLive` on the Worker Effect. Signature failures are 401 (`catchTag` on `StripeWebhookSignatureError` | `StripeWebhookPayloadParseError`); handler defects die (500), never 401.

### Cloudflare engine (needs Cloudflare owner)

This PR also changes Worker engine behavior used by the webhook DAG. Please review before merge:

- **Precreate URL** — when `workersDev` is enabled, precreate now publishes `https://{name}.{subdomain}.workers.dev` instead of `url: undefined`, so cycle peers interpolating `worker.url` get a valid HTTPS URL. Reconcile still overwrites `url` with the final value. `workersDev: false` keeps `url` undefined.
- **`Worker["Binding"].env`** — `worker.bind({ env })` merges into `news.env` (`Redacted` → `secret_text`). Lambda already had this; Worker did not.
- **Fetch+listen dispatch** — when more than one listener returns an Effect, the first `Response` wins (previously `discard: true` dropped webhook 401s). All listeners still run; event-source paths should not also be handled by user `fetch`.

### Runtime

Stripe's Node SDK (`stripe`) runs on Workers via Cloudflare's fetch + WebCrypto support. Alchemy does not wrap that package. Runtime is the distilled HTTP client: `Retrieve*` / `Create*` bindings inject the host token and call `@distilled.cloud/stripe`.

`@stripe/stripe-js` is the PCI browser SDK (Elements / Checkout, publishable key, must load from `js.stripe.com`). It is not a Worker/Lambda binding and is out of scope here.

### Product

```ts
const product = yield* Stripe.Product("pro-plan", {
  name: "Pro Plan",
  description: "Billed monthly",
});
```

```ts
const retrieve = yield* Stripe.RetrieveProduct(product);
const live = yield* retrieve();
```

`RetrieveProduct` also accepts a `prod_…` id string or an `Effect<Product>`.

### Price

```ts
  product: product.id,
  currency: "usd",
  unitAmount: 2000,
});
```

```ts
const retrieve = yield* Stripe.RetrievePrice(price);
const live = yield* retrieve();
```

### Coupon

```ts
const coupon = yield* Stripe.Coupon("launch-20", {
  percentOff: 20,
  duration: "once",
  name: "Launch 20%",
});
```

```ts
const retrieve = yield* Stripe.RetrieveCoupon(coupon);
const live = yield* retrieve();
```

### PromotionCode

```ts
const promo = yield* Stripe.PromotionCode("launch", {
  coupon: coupon.id,
  code: "LAUNCH20",
});
```

```ts
const retrieve = yield* Stripe.RetrievePromotionCode(promo);
const live = yield* retrieve();
```

### Customer

```ts
const create = yield* Stripe.CreateCustomer();
const created = yield* create({ email: "alice@example.com" });
```

### PaymentLink

```ts
const checkout = yield* Stripe.PaymentLink("checkout", {
  lineItems: [{ price: price.id, quantity: 1 }],
  allowPromotionCodes: true,
});
```

### Connect: Account, AccountPerson, AccountExternalAccount

Merchants arrive one at a time, so connected accounts are created from a request handler with the `CreateAccount` / `CreateAccountLink` bindings:

```ts
const createAccount = yield* Stripe.CreateAccount();
const createAccountLink = yield* Stripe.CreateAccountLink();

const account = yield* createAccount({ type: "express", country: "US", email });
const link = yield* createAccountLink({
  account: account.id,
  type: "account_onboarding",
  return_url: `${url.origin}/onboarded`,
});
```

Accounts you already know are Stack resources:

```ts
const person = yield* Stripe.AccountPerson("Owner", {
  account: merchant.id,
  firstName: "Jane",
  lastName: "Diaz",
  relationship: { representative: true, title: "CEO" },
});
```

### WebhookEndpoint

`consumeEvents` creates the endpoint for a Worker automatically. Declare one directly only when you own the delivery URL:

```ts
const webhook = yield* Stripe.WebhookEndpoint("Events", {
  url: "https://example.com/webhooks/stripe",
  enabledEvents: [Stripe.CustomerCreated, Stripe.CheckoutSessionCompleted],
});
```


sam-goodwin:
Build an example with effectful bindings showing common use-cases.
Update PR description with a section and code snippet per resource and also show bindings.
Does stripe have any runtime component? I vaguely recall them having their own subset of JS?

Mkassabov:
Stripe's Node SDK (`stripe`) is what runs on Workers — Cloudflare added fetch + WebCrypto support so `stripe.customers.create()` works at the edge instead of raw REST. Alchemy does not wrap that package.

Runtime here is the distilled HTTP client: `Retrieve*` / `Create*` bindings inject `STRIPE_API_KEY` onto the host (Worker `secret_text`, Lambda/Fly env) and call `@distilled.cloud/stripe`. Same Effect-typed operations as the provider.

`@stripe/stripe-js` is the PCI browser SDK (Elements / Checkout, publishable key, must load from `js.stripe.com`). Not a Worker/Lambda binding; out of scope for this PR.

packages/alchemy/src/Stripe/StripeHttp.ts
```diff
@@ -0,0 +1,200 @@
...
+export const makeStripeAuth = (
+  ambient: Context.Context<Credentials | HttpClient.HttpClient>,
+): StripeAuth => ({
+  authorize: <A, E>(
+    eff: Effect.Effect<A, E, Credentials | HttpClient.HttpClient>,
+  ): Effect.Effect<A, E, RuntimeContext> => {
+    if (globalThis.__ALCHEMY_RUNTIME__) {
+      return eff.pipe(
+        Effect.provide(CredentialsFromEnv),
+        Effect.provide(FetchHttpClient.layer),
```
sam-goodwin: Don't do subsequent Effect.provode. Build the layer once. WE really need a linter rule for this

packages/alchemy/src/Stripe/StripeHttp.ts
```diff
@@ -0,0 +1,200 @@
...
+
+export const stripeApiKey = (
+  ambient: Context.Context<Credentials | HttpClient.HttpClient>,
+): Effect.Effect<string> =>
+  Credentials.pipe(
+    Effect.provideContext(ambient),
+    Effect.flatMap((resolve) => resolve),
+    Effect.map((cfg) => Redacted.value(cfg.apiKey)),
+    Effect.orDie,
+  );
```
sam-goodwin: Can we create one instead? Is there a resource to create a Stripe API token instead of using the User's one.

Does stripe have a notion of scopes/permissions on their tokens we can use. Ideally, each binding uses `Token.bind` to inject its scopes. 

Look at how cloudflare http bindings work with AccountApiToken

packages/alchemy/src/Stripe/StripeHttp.ts
Mkassabov: Done — `CredentialsFromEnv` and `FetchHttpClient.layer` are merged once at module scope, then provided as a single layer:

```ts
const RuntimeLayer = CredentialsFromEnv.pipe(
  Layer.provideMerge(FetchHttpClient.layer),
);

// runtime
eff.pipe(Effect.provide(RuntimeLayer))
```

packages/alchemy/src/Stripe/StripeHttp.ts
Mkassabov: Added `Stripe.RestrictedApiKey`, same shape as Cloudflare `AccountApiToken`. Each HTTP binding does `token.bind(tag, { permissions })` so scopes collect onto one host token.

Stripe does not expose a public mint for restricted keys — `POST /v2/iam/api_keys` is Stripe Apps private preview and 404s on this account. Until that API is GA, the token value is a Dashboard `rk_…` (if supplied) or the account secret. Bindings still declare the permissions they need so a mint can honor them later.

Mkassabov: stripe does have scoped permissions and they can issue tokens, but it seems you can't do it over api only via the dash.

packages/alchemy/src/Stripe/StripeHttp.ts
```diff
@@ -0,0 +1,252 @@
...
+    eff: Effect.Effect<A, E, Credentials | HttpClient.HttpClient>,
+  ): Effect.Effect<A, E, RuntimeContext> => {
+    if (options.credentials === undefined || options.http === undefined) {
+      return Effect.die(
+        "Stripe HTTP binding missing Credentials or HttpClient at plan time",
+      ) as Effect.Effect<A, E, RuntimeContext>;
+    }
+    return eff.pipe(
+      Effect.provideService(Credentials, options.credentials),
+      Effect.provideService(HttpClient.HttpClient, options.http),
```
sam-goodwin: Don't do double `Effect.provide*`, use ast-grep to identify these mistakes and replace with a single Effect.provide and `Layer.*` to build a single provided layer

packages/alchemy/src/Stripe/StripeHttp.ts
```diff
@@ -0,0 +1,252 @@
...
+      Effect.flatMap((apiKey) =>
+        eff.pipe(
+          Effect.provideService(
+            Credentials,
+            Effect.succeed({
+              apiKey,
+              apiBaseUrl: DEFAULT_API_BASE_URL,
+            } satisfies StripeCredentialsConfig),
+          ),
+          Effect.provide(FetchHttpClient.layer),
```
sam-goodwin: Double Effect.provide ...

packages/alchemy/src/Stripe/StripeHttp.ts
```diff
@@ -0,0 +1,252 @@
...
+        permissions: [...permissions],
+      });
+      // Lambda/ECS/Fly ship env through their binding contract. Workers
+      // pick the secret up from RuntimeContext.set via `yield* token.value`
+      // below — do not host.bind({ env }) or hand-build secret_text.
+      if (isBindingHost(host) || isFlyHost(host)) {
+        yield* asEnvHost(host).bind`${host}`({
+          env: { [STRIPE_API_KEY_ENV]: token.value },
+        });
+      }
```
sam-goodwin: You don't need to do this if `token.value` is an `Output`

You just do
```ts
const valueAccessor = yield* token.value`
```

Outputs are alreayd generalized so yielding them can work in any platform. That's best way to transport values from plan to runtime. And it means you don't need this asEnvHost stuff

website/src/content/docs/stripe/tutorial/part-2.mdx
```diff
@@ -0,0 +1,140 @@
...
++    currency: "usd",
++    unitAmount: 2000,
++  });
+
+  return {
+    productId: product.id,
+    name: product.name,
+  };
+}),
+```
```
sam-goodwin, line 126: I don't like putting it in an `Effect.gen`, show it in a Cloudflare.Worker, not a dangling Effect gen

website/src/content/docs/stripe/tutorial/part-3.mdx
```diff
@@ -0,0 +1,321 @@
...
+    state: Alchemy.localState(),
+  },
+  Effect.gen(function* () {
+-    const product = yield* Stripe.Product("Pro", {
+-      name: "Pro",
+-      description: "Billed monthly",
+-    });
++    const product = yield* Pro;
+    const price = yield* Stripe.Price("ProMonthly", {
+      product: product.id,
```
sam-goodwin: nit: ideally you'd be able to just do `product: Pro`

sam-goodwin: `product: string | Product | Effect<Product>`

website/src/content/docs/stripe/tutorial/part-3.mdx
```diff
@@ -0,0 +1,321 @@
...
+
+```diff lang="typescript"
+-  }).pipe(Effect.provide(Stripe.RetrieveProductHttp)),
++  }).pipe(
++    Effect.provide(
++      Layer.mergeAll(
++        Stripe.RetrieveProductHttp,
++        Stripe.CreateCustomerHttp,
++      ),
++    ),
```
sam-goodwin: Don't use Layer.mergeAll here, do `Effect.provide([..])`

website/src/content/docs/stripe/tutorial/part-3.mdx
```diff
@@ -0,0 +1,321 @@
...
++  const customer = yield* createCustomer({
++    email: body.email,
++    ...(body.name !== undefined ? { name: body.name } : {}),
++  }).pipe(Effect.orDie);
++  return yield* HttpServerResponse.json(
++    { id: customer.id, email: customer.email },
++    { status: 201 },
++  );
++}
+return HttpServerResponse.text("ok");
```
sam-goodwin: I don't really like code snippets without the surrounding context. Like, where is this code?

website/src/content/docs/stripe/index.mdx
```diff
@@ -0,0 +1,67 @@
...
+});
+
+const price = yield* Stripe.Price("ProMonthly", {
+  product: product.id,
+  currency: "usd",
+  unitAmount: 2000,
+  recurring: { interval: "month" },
+});
+```
+
```
sam-goodwin, line 58: This doesn't seem very complete?

website/src/content/docs/stripe/setup.mdx
```diff
@@ -0,0 +1,60 @@
...
+shape as a Cloudflare `AccountApiToken`. Stripe has no public API to
+mint restricted keys (`POST /v2/iam/api_keys` is Apps preview), so
+the resource stores those permissions and injects either:
+
+- a Dashboard-created restricted key (`rk_test_…` / `rk_live_…`)
+  passed as `value`, or
+- the account secret from the profile / `STRIPE_API_KEY`
+
+Until Stripe ships a public mint, treat the collected permissions as
+the intended scope, not as an enforced key policy.
```
sam-goodwin: This is so sad

website/src/content/docs/stripe/guides/bindings.mdx
```diff
@@ -0,0 +1,93 @@
+---
+title: Bindings
+description: Retrieve and create Stripe objects from a Worker — Binding.Service, *Http layers, RestrictedApiKey, and RuntimeContext.
```
sam-goodwin: This doesn't look very complete and is a bit generic "bindings"

Guides should probably be use-case based, not mechanic/implementation detail based.

Think from: "what does the user want to acheive"

website/src/content/docs/stripe/guides/webhooks.mdx
sam-goodwin, line 1: Webhooks should be two levels:
1. is whatever resource is required to manage its lifecycle
2. the other is an Event Soruce
```ts
yield* Stripe.consumeEvents("Events", { enabledEvents: ["customer.created", ..] }, (event: EFfect | Stream) => ..) 
```

Effect or Stream is use-case dependent. I'd say Effect probably makes mpore sense here (one effect evaluated per event received through the web hook)

packages/alchemy/src/Stripe/EventSource.ts
```diff
@@ -0,0 +1,241 @@
...
+      props: ConsumeEventsProps,
+      process: (
+        event: StripeEventInstance,
+      ) => Effect.Effect<void, never, never>,
+    ) {
+      const path = webhookPath(props.path);
+      const secretKey = webhookSecretEnvName(path);
+      const byType = new Map(
+        props.events.map((event) => [event.type, event] as const),
+      );
```
sam-goodwin, line 159: What is this? Where does it create the WebhookEndpoint resource?

packages/alchemy/src/Stripe/EventSource.ts
```diff
@@ -0,0 +1,241 @@
...
+      [webhookSecretEnvName(path)]: secret,
+    },
+  });
+
+/**
+ * Subscribe to Stripe webhook events on the host Worker.
+ *
+ * `consumeEvents` only listens and verifies — it does not create the
+ * {@link WebhookEndpoint}. Declare that resource in the Stack after the
+ * Worker (it needs `worker.url`), then {@link bindWebhookSecret}.
```
sam-goodwin, line 63: This is wrong. It MUST create the WebhookEndpoint

packages/alchemy/src/Stripe/EventSource.ts
```diff
@@ -0,0 +1,241 @@
...
+  EventSourceService
+>()("Stripe.EventSource") {}
+
+/**
+ * Cloudflare Worker implementation of {@link consumeEvents}.
+ *
+ * @layer
+ * @provides Stripe.EventSource
+ */
+export const ConsumeEventsLive = Layer.effect(
```
Mkassabov, line 144: this file should probably be called ConsumeEvents.ts to follow conventions of the codebase

examples/stripe-billing/alchemy.run.ts
```diff
@@ -0,0 +1,27 @@
...
+  {
+    providers: Layer.mergeAll(Cloudflare.providers(), Stripe.providers()),
+    state: Alchemy.localState(),
+  },
+  Effect.gen(function* () {
+    const api = yield* Api;
+    const webhook = yield* Stripe.WebhookEndpoint("Events", {
+      url: Output.interpolate`${api.url}/webhooks/stripe`,
+      enabledEvents: [Stripe.CustomerCreated, Stripe.CheckoutSessionCompleted],
+    });
```
Mkassabov, line 20: this is wrong; this should get automatically created

website/src/content/docs/stripe/guides/sell-from-a-worker.mdx
Mkassabov, line 1: this is essentially just a code snippet and doesn't explain much

website/src/content/docs/stripe/guides/connect.mdx
Mkassabov, line 1: nobody actually uses connect this way. we should make an example on how to set up connect and endpoints so users can actually connect accounts with stripe account (using d1 as a simple database).

then this doc should show how to build the stripe parts of that piece by piece

PR #1300, opened 2026-08-21
fix(cloudflare/workers): preserve adoption checks with unresolved bindings

### LINKED ISSUE #1299: Unresolved Worker bindings skip ownership checks during adoption
## Summary

A fresh `Cloudflare.Worker` with unresolved resource-backed bindings can skip its cold ownership read and plan `create`. Worker precreate/reconcile later resolves the explicit physical name and can upsert an existing foreign script even though `AdoptPolicy` never cleared ownership.

This blocks a safe Wrangler-to-Alchemy transfer because the plan cannot prove that the existing Worker is owned or explicitly adopted.

## Environment

- `alchemy@2.0.0-beta.72`
- Bun 1.3.7
- Existing Cloudflare Worker managed by Wrangler
- Remote `Cloudflare.state()`
- Existing physical Worker name: `www`
- Fresh stack: `MaelstromWww/prod`

The same planner gate is present on current `main`.

## Reproduction

Deploy a Worker named `www` outside Alchemy. In a fresh stack, declare the same explicit name with resource-backed bindings:

```ts
export const Website = Cloudflare.Worker(
  "Website",
  Effect.gen(function* () {
    const database = yield* ExistingDatabaseRef;
    const secret = yield* ExistingSecretRef;

    return {
      name: "www",
      main: "./dist/server/entry.mjs",
      bundle: false,
      assets: { directory: "./dist/client" },
      env: {
        DATABASE: database,
        SECRET: secret,
      },
    };
  }),
);
```

Run:

```sh
bun alchemy deploy --dry-run --adopt --yes --stage prod
```

## Actual behavior

```text
Plan: 1 to create
[Website] create
[Website/DATABASE] create
[Website/SECRET] create
```

The existing script is healthy and readable through Cloudflare's API. No apply was performed.

## Root cause

Fresh referenced resources remain unresolved `ResourceExpr` values in desired Worker props. The cold-adoption probe currently requires the complete `news` object to satisfy `isResolved(news)`, so it skips `provider.read` before the Worker provider can inspect the explicit name.

Relevant paths on current main:

- unresolved resource expressions: `packages/alchemy/src/Plan.ts`
- full-object cold-adoption gate: `packages/alchemy/src/Plan.ts`
- Worker read and upsert lifecycle: `packages/alchemy/src/Cloudflare/Workers/WorkerProvider.ts`

The explicit name is available. If `Worker.read` runs, the planner passes projected desired props as `olds` and the provider can return `Unowned(attrs)`.

## Expected behavior

Unresolved non-identity inputs must not suppress the ownership check when the provider can safely project a resolved physical identity.

- Without `--adopt`, planning must fail with `OwnedBySomeoneElse`.
- With `--adopt`, planning must produce an adopted `update`, never a blind create.
- Unresolved identity inputs must remain fail-closed.
- Provider read failures must propagate instead of degrading into create.

## Fix

PR #1300 adds an opt-in provider read projection and a Worker projection that requires resolved script-name, namespace, and version-parent identities while omitting unresolved bindings and observation inputs.

## Regression coverage

- Providers without projections retain the existing unresolved-input gate.
- Projected props enforce ownership rejection and explicit adoption.
- Projection `None` skips reads for unresolved identity.
- Invalid unresolved projections fail before read.
- Provider read errors remain fatal.
- Worker name, nested namespace, version parent, domain, route, and cron inputs are covered.

NachoVazquez:
## TL;DR

A fresh Worker with resource-backed `env` bindings skipped its ownership read and planned `create`, allowing reconcile to upsert an existing foreign script without `AdoptPolicy`. This adds provider-owned read projections so Workers can probe a resolved physical identity while unresolved non-identity bindings remain deferred.

**Files to review (6, +359 / -13):**

| File | Why |
|---|---|
| `packages/alchemy/src/Plan.ts` *(start here)* | Applies projections only during unresolved cold-adoption reads and validates that projected props are fully resolved. |
| `packages/alchemy/src/Provider.ts` | Defines the optional provider projection contract. |
| `packages/alchemy/src/Cloudflare/Workers/WorkerProvider.ts` | Projects Worker identity and safe observation inputs without resolving bindings. |
| `packages/alchemy/test/plan.test.ts` | Covers ownership rejection, explicit adoption, default behavior, read failures, and invalid projections. |
| `packages/alchemy/test/Cloudflare/Workers/WorkerProvider.test.ts` | Covers Worker name, namespace, version-parent, and optional observation inputs. |
| `packages/alchemy/test/test.resources.ts` | Exposes the provider hook to planner lifecycle tests. |

## Why

Before this change, an existing Worker named `www` with fresh D1, Images, Email, and secret bindings produced:

```text
Plan: 1 to create
[Website] create
[Website/EMAIL] create
[Website/IMAGES] create
[Website/TURNSTILE_SECRET_KEY] create
[Website/WAITLIST_DB] create
```

The unresolved bindings made the complete desired props fail `isResolved`, so planning skipped `provider.read`. Worker precreate and reconcile later resolved the explicit name and could overwrite the foreign script through the normal upsert path.

## How

1. Providers may opt into unresolved cold reads by projecting desired props to the resolved subset their `read` operation needs.
2. Planning rejects any `Some` projection that still contains unresolved values, and provider read errors propagate unchanged.
3. The Worker provider requires concrete script-name, dispatch-namespace, and version-parent identities. It strips unresolved bindings and omits optional domain, route, or cron observation surfaces only when those declarations are unresolved.

## Reviewer notes

- **Existing providers keep the old gate.** Providers without `resolveReadProps` still skip cold reads whenever desired props are unresolved.
- **Identity stays fail-closed.** An unresolved Worker name, nested namespace name, or version parent returns `None`; the planner never guesses a physical resource.
- **Reconcile keeps full desired props.** The projection is read-only. Adoption still forces an update using the original desired configuration and bindings.
- **No read-error fallback.** Authentication, permission, schema, and API failures stop planning instead of degrading to create.

## Tests

- `pnpm exec tsc -b`
- `timeout 240 pnpm test test/plan.test.ts --profile testing` — 163 passed
- `timeout 240 pnpm test test/Cloudflare/Workers/WorkerProvider.test.ts --profile testing` — 44 passed
- `pnpm lint:providers`
- `pnpm format:check`

No live Cloudflare test was run. Generic planner lifecycle tests and Worker projection tests cover the ownership and provider-specific halves independently.

## Links

- Closes #1299

### DIFF
--- a/packages/alchemy/src/Cloudflare/Workers/WorkerProvider.ts
+++ b/packages/alchemy/src/Cloudflare/Workers/WorkerProvider.ts
@@ -18,12 +18,19 @@ import { Unowned } from "../../AdoptPolicy.ts";
 import * as Artifacts from "../../Artifacts.ts";
 import type { ScopedPlanStatusSession } from "../../Cli/Cli.ts";
 import { hashDirectory, type MemoOptions } from "../../Command/Memo.ts";
-import { havePropsChanged, isResolved, stripEffects } from "../../Diff.ts";
+import {
+  havePropsChanged,
+  isResolved,
+  stripEffects,
+  stripUnresolved,
+} from "../../Diff.ts";
+import type { Input } from "../../Input.ts";
 import * as ProviderLayer from "../../Local/ProviderLayer.ts";
 import * as Provider from "../../Provider.ts";
 import { type ResourceBinding } from "../../Resource.ts";
 import { Stack } from "../../Stack.ts";
 import { cachedFunction } from "../../Util/cached-function.ts";
+import { isPlainData } from "../../Util/data.ts";
 import { initialCwd } from "../../Util/Node.ts";
 import { sha256Object } from "../../Util/sha256.ts";
 import { CloudflareEnvironment } from "../CloudflareEnvironment.ts";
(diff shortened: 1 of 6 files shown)


NachoVazquez:
Validated this fix against a production Worker using the published `alchemy@2.0.0-beta.72` package with the three runtime files backported as a Bun patch.

Before the patch, the fresh stack planned the explicitly named Worker and its resource-backed bindings as creates. With this fix, the reviewed adoption plan became:

```text
Plan: 1 to update
[Website] update
[Website/EMAIL] create
[Website/IMAGES] create
[Website/TURNSTILE_SECRET_KEY] create
[Website/WAITLIST_DB] create
```

The adoption applied successfully, including 29 static assets and two custom domains. A second dry run converged to:

```text
Plan: 1 to noop
[Website] noop
```

Production HTTP checks passed, and a non-writing waitlist probe confirmed the D1, Email, and Turnstile bindings are present. This covers the original resource-backed `env` reproduction end to end.

NachoVazquez:
Closing in favor of #1317, which carries the same fix ported from the patch we have been running in production, plus focused unit tests for `resolveWorkerReadProps`.

This PR additionally added engine-level adoption tests in `plan.test.ts` covering the full reconcile path rather than just the read projection. Happy to rebase those onto #1317 if reviewers want that coverage there.

PR #1306, opened 2026-08-21
fix(test): use local providers in actions when local is forced

Vicentesan:
## Summary

- Keep the Floci AWS override active during stack plan/apply.
- Ensure AWS data-plane calls inside `Action` use Floci when `ALCHEMY_TEST_DEV=1`.
- Add an S3 `PutObject` regression test with a sentinel live AWS environment.

## Root cause

`scratchStack` applied the Floci override to the test body and initial stack program, but `Action.Run` executes later during `Plan.make → apply` under `compiled.services`. The closer live AWS layer could therefore win, causing data-plane calls inside Actions to target real AWS while resource lifecycle operations remained local.

## Testing

- Focused Action/S3 regression test
- Full S3 Floci data-plane test suite: 10/10 passing
- Workspace TypeScript build
- Formatting and `git diff --check`

### DIFF
--- a/packages/alchemy/src/Test/Core.ts
+++ b/packages/alchemy/src/Test/Core.ts
@@ -530,11 +530,12 @@ export const scratchStack = <ROut>(
       ? Layer.succeed(State.State, State.InMemoryService({}))
       : Layer.provide(State.localState(), PlatformServices);
 
-  // `withProviders` already pins test-body distilled to Floci, but the stack
-  // *program* is composed under `AWS.providers()`'s live Endpoint. Dual
-  // providers still hit Floci in lifecycle ops; user-program distilled
-  // (e.g. ECS Service `findHostedZoneId`) would otherwise query real AWS.
-  const pinProgramToFloci = <A, E, R>(
+  // `withProviders` already pins the test body to Floci, but the stack program
+  // and its later plan/apply phase run under `AWS.providers()`'s live services.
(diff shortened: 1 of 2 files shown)


packages/alchemy/src/Test/Core.ts
```diff
@@ -530,27 +530,27 @@ export const scratchStack = <ROut>(
...
+    (pinToFloci(effect) as Effect.Effect<any, any, never>).pipe(
       makeStack({
         name: stackName,
         providers: options.providers,
         state: stateLayer,
       } as any) as any,
       Effect.flatMap((compiled: any) =>
-        Plan.make(compiled).pipe(
-          Effect.flatMap(apply),
+        pinToFloci(Plan.make(compiled).pipe(Effect.flatMap(apply))).pipe(
```
sam-goodwin: What is going on here lol? .pipe .pipe?

Vicentesan: codex was high lol

sam-goodwin:
Thanks @Vicentesan — your root-cause analysis was spot on. While verifying it end-to-end we found the same gap exists in the production `alchemy dev` path (not just the test harness), so we've superseded this with #1308, which routes every binding client to its bound resource's data plane (local → emulator, `Alchemy.remote()` → real cloud) at the engine level. Your harness fix and regression test are included there with authorship preserved. Closing in favor of #1308.

PR #1313, opened 2026-08-22
feat(cloudflare): deploy Foldkit's fetch handler as the Worker and derive asset routing from its build

filipfalcon:
`Cloudflare.Website.Foldkit` deploys the Worker handler and assets emitted by Foldkit's Vite plugin through the existing `props.vite` path, the same path used for TanStack Start. Client-only apps get SPA fallback; SSR and prerendered builds use `foldkit.build.json` so unmatched requests reach Foldkit's handler instead of an unrendered HTML template. Explicit asset configuration takes precedence.

```ts
const site = yield* Cloudflare.Website.Foldkit("Website");
```

`Website.Foldkit` sets `vite.framework: "foldkit"`. After the normal Vite build completes, the source reads the manifest, applies routing defaults, and rejects a conflicting custom `main` when Foldkit emitted its SSR handler. These rules live in `Cloudflare/Website/FoldkitBuild.ts`; build, hashing, deployment, and development reuse the existing Vite machinery.

The change also:

- Updates Foldkit to 0.163.0 and its Vite plugin to 0.24.0, with Effect rc.116 required by Foldkit's peers.
- Routes built-in Vite workers through the source pipeline instead of a duplicate WorkerProvider build path.
- Makes a custom `main` replace the configured Vite entry instead of merging entry points.
- Keeps Foldkit SSR runnable in Vite during development and removes the obsolete Worker wrapper from the SSR example.
- Adds contracts covering generated handlers/manifests, asset overrides, custom output directories, entry conflicts, and SSR/custom-Worker development.

Validation uses the repository's unchanged PR workflows. Integration checks requiring contributor credentials are left to contributors. Local tests and typechecks are intentionally not run.


sam-goodwin:
Should we move this out of draft?

packages/alchemy/src/Cloudflare/Website/Foldkit.ts
```diff
@@ -66,32 +66,46 @@ export interface FoldkitProps<
...
+   *   `htmlHandling` already resolves those, and `"single-page-application"`
+   *   is wrong here: it answers an unknown path with the rendered `/` at 200
+   *   instead of a 404.
+   * - **Server-rendered.** Nothing is prerendered, so `index.html` is a bare
+   *   template no browser should receive. Set `htmlHandling: "none"` and
+   *   `notFoundHandling: "none"` so `/` and every other page fall through to
+   *   the Worker named by {@link FoldkitProps.main}.
+   * - **Hybrid.** Prerendered routes are files and the rest fall through:
+   *   the default `htmlHandling` plus `notFoundHandling: "none"`. Prerender
+   *   `/` as well, or `htmlHandling` resolves it to the bare template.
```
sam-goodwin, line 90: This is too much text. Break it up with illustrative code snippets instead of dumping bullet point text describing code in natural language

filipfalcon: Yeah I'm still working on it. Please leave it in draft if possible.

PR #1316, opened 2026-08-23
fix(bundle): preserve namespace initialization with Rolldown 1.2.5

aryasaatvik:
## Summary

Upgrade Rolldown to 1.2.5 to prevent generated Lambda bundles from consuming nested namespace exports before their initializer has run.

## Reproduction

A Lambda entry that accesses a resource through Alchemy's broad Cloudflare namespace can trigger the failure:

```ts

export const infrastructure = Effect.gen(function* () {
  const app = yield* Cloudflare.Flagship.App("Flags", {
    name: "example-flags",
  });

  return app;
});
```

With Rolldown 1.1.5, the generated bundle can defer the `App` initializer but consume `App` without invoking that initializer first. The relevant output is effectively:

```js
var App;

var init_App = __esm(() => {
  App = Resource("Cloudflare.Flagship.App");
});

const infrastructure = Effect.gen(function* () {
  // Missing init_App() before this use.
  const app = yield* App("Flags", {
    name: "example-flags",
  });

  return app;
});
```

Executing that bundle fails during initialization:

```text
TypeError: App is not a function
```

Rolldown 1.2.5 emits the required initialization before the namespace member is consumed:

```js
init_App();

const infrastructure = Effect.gen(function* () {
  const app = yield* App("Flags", {
    name: "example-flags",
  });

  return app;
});
```

## Before / After

| Rolldown | Generated Lambda bundle |
| --- | --- |
| 1.1.5 | Fails during execution with `TypeError: App is not a function` |
| 1.2.5 | Initializes `Cloudflare.Flagship.App` before use and executes successfully |

## Changes

- Pin Rolldown 1.2.5 in the build catalog and workspace override.
- Add an executable regression covering Alchemy's real `sideEffects: false` Cloudflare namespace.
- Generate the test bundle through `makeFunctionBundler.resolveBundlePlan` so it exercises the Lambda bundling configuration.
- Execute the emitted ESM rather than relying only on generated-text assertions.

The regression test remains in `packages/alchemy/test/Bundle/NamespaceInitialization.test.ts`. Although this is closer to an integration-level bundling regression than a conventional unit test, keeping it in the Bundle suite protects the exact failure boundary.

## Validation

- `pnpm test test/Bundle/NamespaceInitialization.test.ts`
- `pnpm test test/Bundle`
  - 103 passed
  - 2 skipped


packages/alchemy/package.json
```diff
@@ -439,7 +439,7 @@
     "@octanejs/vite-plugin": "catalog:frontend",
     "@opennextjs/aws": "catalog:frontend",
     "@opennextjs/cloudflare": "catalog:frontend",
-    "@oxc-project/types": "^0.139.0",
+    "@oxc-project/types": "0.146.0",
```
sam-goodwin: Why pin the version?

PR #1317, opened 2026-08-23
fix(cloudflare): strip unresolved input leaves from cold Worker adoption reads

### LINKED ISSUE #1299: Unresolved Worker bindings skip ownership checks during adoption

NachoVazquez:
A fresh `Cloudflare.Worker` whose props contain unresolved Outputs (resource-backed bindings resolved later in the same plan) skips the cold adoption read entirely, so an existing foreign script can be planned as `create` without any ownership check.

Fixes #1299.

- Add an optional `resolveReadProps` hook to `ProviderService`. `Plan.ts` calls it when desired props are unresolved and would otherwise skip the adoption probe; a projection that is still unresolved dies loudly instead of reaching the provider.
- The Cloudflare Worker provider projects cold-read props through `resolveWorkerReadProps`: identity fields (`name`, dispatch `namespace`, `version.parent`) must be concrete or the read returns `None`; other unresolved leaves (e.g. bindings in `env`) are stripped; unresolved `domain`/`routes`/`crons` surfaces are dropped since reconcile performs the forced adoption update from the original desired props afterward.

```diff
 const readProps =
   oldState === undefined && provider.read && !reusesMigratedFqn
     ? isResolved(news)
       ? Option.some(news)
       : (provider.resolveReadProps?.(news) ?? Option.none())
     : Option.none();
```

### DIFF
--- a/packages/alchemy/src/Cloudflare/Workers/WorkerProvider.ts
+++ b/packages/alchemy/src/Cloudflare/Workers/WorkerProvider.ts
@@ -18,12 +18,19 @@ import { Unowned } from "../../AdoptPolicy.ts";
 import * as Artifacts from "../../Artifacts.ts";
 import type { ScopedPlanStatusSession } from "../../Cli/Cli.ts";
 import { hashDirectory, type MemoOptions } from "../../Command/Memo.ts";
-import { havePropsChanged, isResolved, stripEffects } from "../../Diff.ts";
+import {
+  havePropsChanged,
+  isResolved,
+  stripEffects,
+  stripUnresolved,
+} from "../../Diff.ts";
+import type { Input } from "../../Input.ts";
 import * as ProviderLayer from "../../Local/ProviderLayer.ts";
 import * as Provider from "../../Provider.ts";
 import { type ResourceBinding } from "../../Resource.ts";
 import { Stack } from "../../Stack.ts";
 import { cachedFunction } from "../../Util/cached-function.ts";
+import { isPlainData } from "../../Util/data.ts";
 import { initialCwd } from "../../Util/Node.ts";
 import { sha256Object } from "../../Util/sha256.ts";
 import { CloudflareEnvironment } from "../CloudflareEnvironment.ts";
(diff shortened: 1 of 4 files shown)


NachoVazquez:
Supersedes #1300 (closed in its favor). That PR carried the same fix plus broader engine-level adoption tests in `plan.test.ts`; those can be rebased here if reviewers want reconcile-path coverage.

PR #1318, opened 2026-08-23
fix(deps): update to effect 4.0.0-rc.111

Mkassabov:
Bump Effect from `4.0.0-rc.110` to `4.0.0-rc.111` (catalog, overrides, lockfile, and the remaining hard pins).

```yaml
effect: ">=4.0.0-rc.111 || >=4.0.0"
```

Also points `distilled` at the matching bump: [alchemy-run/distilled#474](https://github.com/alchemy-run/distilled/pull/474)

### DIFF
--- a/packages/cloudflare-runtime/package.json
+++ b/packages/cloudflare-runtime/package.json
@@ -779,7 +779,7 @@
   "inlinedDependencies": {
     "capnp-es": "0.0.14",
     "capnweb": "0.8.0",
-    "effect": "4.0.0-rc.111",
+    "effect": "catalog:effect",
     "http-cache-semantics": "4.2.0",
     "postal-mime": "2.7.5"
   }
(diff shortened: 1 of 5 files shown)


packages/cloudflare-runtime/package.json
```diff
@@ -779,7 +779,7 @@
   "inlinedDependencies": {
     "capnp-es": "0.0.14",
     "capnweb": "0.8.0",
-    "effect": "4.0.0-rc.110",
+    "effect": "4.0.0-rc.111",
```
sam-goodwin: Should use the catalog

sam-goodwin: Done in e1b9d8580c — `inlinedDependencies.effect` now reads `catalog:effect`.

PR #1329, opened 2026-08-23
feat(cloudflare/r2): enable the r2.dev managed domain on Bucket

### LINKED ISSUE #1328: Cloudflare.R2.Bucket cannot enable the r2.dev managed domain, though the SDK ships both operations
## Summary

`Cloudflare.R2.Bucket` cannot enable Cloudflare's R2-managed (`r2.dev`) domain, and exposes no attribute carrying the resulting hostname. A bucket therefore has no publicly readable URL unless a custom domain is attached to it.

The gap is only at the resource layer: `@distilled.cloud/cloudflare/r2` — which `Bucket.ts` already imports on line 1 — ships both the read and the write operation for the managed domain, fully typed. Nothing calls them.

Use cases:

- serving objects from a plain `` or `<a href=…>`, where the URL is persisted and rendered later so a signed S3 URL is not an option
- environments created freely (per branch, per developer, per preview), where a zone subdomain per stack instance is not workable because their number is unbounded

`wrangler r2 bucket dev-url enable` and the Cloudflare dashboard both expose this, so it is not an intentionally restricted capability of the platform.

## Current behavior

`BucketProps` carries only `domains`, whose members are custom domains (`{ name, zone?, enabled?, ciphers?, minTLS? }`). There is no managed-domain prop:

https://github.com/alchemy-run/alchemy/blob/208623a25390f38ac4a16fd593bfa399b06c0604/packages/alchemy/src/Cloudflare/R2/Bucket.ts#L157-L161

None of the resource's attributes can yield the `r2.dev` hostname:

https://github.com/alchemy-run/alchemy/blob/208623a25390f38ac4a16fd593bfa399b06c0604/packages/alchemy/src/Cloudflare/R2/Bucket.ts#L194-L207

`reconcile` converges custom domains, lifecycle rules and CORS. The managed domain is never observed and never written:

https://github.com/alchemy-run/alchemy/blob/208623a25390f38ac4a16fd593bfa399b06c0604/packages/alchemy/src/Cloudflare/R2/Bucket.ts#L1123-L1141

A grep over `packages/` for `BucketDomainManaged` returns nothing. The only mentions of `r2.dev` anywhere in the repository are two comments, in the `Bucket.ts` CORS docblock and in `Cloudflare/CloudConnector/Rules.ts`.

## The SDK operations already exist

Both are typed, and both responses carry `domain`, documented as *"Domain name of the bucket's `r2.dev` domain"* — exactly the value missing from the attributes:

https://github.com/alchemy-run/distilled/blob/dd2a32258180b1176513dc2bb2665cfa13f10ebf/packages/cloudflare/src/services/r2.ts#L5475-L5487
https://github.com/alchemy-run/distilled/blob/dd2a32258180b1176513dc2bb2665cfa13f10ebf/packages/cloudflare/src/services/r2.ts#L5679-L5691
https://github.com/alchemy-run/distilled/blob/dd2a32258180b1176513dc2bb2665cfa13f10ebf/packages/cloudflare/src/services/r2.ts#L2567-L2575

Their requests take the same `accountId` / `bucketName` / `jurisdiction` triple the reconciler already passes around, and their error unions (`NoSuchBucket | InvalidRoute | CloudflareOpError`) are the ones `reconcileCustomDomains` already handles.

## Requested behavior

- A prop such as `publicAccess?: boolean`, defaulting to `false` to match Cloudflare's default.
- An attribute such as `publicDomain: string | undefined` carrying the `r2.dev` hostname.
- One more converging step in `reconcile` alongside the existing three — read with `listBucketDomainManageds`, write with `putBucketDomainManaged` when it differs.

Two behaviours seem worth deciding explicitly rather than falling out of the implementation:

- **Should `delete` disable it?** Arguably not — the managed domain is a property of the bucket and goes away with the bucket, whereas disabling it while the bucket lives breaks every URL already persisted elsewhere, with no visible cause.
- **Should the hostname be reported while disabled?** The read returns `domain` regardless of `enabled`, so the attribute can be populated either way — but reporting a hostname that serves 401 may be worse than reporting `undefined`.

Happy to open a PR, with tests in the style of `test/Cloudflare/R2/R2Bucket.test.ts` — let me know whether the shape above is how you'd want it to land.

## Cloudflare references

- Public buckets and the managed `r2.dev` domain: https://developers.cloudflare.com/r2/buckets/public-buckets/
- `wrangler r2 bucket dev-url`: https://developers.cloudflare.com/workers/wrangler/commands/#r2-bucket-dev-url

Mkassabov:
`Cloudflare.R2.Bucket` can now enable Cloudflare's managed `r2.dev` domain and report the hostname. The distilled SDK already shipped `listBucketDomainManageds` / `putBucketDomainManaged`; the resource never called them.

```ts
const bucket = yield* Cloudflare.R2.Bucket("Assets", {
  publicAccess: true,
});
// objects are at https://${bucket.publicDomain}/<key>
```

- `publicAccess` defaults to `false`, matching Cloudflare
- `publicDomain` is the `r2.dev` hostname while enabled, `undefined` while disabled (a disabled domain serves 401)
- Reconcile observes live managed-domain state and PUTs only on drift
- Delete does not disable the domain — it goes away with the bucket
- The local/dev provider reports `publicDomain: undefined` (no `r2.dev` in the simulator)

Closes #1328

### DIFF
--- a/packages/alchemy/test/Cloudflare/R2/Bucket.test.ts
+++ b/packages/alchemy/test/Cloudflare/R2/Bucket.test.ts
@@ -822,14 +822,24 @@ test.provider("managed r2.dev domain converges drift and adoption", (stack) =>
 
     yield* stack.destroy();
 
+    // Pin the physical name on every deploy so adding `name` cannot be
+    // what schedules the update — same shape as the CORS drift case
+    // above. A PR/user suffix keeps concurrent runs from colliding.
+    const suffix = (process.env.PULL_REQUEST ?? process.env.USER ?? "local")
+      .toLowerCase()
+      .replace(/[^a-z0-9]/g, "")
+      .slice(0, 24);
+    const bucketName = `alchemy-test-r2-pub-drift-${suffix}`;
+
     const initial = yield* stack.deploy(
       Effect.gen(function* () {
         return yield* Cloudflare.R2.Bucket("DriftPublicBucket", {
           forceDestroy: true,
+          name: bucketName,
+          publicAccess: false,
         });
       }),
     );
-    const bucketName = initial.bucketName;
     expect(initial.publicDomain).toBeUndefined();
 
     // Drift: enable the managed domain out of band.
@@ -839,16 +849,22 @@ test.provider("managed r2.dev domain converges drift and adoption", (stack) =>
       enabled: true,
     });
 
-    // Re-deploy without publicAccess. Reconcile diffs desired against
-    // observed cloud state, so the leaked enable is turned back off.
+    // Re-deploy with publicAccess still false. The storage-class change
+    // is what makes Plan run reconcile (unchanged props skip it); the
+    // managed-domain sync must still disable the leaked enable because
+    // it diffs desired against *observed* cloud state, not olds.
     const repaired = yield* stack.deploy(
       Effect.gen(function* () {
         return yield* Cloudflare.R2.Bucket("DriftPublicBucket", {
           forceDestroy: true,
           name: bucketName,
+          publicAccess: false,
+          storageClass: "InfrequentAccess",
         });
       }),
     );
+    expect(repaired.bucketName).toEqual(bucketName);
+    expect(repaired.storageClass).toEqual("InfrequentAccess");
     expect(repaired.publicDomain).toBeUndefined();
 
     const afterRepair = yield* r2.listBucketDomainManageds({
--- a/packages/alchemy/test/Cloudflare/R2/Bucket.test.ts
+++ b/packages/alchemy/test/Cloudflare/R2/Bucket.test.ts
@@ -54,6 +54,7 @@ test.provider("create and delete bucket with default props", (stack) =>
     expect(bucket.bucketName).toBeDefined();
     expect(bucket.storageClass).toEqual("Standard");
     expect(bucket.jurisdiction).toEqual("default");
+    expect(bucket.publicDomain).toBeUndefined();
 
     const actualBucket = yield* getBucketWhenReady(
       bucket.bucketName,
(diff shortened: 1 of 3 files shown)


xiaoyu2er:
Filed #1328 — thanks for picking it up, and for settling both of the open questions I'd flagged there in the same direction.

One note left inline on the drift half of `managed r2.dev domain converges drift and adoption`: I think the redeploy's `name: bucketName` is what's triggering the update, rather than drift detection, so the test may not pin what its comment claims. Details and how to falsify it are on the line.

packages/alchemy/test/Cloudflare/R2/Bucket.test.ts
```diff
@@ -766,6 +767,181 @@ test.provider("cors is applied to the new bucket on replacement", (stack) =>
...
+      enabled: true,
+    });
+
+    // Re-deploy without publicAccess. Reconcile diffs desired against
+    // observed cloud state, so the leaked enable is turned back off.
+    const repaired = yield* stack.deploy(
+      Effect.gen(function* () {
+        return yield* Cloudflare.R2.Bucket("DriftPublicBucket", {
+          forceDestroy: true,
+          name: bucketName,
```
xiaoyu2er, line 860: This `name` is doing more than it looks like it is: it's what makes the redeploy do anything at all.

The first deploy passes `{ forceDestroy: true }`, so:

- `Bucket.ts`'s `diff` falls through every branch and returns `undefined` — `publicAccess` is `false` on both sides, everything else equal.
- `Plan.ts` L684 falls back to `havePropsChanged(oldProps, props)`.
- Persisted props are the user-supplied object as given (`Apply.ts` L857, `props: news`), so `oldProps` genuinely has no `name`, and the canonical-JSON comparison in `Diff.ts` L184 reports a change.
- `Apply.ts` L1585 — `if (!propsChanged && !bindingsChanged) continue;` — is what would otherwise skip `reconcile` entirely.

So dropping this line should turn the test red at `expect(afterRepair.enabled).toEqual(false)` below: reconcile never runs and the out-of-band enable survives. Worth confirming rather than taking from me, since these tests don't execute in PR CI — `test:cloudflare-packages` filters to `cloudflare-runtime` / `frontend-frameworks` / `fixtures`, so `packages/alchemy/test/**` is typechecked here but not run.

What makes me raise it rather than assume it's the house pattern is `cors reconciliation converges drift and adoption` just above: that one passes `name` in *both* deploys and changes `maxAgeSeconds` `3600 -> 7200` deliberately, and its comment claims exactly what that setup proves — reconcile diffs against *observed* rather than *olds*. Here the trigger is incidental, while the comment ("Re-deploy without publicAccess ... the leaked enable is turned back off") reads as the broader claim that a redeploy repairs drift.

Not arguing the behaviour is wrong — skipping reconcile on unchanged props is consistent with `domains` / `cors` / `lifecycleRules`. It's that for this prop the un-repaired state is "a bucket that should be private is publicly readable", so it seemed worth having the test say plainly which of the two properties it pins.

(The adoption half below is unaffected — it wipes the state row first, so `reconcile` runs regardless.)

Mkassabov: You're right — adding `name` on the second deploy was the update trigger, so the test wasn't pinning observe-vs-olds.

Fixed to match the CORS case: `name` is on every deploy, `publicAccess` stays `false`, and a storage-class change is what runs reconcile. The assertion is still that the leaked enable gets turned off because reconcile diffs against observed cloud state.

Ran the three new tests against `--profile alchemy-testing`; all passed.

xiaoyu2er: Confirmed against `1eb602c` — that does it, and thanks for actually running them rather than taking my word for it.

Checking the new setup: `name` is on both deploys so it can't be the trigger, `publicAccess: false` is explicit on both sides, and the storage-class change lands on `diff`'s `{ action: "update" }` branch rather than `replace` (replace only fires on accountId / name / jurisdiction / locationHint), so `reconcile` runs against the same bucket — with `expect(repaired.bucketName).toEqual(bucketName)` there to catch it if that ever stops being true. The assertion is load-bearing now.

The part I like most is that the new comment states the limitation outright — "the storage-class change is what makes Plan run reconcile (unchanged props skip it)" — instead of routing around it. That's the bit a future reader needs, and it's now written down at the only place it matters. The `PULL_REQUEST ?? USER` suffix is a good catch too; I hadn't thought about concurrent runs colliding on the pinned name.

Nothing further from me — resolve whenever.

xiaoyu2er (approved):

PR #1330, opened 2026-08-23
fix(aws/dev): ECS Task/Service and Lambda Function declare the floci data plane

sam-goodwin:
In an `alchemy dev` run this four-line program died before provisioning anything:

```ts
const cluster = yield* AWS.ECS.Cluster("Cluster");
const task = yield* AWS.ECS.Task("Task", { image: "busybox:stable", command: ["true"] });
yield* AWS.ECS.RunTask(cluster, task);
```

```
Binding client spans mixed data planes: [Cluster, Task] resolve to different targets (some local, some live).
```

Nothing was live. `ECS.Cluster` is a `flociDual`, which registers `dataPlane: flociServices`; `ECS.Task`, `ECS.Service` and `Lambda.Function` are hand-written `ProviderLayer.dual` registrations over the floci providers that never declared a data plane, so `resolveLocalDataPlane` returned `undefined` for them — and the #1308 check can't tell "undeclared" from "live". The single-resource case is worse: `InvokeFunction(fn)` in a dev Action didn't die, it was left unrouted and silently called the real cloud.

```diff
  ProviderLayer.dual(ECS.Task, {
    live: () => ECS.TaskProvider(),
    local: () => ECS.FlociTaskProvider(),
+   dataPlane: flociServices,
  }),
```

Same for `ECS.Service` and `Lambda.Function`.

The mixed-plane error now says where each resource lands and why, via a new `describeDataPlane`:

```
Binding client spans mixed data planes: DataPlaneCluster → local emulator; DataPlaneTask → real cloud (its provider AWS.ECS.Task is dual-mode but registers no local data plane). …
```

### Tests

- `test/AWS/Local/EcsDataPlane.local.test.ts` — the report, end to end in dev: an Action binds `RunTask(cluster, task)` and launches the task; the task ARN must be emulator-shaped. Fails on `main` with the exact reported message.
- `test/AWS/LocalDataPlane.test.ts` — walks the provider registry and pins that every floci-backed dual declares its data plane (names `AWS.ECS.Service`, `AWS.ECS.Task`, `AWS.Lambda.Function` on `main`). `Command.Dev` / `AWS.Website.Server` are the documented exception: process-hosted locals with no emulator API.

Why existing tests missed it: #1308's only routing test is on `S3.Bucket` (a `flociDual`, which declares the plane automatically), and the ECS dev suites drive floci over raw HTTP rather than through the `RunTask` binding, so the bind step never ran in dev for a hand-written dual.

Green: `test/AWS/Local`, `test/AWS/S3/Bucket.data-plane.test.ts`, `test/provider-mode.test.ts`, the two new suites.

### DIFF
--- a/packages/alchemy/src/AWS/Providers.ts
+++ b/packages/alchemy/src/AWS/Providers.ts
@@ -22,7 +22,7 @@ import * as Command from "../Command/index.ts";
 import { DockerLive } from "../Docker/Docker.ts";
 import { KeyPair, KeyPairProvider } from "../KeyPair.ts";
 import * as ProviderLayer from "../Local/ProviderLayer.ts";
-import { flociDual } from "./Local/FlociServices.ts";
+import { flociDual, flociServices } from "./Local/FlociServices.ts";
 import * as Provider from "../Provider.ts";
 import { Random, RandomProvider } from "../Random.ts";
 import * as AccessAnalyzer from "./AccessAnalyzer/index.ts";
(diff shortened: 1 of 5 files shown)


sam-goodwin:
Folded into #1324 (merged main in and cherry-picked the fix there) — one PR, per Sam.

PR #1332, opened 2026-08-23
fix(cloudflare): flush event telemetry before settlement

### LINKED ISSUE #1331: Cloudflare event-scoped OTLP batches can be dropped during background scope close
## Summary

Alchemy's per-event OTLP exporters can lose an entire Worker or Durable Object event batch because the bridge lets the event promise settle before the exporter's final POST has completed.

The bridge currently closes the request/call scope under `ctx.waitUntil()` / `state.waitUntil()`. The exporter finalizer starts its OTLP request during that background scope close. If workerd destroys the owning I/O context before the request settles, the whole batch disappears. In a distributed trace this leaves unresolved parent IDs and unmatched HTTP client/server boundaries.

## Minimal real-workerd reproduction

https://github.com/dmmulroy/alchemy-workerd-otlp-flush-repro

```sh
pnpm install --ignore-scripts
pnpm reproduce
```

The repository pins:

- `alchemy@2.0.0-beta.72`
- `effect@4.0.0-beta.107`

It runs an actual local workerd Worker that makes a traced HTTP request to a Durable Object. A delayed local OTLP endpoint records only completed requests and minimal span topology. After the sole trigger request returns, the test tears workerd down immediately, so no later event can take over background work.

Expected output:

```text
== Unpatched alchemy@2.0.0-beta.72 (expected RED) ==
... expected 0 to be 2
Observed the expected missing-export failure.

== Flusher lifecycle fix (expected GREEN) ==
... 1 passed
```

The green phase applies the candidate patch included in the repository. In addition to both event batches arriving, it asserts parent closure and the Worker HTTP client -> Durable Object HTTP server relationship.

## Lifecycle sequence

Targeted, payload-free instrumentation against the original distributed E2E isolated this sequence:

```text
exporter received event spans
scope finalizer ran
single-target OTLP POST started
target did not settle
exporter shutdown completed without that batch
```

This ruled out trace-context propagation, parent-ID serialization, viewer rendering, multi-destination fanout, and collector persistence after acknowledgement.

The relevant ownership constraint is in workerd: outgoing I/O belongs to an `IoContext` and is canceled when that context is destroyed. `waitUntil` extends background work, but the final exporter fetch is still started from scope finalization after the bridge has produced the event result.

Relevant workerd source:

- [`ExecutionContext::waitUntil` and module fetch response handling](https://github.com/cloudflare/workerd/blob/main/src/workerd/api/global-scope.c++)
- [`IoContext`, `addWaitUntil`, request draining, and I/O cancellation](https://github.com/cloudflare/workerd/blob/main/src/workerd/io/io-context.c++)
- [`IoContext` ownership documentation](https://github.com/cloudflare/workerd/blob/main/src/workerd/io/io-context.h)
- [`DurableObjectState::waitUntil`](https://github.com/cloudflare/workerd/blob/main/src/workerd/api/actor-state.c++)

## Why this belongs in the bridge

Application code cannot correctly repair this lifecycle:

- Alchemy creates the exporters and event scopes.
- Alchemy owns the bridge promise returned to workerd.
- Effect's HTTP server root span ends after the handler Effect resolves.
- Alchemy owns Worker, Durable Object/RPC, error, and stream scope-ejection paths.

Waiting for all of `Scope.close` before returning avoids loss, but incorrectly turns unrelated application finalizers into response latency. Registering scope close with `waitUntil` earlier does not make the exporter POST foreground-owned.

## Candidate fix

The minimal fix in the reproduction:

1. captures the event telemetry `Context` returned by `buildEventTelemetry`;
2. yields one macrotask so Effect's late HTTP server root span reaches the exporter buffer;
3. gets `OtlpExporter.Flusher` from that context and awaits only its bounded `flush` before the bridge event promise settles;
4. retains the existing background `Scope.close` for ordinary request/call finalizers;
5. leaves stream-ejected scopes under stream completion ownership.

This was verified repeatedly against local workerd and deployed Cloudflare Worker/Durable Object traffic. The resulting traces had zero missing parents, zero unpaired HTTP clients, and zero server spans without client parents. Ordinary background-finalizer behavior remained unchanged.

I searched open and closed Alchemy issues/PRs for telemetry, OTLP, `OtlpExporter`, `waitUntil`, `WorkerBridge`, `DurableObjectBridge`, flushing, scope finalization, missing spans, and workerd/`IoContext`. The closest history is #860 (telemetry introduction), #774/#752 (event scopes and `waitUntil`), and #1229 (a different background-lifetime defect); none covers this export-settlement race.

The investigation and remediation were guided by Dillon Mulroy ([@dmmulroy](https://github.com/dmmulroy)).

dmmulroy:
Flush event-scoped OTLP buffers before Worker and Durable Object promises settle, while keeping ordinary scope finalizers under `waitUntil`.

```ts
await flushEventTelemetry(eventTelemetry)
ctx.waitUntil(Effect.runPromise(Scope.close(scope, Exit.void)))
```

The bridge still yields one macrotask for Effect's late HTTP server span. Stream-ejected scopes keep their existing completion ownership.

Adds a real local-workerd regression that is red without the fix (`0` completed event batches) and green with it (`2` completed Worker/DO batches). Reproduction: https://github.com/dmmulroy/alchemy-workerd-otlp-flush-repro

Closes #1331.

Investigation and remediation guided by Dillon Mulroy ([@dmmulroy](https://github.com/dmmulroy)).

### DIFF
--- a/packages/alchemy/src/Cloudflare/Workers/WorkerBridge.ts
+++ b/packages/alchemy/src/Cloudflare/Workers/WorkerBridge.ts
@@ -19,7 +19,7 @@ import {
 } from "../../Runtime.ts";
 import { Self } from "../../Self.ts";
 import { Stack } from "../../Stack.ts";
-import { buildEventTelemetry } from "../../Telemetry.ts";
+import { buildEventTelemetry, flushEventTelemetry } from "../../Telemetry.ts";
 import { CloudflareEnvironment } from "../CloudflareEnvironment.ts";
 import cloudflare_workers from "./cloudflare_workers.ts";
 import { isScopeEjected } from "./HttpServer.ts";
@@ -88,6 +88,7 @@ export const makeWorkerBridge = (
     ) => T | Promise<T>,
   ): Promise<T> => {
     const scope = Scope.makeUnsafe();
+    let eventTelemetry = Context.empty();
     return build((promise) => ctx.waitUntil(promise as Promise<any>))
       .then(
         (built) => {
@@ -109,13 +110,22 @@ export const makeWorkerBridge = (
                 Layer.succeed(Scope.Scope, scope),
                 // The configured telemetry exporters. Constructed as part
                 // of this per-event layer, but `buildEventTelemetry`
-                // attaches their batching fibers and flush finalizers to
-                // the request `scope` (not this build's transient scope),
-                // so buffered telemetry flushes when the scope closes into
-                // `ctx.waitUntil` below — never on workerd's ephemeral
-                // isolate scope.
+                // attaches their batching fibers and finalizers to the
+                // request `scope` (not this build's transient scope). The
+                // captured Context is explicitly flushed before this event
+                // settles; ordinary scope finalizers close via `waitUntil`.
                 Layer.effectContext(
-                  buildEventTelemetry(built.context, scope, built.telemetry()),
+                  buildEventTelemetry(
+                    built.context,
+                    scope,
+                    built.telemetry(),
+                  ).pipe(
+                    Effect.tap((context) =>
+                      Effect.sync(() => {
+                        eventTelemetry = context;
+                      }),
+                    ),
+                  ),
                 ),
               ).pipe(
                 Layer.provideMerge(Layer.succeedContext(services)),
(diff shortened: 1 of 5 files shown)


packages/alchemy/src/Cloudflare/Workers/WorkerBridge.ts
```diff
@@ -130,20 +140,18 @@ export const makeWorkerBridge = (
...
-            ),
-      );
+      .finally(() => {
+        if (isScopeEjected(scope)) return undefined;
+        // The HttpMiddleware tracer ends the request's root span in a
+        // dispatcher task scheduled after the handler effect resolves.
+        // Yield one macrotask, flush telemetry before the event settles, then
+        // close ordinary request resources in the background via waitUntil.
+        return new Promise((resolve) => setTimeout(resolve, 0))
+          .then(() => Effect.runPromise(flushEventTelemetry(eventTelemetry)))
```
sam-goodwin, line 150: Does this mean the request will wait for the flush to complete? Does this affect latency?

dmmulroy: i thought not, but it does - this is not the right pr/fix. will close and keep iterating

sam-goodwin: I am also looking into it. Problem looks real!

PR #1336, opened 2026-08-24
feat(gcp): add GCP resource providers, HTTP bindings, and nuke

Mkassabov:
GCP resource factory coverage: 1051 `Resource` exports (each with a dedicated `{Resource}.test.ts`), 192 HTTP `Binding.Service`s, live tests under `packages/alchemy/test/GCP`, and `pnpm nuke:gcp`.

### Effect-native Function with bindings

Cloud Run is the GCP analog of Lambda/Worker. Bindings grant IAM onto the runtime service account (AWS `policyStatements` / Cloudflare `AccountApiToken`) and inject env. Memorystore speaks RESP through the shared `alchemy/Redis` client.

```ts
const tweets = yield* GCP.PubSub.Topic("tweets", {});
const cache = yield* GCP.Redis.Instance("Cache", { memorySizeGb: 1 });

export class TweetBot extends GCP.Function<TweetBot>()(
  "TweetBot",
  { main: import.meta.url },
  Effect.gen(function* () {
    const publish = yield* GCP.PubSub.Publish(tweets);
    const redis = yield* GCP.Redis.ReadWriteRedis(cache);
    return {
        const text = (yield* redis.get("last-tweet")) ?? "hello twitter";
        yield* publish({
          body: { messages: [{ data: btoa(text) }] },
        });
        return HttpServerResponse.text("queued");
      }),
    };
  }).pipe(
    Effect.provide(GCP.PubSub.PublishHttp),
    Effect.provide(GCP.Redis.ReadWriteRedisHttp),
  ),
) {}
```

`GCP.Function` is `GCP.Run.Service`. Jobs are the same shape with `GCP.Run.Job` (runs to completion). Zip-source Cloud Functions stay on `GCP.CloudFunctions.Function` and still receive binding env + IAM on the runtime SA.

```ts
const bucket = yield* GCP.Storage.Bucket("data", { forceDestroy: true });
const topic = yield* GCP.PubSub.Topic("events", {});
```

Observe-ensure-sync reconcilers, idempotent delete, `list()` of alchemy-labeled rows. HTTP bindings yield distilled ops once at Layer construction (no double-provide of Credentials/HttpClient) and grant least-privilege IAM on the host. Distilled pin: `c7b11bc6b3adb65746cfa9c16bf9627f775ce25b` (`@distilled.cloud/gcp` 1.0.0-rc.8).

### Catalog

Every distilled create+get+delete trio is implemented or out-of-scope (`{SCRATCH}/gcp-catalog.txt`): 1127 implemented rows, 0 missing, 3 out-of-scope (Cloud Tasks tasks, Spanner sessions, Storage objects). Vertex/Apigee/consumer APIs are implemented with skipIf-gated live tests, not catalogued away.

Every registered Resource has `packages/alchemy/test/GCP/{Service}/{Resource}.test.ts` (1051/1051).

### Live recapture (188 dirs)

Per-service `timeout 240 pnpm test test/GCP/{dir} --profile testing --retry 0`. The full `test/GCP` tree cannot finish in a 240s wall (~2124 tests).

- **188/188 dirs PASS** after targeted retests (skipIf FAST/creds/entitlement todos are not failures)
- **Compute** whole-dir hits the 240s wall. File-batch recapture @ `6495a58b9`: 123 pass / 2 fail / 27 gated. Both fails fixed on `e72859c26` (RegionCompositeHealthCheck wait ~60s, 156.6s pass under 240s; RegionNetworkFirewallPolicy replace via `toName`)
- Core slice green: Storage 9/9, PubSub 6/6, Redis 3+2 gated, Run lifecycle 4/4

### Bindings census

192 `Binding.Service` exports, each with a live Action/HTTP invocation in `Bindings.test.ts`. Memorystore also has `ReadRedis` / `WriteRedis` / `ReadWriteRedis` over `alchemy/Redis`.

### Nuke residue (catalogued)

`pnpm nuke:gcp --dry-run` after the recapture:

```
GCP.Analyticshub.DataExchange  2 to delete
GCP.Analyticshub.DataExchangesListing  2 to delete
GCP.KMS.CryptoKey  2 to delete
6 resource(s) to delete.
```

- Analytics Hub listings/exchanges: GCP HTTP 500 INTERNAL when `replicaState: UNAVAILABLE`
- KMS crypto keys: `DESTROY_SCHEDULED` (24h; no immediate delete API)

### Examples and docs

Two runnable examples, each a thing someone would actually build:

- **`examples/gcp-cloud-run-api`** — a link shortener on one Cloud Run service. `POST /links` mints a base62 code, `GET /l/:code` redirects and counts the click, `DELETE /links/:code` removes it. State lives in Firestore, the API key in Secret Manager, both reached through bindings, so the host's service account ends up with `roles/datastore.user` and `roles/secretmanager.secretAccessor` and nothing else.
- **`examples/gcp-event-pipeline`** — an ingest service publishes to Pub/Sub and triggers a Cloud Run Job through a `Run.RunJob` binding; the job pulls the subscription and streams rows into BigQuery with `InsertAll`. Shows one host holding IAM on another, and a Job used as a batch drain rather than a request handler.

Both have `test/integ.test.ts` that deploys the stack, drives it over HTTP, checks the backing GCP resource out of band, and destroys. They skip unless Docker and GCP credentials are both present, because the `main`-form hosts build a local image.

The four-part tutorial is replaced by docs organised around what you are building:

| Page | Covers |
| --- | --- |
| `/gcp` | What the provider gives you, routed by use case |
| `/gcp/setup` | Project, service-account key, APIs to enable, Docker, CI env vars, teardown |
| `/gcp/guides/cloud-run-api` | The link shortener, file by file |
| `/gcp/guides/event-pipeline` | Pub/Sub to BigQuery through a Job |
| `/gcp/guides/memorystore` | Redis over Direct VPC egress, and what create/destroy cost in wall time |
| `/gcp/guides/bindings` | What `yield*` of a binding does to IAM, and what removing one revokes |

Sidebar and `website/scripts/generate-llms-txt.ts` follow. `pnpm docs:check` passes (6187 pages, no dead links).

### Remaining (draft)

HEAD `205719233`, with `main` merged in.

Merging `main` brought in `pkg.yml`, which builds the packages cold instead of restoring a Turbo cache. That exposed a memory ceiling: cold-building the `alchemy` package peaks near 17 GB of compiler heap on this branch against about 12 GB on `main`, which starves the runner agent on its 32 GB machine and drops the job with no log. A `GOMEMLIMIT` soft limit on that step holds the peak near 11 GB for roughly twice the compile time; Turbo needs the variable declared as a pass-through to hand it to the task.

The effect-native Function fixture (`test/GCP/Run/Service.test.ts` T1) is `skipIf` without Docker, so the recapture environment did not live-run out-of-band SA/IAM assertions. Image-build failures after minting the per-host `alch-*` SA can leak the account (`tapError` today wraps only Cloud Run create, not `images.resolve`); that patch is held until Check reports. A Job `run` e2e fixture is a follow-up, not a merge blocker.

The two new example integration tests are gated the same way and were not live-run in the environment this branch was prepared in (no Docker daemon available there). Everything else was checked: full-repo `tsc -b`, `oxfmt`, and `pnpm docs:check` are clean.


sam-goodwin:
Please update PR description with DX (heading + code snippet showing resources, effectful sevices, bindings) etc.

sam-goodwin:
Can you show me a GCP.Function with bindings (something for twitter)

packages/alchemy/src/GCP/Redis/Instance.ts
sam-goodwin, line 1: I recently commited src/Redis used in Fly/Railway redis bindings. Might be worth adopting

packages/alchemy/src/GCP/AIPlatform/BindingHttp.ts
```diff
@@ -0,0 +1,48 @@
...
+    GcpOpContext
+  > &
+    ((input: I) => Effect.Effect<A, E, GcpOpContext>);
+}) =>
+  Effect.gen(function* () {
+    const credentials = yield* Credentials;
+    const httpClient = yield* HttpClient.HttpClient;
+    const run = yield* options.operation.pipe(
+      Effect.provideService(Credentials, credentials),
+      Effect.provideService(HttpClient.HttpClient, httpClient),
```
sam-goodwin: no doublr provide

packages/alchemy/src/GCP/AIPlatform/BindingHttp.ts
```diff
@@ -0,0 +1,48 @@
...
+      const name = yield* resource.name;
+      return Effect.fn(`${options.tag}(${resource.LogicalId})`)(function* (
+        request?: Omit<I, "name">,
+      ) {
+        const resourceName = yield* name;
+        return yield* run({
+          ...(request ?? {}),
+          name: resourceName,
+        } as I);
+      });
```
sam-goodwin: I'd expect this to be like how AWS IAM grants IAM policies, or cloudflare's HTTP bindings create an AccountApiToken and bind specific policies on to it.

packages/alchemy/src/GCP/Run/Service.ts
```diff
@@ -0,0 +1,1032 @@
...
+ *   template: {
+ *     timeout: "60s",
+ *     containers: [
+ *       {
+ *         image: "us-docker.pkg.dev/cloudrun/container/hello",
+ *         env: [{ name: "ENV", value: "prod" }],
+ *       },
+ *     ],
+ *   },
+ * });
```
sam-goodwin, line 348: Missing effectful platforms with bindings

packages/alchemy/src/GCP/Run/Job.ts
sam-goodwin, line 1: Needs to be effectful?

packages/alchemy/src/GCP/Redis/Instance.ts
Mkassabov: Adopted `alchemy/Redis` (shared RESP client). Memorystore now has `ReadRedis` / `WriteRedis` / `ReadWriteRedis` that pack `REDIS_URL` and grant `roles/redis.viewer|editor` on the Cloud Run/Function host.

packages/alchemy/src/GCP/AIPlatform/BindingHttp.ts
Mkassabov: Dropped the extra `provideService(Credentials)` / `HttpClient`. Bindings yield the distilled op once at Layer construction; the host Layer already provides GcpOpContext.

packages/alchemy/src/GCP/AIPlatform/BindingHttp.ts
Mkassabov: HTTP bindings now `host.bind` `{ iam, env }` onto Cloud Run Service/Job and Cloud Functions. Reconcile grants the role on the runtime service account via `projects.setIamPolicy` (AWS policyStatements / Cloudflare AccountApiToken analog).

packages/alchemy/src/GCP/Run/Service.ts
Mkassabov: `GCP.Run.Service` is a Platform. Pass `main` + an Effect `{ fetch }` impl; Artifact Registry image + Cloud Run bootstrap. Bindings attach env + IAM. Also exported as `GCP.Function`.

packages/alchemy/src/GCP/Run/Job.ts
Mkassabov: `GCP.Run.Job` is a Platform too — `main` + Effect impl, runs to completion (CloudRunJob bootstrap). Same `{ env, iam }` contract.

Mkassabov:
Updated the PR description with an Effect-native \`GCP.Function\` (Cloud Run Service) that binds Pub/Sub + Memorystore Redis — the twitter-bot shape. \`GCP.Function\` is an alias of \`GCP.Run.Service\`.

PR #1338, opened 2026-08-24
fix(cloudflare/r2): register catalog credential first

### LINKED ISSUE #1337: R2 DataCatalog applies maintenance before registering its credential
## Summary

A fresh `Cloudflare.R2.DataCatalog` deployment fails when `token` and non-default catalog maintenance settings are supplied together. The provider updates maintenance before registering the credential, while Cloudflare requires the credential before accepting the maintenance update.

## Version

`alchemy@2.0.0-beta.74`

## Reproduction

```ts
const bucket = yield* Cloudflare.R2.Bucket("DataLakeBucket");

const token = yield* Cloudflare.ApiToken.AccountApiToken("DataCatalogToken", {
  accountId,
  policies: [
    {
      effect: "allow",
      permissionGroups: [
        "Workers R2 Storage Write",
        "Workers R2 Data Catalog Write",
      ],
      resources: {
        [`com.cloudflare.api.account.${accountId}`]: "*",
      },
    },
  ],
});

const catalog = yield* Cloudflare.R2.DataCatalog("DataCatalog", {
  bucketName: bucket.bucketName,
  compaction: {
    state: "enabled",
    targetSizeMb: "256",
  },
  token: token.value,
});
```

Deploy this into an account where the bucket/catalog do not already exist.

## Actual behavior

The catalog is enabled, but reconciliation fails while updating maintenance:

```text
BadRequest: A credential is not present for the warehouse=<account>_<bucket>
```

The provider currently performs these operations in `src/Cloudflare/R2/DataCatalog.ts`:

1. Enable the catalog.
2. Update maintenance configuration when dirty.
3. Register the credential.

Because step 2 throws, step 3 is never reached. The deployment remains partial.

Registering the same generated token through `POST /accounts/{account_id}/r2-catalog/{bucket_name}/credential` and rerunning the deployment allows it to complete.

## Expected behavior

When `token` is supplied, register or rotate the credential before applying catalog-level maintenance settings:

1. Enable/observe the catalog.
2. Register the credential if absent or rotated.
3. Update maintenance configuration.

## Suggested regression coverage

Add a provider test where a newly enabled catalog reports `credentialStatus: "absent"`, and both `token` and dirty `compaction` settings are supplied. Assert that `createCredential` runs before `updateMaintenanceConfig`.

dmmulroy:
Register an absent or rotated R2 Data Catalog credential before applying maintenance configuration changes.

Adds a live lifecycle regression that deploys a fresh catalog with a credential and dirty compaction settings in the same reconciliation, then verifies both through Cloudflare.

Fixes #1337.


packages/alchemy/test/Cloudflare/R2DataCatalog/DataCatalogProvider.test.ts
sam-goodwin, line 1: Don't mock stuff. Deploy a real fixture and test it works e2e

dmmulroy: Replaced the mocked transport test with the live R2 Data Catalog lifecycle test. It now deploys a fresh catalog with an API token and dirty compaction in the initial reconciliation, then verifies the credential and maintenance config through Cloudflare. The focused case and full R2 Data Catalog suite pass live.

PR #1339, opened 2026-08-24
test(cloudflare): pin OTLP event-telemetry contracts; fix teardown-hook cleanup leaks

sam-goodwin:
Resolves the investigation behind #1331 / #1332: the runtime bridges need **no change** — the existing `waitUntil`-deferred background export delivers every batch, verified empirically on local and deployed workerd for Worker fetch, DO fetch, and DO RPC events. The loss in #1331's repro was the test harness tearing workerd down before background work ran (#1341 fixes the Vitest adapter). This PR pins the telemetry contracts as tests, and fixes the same teardown-leak class in the two other test adapters.

### Telemetry contract tests (no bridge changes)

| environment | Worker fetch | DO fetch | DO RPC |
| --- | --- | --- | --- |
| local workerd (dev sidecar, alive) | ✓ | ✓ | ✓ |
| deployed workerd | ✓ | ✓ | ✓ |

- **Delivery** — every event's OTLP batch reaches the collector while the runtime is alive. Locally via a 500ms-delayed `127.0.0.1` collector counting only fully-written responses; live via the DO-backed collector on the test zone (`Telemetry.test.ts` now also deploys the Worker→DO fixture and asserts the DO fetch and DO RPC event spans arrive from deployed workerd — `state.waitUntil` is a no-op in DOs, so this is the path #1331 believed was broken).
- **Latency** — an event's response never waits on its own telemetry export:

```ts
const completedAtResponse = collector.completedRequests.value;
expect(completedAtResponse).toBeLessThanOrEqual(1); // own batch still in flight
```

A foreground-flush design fails the latency bound; a lossy design fails the delivery poll.

### Teardown-hook leak fixes (companion to #1341)

A failing teardown hook silently dropped later cleanup — leaking the shared scope and local provider sidecar for the rest of the test process. Each adapter had it with a different trigger; #1341 covers Vitest, this covers the other two:

- **alchemy-test runner**: `runAfterAll` short-circuited on the first failing hook. Every `afterAll` now runs; failures aggregate. (`beforeAll` keeps short-circuit semantics — a failed setup invalidates what follows.)
- **Test/Bun**: bun:test stops the `afterAll` chain once a hook throws (probed empirically; registration order and afterAll-after-failed-`beforeAll` are fine), skipping the microtask-registered fallback. User teardowns are now guarded:

```ts
const guardTeardown = (eff) => () =>
  runEff(eff).catch(async (error) => {
    await Effect.runPromise(closeAll); // idempotent
    throw error;                       // still fails the suite
  });
```

Both regression tests are two-sided and red against the previous code: the runner test asserts a second teardown's sentinel after a failing first; the Bun test spawns real `bun test` on a fixture using the public adapter and asserts a shared-scope finalizer sentinel after the throwing user teardown.

Builds on @dmmulroy's investigation and repro topology from #1331/#1332.

### DIFF
--- a/packages/alchemy-test/src/Runner.ts
+++ b/packages/alchemy-test/src/Runner.ts
@@ -670,13 +670,24 @@ const runSuite: (suite: Suite, ctx: ExecContext) => Effect.Effect<void> =
 const runAfterAll = Effect.fn(function* (suite: Suite, ctx: ExecContext) {
   if (suite.afterAll.length === 0) return;
   yield* ctx.emit({ _tag: "HookStart", file: ctx.file, hook: "afterAll" });
-  const exit = yield* runHooks(suite.afterAll, ctx.options.timeout).pipe(
-    withCapture(ctx.fileLogs),
-    Effect.exit,
-  );
+  // Unlike beforeAll (where a failure invalidates everything after it),
+  // every teardown hook runs even when an earlier one fails: a failing
+  // teardown assertion must not drop later cleanup — in particular
+  // Test.make's fallback hook that closes the shared scope and local
+  // provider sidecar, which registers last and would otherwise leak the
+  // sidecar for the rest of the process. Failures aggregate.
+  const exits = yield* Effect.forEach(suite.afterAll, (hook) =>
+    Effect.suspend(hook.body).pipe(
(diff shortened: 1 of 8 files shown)


sam-goodwin:
cc @dmmulroy

PR #1340, opened 2026-08-24
feat(cloudflare/containers): expose placement jurisdiction constraints

yovanoc:
## Summary

- expose typed regional placement constraints on `Cloudflare.Container` applications
- expose the Cloudflare Containers `jurisdiction` placement boundary (`eu`/`fedramp`)
- document the placement fields without changing defaults or runtime behavior

## Context

A downstream application needs a typed, fail-closed way to request an EU-resident native Container. The current Alchemy surface leaves `constraints` as an untyped object, while the Containers API accepts placement constraints. This keeps the change at the Alchemy API boundary; it does not claim that every account/plan has EU Container capacity.

The async binding collision is separately fixed by #1271.

## Validation

- `git diff --check`
- `bunx oxfmt --check packages/alchemy/src/Cloudflare/Containers/ContainerApplication.ts`
- Container integration tests require the repository's full `alchemy-test`/distilled workspace and were not runnable from this cached checkout.

### DIFF
--- a/packages/alchemy/src/Cloudflare/Containers/ContainerApplication.ts
+++ b/packages/alchemy/src/Cloudflare/Containers/ContainerApplication.ts
@@ -53,6 +53,24 @@ export namespace ContainerApplication {
   >[number];
   export type Constraints = {
     tier?: number;
+    /**
+     * Limit container placement to specific geographic regions.
+     */
+    regions?: (
+      | "ENAM"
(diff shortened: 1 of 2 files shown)


sam-goodwin:
update PR description to show DX, not bullet point text

sam-goodwin:
Missing a test fixture

PR #1345, opened 2026-08-25
feat(cloudflare/hyperdrive): add read-only Ref resource for existing configs on v2

### LINKED ISSUE #1343: no v2 equivalent of v1 HyperdriveRef
In v1, HyperdriveRef binds an existing Hyperdrive config to a Worker without Alchemy ever owning it. The config is never created, never updated on reconcile, and never deleted on destroy. It's the right tool when the config was created out-of-band and Alchemy shouldn't manage its lifecycle.

As of v2.0.0-beta.74 I can't find an equivalent

On v2 the closest I can get is adopting the config into a dedicated stack that I never destroy, then binding it from the app stack through a cross-stack ref. That contains the destroy risk but not the update risk

A read-only reference to an existing Hyperdrive config would be appreaciated

Franz1241:
Closes #1343

Adds `Cloudflare.Hyperdrive.Ref`, a read-only reference to an existing Hyperdrive config (created in the dashboard or owned by another system) with v1 `HyperdriveRef` semantics: deploys only observe the config, destroy only drops the state row, and it has no `list`, so `nuke`/orphan GC can never touch the referenced config


Franz1241:
Hey @sam-goodwin would love your help merging this when you get a chance. It brings back v1's HyperdriveRef on v2

PR #1346, opened 2026-08-25
fix(test): run cleanup when a teardown hook fails (runner + Bun adapter)

sam-goodwin:
Companion to #1341: the Vitest adapter's teardown-ordering fix revealed a broader class — a **failing teardown hook silently drops later cleanup**, leaking the shared scope and local provider sidecar for the rest of the test process. The other two adapters each had it with a different trigger.

**alchemy-test runner** — `runAfterAll` short-circuited on the first failing hook:

```ts
// before: Effect.forEach on the error channel — first failure skips the rest
// after: every afterAll runs; failures aggregate into fileErrors
const exits = yield* Effect.forEach(suite.afterAll, (hook) =>
  Effect.suspend(hook.body).pipe(Effect.timeout(...), Effect.exit),
).pipe(withCapture(ctx.fileLogs));
```

`beforeAll` keeps short-circuit semantics (a failed setup invalidates what follows); teardowns all run, matching every mainstream runner's contract.

**Test/Bun** — probed bun:test empirically: `afterAll` runs in registration order ✓, and still runs after a failed `beforeAll` ✓, but the chain **stops once a hook throws** — skipping the microtask-registered fallback. The adapter now guards user teardowns:

```ts
const guardTeardown = (eff) => () =>
  runEff(eff).catch(async (error) => {
    await Effect.runPromise(closeAll); // idempotent
    throw error;                       // still fails the suite
  });
```

Both regression tests are two-sided and red against the previous code:

- `alchemy-test/test/Runner.test.ts` — a file with a failing first `afterAll` must still run the second (sentinel asserted from the spawned runner's output) and exit non-zero.
- `test/Test/Bun.test.ts` — spawns real `bun test` on a fixture using the public adapter; a shared-scope finalizer sentinel proves the cleanup ran after the throwing user teardown.

### DIFF
--- a/packages/alchemy-test/src/Runner.ts
+++ b/packages/alchemy-test/src/Runner.ts
@@ -670,13 +670,24 @@ const runSuite: (suite: Suite, ctx: ExecContext) => Effect.Effect<void> =
 const runAfterAll = Effect.fn(function* (suite: Suite, ctx: ExecContext) {
   if (suite.afterAll.length === 0) return;
   yield* ctx.emit({ _tag: "HookStart", file: ctx.file, hook: "afterAll" });
-  const exit = yield* runHooks(suite.afterAll, ctx.options.timeout).pipe(
-    withCapture(ctx.fileLogs),
-    Effect.exit,
-  );
+  // Unlike beforeAll (where a failure invalidates everything after it),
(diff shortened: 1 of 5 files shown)


sam-goodwin:
Folded into #1339.

PR #1349, opened 2026-08-25
fix(cloudflare/email): retry rule writes while a fresh Worker propagates

### LINKED ISSUE #1348: Email.Rule fails on first deploy when its worker action names a Worker created in the same apply
### What happens

Deploying a stack that creates a Worker and an `Email.Rule` forwarding to it, into a stage that has never existed before, fails at the rule:

```
NotFound: Workers Script Info not found
```

Re-running the same deploy succeeds immediately (measured: 460 ms). Nothing about the stack changes between the two runs.

### Why

`Cloudflare/Email/Rule`'s `reconcile` never looks the Worker up. It builds a body whose action is `{ type: "worker", value: [scriptName] }` and calls `createRule`. Cloudflare validates that script reference while handling the request, and a Worker created earlier in the same apply is not yet visible to that validation. The window closes within seconds.

So the call that needs to ride out the delay is `createRule` itself, rather than any lookup performed before it.

The failure only meets a stage whose Worker did not already exist. Redeploys, and any stage deployed more than once, create the rule against a Worker that is already there — which is why it stays invisible until someone deploys a genuinely fresh stage. In our case every PR preview stage meets it, and every one of them is fixed by re-running.

### Why this looks like an oversight rather than a missing feature

alchemy already rides out this class of Cloudflare latency in several places, and `Fetcher` describes the same condition in the same words — "A freshly-deployed Durable Object / service script is eventually consistent … it is safe to retry" — with a bounded backoff:

```ts
Effect.retry({
  while: (error) => error instanceof HandlerNotReady,
  schedule: Schedule.exponential("100 millis"),
  times: 8,
})
```

`Alerting/Webhook` ("rides out edge propagation — fresh workers.dev URLs"), `Alerting/Silence` ("created silence not yet visible in the list — retried internally") and `Snippets/Snippet` carry comparable handling. `Email/Rule` has none.

### Suggested fix

Retry `createRule` (and `updateRule`) on the script-not-found error using the shape `Fetcher` already uses: bounded exponential backoff, then re-raise the original error unchanged, so a genuinely missing Worker still fails and says so rather than hanging.

`Fetcher`'s own justification for retrying transfers directly. There, the request never reached a handler, so nothing was committed. Here, a `createRule` that fails validation creates no rule — so a retry cannot produce a duplicate.

### Environment

- `alchemy@2.0.0-beta.74`
- Cloudflare Workers + Email Routing, with the Worker and the rule declared in one stack
- Reproduces on any stage name that has never been deployed before; never on a redeploy

### Scope note

This is reported against v2 because that is what we run. The error originates in Cloudflare's `createRule` validation rather than in any particular alchemy version's code, so it is probably not v2-specific — but we have not measured it against v1 and are claiming nothing about it.

Mkassabov:
Cloudflare validates a rule's `{ type: "worker", value: [script] }` action while handling the write and rejects it until a freshly-uploaded script propagates. `createRule` declared no typed errors at all, so there was nothing to catch.

```ts
// before
export type CreateRuleError = CloudflareOpError;
// after
export type CreateRuleError = WorkerScriptNotFound | CloudflareOpError;
```

```ts
// Email/retry.ts — the shape Fetcher already uses for a fresh script
Effect.retry({
  while: (error) => error._tag === "WorkerScriptNotFound",
  schedule: Schedule.exponential("100 millis"),
  times: 8,
})
```

Applied to `createRule` / `updateRule` in `Email.Rule` and to `putRuleCatchAll` in `Email.CatchAll` — the catch-all takes a `worker` action through a different endpoint and hits the same validation. After the budget (~25s) the original error is re-raised unchanged, so a genuinely missing Worker still fails and says so.

Depends on alchemy-run/distilled#476, which adds the typed error and is picked up by the submodule bump here.

Closes #1348

### Observed live

The error is real and now tagged. Probed against the API directly — both endpoints return Cloudflare code **2016**:

```
createRule       -> WorkerScriptNotFound: Workers Script Info not found
putRuleCatchAll  -> WorkerScriptNotFound: Workers Script Info not found
```

The distilled matcher keys on `{ "code": 2016 }` rather than a status + message substring.

### The regression guard is `retry.test.ts`, not the live suite

`test/Cloudflare/Email/retry.test.ts` injects the failure rather than trying to race the real API, so it fails when the retry is removed:

| | retry as written | retry replaced with `effect` |
| --- | --- | --- |
| `EmailRuleWorkerTarget.test.ts` (live) | 2 passed | 2 passed |
| `retry.test.ts` (injected) | 3 passed | **2 failed** |

It covers retrying while `WorkerScriptNotFound` then succeeding (3 attempts), not retrying an unrelated tag (1 attempt), and re-raising the original error unchanged after the bounded budget (9 attempts = initial + `times: 8`).

### What this does not claim

**The reported failure does not reproduce on current `main`.** `test/Cloudflare/Email/EmailRuleWorkerTarget.test.ts` covers the reported shape — a Worker plus a rule, and a Worker plus a catch-all, each deployed into a torn-down stack so the Worker really is created in that apply — and it passes. It also passes with the retry removed, so it is a smoke test, not a regression guard, and the test says so.

#1348's reporter confirmed this independently on their own Email-Routing-capable account: they mutated the retry away and got three green runs, and verified from the deploy log that a never-before-deployed stage really is what runs. Credit to @xiaoyu2er — the deterministic test above exists because of that review.

The reason is visible in a deploy log: the Worker provider pre-creates a stub script before the rule is validated, so the name already resolves.

```
[EmailRuleTargetWorker] pre-creating      <- stub uploaded
[EmailRuleTargetWorker] creating          <- real upload
[WorkerRule] creating                     <- name already resolves
```

The reporter is on `2.0.0-beta.74`. The retry is kept as defensive handling for paths where that stub is absent or slow: it is free when it does not fire, and it only fires on a tag that means precisely "script not visible yet".

`getRule` also gains the already-defined `Forbidden` / `EmailRoutingRuleNotFound`, which it declared no errors for.

### DIFF
--- a/packages/alchemy/src/Cloudflare/Email/CatchAll.ts
+++ b/packages/alchemy/src/Cloudflare/Email/CatchAll.ts
@@ -8,6 +8,7 @@ import { CloudflareEnvironment } from "../CloudflareEnvironment.ts";
 import type { Providers } from "../Providers.ts";
 import { resolveZoneId, type Reference } from "../Zone/index.ts";
 import { listAllZones } from "../Zone/lookup.ts";
+import { retryWorkerScriptNotFound } from "./retry.ts";
 import type { Action } from "./Rule.ts";
 
 const CatchAllTypeId = "Cloudflare.Email.CatchAll" as const;
@@ -222,17 +223,19 @@ export const CatchAllProvider = () =>
       ) {
         return toAttributes(zoneId, observed, initial);
       }
-      const result = yield* emailRouting.putRuleCatchAll({
-        zoneId,
-        matchers: [{ type: "all" }],
-        actions: news.actions.map((a) =>
-          a.type === "drop"
-            ? { type: a.type }
-            : { type: a.type, value: a.value },
-        ),
-        enabled: desiredEnabled,
-        name: desiredName,
-      });
+      const result = yield* emailRouting
+        .putRuleCatchAll({
+          zoneId,
+          matchers: [{ type: "all" }],
+          actions: news.actions.map((a) =>
+            a.type === "drop"
+              ? { type: a.type }
+              : { type: a.type, value: a.value },
+          ),
+          enabled: desiredEnabled,
+          name: desiredName,
+        })
+        .pipe(retryWorkerScriptNotFound);
       return toAttributes(zoneId, result, initial);
     }),
 
--- a/packages/alchemy/src/Cloudflare/Email/Rule.ts
+++ b/packages/alchemy/src/Cloudflare/Email/Rule.ts
@@ -8,6 +8,7 @@ import { CloudflareEnvironment } from "../CloudflareEnvironment.ts";
 import type { Providers } from "../Providers.ts";
 import { resolveZoneId, type Reference } from "../Zone/index.ts";
 import { listAllZones } from "../Zone/lookup.ts";
+import { retryWorkerScriptNotFound } from "./retry.ts";
 
 export type Matcher =
   | { type: "all" }
@@ -174,16 +175,19 @@ export const RuleProvider = () =>
             ...body,
           })
           .pipe(
+            retryWorkerScriptNotFound,
             Effect.catch(() =>
               emailRouting
                 .createRule({ zoneId, ...body })
-                .pipe(Effect.map((r) => r)),
+                .pipe(retryWorkerScriptNotFound),
             ),
           );
         return normalize(result, zoneId);
       }
 
-      const result = yield* emailRouting.createRule({ zoneId, ...body });
+      const result = yield* emailRouting
+        .createRule({ zoneId, ...body })
+        .pipe(retryWorkerScriptNotFound);
       return normalize(result, zoneId);
     }),
     delete: Effect.fn(function* ({ output }) {
--- a/packages/alchemy/src/Cloudflare/Email/retry.ts
+++ b/packages/alchemy/src/Cloudflare/Email/retry.ts
@@ -1,0 +1,29 @@
+import * as Effect from "effect/Effect";
+import * as Schedule from "effect/Schedule";
+
+/**
+ * Ride out the window in which a Worker created earlier in the same apply is
+ * not yet visible to Email Routing's action validation.
+ *
+ * Cloudflare validates a `{ type: "worker", value: [scriptName] }` action
+ * while handling the create/update request and rejects it with
+ * `WorkerScriptNotFound` ("Workers Script Info not found") until the script
+ * propagates — typically within a few hundred milliseconds. The rejected call
+ * creates nothing, so a retry cannot produce a duplicate rule.
+ *
+ * Mirrors `Cloudflare/Fetcher`'s treatment of a freshly-deployed script:
+ * bounded exponential backoff (~25s total), then re-raise the original error
+ * unchanged so a genuinely missing Worker still fails and says so.
+ *
+ * @see https://github.com/alchemy-run/alchemy/issues/1348
+ */
+export const retryWorkerScriptNotFound = <A, E extends { _tag: string }, R>(
+  effect: Effect.Effect<A, E, R>,
+): Effect.Effect<A, E, R> =>
+  effect.pipe(
+    Effect.retry({
+      while: (error) => error._tag === "WorkerScriptNotFound",
+      schedule: Schedule.exponential("100 millis"),
+      times: 8,
+    }),
+  );
--- a/packages/alchemy/test/Cloudflare/Email/EmailRuleWorkerTarget.test.ts
+++ b/packages/alchemy/test/Cloudflare/Email/EmailRuleWorkerTarget.test.ts
@@ -1,0 +1,147 @@
+import * as Cloudflare from "@/Cloudflare";
+import { CloudflareEnvironment } from "@/Cloudflare/CloudflareEnvironment";
+import { findZoneByName } from "@/Cloudflare/Zone/lookup";
+import * as Test from "@/Test/Alchemy";
+import * as emailRouting from "@distilled.cloud/cloudflare/email-routing";
+import { describe, expect } from "alchemy-test";
+import * as Effect from "effect/Effect";
+import { MinimumLogLevel } from "effect/References";
+import * as Schedule from "effect/Schedule";
+import RuleTargetWorker from "./fixtures/rule-target-worker.ts";
+
+const { test } = Test.make({ providers: Cloudflare.providers() });
+
+const logLevel = Effect.provideService(
+  MinimumLogLevel,
+  process.env.DEBUG ? "Debug" : "Info",
+);
+
+const zoneName =
+  process.env.CLOUDFLARE_TEST_DNS_ZONE_NAME ?? "alchemy-test-2.us";
+
+const resolveZoneId = Effect.gen(function* () {
+  const { accountId } = yield* yield* CloudflareEnvironment;
+  const zone = yield* findZoneByName({ accountId, name: zoneName });
+  if (!zone) {
+    return yield* Effect.die(
+      new Error(`zone "${zoneName}" not found in account`),
+    );
+  }
+  return zone.id;
+});
+
+// The scoped API token the test harness mints propagates eventually-
+// consistently across Cloudflare's edge — a fresh token intermittently 403s.
+const enableRouting = (zoneId: string) =>
+  emailRouting.enableEmailRouting({ zoneId }).pipe(
+    Effect.retry({
+      while: (e) => e._tag === "Forbidden",
+      schedule: Schedule.exponential("500 millis"),
+      times: 8,
+    }),
+  );
+
+describe.sequential("Email.Rule -> Worker in one apply", () => {
+  // Coverage for the shape reported in #1348: a Worker plus a routing
+  // resource whose `{ type: "worker", value: [scriptName] }` action names
+  // it, both created in the same apply, into a stage that did not exist.
+  //
+  // NOTE: this does NOT reproduce the #1348 failure on current main, and
+  // was confirmed to still pass with the provider's retry removed. The
+  // Worker provider pre-creates a stub script before the rule is created
+  // (see "pre-creating" in a deploy log), so by the time Cloudflare
+  // validates the action the name already resolves. The retry stays as
+  // defensive handling for paths where that stub is absent or slow — the
+  // error itself is real and typed (`WorkerScriptNotFound`, Cloudflare
+  // code 2016, observed live on both createRule and putRuleCatchAll).
+  test.provider(
+    "first deploy of a fresh stage creates the rule targeting the worker",
+    (stack) =>
+      Effect.gen(function* () {
+        const zoneId = yield* resolveZoneId;
+
+        // A never-before-deployed stage is the precondition — start from a
+        // torn-down stack so the Worker really is created in this apply.
+        yield* stack.destroy();
+        yield* enableRouting(zoneId);
+
+        const { rule, workerName } = yield* stack.deploy(
+          Effect.gen(function* () {
+            const routing = yield* Cloudflare.Email.Routing("Routing", {
+              zone: zoneName,
+            });
+            const worker = yield* RuleTargetWorker;
+            const rule = yield* Cloudflare.Email.Rule("WorkerRule", {
+              zone: { zoneId: routing.zoneId },
+              name: "alchemy worker-target test",
+              matchers: [
+                {
+                  type: "literal",
+                  field: "to",
+                  value: `worker-target@${zoneName}`,
+                },
+              ],
+              actions: [{ type: "worker", value: [worker.workerName] }],
+            });
+            return { rule, workerName: worker.workerName };
+          }),
+        );
+
+        expect(rule.ruleId).not.toEqual("");
+        expect(rule.zoneId).toEqual(zoneId);
+        expect(rule.actions).toEqual([{ type: "worker", value: [workerName] }]);
+
+        // Verify out-of-band that Cloudflare really stored the worker action.
+        const live = yield* emailRouting
+          .getRule({ zoneId, ruleIdentifier: rule.ruleId })
+          .pipe(
+            Effect.retry({
+              while: (e) => e._tag === "Forbidden",
+              schedule: Schedule.exponential("500 millis"),
+              times: 8,
+            }),
+          );
+        expect(live.actions?.[0]?.type).toEqual("worker");
+        expect(live.actions?.[0]?.value).toEqual([workerName]);
+
+        yield* stack.destroy();
+      }).pipe(logLevel),
+    { timeout: 180_000 },
+  );
+
+  // The catch-all rule takes a `worker` action through a different endpoint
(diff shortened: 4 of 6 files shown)


xiaoyu2er:
@Mkassabov  I filed #1348, and I have the thing your note says is missing — an API-token profile with Email Routing permissions. Ran your test on this branch. **The 403 blocker is gone: the suite runs.** But the headline result is a negative one you will want before this merges.

## The test passes — and it passes just as well with the fix removed

| run | `retryWorkerScriptNotFound` | result |
| --- | --- | --- |
| 1 | as written | 2 passed (26.3s) |
| 2 | **body replaced with `effect`** (retry disabled) | 2 passed (24.7s) |
| 3 | same mutation | 2 passed (24.2s) |
| 4 | same mutation | 2 passed (24.3s) |

Three mutated runs, all green. So the suite does not currently discriminate: it would look identical if the bug were unfixed. As a regression test for #1348 it is not yet load-bearing.

To be clear about what this is and is not evidence for — the fix's *logic* is not in question, and I am not claiming it does not work. What is established is only that **this test does not exercise the window**, so a green run of it says nothing either way.

## The precondition is met, so that is not the explanation

I checked the log rather than assuming. A never-before-deployed stage really is what runs:

```
Plan: 3 to create
[EmailRuleTargetWorker] create
[Routing] create
[WorkerRule] create
...
Cloudflare Worker create: uploading script for cloudflare-email-emailruleworkytv2umxkaaksmiglr5nppbrq
[EmailRuleTargetWorker] Uploading worker (1.36 MB) ...
[WorkerRule] creating          <- rule create begins here
[EmailRuleTargetWorker] created
[WorkerRule] created
```

The Worker is uploaded fresh in that apply, the script name differs on every run (`…gkpws5pj…`, `…ytv2umxk…`), and `[WorkerRule] creating` starts while the Worker is still finishing (`Enabling workers.dev subdomain…`). `stack.destroy()` does its job. The rule create genuinely races the Worker — Cloudflare's validation simply never lost that race here.

## A hypothesis, stated as one

Where I hit this, it reproduces **every time** on a fresh stage. The difference I would look at first is what the Worker is:

- your fixture is a minimal Worker, 1.36 MB, no bindings;
- ours is a full application Worker — Hyperdrive, three queue producers, a KV namespace, Durable Objects, several secrets.

If the propagation window is roughly fixed, then the *slower* path is the safer one: a large Worker with bindings takes long enough to finish that the rule create lands well after the script is visible. That predicts the opposite of intuition — the harder a Worker is to create, the less likely this fires — and it does not match our 100% reproduction rate. So I do not think that is it either, and I would not want you to take the hypothesis as more than a starting point.

The other candidates I could not separate: account/colo differences in propagation, or something about a Worker that already occupied that *name* recently (our stage names recur across preview deploys; your fixture's name is unique per run).

## Suggestion

Consider splitting the two concerns, since one is testable deterministically and one is not:

- **The retry logic itself** — that `WorkerScriptNotFound` is retried, that the budget is bounded, and that the original error is re-raised unchanged after exhaustion — can be asserted against an injected failure, with no live API and no flakiness. That is the part that would actually catch a regression.
- **The live test** then stands as a smoke test that the happy path works end to end, which is worth having but should probably not be described as the regression guard for #1348.

If a deterministic version would help, I am glad to write it.

## Environment

- branch `fix/email-rule-script-not-found` at `463c20a`, submodule `distilled` at `302dcb6` (carries `WorkerScriptNotFound`, so the branch is self-contained)
- `pnpm@11.21.0`; note `@alchemy.run/cloudflare-test-tools:build` warns `This project requires Bun 1.3.13. Your current Bun is v1.3.11` — the build and the tests ran anyway, flagging it in case it matters
- API-token profile, `CLOUDFLARE_TEST_DNS_ZONE_NAME` set to a zone whose Email Routing was `unconfigured` with no rules
- I confirmed afterwards that the zone returned to `enabled=False / status=unconfigured` with its original single disabled catch-all, and that no `emailrulework*` Worker was left on the account. Cleanup is correct.
- the mutation was reverted and verified clean (`git diff` empty) before writing this

Happy to re-run anything, including against a different zone or with `DEBUG` set, if that would help narrow the propagation question.

Mkassabov:
Thanks — the mutation runs are the useful part, and you're right. I'd reached the same conclusion from the deploy log (the Worker provider pre-creates a stub script, so the name resolves before the rule is validated) but I left the live test carrying a load it can't bear. Your three green runs settle it.

I've taken the split you suggested. `test/Cloudflare/Email/retry.test.ts` injects the failure instead of hoping to race the real API:

- retries while `WorkerScriptNotFound`, then succeeds — 3 attempts
- does **not** retry an unrelated tag — 1 attempt, no backoff burned
- after the bounded budget, re-raises the original error unchanged — 9 attempts (initial + `times: 8`)

I ran your mutation against it. Body replaced with `effect`:

| run | `retryWorkerScriptNotFound` | live suite | `retry.test.ts` |
| --- | --- | --- | --- |
| 1 | as written | 2 passed | 3 passed |
| 2 | replaced with `effect` | 2 passed | **2 failed** |

So it discriminates where the live suite doesn't. `EmailRuleWorkerTarget.test.ts` is now labelled a smoke test in its own comment, pointing at the deterministic one, and the PR description says the same — no green run there should be read as evidence about #1348 either way.

On the propagation question: I can't reproduce it either, and I don't think your Worker-size hypothesis is wrong so much as untested — you're right that it predicts the opposite of your 100% rate. The name-reuse angle interests me more. Your stage names recur across preview deploys, so the script name recurs; my fixture's is unique per run. If Cloudflare caches a negative lookup keyed by script name, a name that recently existed and was deleted is exactly the case that would 2016, and a never-before-seen name would not. That's a guess, but it's the one that fits both our observations rather than contradicting one of them.

If you're up for one more run, the discriminating experiment would be: deploy to a fresh stage whose name you've used before (so the script name recurs), versus one with a random suffix, same Worker. If only the recurring name fails, that's the mechanism.

Either way the retry is cheap and only fires on a tag that means precisely "script not visible yet", so I'd rather land it than block on the root cause. The distilled side (alchemy-run/distilled#476) is what makes it a tag at all — `createRule` previously declared no typed errors, so there was nothing to catch.

Also noted your Bun version warning; it's unrelated to this change but worth someone looking at.

xiaoyu2er:
Follow-up now that this is merged, because a control run on our side contradicts the explanation in the merged test comment. Short version: **the stub pre-create happens here too, and the rule create fails anyway** — so the retry is load-bearing on this path, not defensive.

## The measurement

A stage name that had never been deployed, on `alchemy@2.0.0-beta.74` — i.e. **without** this PR — deploying a Worker and an `Email.Rule` whose worker action names it, in one apply:

| | plan | the rule | exit |
| --- | --- | --- | --- |
| first deploy | `Plan: 16 to create` | create → pending → creating → **fail** | 1 |
| re-run, same stage | `Plan: 1 to create, 2 to update, 13 to noop` | create → pending → creating → **created** | 0 |

Stack trace at `Cloudflare/Email/Rule.ts:147:23` → `provider.create`. The re-run's plan shows the one remaining resource was exactly that rule.

So the failure still reproduces on the unfixed version, and the re-run still clears it.

## Where it disagrees with the merged comment

> NOTE: this does NOT reproduce the #1348 failure on current main, and was confirmed to still pass with the provider's retry removed. The Worker provider pre-creates a stub script before the rule is created (see "pre-creating" in a deploy log), so by the time Cloudflare validates the action the name already resolves.

Grepping our **failing** deploy log for exactly that marker:

```
pre-creating   9 occurrences, all [app]
```

The stub pre-create ran, and Cloudflare still rejected the rule's worker action. So `pre-creating` being present is not sufficient for the name to resolve, and its presence is not what separates your test from this failure.

That leaves your non-reproduction unexplained rather than explained. Candidates I can see but cannot separate from here:

- **what the Worker is.** Yours is a minimal fixture; ours uploads 17.01 MB with Hyperdrive, three queue producers, KV, Durable Objects and cron triggers, and the log shows the rule create beginning while the Worker is still reconciling custom domains and cron triggers. If the stub only makes the *name* resolve and validation additionally needs something the real upload completes, a fixture that finishes quickly would never show it.
- **account or colo differences in propagation.**

I am not asking you to reopen anything — the fix is right either way. But I would not describe the retry as defensive: on this path it is the difference between a deploy that works and one that needs a manual re-run, and that is the shape the original report came from.

## One suggestion, unchanged from before

Since neither of us can make the live test fail on demand, the retry logic itself — retried on `WorkerScriptNotFound`, budget bounded, original error re-raised unchanged after exhaustion — is worth asserting against an injected failure. That would catch a regression; the live test currently cannot, as the three mutated runs showed. Still happy to write it if useful.

## What I could not capture

The error message text. alchemy's TUI reprints its full state on every repaint and elides the rest, so the log carries `[124 lines elided]` where the message was, and `.alchemy/log/out` records only INFO lines — `Workers Script Info not found` appears nowhere in the captured output. The stack trace is what identified the failure. A way to force non-TUI output for a captured deploy would have made this cheaper; if one exists and I missed it, I would like to know.

xiaoyu2er:
Heads-up on the PR build packages, reported because the bot's own command is the thing that fails: **`bun add https://pkg.ing/alchemy/<sha>` cannot be installed by bun**, in a clean directory, with nothing else involved.

I hit this trying to verify the merged fix against our deploy before a release — we reproduce #1348 100% of the time on a fresh stage, so a pre-release check was worth doing.

## Reproduction

```
$ mkdir /tmp/probe && cd /tmp/probe
$ echo '{"name":"probe","version":"0.0.0"}' > package.json
$ bun add https://pkg.ing/alchemy/b242389
bun add v1.3.11
Resolving dependencies
Resolved, downloaded and extracted [342]
error: @distilled.cloud/core@https://pkg.ing/@distilled.cloud/core/8b13420137320cb48af4c5efd6a1722c793fd090 failed to resolve
  … (x5)
error: @alchemy.run/node-utils@https://pkg.ing/@alchemy.run/node-utils/b242389d235d268d2560ae72a092ccabe80ab7cc failed to resolve
$ ls node_modules/alchemy
ls: node_modules/alchemy: No such file or directory
```

## What it is not

I narrowed it before reporting, because my first two guesses were both wrong:

| probe | result |
| --- | --- |
| the URLs are expired / 404 | **no** — all 200. My first probe said 403 for every one of them, including URLs bun had already fetched; that was my own client missing a `User-Agent`, not the server |
| the 40-char sha form is the problem | **no** — `bun add https://pkg.ing/@distilled.cloud/core/8b13420137320cb48af4c5efd6a1722c793fd090` succeeds on its own |
| scoped names are the problem | **no** — `@distilled.cloud/core` and `@alchemy.run/node-utils` both install fine on their own |
| our monorepo / catalog / overrides | **no** — fails identically in an empty directory |

## What it is

The published `alchemy` tarball declares **10 of its 38 dependencies as `https://pkg.ing/...` URL specifiers**:

```
@alchemy.run/floci, @alchemy.run/cloudflare-runtime, @alchemy.run/node-utils,
@distilled.cloud/{aws, axiom, cloudflare, core, hetzner, neon, planetscale}
```

bun resolves a URL specifier at the **top level** (that is why `bun add <url>` gets as far as downloading and extracting 342 packages) but not one that appears **inside a dependency's own `package.json`**. Overrides do not help: bun kept resolving the URL the package declares, not the one I overrode it with.

Since this is how the packages are published, every one of them is uninstallable by bun for the same reason.

## Why this is worth a moment

The bot comment on every PR says `bun add https://pkg.ing/<pkg>/<sha>`, and alchemy's own docs use bun. Whatever the publish step does when it rewrites `workspace:*`, resolving those to a registry-installable specifier — or publishing the whole set under one tag — would make PR builds usable by bun consumers. Right now the pnpm path works (your own repo is pnpm) and the bun path does not, which is the case the comment advertises.

## Not blocking anything for us

We will verify against a release instead. For the record, the tarball does carry the fix — I extracted it and `lib/Cloudflare/Email/retry.js` contains `retryWorkerScriptNotFound` and the exponential schedule — so this is purely about installability, not about the build being wrong.

And the reason we wanted an early check is worth passing on: our retry budget question is open. The merged schedule is `100ms` exponential × 8 ≈ **25.5s total**. Our failing deploy uploads a 17 MB Worker with 292 assets, Hyperdrive, three queue producers, KV, DOs and cron triggers, and the rule create begins while the Worker is still reconciling. Whether 25.5s covers that window is exactly what we cannot answer until we can run the fix. If it turns out not to, that is a number worth knowing — and I will report it either way once a release lands.

xiaoyu2er:
Your name-reuse hypothesis is testable against data I already have, and it does not survive it. Sorry — I posted the packaging note without reading this comment first, so that one talks past you in places.

## A never-before-used script name fails the same way

I ran a control on the unfixed version today (before a release, to make our eventual verification discriminate). The stage name was invented for the run and had never been deployed:

| | plan | rule | exit |
| --- | --- | --- | --- |
| first deploy | `Plan: 16 to create` | create → pending → creating → **fail** | 1 |
| re-run, same stage | `Plan: 1 to create, 2 to update, 13 to noop` | **created** | 0 |

The "never used" part is not me remembering — the deploy runs with `--adopt`, so a Worker that already existed under that name would have been adopted and shown as an update. **All 16 resources planned as `create`.** `alchemy state stages` listed only `prod` and `staging` beforehand.

So a script name Cloudflare has never seen still gets `WorkerScriptNotFound` on the rule create. A cached negative lookup keyed by name cannot be the necessary condition.

It could still be an *aggravating* one — I have not tested a recurring name against a fresh one side by side, and that experiment is now cheap for me because that stage was destroyed afterwards, so its name is in exactly the "recently existed, then deleted" state your hypothesis is about. Happy to run it if you still want the comparison; I just don't think it can explain the 100% rate any more.

## What the timeline does show

alchemy's own timestamped log, from the failing run:

```
15:09:54.347  Cloudflare Worker precreate: starting postflare-app-<stage>
15:09:58.406  Cloudflare Worker reconcile: starting
15:10:04.932  Cloudflare Worker create: uploading assets   (292 assets)
15:10:14.631  Cloudflare Worker create: uploading script   (17.01 MB)
15:10:26      deploy exits 1
```

The stub had existed for **32 seconds** before the deploy gave up, and the rule create still failed. Whatever makes the name resolvable for Email Routing's validation, `precreate` completing is not it — which is the same conclusion your deterministic split reached from the other direction, since you no longer rely on the live test to say anything about the window.

The other thing that line ordering shows: the rule create begins while the Worker is still reconciling custom domains and cron triggers. It is not waiting on the Worker resource, because it only names the script as a string.

## On landing it anyway

Agreed, and the mutation table you ran on `retry.test.ts` is the part that makes it safe to land without the root cause — it discriminates, so a future refactor cannot quietly remove the retry. That is exactly what the live suite could not do.

One number for whenever the root cause does come up: the budget is `100ms` exponential × 8 ≈ **25.5s**. I cannot yet tell whether that covers our window, because our window's length is unknown — we only know it is longer than the 32s the stub had been alive, and shorter than the gap before a manual re-run succeeded. If a release lands and a fresh stage still fails for us, that is not a regression, it means 25.5s is short for this shape, and I will report it with the timings rather than as an impression.

(Also: I could not try this against the PR build, for a bun packaging reason I described in the next comment — unrelated to the change itself.)

PR #1356, opened 2026-08-25
fix(aws-sqs): reconcile adoption and queue Standrd/FIFO changes

aryasaatvik:
Make SQS queue mode changes and explicit adoption converge when mutable settings depend on other resources.

```ts
const orders = yield* AWS.SQS.Queue("Orders", {
  queueName: "existing-orders",
  tags: { auditQueue: audit.queueArn },
}).pipe(Alchemy.adopt(true));
```

- Honor resource, stack, and default adoption settings during precreation. Explicit adoption preserves the queue and its messages while reconciling settings, policies, and tags; opting out still protects foreign queues.
- Add or remove the reserved `.fifo` suffix according to `fifo`, including for explicit names. Mode changes create a distinct replacement and update downstream references before deleting the old queue. Existing messages are not migrated between queue types.
- Detect identity changes independently of unresolved mutable settings, and cover AWS's 60-second deletion/recreation window with bounded retries.

Builds on the self-referential queue support merged in #1357, without core or planning-engine changes.

### DIFF
--- a/packages/alchemy/src/AWS/SQS/Queue.ts
+++ b/packages/alchemy/src/AWS/SQS/Queue.ts
@@ -5,6 +5,7 @@ import * as Effect from "effect/Effect";
 import * as Schedule from "effect/Schedule";
 import * as Stream from "effect/Stream";
 import { Unowned } from "../../AdoptPolicy.ts";
+import type { ScopedPlanStatusSession } from "../../Cli/Cli.ts";
 import { isResolved } from "../../Diff.ts";
 import { createPhysicalName } from "../../PhysicalName.ts";
 import * as Provider from "../../Provider.ts";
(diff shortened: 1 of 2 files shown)


packages/alchemy/src/AWS/SQS/Queue.ts
```diff
@@ -486,6 +498,108 @@ export const QueueProvider = () =>
           }
           // Return undefined to allow update function to be called for other attribute changes
         }),
+        precreate: Effect.fn(function* ({ id, news = {}, session }) {
```
sam-goodwin, line 632: can we share any logic with `create`? Looks a bit redundant.

aryasaatvik: I pulled the shared queue logic into helpers; precreate still only creates the shell. Pushed in 6c6d3f351

PR #1357, opened 2026-08-25
fix(aws): grant exact runtime capability actions

aryasaatvik:
Support versioned S3 operations and event delivery through runtime bindings, with least-privilege Secrets Manager reads.

```ts
const presign = yield* AWS.S3.PresignGetObject(bucket);
const url = yield* presign({
  key: "reports/annual.pdf",
  versionId: previousVersionId,
});
```

- Grant version reads for presigned GETs, CopyObject, and UploadPartCopy. Cross-bucket copying composes a separate source read binding.
- Preserve event version IDs and sequencers, decode object keys, and normalize event names to the declared `s3:` form.
- Precreate SQS identities to resolve circular policy bindings; `queueName` and `fifo` must be known beforehand. Reconcile Lambda, SQS, SNS, and EventBridge notification additions, updates, drift, and removal while preserving external configurations and migrating older managed Lambda targets.
- Run the Server SQS consumer without a deployment-only AWS environment dependency.
- Keep HeadObject's bucket-scoped listing permission and GetSecretValue's value-only permission. Conditional tag, ACL, KMS, and governance-bypass permissions remain separate.

Uses [Distilled #627](https://github.com/alchemy-run/distilled/pull/627) for signed version selection, typed S3/SQS errors, bodyless-request metadata headers, and empty XML configuration decoding.


packages/alchemy/test/AWS/S3/HeadObjectHttp.test.ts
```diff
@@ -0,0 +1,64 @@
...
+        "s3:GetObjectVersion",
+      ]);
+      expect(captured.policyStatements[1].Action).toEqual(["s3:ListBucket"]);
+    }).pipe(
+      Effect.provide(HeadObjectHttp),
+      Effect.provide(Credentials.mock),
+      Effect.provide(Region.of("us-east-1")),
+      Effect.provide(FetchHttpClient.layer),
+      Effect.provide(Layer.succeed(Self, host)),
+      Effect.provide(Layer.succeed(RuntimeContext, runtime)),
```
sam-goodwin: Never do subsequent Effect.provide. You need to provide once, use Layer.* to build the layer

packages/alchemy/test/AWS/S3/HeadObjectHttp.test.ts
sam-goodwin, line 1: These tests should be in fixtures that actually exercise the binding through a lambda function

sam-goodwin: Don't locally mock. Actually test it

sam-goodwin: We already have the pattern established, just add tests

aryasaatvik: I removed the mocks and moved coverage into the existing Lambda fixtures, including HeadObject. Pushed in 46ebfa109; targeted tests pass

PR #1359, opened 2026-08-25
test(prisma): cover database branch attachment lifecycle

kristof-siket:
Add live regression coverage for `Prisma.Database` branch attachment. The detachment fix is already on main; omitting branch props preserves existing attachments rather than moving databases back to the default branch.

- Cover default-branch attachment for named and generated databases, updates, and repeat deployments.
- Cover explicit branch IDs and git names, removing branch props, and preserving out-of-band attachment changes.
- Verify database identity stays stable and databases, branches, and projects are deleted.

Replace the old mocked suite with real deployments and independent reads through the Prisma SDK.

### DIFF
--- a/packages/alchemy/src/Prisma/App.ts
+++ b/packages/alchemy/src/Prisma/App.ts
@@ -142,13 +142,13 @@ const findApp = Effect.fn(function* (
     limit: 100,
   })).filter((app) => app.name === displayName);
   if (candidates.length === 0) return undefined;
-  const branch = yield* desiredBranchId(client, projectId, props);
-  if (!branch.resolved) return undefined;
-  const matches = candidates.filter((app) => app.branchId === branch.id);
+  const branchId = yield* desiredBranchId(client, projectId, props);
+  if (branchId === undefined) return undefined;
+  const matches = candidates.filter((app) => app.branchId === branchId);
   if (matches.length > 1) {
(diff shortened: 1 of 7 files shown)


sam-goodwin:
The original detachment bug was already fixed in [#1582](https://github.com/alchemy-run/alchemy/pull/1582), merged September 10. That change stopped sending `branchId: null` when branch props are omitted and rejected explicit `null`, which Prisma's API does not support.

I merged current main into this branch and changed this PR to live regression coverage rather than another provider behavior change:

- New databases with omitted branch props attach to the project's default branch through the Management API.
- Existing databases keep their observed attachment when branch props are omitted—even if it is a non-default branch or was changed outside Alchemy. This avoids the earlier version of this PR implicitly moving them back to the default branch.
- Replaced the mocked database suite with real deployments covering named/generated databases, branch IDs/git names, updates, repeat deployments, stable IDs, and deletion checks through the SDK.

The remaining diff is only `packages/alchemy/test/Prisma/Database.test.ts`; production provider code matches main. I updated the title and description accordingly.

Validation: 74 tests across the relevant Prisma suites passed, including both live lifecycle tests; workspace type checking and CI passed. Temporarily restoring either null-detachment updates or implicit default-branch reassignment made the new tests fail as expected. Those temporary changes were removed, and cleanup completed.

PR #1363, opened 2026-08-25
chore(website): refine logo assets

BlankParticle:
## Summary

- render the light, dark, and white background logo assets at 2048px
- reduce and center the yantra mark for more balanced surrounding space
- keep the generated outputs deterministic and sourced from the brand asset script

## Validation

- regenerated assets twice and verified identical checksums
- oxfmt check
- git diff check


website/public/alchemy-logo-light-bg.png
sam-goodwin, line 1: Why did the size increase? Does that affect anything else we haven't considered?

BlankParticle: its being rendered at 4x density, so the actual size is 2048px. 512px wasn't very good resolution for a lot of things, this should be fine, i used this in the github org icon, and these images are not used just uploaded manually whenever needed

PR #1375, opened 2026-08-26
fix(cloudflare/workers): simulate inbound email in alchemy dev

Mkassabov:
`Cloudflare.email().subscribe(...)` did not work under `alchemy dev`, in two independent ways.

### Handlers have to live on the bridge prototype

workerd resolves JSRPC methods only on the target entrypoint's *prototype chain*. An own instance property of the same name shadows the prototype entry and the lookup fails outright — it does not fall back:

```
TypeError: The RPC receiver does not implement the method "email".
```

`WorkerBridge` assigned the whole handler set in the constructor, leaving throwing stubs on the prototype. Move the real dispatch onto the prototype and drop the stubs:

```diff
-      for (const methodName of ExportedHandlerMethods) {
-        (this as any)[methodName] = async (input: any) => processEvent(...);
-      }
-
       return new Proxy(this, { /* user-shape RPC fallback */ });
```

```diff
 for (const method of ExportedHandlerMethods) {
   Object.defineProperty(WorkerBridge.prototype, method, {
-    value: function () {
-      throw new Error(`Bridge method '${method}' was called before instance setup`);
-    },
+    value: function (this: any, input: any) {
+      return processEvent(
+        (built) => built.export[method](input, this.env, this.ctx),
+        this.ctx,
+        this.env,
+        (exit) => exit._tag === "Success"
+          ? Promise.resolve(exit.value)
+          : Promise.reject(Cause.squash(exit.cause)),
+      );
+    },
```

`fetch`/`scheduled`/`queue` never hit this path — workerd dispatches those as built-in events. The local runtime is the exception: its entry worker forwards the `/cdn-cgi/handler/email` trigger route to the user worker as `env[USER_WORKER].email(message)`, a plain JSRPC call. So inbound email was the one handler that worked deployed and 500'd in dev.

Isolated with four workerd probes, calling `.email()` over a service binding:

| entrypoint shape | result |
| --- | --- |
| prototype method | resolves |
| own property assigned in constructor | `does not implement the method` |
| own property **+** prototype stub | `does not implement the method` |
| prototype method + constructor `Proxy` | resolves |

### Dev must not touch real Email Routing

Email resources have no local providers, so `email({ zone }).subscribe(...)` provisioned a real `Email.Routing` toggle and `Email.CatchAll` during `alchemy dev` — pointing a live zone's catch-all at a script that was never uploaded. If that landed it would silently take over inbound mail for the whole zone and drop it.

```diff
+const dev = yield* Effect.serviceOption(AlchemyContext).pipe(
+  Effect.map((ctx) => (ctx._tag === "Some" ? ctx.value.dev : false)),
+);
+
-if (!globalThis.__ALCHEMY_RUNTIME__ && props.zone !== undefined) {
+if (!globalThis.__ALCHEMY_RUNTIME__ && props.zone !== undefined && !dev) {
```

Local inbound is driven by the trigger route instead, which reaches the same registered listener.

`EmailEventSource.local.test.ts` covers the accept path and `setReject` end-to-end against the local simulator, and guards the prototype placement — `CronEventSource.local.test.ts` cannot, since `scheduled` never goes over RPC.

### DIFF
--- a/packages/alchemy/src/Cloudflare/Workers/EmailEventSource.ts
+++ b/packages/alchemy/src/Cloudflare/Workers/EmailEventSource.ts
@@ -264,67 +264,75 @@ export const EmailEventSourceLive = Layer.effect(
         message: ForwardableEmailMessage,
       ) => Effect.Effect<void, E, Req>,
     ) {
-      // Under `alchemy dev` the Worker only exists locally, so Cloudflare's
-      // mail pipeline has nothing to deliver to. Pointing a real zone's
-      // catch-all at a script that was never uploaded would fail — and if it
-      // did land, it would silently take over inbound mail for the whole zone
-      // and drop it. Local inbound is driven by the runtime's
-      // `POST /cdn-cgi/handler/email?from=&to=` trigger route instead, which
-      // dispatches to the same listener registered below.
-      const dev = yield* Effect.serviceOption(AlchemyContext).pipe(
-        Effect.map((ctx) => (ctx._tag === "Some" ? ctx.value.dev : false)),
-      );
+      // Deploy-time only: provision the Email.Routing toggle plus the
+      // routing resource that hands matched mail to this Worker.
+      //
+      // The guard is a bare `__ALCHEMY_RUNTIME__` check with every other
+      // condition nested inside, so the bundler folds it away and drops the
+      // whole block — and with it `AlchemyContext`, `Namespace`, `Routing`,
+      // `CatchAll` and `Rule` — from the deployed Worker. Hoisting the `dev`
+      // lookup out into a compound condition would keep it (and the context
+      // it reaches for) live in the runtime bundle.
+      if (!globalThis.__ALCHEMY_RUNTIME__) {
+        // Under `alchemy dev` the Worker only exists locally, so Cloudflare's
+        // mail pipeline has nothing to deliver to. Pointing a real zone's
+        // catch-all at a script that was never uploaded would fail — and if it
+        // did land, it would silently take over inbound mail for the whole zone
+        // and drop it. Local inbound is driven by the runtime's
+        // `POST /cdn-cgi/handler/email?from=&to=` trigger route instead, which
+        // dispatches to the same listener registered below.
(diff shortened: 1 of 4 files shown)


packages/alchemy/src/Cloudflare/Workers/EmailEventSource.ts
```diff
@@ -263,12 +264,23 @@ export const EmailEventSourceLive = Layer.effect(
...
       // Deploy-time: provision the Email.Routing toggle plus the routing
       // resource that hands matched mail to this Worker. Skipped once
-      // running inside the deployed Worker (the global guard) and when
-      // `zone` is omitted (bring-your-own routing). Namespaced under the
-      // host so logical identity is stable per Worker.
-      if (!globalThis.__ALCHEMY_RUNTIME__ && props.zone !== undefined) {
+      // running inside the deployed Worker (the global guard), when `zone`
+      // is omitted (bring-your-own routing), and in dev (above). Namespaced
+      // under the host so logical identity is stable per Worker.
+      if (!globalThis.__ALCHEMY_RUNTIME__ && props.zone !== undefined && !dev) {
```
sam-goodwin: not a regression but this won't be tree shaken. Might want to wrap `if (!globalThis.__ALCHEMY_RUNTIME__ ) { .. }` otuer and move the `if (props.zone != .. )` stiuff inside

Mkassabov: Good catch — fixed in bbcb7833f.

The compound condition was the smaller half of the problem: the `dev` lookup was hoisted *above* the guard, so `AlchemyContext` stayed live in the runtime bundle no matter what the bundler did with the `if`. Now it's a bare `__ALCHEMY_RUNTIME__` guard with everything nested inside, matching the AWS `*BindingHttp` idiom, so the whole block — `AlchemyContext`, `Namespace`, `Routing`, `CatchAll`, `Rule` — drops out of the Worker.

Same conditions, only the nesting changed. Local email dev-mode tests still green (4 passed).

PR #1379, opened 2026-08-26
fix(core): reuse pure Action outputs when inputs are unchanged

### LINKED ISSUE #1378: Worker plans never converge with Better Auth migrations
## Summary

`@alchemy.run/better-auth` places its migration Action output in the host env only to create a migration → Worker ordering edge. Action outputs remain unresolved during planning, so `WorkerProvider.diff` falls back to raw props and compares the expression with its persisted value. Every later plan proposes the same Worker update.

## Reproduction

Deploy a Worker using Better Auth automatic migrations, then run `alchemy plan` twice. The migration is unchanged, but the Worker remains dirty.

Observed with `alchemy@2.0.0-beta.74`, PlanetScale Postgres, and a Cloudflare Worker.

## Expected

The second plan is a noop while preserving the migration ordering edge.

focux:
Treat completed Actions as pure functions of their inputs: when the inputs are known and unchanged, resolve the saved output during planning so downstream resources can also plan as no-ops.

This fixes repeated Worker updates after Better Auth migrations without introducing a separate dependency API. Changed, unfinished, and forced Actions keep their outputs unresolved, preserving execution ordering and failure propagation through existing output references.

- Recheck inputs during apply when they were unknown during planning; skip the Action body if they resolve to the last successful input, unless forced.
- Distinguish valid falsy outputs from missing outputs.
- Add regression coverage in `plan.test.ts` and `apply.test.ts` for chained Actions, props and bindings, invalidation, forced runs, replacement, failure recovery, and recursive planning.

Closes #1378.

### DIFF
--- a/packages/alchemy/src/Plan.ts
+++ b/packages/alchemy/src/Plan.ts
@@ -933,13 +933,21 @@ export const make = <A>(
     const newUpstreamDependencies: {
       [fqn: string]: string[];
     } = Object.fromEntries([
-      ...resources.map(
-        (resource) =>
-          [
-            resource.FQN,
-            Object.values(Output.upstreamAny(resource.Props)).map((r) => r.FQN),
-          ] as const,
-      ),
+      ...resources.map((resource) => {
+        // Resource proxies synthesize attribute Outputs for missing keys, so
+        // only read explicit dependencies when the resource owns the field.
+        const explicitDependencies = Object.hasOwn(resource, "Dependencies")
+          ? resource.Dependencies
+          : [];
+        const dependencies = [
+          ...Object.values(Output.upstreamAny(resource.Props)),
+          ...Object.values(Output.upstreamAny(explicitDependencies ?? [])),
+        ];
+        return [
+          resource.FQN,
+          Array.from(new Set(dependencies.map((resource) => resource.FQN))),
+        ] as const;
+      }),
       ...actions.map(
         (action) =>
           [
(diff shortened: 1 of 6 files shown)


packages/alchemy/src/Plan.ts
sam-goodwin, line 1: if you change plan, you gotta add corresponding `plan.test.ts` cases to document and enforce this

sam-goodwin:
Thanks for tracking this down. I took a different approach to the fix: instead of removing the migration output from the host environment and introducing a separate `Dependencies` API, this fixes how the engine resolves Action outputs.

The planner already knew an Action was a no-op when its inputs were unchanged, but it still left the Action's output unresolved. That made downstream resources look changed on every deployment.

Actions are assumed pure in their inputs, so a completed Action with known, unchanged inputs can expose its saved output during planning. Consumers can then also plan as no-ops. Changed, unfinished, and forced Actions remain unresolved and retain the existing output-based execution dependencies. Apply also rechecks inputs that were unknown during planning: if an upstream reruns but produces the same input, the downstream Action reuses its previous result unless forced.

This also avoids a safety gap in the original approach: the new `Dependencies` field added graph edges in Plan, but Apply still derived its waits only from props and bindings. The regression tests showed that a host could reconcile before the Action completed, or even after it failed. Keeping the existing output references preserves those waits and failure propagation.

The updated `plan.test.ts` and `apply.test.ts` coverage includes repeated no-op deployments, chained Actions, props and bindings, forced runs, replacement, failure recovery, and falsy outputs. After merging current main, all 648 tests in the 12 engine suites and the full workspace type-check pass.

PR #1383, opened 2026-08-27
fix(cloudflare): validate AI Search custom metadata types

aryasaatvik:
## Summary

- expose AI Search custom metadata as the closed Cloudflare contract: `text | number | boolean | datetime`
- reject widened JavaScript/TypeScript values before the provider makes a Cloudflare request
- exercise accepted and rejected values through the existing live provider lifecycle

## Validation

- `mise exec bun@1.3.13 -- bun alchemy-test test/Cloudflare/AI/SearchInstance.test.ts -t "create, update mutable props|rejects custom metadata" --profile testing --sequential --retry 0`
- 2/2 focused live-provider tests passed
- post-test Cloudflare readback found no test AI Search instances or matching R2 buckets
- `pnpm exec oxfmt packages/alchemy/src/Cloudflare/AI/SearchInstance.ts packages/alchemy/test/Cloudflare/AI/SearchInstance.test.ts`
- changed-file TypeScript diagnostics and `git diff --check` passed


sam-goodwin:
is the AU.CustomMetadata type wrong? Seems like something we can catch at the type level.

aryasaatvik:
> is the AU.CustomMetadata type wrong? Seems like something we can catch at the type level.

Yes—the generated type is too wide for this field. Distilled [reopens request enums here](https://github.com/alchemy-run/distilled/blob/cc93bf364f93840f008e0a50e8edbaab63348a9d/packages/core/src/codegen/generator.ts#L512-L558) and emits S.String. We could support a closed enum there, emit literal validation, and use that generated type in Alchemy.

sam-goodwin:
We leave those open intentionally for backwards compatibility. So that would mean a runtime or type level check would be incompatible. Does the service already provide the error? If so, why not just rely on that?

PR #1403, opened 2026-08-29
fix(cloudflare/workers): provide workerd-safe runtime services

harrysolovay:
## Summary

- replace the aggregate `NodeServices.layer` with the specific Node platform services supported by workerd, excluding `NodeTerminal`
- compose `NodeChildProcessSpawner.layer` with `FileSystem` and `Path` via `Layer.provideMerge` so its requirements are satisfied
- provide the runtime `Alchemy.Stage` service from `stack.stage`

## Background

`NodeServices.layer` initializes `NodeTerminal`, which calls `process.stdin.once(...)`. Cloudflare's Node compatibility shim does not implement `.once()`, so Worker startup failed with `t.once is not a function` before the user handler ran.

Simply merging the remaining platform layers as siblings is insufficient because `Layer.mergeAll` does not use sibling outputs to satisfy sibling requirements. In particular, `NodeChildProcessSpawner.layer` requires `FileSystem` and `Path`, which led to `Service not found: effect/platform/FileSystem`. This change retains dependency-aware composition while omitting the unsupported terminal service.

## Verification

- commit hook formatting passed
- further checks skipped as requested
- `pnpm install --frozen-lockfile` reached the repository prepare step, which could not find the uninitialized `distilled/packages/cloudflare` submodule in the fresh clone

### DIFF
--- a/packages/alchemy/src/Cloudflare/Workers/WorkerBridge.ts
+++ b/packages/alchemy/src/Cloudflare/Workers/WorkerBridge.ts
@@ -1,5 +1,9 @@
 import type * as cf from "@cloudflare/workers-types";
-import * as NodeServices from "@effect/platform-node/NodeServices";
+import * as NodeChildProcessSpawner from "@effect/platform-node/NodeChildProcessSpawner";
+import * as NodeCrypto from "@effect/platform-node/NodeCrypto";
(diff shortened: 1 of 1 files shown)


sam-goodwin:
Can you provide a reproduction? I am confused about this error.

PR #1405, opened 2026-08-29
fix(cloudflare): route asset upload sessions to the dispatch-namespace endpoint

pltoledo:
Asset upload sessions for Workers deployed into a dispatch namespace were created on the account-level endpoint, so the completion JWT comes back scoped to the wrong script identity and the namespaced script PUT is rejected (400 on a new script, 500 on an existing one).

`uploadAssets` now takes the namespace and picks the endpoint the same way the namespace-aware helpers in `WorkerProvider` already do (`putWorkerScript`, `getScriptSettings`, `deleteWorkerScript`):

```ts
const session = dispatchNamespace
  ? yield* wfp.createDispatchNamespaceScriptAssetUpload({
      accountId,
      dispatchNamespace,
      scriptName: workerName,
      manifest: assets.manifest,
    })
  : yield* workers.createScriptAssetUpload({
      accountId,
      scriptName: workerName,
      manifest: assets.manifest,
    });
```

The version-upload call site is unchanged: namespaced scripts have no versions API and the provider already rejects `version` + `namespace`.

We run this change in production as a patch over `2.0.0-beta.72` (Workers for Platforms platform publishing sites into a dispatch namespace, static assets + prebuilt SSR workers); with it, `Worker` with `namespace` and `assets` deploys end to end.

### DIFF
--- a/packages/alchemy/src/Cloudflare/Workers/Assets.ts
+++ b/packages/alchemy/src/Cloudflare/Workers/Assets.ts
@@ -1,4 +1,5 @@
 import * as workers from "@distilled.cloud/cloudflare/workers";
+import * as wfp from "@distilled.cloud/cloudflare/workers-for-platforms";
 import * as Data from "effect/Data";
 import * as Effect from "effect/Effect";
 import * as FileSystem from "effect/FileSystem";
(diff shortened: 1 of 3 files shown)


sam-goodwin (commented):
Please add a regression test

pltoledo:
Added a regression test in `test/Cloudflare/WorkersForPlatforms/DispatchNamespace.test.ts`. It deploys a Worker with assets into a namespace, then changes an asset and redeploys. Ran it live on a WFP-entitled account: green on this branch, fails on main.

PR #1409, opened 2026-08-30
feat(cloudflare/stream): surface live input WebRTC (WHIP/WHEP) endpoints

Mkassabov:
Reported on Discord as "cloudflare seems to be missing a webrtc/signaling resource". There is no separate resource to add — WHIP and WHEP *are* the signaling, carried over plain HTTP, and every Stream live input is already reachable over both. The actual gap was that `LiveInput` surfaced **no ingest or playback endpoints at all**: you could create a live input and had no way to get a URL to stream to.

distilled already types all of them on the create/get/update responses, so no SDK patch was needed.

```diff
  export type LiveInputAttributes = {
    liveInputId: string;
    // …
+   webRTCUrl: Redacted.Redacted<string> | undefined;         // WHIP publish
+   webRTCPlaybackUrl: string | undefined;                    // WHEP playback
+   rtmps: LiveInputRtmpsEndpoint | undefined;
+   rtmpsPlayback: LiveInputRtmpsEndpoint | undefined;
+   srt: LiveInputSrtEndpoint | undefined;
+   srtPlayback: LiveInputSrtEndpoint | undefined;
  };
```

The WHIP URL is `Redacted` because it embeds a secret that grants publish access (`.../<SECRET>/webRTC/publish`); the WHEP URL is what you hand viewers, so it stays a plain string. Same split for `streamKey` / `passphrase`. `list()` does not return endpoint details, hence `| undefined` on those rows rather than a fabricated URL.

### Example

`examples/cloudflare-stream-webrtc` — a live input plus a Worker that demonstrates the asymmetry between the two URLs:

```ts
env: {
  WHIP_URL: broadcast.webRTCUrl,          // Redacted -> secret_text
  WHEP_URL: broadcast.webRTCPlaybackUrl,  // plain string
}
```

The browser never sees `WHIP_URL`. It posts its SDP offer to the Worker's `/whip` route, which authorizes the caller and forwards to Cloudflare. Playback fetches the WHEP URL from `/whep` and negotiates with Cloudflare directly.

@see https://developers.cloudflare.com/stream/webrtc-beta/

### DIFF
--- a/examples/cloudflare-stream-webrtc/src/Api.ts
+++ b/examples/cloudflare-stream-webrtc/src/Api.ts
@@ -1,0 +1,94 @@
+import * as Cloudflare from "alchemy/Cloudflare";
+import * as Config from "effect/Config";
+import * as Effect from "effect/Effect";
+import * as Redacted from "effect/Redacted";
+import * as HttpClient from "effect/unstable/http/HttpClient";
+import * as HttpClientRequest from "effect/unstable/http/HttpClientRequest";
+import { HttpServerRequest } from "effect/unstable/http/HttpServerRequest";
+import * as HttpServerResponse from "effect/unstable/http/HttpServerResponse";
+
+import { Broadcast } from "./Broadcast.ts";
+import { PAGE } from "./page.ts";
+
+const whipUrl = Config.redacted("WHIP_URL");
+const whepUrl = Config.string("WHEP_URL");
+const broadcastKey = Config.redacted("BROADCAST_KEY");
+
+/**
+ * A Worker that fronts a Stream live input's WebRTC endpoints.
+ *
+ * The two URLs are handled very differently on purpose:
+ *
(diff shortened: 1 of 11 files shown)


examples/cloudflare-stream-webrtc/src/Api.ts
```diff
@@ -0,0 +1,94 @@
...
+        WHIP_URL: broadcast.webRTCUrl,
+        WHEP_URL: broadcast.webRTCPlaybackUrl,
+        // Whatever you already use to decide who may broadcast. A shared
+        // key keeps the example to one moving part.
+        BROADCAST_KEY: Redacted.make("dev-broadcast-key"),
+      },
+    };
+  }),
+  Effect.gen(function* () {
+    const client = yield* HttpClient.HttpClient;
```
sam-goodwin, line 47: If using an Effectful worker, don't use `env`, use this in the initialize phase:
```ts
const WHIP_URL = yield* Config.redacted("WHIP_URL");
// etc.
```

PR #1410, opened 2026-08-30
fix(cloudflare): make vite main authoritative over an app-declared entry input

filipfalcon:
Vite folds every plugin's `config` result into the config the next plugin sees, so the entry input this plugin returns is unioned with an input the app's own plugins declared for the same environment. A framework that names its own server entry — Foldkit's `ssr.build`, for one — left two entry chunks in the bundle, and the deployed Worker became whichever one the bundle listed first: the framework's server entry rather than the module `main` names. That Worker exports no handler, and nothing warned.

Rewrite the entry environment's input from the per-environment `options` hook, which runs after config resolution, where nothing merges over it. Only when `main` is set, and only for the entry environment — children keep the entries their framework gave them.

### DIFF
--- a/packages/cloudflare-runtime/src/rolldown/plugins/options.ts
+++ b/packages/cloudflare-runtime/src/rolldown/plugins/options.ts
@@ -248,6 +248,28 @@ export const optionsPlugin = createPlugin<"options", OptionsApi>(
             },
           };
         },
+        // Vite merges every plugin's `config` result into the config the next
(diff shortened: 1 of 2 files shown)


sam-goodwin:
It's not clear to me what problem this solves and why a blanket solution is the correct approach? Can you elaborate?

filipfalcon:
I will adress it in https://github.com/alchemy-run/alchemy/pull/1313/.

PR #1412, opened 2026-08-30
fix(cli): only load floci CA when it exists

BlankParticle:


packages/alchemy/src/Cli/commands/dev.ts
```diff
@@ -39,10 +40,14 @@ export const devCommand = Command.make(
...
       });
+
+      const fs = yield* FileSystem.FileSystem;
       // Set on THIS process too, so the RPC spawner's sidecars (and the workerd
       // they launch) inherit it — they are forked from here, not from the exec
       // child below.
-      process.env.NODE_EXTRA_CA_CERTS = NODE_EXTRA_CA_CERTS;
+      if (yield* fs.exists(Floci.FLOCI_CA_PATH)) {
+        process.env.NODE_EXTRA_CA_CERTS ??= Floci.FLOCI_CA_PATH;
+      }
```
sam-goodwin, line 50: floci certificate should probably be managed in AWS.providers(). How can we have it so `alchemy dev` should not understand floci. We can merge this for now as it fixes an issue, but we need to invert it.

PR #1415, opened 2026-08-31
fix(cloudflare/workers): type named entrypoint bindings with their RPC surface

Mkassabov:
Type named Worker entrypoint bindings with Cloudflare's existing `Service<Api>`.

```ts
import type { Api } from "./target/src/worker.ts";

env: {
  API: Cloudflare.WorkerEntrypoint<Api>(target, "Api"),
}

// InferEnv exposes Service<Api>:
const greeting = await env.API.greet("alice");
```

The generic takes the entrypoint's instance type (`Api`), not its constructor (`typeof Api`). `InferEnv` delegates directly to `Service<Entrypoint>` without additional RPC conversion helpers.

Untyped entrypoint bindings remain `Fetcher`s and do not inherit the target Worker's default-entrypoint methods. Direct Effect-native Worker bindings retain their existing wire types.

Run affinity lifecycle tests sequentially against their shared zone ruleset, and check readiness on the actual JSON reads in the Durable Object abort test.

### DIFF
--- a/packages/alchemy/src/Cloudflare/Workers/InferEnv.ts
+++ b/packages/alchemy/src/Cloudflare/Workers/InferEnv.ts
@@ -58,8 +58,8 @@ export type GetBindingType<T> =
   // A named-entrypoint service binding (`Cloudflare.WorkerEntrypoint`).
   // Tested first: the marker *contains* a Worker, so the later Worker
   // branches must never see it.
-  T extends WorkerEntrypointBinding
-    ? Fetcher
+  T extends WorkerEntrypointBinding<infer Entrypoint, infer Target>
+    ? EntrypointStub<Entrypoint, Target>
     : // A Container bound in `env` is a container-backed Durable Object class —
       // the runtime binding is the class's namespace. Must be tested BEFORE the
       // generic Effect unwrap: a Container declaration is itself an Effect.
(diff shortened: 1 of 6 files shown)


packages/alchemy/test/types/WorkerEntrypointEnv.ts
```diff
@@ -0,0 +1,40 @@
...
+  /** Non-promise returns are promisified by the RPC stub type. */
+  count(): number;
+}
+
+declare const target: Cloudflare.Worker;
+
+export const Worker = Cloudflare.Worker("EntrypointEnvTypeProbe", {
+  script: "export default {}",
+  env: {
+    API: Cloudflare.WorkerEntrypoint<typeof Api>(target, "Api"),
```
sam-goodwin, line 21: Why not `Cloudflare.WorkerEntrypoint<Api>(target, "Api")`

packages/alchemy/src/Cloudflare/Workers/InferEnv.ts
```diff
@@ -153,6 +153,37 @@ export type GetBindingType<T> =
...
+      ? EntrypointStubOf<Instance>
+      : EntrypointStubOf<Entrypoint>;
+
+/** {@link EntrypointStub} for an entrypoint *instance* type. */
+type EntrypointStubOf<Instance> = [Instance] extends [
+  Rpc.WorkerEntrypointBranded,
+]
+  ? Service<Extract<Instance, Rpc.WorkerEntrypointBranded>>
+  : // A plain method-bag shape (no `cloudflare:workers` brand).
+    Fetcher & { [K in keyof Instance]: Instance[K] };
```
sam-goodwin, line 185: Did we not already have something for this?

PR #1425, opened 2026-09-01
feat(forgejo): Forgejo provider for self-hosted instances

wyattjoh:
Add a Forgejo provider for self-hosted instances, using `@distilled.cloud/forgejo` for resource lifecycles and provider-independent HTTP bindings.

Based on #1731, which contains the shared Worker routing and webhook acknowledgement changes. `GitHub/Secrets.ts` remains unchanged.

### Resources

```ts
import * as Forgejo from "alchemy/Forgejo";

export default Alchemy.Stack(
  "Repos",
  { providers: Forgejo.providers(), state: Alchemy.localState() },
  Effect.gen(function* () {
    const repo = yield* Forgejo.Repository("api", {
      owner: "acme",
      name: "api",
      private: true,
      autoInit: true,
      topics: ["typescript", "effect"],
    });
    return { url: repo.htmlUrl };
  }),
);
```

Includes repositories, organizations, teams, memberships, labels, branch protection, Actions secrets/variables, webhooks, and API tokens.

- Existing foreign resources require explicit adoption. Repositories and organizations are retained by default.
- Repository renames and observed same-ID transfers preserve identity. Deletion never targets a namesake after a saved ID disappears.
- Reconciliation compares declared settings against observed state and preserves omitted settings.
- Token scopes must be nonempty; empty repository restrictions are rejected before replacement.
- Forgejo 16's immutable webhook secrets use delete-first replacement, with a possible delivery gap. Unapplied organization settings and unrecoverable token plaintext produce explicit errors.

### HTTP bindings

Provide the same Forgejo layers inside a Worker, Lambda function, or another Alchemy runtime host:

```ts
Effect.gen(function* () {
  const repo = yield* Repository;
  const repository = yield* Forgejo.ReadRepository(repo);

  return {
      const details = yield* repository.get().pipe(Effect.orDie);
      return yield* HttpServerResponse.json(details);
    }),
  };
}).pipe(Effect.provide(Forgejo.ReadRepositoryHttp));
```

Repository and issue capabilities each have `Read`, `Write`, and `ReadWrite` contracts with corresponding `*Http` layers under `alchemy/Forgejo`. They use shared host/output binding APIs, without provider-specific client collections or Forgejo code in host provider directories.

Bindings automatically create repository-restricted tokens. The deployment credential stays out of the runtime. Identical managed capabilities reuse credentials; read and write permissions stay separate. Managed token rotation provisions the successor before updating the host and revoking the predecessor. Removing bindings removes their managed auth resources; external credentials remain externally owned.

Repository-restricted tokens cannot update topics on Forgejo 16; that operation preserves the SDK's typed `Forbidden` error rather than broadening permissions.

### Signed repository events

Incoming webhook adapters are separate from the HTTP clients:

```ts
import { RepositoryEventSourceCloudflare } from "alchemy/Forgejo/RepositoryEventSourceCloudflare";

const subscribe = Effect.gen(function* () {
  const repo = yield* Repository;
  yield* Forgejo.RepositoryEventSource(
    repo,
    { events: ["push"] },
    (event) => Effect.log(`Received ${event.payload.ref}`),
  );
}).pipe(Effect.provide(RepositoryEventSourceCloudflare));
```

Yield the subscription during Worker initialization. Lambda uses `RepositoryEventSourceLambda` from its Forgejo leaf module. Both generate persisted signing secrets, configure webhooks, and verify signed push/issue deliveries.

Subscriptions with the same host, repository, and event selection share one receiver and signing secret. It verifies and decodes each request once, runs all subscribers, and returns `202` only after all succeed. Failures or a 30-second timeout return `503`. Delivery is at least once: redelivery runs successful subscribers again, so side effects must be idempotent.

### Authentication and documentation

```sh
alchemy profile edit --add Forgejo
```

The profile stores the instance URL and credential. CI uses `FORGEJO_URL` and `FORGEJO_TOKEN`; explicit `providers({ baseUrl, token })` is also supported. Automatic runtime-token provisioning requires an administrator profile with token-management permissions. Webhook-only subscriptions need hook-management permission, not token-minting permission.

The `/forgejo` guides cover setup, ownership, repositories, Actions configuration, organizations, HTTP bindings, and repository events. API references are generated from source JSDoc.


packages/alchemy/src/Forgejo/Client.ts
sam-goodwin, line 1: This needs to be deleted ands implemented as a generated SDK in https://github.com/alchemy-run/distilled and patched to include tagged errors. We don't accept providers without this.

sam-goodwin: cc @Mkassabov who can help with generating SDKs

wyattjoh: Opened a PR over here https://github.com/alchemy-run/distilled/pull/541, will update this one once it's merged and available.

wyattjoh:
@sam-goodwin `Client.ts` is gone — every resource now calls through the generated `@distilled.cloud/forgejo` SDK, with the tagged errors patched in. Ready for a look when you have time.

PR #1445, opened 2026-09-01
perf(cloudflare/workers): upload asset buckets concurrently

BlankParticle:
Bulk asset buckets were uploaded one at a time. This matches wrangler: three buckets in flight, and the first gateway error (502/503/504) drops the deploy to one bucket at a time for the rest of the run, retrying the failed bucket with backoff.

```ts
const semaphore = yield* Semaphore.make(3);

createAssetUpload({ ... }).pipe(
  // let 502/503/504 surface instead of ~20s of hidden SDK retries
  Retry.policy(surfaceGatewayErrors),
  Effect.tapError((e) => isGatewayError(e) ? semaphore.resize(1) : Effect.void),
  Effect.retry({ while: isGatewayError, schedule: Schedule.exponential("2 seconds"), times: 5 }),
);

yield* Effect.forEach(
  session.buckets,
  (bucket) => uploadBucket(bucket).pipe(semaphore.withPermits(1)),
  { concurrency: 3 },
);
```

- The semaphore is hoisted above the session retry, so a JWT-expiry restart stays degraded instead of bursting back to three.
- Every bucket response is checked for the completion JWT, since under concurrency the last to finish is not necessarily the last in the list.

### DIFF
--- a/packages/alchemy/src/Cloudflare/Workers/Assets.ts
+++ b/packages/alchemy/src/Cloudflare/Workers/Assets.ts
@@ -1,15 +1,11 @@
 import * as workers from "@distilled.cloud/cloudflare/workers";
 import * as wfp from "@distilled.cloud/cloudflare/workers-for-platforms";
-import {
-  BadGateway,
-  GatewayTimeout,
-  Retry,
-  ServiceUnavailable,
-} from "@distilled.cloud/cloudflare";
+import { Retry } from "@distilled.cloud/cloudflare";
 import * as Data from "effect/Data";
 import * as Effect from "effect/Effect";
 import * as FileSystem from "effect/FileSystem";
 import * as Path from "effect/Path";
+import * as Predicate from "effect/Predicate";
 import * as Schedule from "effect/Schedule";
 import * as Semaphore from "effect/Semaphore";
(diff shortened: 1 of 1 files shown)


packages/alchemy/src/Cloudflare/Workers/Assets.ts
```diff
@@ -453,6 +486,71 @@ export const uploadAssets = Effect.fn(function* (
   // concurrent tool's transient chdir (framework source builds).
   const directory = path.resolve(initialCwd, assets.directory);
 
+  // Bucket concurrency, wrangler-style: three in flight, and the first
+  // gateway error resizes it to one for the rest of the deploy. Hoisted out
+  // of the session so a JWT-expiry retry stays degraded instead of bursting
+  // back to three against a gateway that just told us to back off.
+  const semaphore = yield* Semaphore.make(BULK_UPLOAD_CONCURRENCY);
```
sam-goodwin, line 493: Just use `Effect.all` with `concurrency: 3`?

BlankParticle: its matching wrangler behaviour to reduce concurrency to 1 on errors

PR #1448, opened 2026-09-02
fix(package): include agent instruction files

### LINKED ISSUE #1286: Agents.md/docs
Hey mate, recently migrated over to alchemy and I've been loving it so far!

It would be great if there was relevant docs or agents.md/claude.md files bundled in to the pkg in ~/node_modules/alchemy :)

sunxiayi:
Copy the repository's agent instructions into the published `alchemy` package so installed copies expose the same guidance as the source checkout.

```diff
+ AGENTS.md
+ CLAUDE.md
```

Both files are copied by the existing package-file build step and explicitly included in the npm `files` allowlist.

Fixes #1286

### DIFF
--- a/packages/alchemy/package.json
+++ b/packages/alchemy/package.json
@@ -26,6 +26,8 @@
     "!bin/register-dev-mode.js",
     "lib",
     "src",
+    "AGENTS.md",
+    "CLAUDE.md",
     "NOTICE",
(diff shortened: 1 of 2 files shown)


sam-goodwin:
Our instructions are specific to the repo. Why do you want them available? Is it to build custom resources or something?

PR #1450, opened 2026-09-02
feat(gcp): IAM members, service accounts, Router NAT, and GKE/VM props

connyay:
ServiceAccount, Project/ServiceAccount/CryptoKey/Secret IamMember (one locked read-modify-write that Host.ts now reuses), RouterNat as a serialized router.nats entry, and ServiceUsage.ProjectService. Instance, Cluster, and NodePool gain scheduling, identity, Shielded VM, networking, addon, and observability props.

GKE waits are sized for real operations, delete settles in-flight ops first, logging changes pin the monitoring service, and GKE_METADATA follows the host's Workload Identity. Diffs read observed output before persisted olds so declaring GKE-injected defaults is a no-op.


sam-goodwin:
Can we cross reference this with our gcp pr that is generating 100% coverage https://github.com/alchemy-run/alchemy/pull/1336 @Mkassabov

connyay:
This has that branch as a base. I fell down the rabbit hole kicking tires on that branch and stumbled across a few gaps. Feel free to close / ignore if not helpful. Test loops are super slow and painful with GKE so figured this might be worth contributing back.

sam-goodwin:
That is extremely helpful. Thanks! We should just merge them in to it.

connyay:
Will follow up if needed after https://github.com/alchemy-run/alchemy/pull/1336 lands

PR #1453, opened 2026-09-02
feat(dev): dev relay for stable public subdomains

sam-goodwin:
Stable public subdomains for `alchemy dev`, without tunnels: one WebSocket from the dev sidecar to a relay Worker, which routes `https://<name>.<namespace>.<domain>` back down it by `Host`.

```sh
ALCHEMY_DEV_RELAY_TOKEN=… alchemy dev --relay https://dev.alchemy.run
```

```
Connected to the dev relay: http://<name>.dev-sam.dev.alchemy.run
[Api] → https://api.dev-sam.dev.alchemy.run
[Web] → https://web.dev-sam.dev.alchemy.run
```

Quick tunnels (#1449) can't multiplex — one random `*.trycloudflare.com` hostname per `cloudflared` process. The relay is what a managed service looks like: real subdomains, names that survive restarts, no binary, no process per resource, and cookies scoped to `<namespace>.<domain>` work across a session's resources like they will in production.

### How it works

- `Local/Relay/RelayWorker.ts` — an Effect-native `Cloudflare.Worker` on `*.<domain>/*` hosting the `RelaySession` Effect `Cloudflare.DurableObject`. `GET /__relay/connect?namespace=<ns>` (bearer token) is forwarded to the DO named `<ns>`, which upgrades it with `Cloudflare.upgrade()` and keeps the socket (hibernation API). Configuration is plain `Config` at init (`DEV_RELAY_DOMAIN`, `DEV_RELAY_ZONE`, `DEV_RELAY_SCHEME`, `DEV_RELAY_TOKEN`), captured as bindings at deploy — routes included. Every request for `<label>.<ns>.<domain>` is routed to that DO and forwarded down the socket as a multiplexed stream: JSON control frames for heads/end/abort, binary frames (4-byte stream id + bytes, 64 KiB chunks) for bodies, streamed both ways. No connector → a clear 502 page; no answer in 30s → 504.
- `Local/Relay/RelayConnector.ts` — the sidecar side: one socket per session, reconnect with capped backoff, each request answered from the local ingress with the `x-alchemy-ingress-host` route hint (`api` → `api.localhost`), `x-forwarded-host/proto` set to the public values.
- `Local/Relay/Relay.ts` — `yield* DevRelay` deploys the Worker (+ DO, routes) and the wildcard/apex DNS records with Alchemy itself, reading the same `Config`; `stacks/dev-relay.ts` is Alchemy's own.
- `--relay <url>` / `--relay-namespace` on `alchemy dev` (namespace defaults to the stage). The relay URL becomes a Worker's primary `url`; quick tunnels still work alongside.

Verified end to end through the real edge on the test zone: a public `http://api.reltest.relay.alchemy-test-2.us/echo` request → Worker → DO → WebSocket → sidecar → local ingress → local Worker, CORS headers intact, 502 for an unconnected namespace.

### Before hosting it

- Two-level hostnames need a cert covering `*.<namespace>.<domain>`: Advanced Certificate Manager (or Total TLS) on the zone. Tests use plain HTTP for that reason.
- Put `<domain>` on the Public Suffix List (like `workers.dev`) so browsers isolate namespaces as separate sites.
- Auth is a single shared bearer token for now; a managed service swaps that for per-account tokens on connect.
- Not yet: WebSocket passthrough (HMR through the relay), flow control for very large streamed bodies.


sam-goodwin:
Superseded: the relay moves to Fly (the two-level wildcard certificate problem rules out Cloudflare's edge). The protocol, connector, and CLI flags carry over into the Fly relay PR; the Cloudflare Worker + Durable Object server is dropped.

PR #1457, opened 2026-09-02
feat(cloudflare/workflows): export makeWorkflowName

### LINKED ISSUE #1456: cloudflare/workflows: export makeWorkflowName so a stack can reference a Workflow's derived name
### Problem

A stack that needs a Workflow's physical name outside the Workflow itself cannot get it. The name is derived by `makeWorkflowName(scriptName, className)` in `packages/alchemy/src/Cloudflare/Workflows/WorkflowName.ts`, which is `@internal` and not exported from `Cloudflare/Workflows`.

Our case: a Queue subscription whose source is a Workflow that an async Worker binding registers.

```ts
const worker = yield* Cloudflare.Worker("workflows-host", { bindings: { FILE_URL_INGESTION: Cloudflare.Workflow("FileUrlIngestion", { className: "FileUrlIngestionWorkflow" }) } })

yield* Cloudflare.Queues.Subscription("terminal-events-subscription", {
  source: { type: "workflows.workflow", workflowName: ??? },
  ...
})
```

- `WorkflowResource.ref(...)` cannot resolve on the first deployment because the Workflow is created in the same pass.
- Re-registering `WorkflowResource` with the same `workflowName` is idempotent and gives the subscription a current-deploy dependency, but that needs the exact name the binding derived.
- Copying the algorithm into the stack drifts silently if the derivation ever changes; the subscription would then point at a Workflow that does not exist.

Today we ship a `bun patch` that re-exports `makeWorkflowName` from `Cloudflare/Workflows/index.ts`.

### Proposal

Export `makeWorkflowName` from `Cloudflare/Workflows` and drop `@internal`, so a stack can write:

```ts
const workflowName = Cloudflare.Workflows.makeWorkflowName(worker.workerName, "FileUrlIngestionWorkflow")
```

An alternative with the same effect: surface the derived `workflowName` on the Workflow binding result so stacks never derive it themselves.

PR with the export follows.

agcty:
Export `makeWorkflowName` from `Cloudflare/Workflows` and drop its `@internal` marker.

```ts
const workflowName = Cloudflare.Workflows.makeWorkflowName(worker.workerName, "FileUrlIngestionWorkflow")

yield* Cloudflare.Queues.Subscription("terminal-events-subscription", {
  source: { type: "workflows.workflow", workflowName },
  // ...
})
```

A stack that subscribes a Queue to a Workflow registered by an async Worker binding needs the Workflow's physical name in the same deploy pass, where `WorkflowResource.ref` cannot resolve yet. Deriving it through the same function the binding uses keeps both sides in sync.

Closes #1456

### DIFF
--- a/packages/alchemy/src/Cloudflare/Workflows/WorkflowName.ts
+++ b/packages/alchemy/src/Cloudflare/Workflows/WorkflowName.ts
@@ -8,7 +8,8 @@ import { sha256 } from "../../Util/sha256.ts";
  * exported class. The hash preserves uniqueness when the readable prefix must
  * be truncated to Cloudflare's 64-character limit.
  *
- * @internal
+ * Exported so a stack can reference a Workflow's physical name, for
+ * example to subscribe a Queue to it, without re-deriving it.
  */
 export const makeWorkflowName = (
   scriptName: Input<string>,
--- a/packages/alchemy/src/Cloudflare/Workflows/index.ts
+++ b/packages/alchemy/src/Cloudflare/Workflows/index.ts
@@ -1,2 +1,3 @@
 export * from "./Workflow.ts";
 export { makeWorkflowBridge } from "./WorkflowBridge.ts";
+export { makeWorkflowName } from "./WorkflowName.ts";


sam-goodwin:
I don't think we should expect to construct this:
```ts
const workflowName = Cloudflare.Workflows.makeWorkflowName(worker.workerName, "FileUrlIngestionWorkflow")
```
Instead, we should provide it as an output?
```ts
const worker = yield* Cloudflare.Worker("Worker", {
  env: {
     FILE_URL_INGESTION: Cloudflare.Workflow("FileUrlIngestion", { 
       className: "FileUrlIngestionWorkflow"
     })
  }
});

yield* Cloudflare.Queues.Subscription("Subscription", {
  source: { 
    type: "workflows.workflow", 
    workflowName: worker.env.FILE_URL_INGESTION.workflowName
  }
})
```

Or more concisely:
```ts
yield* Cloudflare.Queues.Subscription("Subscription", {
  sources: worker.env.FILE_URL_INGESTION
})
```

agcty:
Agreed, an output on the binding is the better shape, and it would remove a workaround on our side too. Here is what our stack does today and why:

```ts
const worker = yield* workflowsHost; // Worker with env.FILE_URL_INGESTION: Workflow("FileUrlIngestion", { className: "FileUrlIngestionWorkflow" })

// 1. derive the name, because the async binding's ref cannot resolve in this pass
const workflowName = Cloudflare.Workflows.makeWorkflowName(worker.workerName, "FileUrlIngestionWorkflow");

// 2. re-register the same WorkflowResource so the Queue subscription gets a
//    current-deploy dependency instead of a persisted-state ref
const workflow = yield* Cloudflare.Workflows.WorkflowResource("file-url-ingestion", {
  workflowName, className: "FileUrlIngestionWorkflow", scriptName: worker.workerName,
});

yield* Cloudflare.Queues.Subscription("terminal-events", {
  source: { type: "workflows.workflow", workflowName: workflow.workflowName },
  events: ["instance.completed", "instance.errored", "instance.terminated"],
  queueId: terminalEvents.queueId,
});
```

Both steps exist only because the physical name is not reachable from the binding at plan time. Your shape,

```ts
yield* Cloudflare.Queues.Subscription("Subscription", {
  source: { type: "workflows.workflow", workflowName: worker.env.FILE_URL_INGESTION.workflowName },
})
```

solves it as long as `workflowName` on the binding output is an `Input<string>` resolved in the same deploy pass (the `WorkflowResource` the binding registers already computes it from `worker.workerName` and the class name, so it is known before apply). The `sources: worker.env.FILE_URL_INGESTION` sugar on top is nice but the explicit `workflowName` output is the part we need.

One thing the output also fixes: today the Workflow's *first* deployment cannot be subscribed to in the same stack without the re-registration above, because `WorkflowResource.ref` reads persisted state. If the binding output is a current-pass value, that goes away.

I will close this PR and rework it as the output on `Worker.env.<binding>` (and drop the `@internal` marker only where the output needs it). Do you want the `sources:` sugar in the same PR or separately?

agcty:
Reworked as #1505 in the binding-output shape: the Workflow's physical name and script name are Outputs on worker.env.<binding>, resolved in the same deploy pass, so a same-stack Queue subscription works on the Workflow's first deployment; makeWorkflowName stays internal. The sources: sugar is left as a follow-up (the subscription's source is a persisted discriminated union that drives replacement).

Issue #522, opened 2026-06-02
Cloudflare Workflow names are not stage-scoped, causing cross-stage collisions and destructive deletes


V1RE:
Cloudflare Workflows currently appear to use the workflow class/logical name directly as the deployed Cloudflare workflow name.

In `packages/alchemy/src/Cloudflare/Workers/Workflow.ts`, the `name` passed to:

```ts
export default class MyWorkflow extends Cloudflare.Workflow<MyWorkflow>()(
  "MyWorkflow",
  ...
) {}
```

is used as:

```ts
yield* worker.bind`${name}`({
  bindings: [
    {
      type: "workflow",
      name,
      workflowName: name,
      className: name,
    },
  ],
});

yield* WorkflowResource(name, {
  workflowName: name,
  className: name,
  scriptName: worker.workerName,
});
```

The provider then upserts and deletes the Cloudflare Workflow by that same `workflowName`:

```ts
yield* putWorkflow({
  accountId: acct,
  workflowName: news.workflowName,
  className: news.className,
  scriptName: news.scriptName,
});

yield* deleteWorkflow({
  accountId: output.accountId,
  workflowName: output.workflowName,
});
```

Cloudflare's API path is account-scoped:

```txt
PUT    /accounts/{account_id}/workflows/{workflowName}
DELETE /accounts/{account_id}/workflows/{workflowName}
```

So two Alchemy stages/stacks using the same workflow class/logical name target the same Cloudflare Workflow resource.

## Why this is dangerous

If I deploy the same app to `prod` and `preview-123`, both stages define:

```ts
export default class OrderWorkflow extends Cloudflare.Workflow<OrderWorkflow>()(
  "OrderWorkflow",
  ...
) {}
```

Then both stages appear to upsert the same Cloudflare Workflow named `OrderWorkflow`.

The last deployed stage can update the shared Workflow to point at its own Worker script:

```ts
scriptName: worker.workerName
```

Then destroying any one stage runs:

```ts
deleteWorkflow({ workflowName: "OrderWorkflow" })
```

That deletes the shared Workflow backing every Worker bound to `OrderWorkflow`, including prod.

Result:

- prod Worker can remain deployed
- prod Worker binding can still reference `workflow_name = "OrderWorkflow"`
- backing Cloudflare Workflow may now point to another stage's Worker, or be deleted entirely
- `workflow.create(...)` can fail with workflow-not-found or run the wrong stage's implementation

## Comparison with other resources

[...]

PR #1460, opened 2026-09-02
feat(cloudflare/workflows): support explicit physical names

odysseaspapadimas:
Allow both Cloudflare Workflow forms to opt into an account-global physical name while keeping generated, stage-scoped names as the default.

```ts
Cloudflare.Workflow("MyWorkflow", {
  className: "MyWorkflow",
  workflowName: "my-existing-workflow",
});
```

Exact-name cold reads require explicit adoption, Worker bindings use the persisted physical name, and explicit renames replace without overwriting occupied targets.

Related to #522.


odysseaspapadimas:
Thanks!

PR #1476, opened 2026-09-03
docs(website): re-think the onboarding funnel

sam-goodwin:
Re-think the top of the docs funnel so a new reader is walked from the pitch to a deployed Worker to the concept pages, instead of from Getting started straight into the v1 migration guide.

### The path

```
What is Alchemy?  →  Getting started  →  Tutorial Part 2 … Part 6  →  Infrastructure as Effects
```

- **What is Alchemy?** is rewritten as a short narrative: one complete app (bucket + Worker) and the deploy command, then Runtime, Binding, and Layer each in a few lines with a snippet and a link to its page, then the IaC engine underneath. Providers' lifecycle operations, the type-error demo, and the two-styles comparison are gone; those live on their own pages.
- **Getting started** deploys exactly the Stack that tutorial Part 1 builds, so its pager and first card now go to Part 2. The cards lead with the tutorial, then Infrastructure as Effects, then the provider hubs and CLI.
- **Tutorial Part 1** tells Getting started graduates to skip ahead. **Part 6** now sends finishers to the Infrastructure as Effects overview first.
- **Migrating from v1** moves from third in Core to the end of Core, so the sidebar pager no longer lands on it after Getting started.

### DIFF
--- a/website/astro.config.mjs
+++ b/website/astro.config.mjs
@@ -456,7 +456,6 @@ export default defineConfig({
           items: [
             { label: "What is Alchemy?", link: "/what-is-alchemy" },
             { label: "Getting started", link: "/getting-started" },
-            { label: "Migrating from v1", link: "/migrating-from-v1" },
             {
               label: "Infrastructure as Code",
               items: [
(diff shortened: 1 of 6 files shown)


sam-goodwin:
Folded into #1472.

PR #1494, opened 2026-09-04
fix(aws/lambda): flush telemetry before a Lambda timeout

### LINKED ISSUE #1484: Flush telemetry and run finalizers when a Lambda is about to time out
If a Lambda function hits its timeout, you lose the trace for that invocation. The sandbox freezes and is terminated, taking the telemetry exporter's buffer with it, and the span never flushes. This is counterproductive: a timeout is precisely when you need the trace, yet it’s the one scenario where you receive nothing.

The handler already has access to the Lambda context through HandlerContext, and the runtime closes the request scope and flushes inline after the handler resolves. What’s missing is a deadline guard: compare the handler's execution time against the remaining time minus a margin. If it exceeds this limit, interrupt the fiber, close the request scope, and flush before Lambda forcefully terminates the sandbox. This way, the timeout appears as an errored span instead of disappearing.

You can already implement this in user-land using HandlerContext and Effect.timeout, but the key benefit is making it a default feature. This ensures that users are not caught off guard by this issue only after their invocations time out, leaving them without visibility into what occurred.

sam-goodwin:
Closes #1484

## The problem

The Lambda dispatcher gives every invocation a fresh request scope and closes it inline once the handler settles. That close is where the telemetry exporters flush:

```ts
const scope = Scope.makeUnsafe();
const exit = await eff.pipe(
  Effect.provide(
    Layer.mergeAll(
      Layer.succeed(HandlerContext, context),
      Layer.succeed(Scope.Scope, scope),
      Layer.effectContext(buildEventTelemetry(services, scope, ctx.telemetry)),
    ),
  ),
  Effect.runPromiseExit,
);
await Scope.close(scope, exit); // <- spans, logs and metrics ship here
```

When the handler outlives its timeout, Lambda freezes the sandbox mid-flight and `Scope.close` never runs. The exporter buffer and the still-open root span die with it, on exactly the invocation you most want a trace for.

## The fix: flush early, never interrupt

The deadline is known up front from `context.getRemainingTimeInMillis()`, and a sandbox only ever handles one invocation at a time, so a pre-deadline timer is race-free. `withInvocationDeadline` forks a watcher that parks for `remaining - margin` and, if the handler is still running by then, ends the root span, logs, and drains the exporters. The handler is left alone:

```ts
const watcher = yield* Effect.forkChild(
  Effect.sleep(Duration.millis(budgetMs)).pipe(
    Effect.andThen(Effect.uninterruptible(flushBeforeDeadline)),
  ),
);
return yield* self.pipe(Effect.ensuring(Fiber.interrupt(watcher)));
```

```ts
const flushBeforeDeadline = Effect.gen(function* () {
  const error = new InvocationTimeoutError({ functionName, requestId, budgetMs, marginMs });
  const span = yield* Effect.option(Effect.currentSpan); // the http.server root span
  if (Option.isSome(span) && span.value.status._tag === "Started") {
    span.value.attribute("aws.lambda.timeout.imminent", true);
    span.value.end(yield* Clock.currentTimeNanos, Exit.fail(error));
  }
  yield* Effect.logWarning(error.message);            // carries the trace id
  const flusher = yield* Effect.serviceOption(Flusher); // effect's OTLP exporter registry
  if (Option.isSome(flusher)) yield* flusher.value.flush;
});
```

If Lambda then freezes the sandbox, the trace already exists with an errored root span, the child spans that had ended, and every log record. If the handler finishes inside the margin, its response goes out exactly as before, and the HTTP tracer middleware ends the span a second time with the real outcome. That earlier export is the accepted false positive; the `aws.lambda.timeout.imminent` attribute stays on the span so a backend can tell the two apart. This is the same model dd-trace uses (it finishes the invocation span as a timeout and flushes, without cancelling the handler). Nothing about the invocation's outcome changes: no interruption, no synthetic failure, no change to Lambda's `Errors` metric.

The watcher inherits the invocation fiber's context, so it sees the invocation's logger, the exporters and, on the HTTP path, the root span. The flush is uninterruptible, so a handler that finishes mid-flush waits for the export instead of abandoning an in-flight batch. Spans still open at the flush other than the root are not exported, which is what it means to flush a running invocation.

## Why the HTTP listener guards twice

The dispatcher guards every listener effect, but the `http.server` span is created further in, by `HttpMiddleware.tracer`. To be able to end that span, the HTTP listener applies the guard inside the tracer:

```ts
const safeHandler = HttpMiddleware.tracer(
  withInvocationDeadline(Http.safeHttpEffect(handler)),
);
```

Two watchers on one invocation would flush twice, so nested guards hand the deadline down: the guard provides an internal `InvocationDeadline` service to its inner effect, an inner guard claims it, and the outer watcher does nothing when its timer fires:

```ts
const outer = yield* Effect.serviceOption(InvocationDeadline);
if (Option.isSome(outer)) outer.value.claim();
let claimed = false;
// ...
Effect.sleep(budget).pipe(
  Effect.andThen(Effect.suspend(() => (claimed ? Effect.void : flushBeforeDeadline))),
)
```

Queue, stream and other non-HTTP listeners only have the dispatcher's guard; without a span in context it logs and flushes, which still ships everything that had ended.

## Configuring the margin

The margin is a Function prop next to `timeout`, written to `ALCHEMY_LAMBDA_TIMEOUT_MARGIN_MS` at deploy time and read per invocation. Size it for one export round-trip to your telemetry backend, or set it to zero to turn the flush off:

```ts
export default class Api extends AWS.Lambda.Function<Api>()(
  "Api",
  {
    main: import.meta.url,
    timeout: Duration.seconds(30),
    timeoutMargin: Duration.seconds(1), // default 500 millis
  },
  Effect.gen(function* () {
    return { fetch: /* ... */ };
  }),
) {}
```

A margin that leaves no budget logs a warning and runs without the watcher rather than flushing at t=0 on every invocation. `HandlerContext` moved into the new `InvocationDeadline.ts` module unchanged and is still exported from `AWS.Lambda`.

## Tests

`InvocationDeadline.test.ts` covers the guard under the TestClock with a counting `Flusher`. The central case pins that the root span was ended with the timeout at the margin, that the handler still completed with its value afterwards, and that only one flush happened:

```ts
yield* TestClock.adjust("100 millis"); // 5000 remaining - 4900 margin
expect(counter.flushes).toBe(1);
expect(timeoutFailure(span.status.exit)).toBeInstanceOf(InvocationTimeoutError);
expect(span.attributes.get(TIMEOUT_IMMINENT_ATTRIBUTE)).toBe(true);

yield* TestClock.adjust("10 seconds");
expect(yield* Fiber.join(fiber)).toEqual(Exit.succeed("done late"));
expect(counter.flushes).toBe(1);
```

`InvocationDeadline.dispatch.test.ts` drives the real dispatcher export in-process with a fake invocation context, a recording tracer and a counting `Flusher`. A 2.5 s handler with 2 s remaining returns its real 200 after 2.5 s, the flush fired at ~1.5 s, and the `http.server` span was ended twice, first with the timeout as a `Fail` and then with the real success:

```ts
expect(statusOf(result)).toBe(200);
expect(flushes[0]).toBeGreaterThanOrEqual(1_400);
const [early, final] = ends.filter((e) => e.name.startsWith("http.server"));
expect(failureOf(early.exit)).toBeInstanceOf(InvocationTimeoutError);
expect(Exit.isSuccess(final.exit)).toBe(true);
```

`Telemetry.test.ts` (live) now deploys the fixture with `timeout: Duration.seconds(5)` and a `/slow` route that ends a child span, logs, then sleeps 60 s. Lambda kills it; the test asserts the collector still received `AWS.Lambda.InvocationTimeoutError`, `aws.lambda.timeout.imminent`, the child span and its log. I could not run the live test from this session because the AWS SSO token had expired; it needs a CI run.

## Docs

`aws/compute/lambda.mdx` gains a "Timeouts still export their trace" section, and the telemetry page's Lambda flush note mentions the early flush.

### DIFF
--- a/packages/alchemy/src/AWS/Lambda/DurableBridge.ts
+++ b/packages/alchemy/src/AWS/Lambda/DurableBridge.ts
@@ -40,7 +40,7 @@ import {
   type DurableRetryPolicy,
   type DurableStepOptions,
 } from "./Durable.ts";
-import { HandlerContext } from "./Function.ts";
+import { HandlerContext } from "./InvocationDeadline.ts";
 
 /** Module id of AWS's Durable Execution SDK (an optional peer dependency). */
 export const DURABLE_SDK_MODULE = "@aws/durable-execution-sdk-js";
--- a/packages/alchemy/src/AWS/Lambda/DurableFunction.ts
+++ b/packages/alchemy/src/AWS/Lambda/DurableFunction.ts
@@ -24,8 +24,8 @@ import {
   Function,
   type FunctionProps,
   type FunctionServices,
-  type HandlerContext,
 } from "./Function.ts";
+import type { HandlerContext } from "./InvocationDeadline.ts";
 
 type TypeId = "AWS.Lambda.DurableFunction";
 const TypeId = "AWS.Lambda.DurableFunction" as const;
--- a/packages/alchemy/src/AWS/Lambda/Function.ts
+++ b/packages/alchemy/src/AWS/Lambda/Function.ts
@@ -60,6 +60,12 @@ import {
   makeFunctionImage,
 } from "./FunctionImage.ts";
 import { makeFunctionHttpHandler } from "./HttpServer.ts";
+import {
+  HandlerContext,
+  TIMEOUT_MARGIN_ENV,
+  toTimeoutMarginMillis,
+  withInvocationDeadline,
+} from "./InvocationDeadline.ts";
 
 export type { FunctionImageSource } from "./FunctionImage.ts";
 
@@ -74,11 +80,6 @@ class FunctionUpdateFailed extends Data.TaggedError("FunctionUpdateFailed")<{
   functionName: string;
   reason?: string;
 }> {}
-
-export class HandlerContext extends Context.Service<
-  HandlerContext,
-  lambda.Context
->()("AWS.Lambda.HandlerContext") {}
 
 export const isFunction = (value: any): value is Function => {
   return (
@@ -266,6 +267,26 @@ export interface FunctionCommonProps extends PlatformProps {
    * @default 3 seconds (AWS Lambda default)
    */
   timeout?: Duration.Duration;
+  /**
+   * How long before {@link timeout} telemetry is flushed for an invocation
+   * that is still running. A Lambda that hits its timeout is frozen
+   * mid-flight and its buffered spans and logs are lost; at
+   * `timeout - timeoutMargin` the runtime ends the invocation's root span
+   * with an `AWS.Lambda.InvocationTimeoutError`, logs a warning and drains
+   * the exporters, so the trace of the slow invocation is exported instead.
+   *
+   * The handler is never interrupted and the invocation's outcome is never
+   * changed. If it does finish inside the margin, its response goes out
+   * as normal and the span is re-exported with the real outcome; the
+   * earlier export (marked `aws.lambda.timeout.imminent`) is the accepted
+   * false positive.
+   *
+   * Size it for one export round-trip to your telemetry backend. Set to
+   * `Duration.zero` to disable.
+   *
+   * @default 500 millis
+   */
+  timeoutMargin?: Duration.Duration;
   /**
    * Maximum number of concurrent executions reserved for this function.
    * Omit to remove the function-level reserved concurrency limit.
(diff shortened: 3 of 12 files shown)


anatolzak:
wanted to flag one thing about which spans actually survive the flush.

Very often a Lambda times out because some downstream call hung. In that case the span of the hanging request is still open when the deadline fires. Since the flush only ends the root and doesn't export open spans, that span never ships. So the trace shows the errored http.server root with timeout.imminent on it, but the span that actually explains the timeout isn't there.

In contrast, dd-trace handles this differently. dd-trace finishes every unfinished span so the hanging call does get exported. So on a timeout that's usually the most useful thing in the trace.

I am assuming the reason we only close the root span is that there is no easy way to end all open spans in Effect. It's also worth pointing out this is a side effect of not interrupting. In my case, I just use Effect.timeout() which does interrupt and close all the open spans. But I realize this may not be a good default to actually interrupt a user's handler while it could have still succeeded.

Curious what you think the best move is. Personally I'd take interruption on an imminent timeout and I'd rather have the trace of what hung. But that's just me, and I'm not sure it's a safe default to impose on everyone. At the same time I don't love making it opt-in. The whole value is that it just works without anyone having to find out the hard way that their timeout traces are empty and/or don't explain what caused the timeout.

PR #1499, opened 2026-09-04
feat(cloudflare): let a class-form Durable Object keep a distinct class name

agcty:
## What

The class form `Cloudflare.DurableObject<Self, Shape>()("Name", props)` accepts `className` in its props, so the physical class can keep a name distinct from the logical id (the binding's `env` key), and the physical name flows through every place it is read.

## Why

The class form bound the logical id and the class name to the same string: `("Name")` registered a binding with `className: "Name"`, exported the class under `"Name"` and read the namespace map by `"Name"`. A Worker whose namespaces were created by the async form (`env.LENS_DO = Cloudflare.DurableObject("lens", { className: "LensServer" })`) could therefore not convert to the class form in place: the migration engine keys a class on `(logical id, class name)`, so the conversion was a delete plus a create, and a namespace with stored objects would be lost.

## How

- `DurableObject.ts`: `className` in the class-form props; the local and cross-script bindings carry it; the namespace accessor reads `durableObjectNamespaces[className]` (the provider keys that map by class name); the export is registered under the class name (the generated entry emits `export class <key>` and the bridge resolves its export by that key). A JSDoc section shows the conversion.
- `WorkerProvider.ts`: the class-identity half of the migration derivation moves into a pure `planDurableObjectClassIdentity` (kept / renamed / unresolved / delete candidates), called from `put`; the transfer resolution and the observed-namespace validation stay in the provider. The precreate placeholder's export-derived entries are documented as keyed by class name (the binding-derived entry, which carries the logical id, wins the merge as before).
- Tests: `WorkerProvider.test.ts` gains the identity matrix (same logical id and class name: no migration; changed class name: rename; fresh stage: one class to create; vanished logical id: delete candidate; untagged worker: adoption by binding name). `DurableObjectClassName.test.ts`: a cross-script `.from()` binding carries the physical class and resolves its namespace id by it.

## Checks

- `pnpm test test/Cloudflare/Workers/WorkerProvider.test.ts test/Cloudflare/Workers/DurableObjectClassName.test.ts`: 46 passed (bun 1.3.13).
- `tsc -b packages/alchemy` and `tsc -p packages/alchemy/tsconfig.test.json --noEmit` clean (apart from the pre-existing `@alchemy.run/floci` resolution errors on a checkout without the floci submodule).
- `oxfmt` clean.
- Not run: a cloud deploy that converts an async-form host in place; the identity matrix is proven on the pure planner.

### DIFF
--- a/packages/alchemy/test/Cloudflare/Workers/DurableObjectClassName.test.ts
+++ b/packages/alchemy/test/Cloudflare/Workers/DurableObjectClassName.test.ts
@@ -1,0 +1,80 @@
+import * as Cloudflare from "@/Cloudflare";
+import {
+  WorkerEnvironment,
+  WorkerTypeId,
+} from "@/Cloudflare/Workers/Worker.ts";
+import * as Output from "@/Output";
+import { Self } from "@/Self.ts";
+import { describe, expect, it } from "alchemy-test";
+import * as Effect from "effect/Effect";
+
+/**
+ * A class-form Durable Object whose physical class keeps a name distinct from
+ * its logical id, the shape a host converted from the async form declares to
+ * keep its namespace in place.
+ */
+interface LensShape {
+  readonly ping: () => Effect.Effect<string>;
+}
+
+class LensDo extends Cloudflare.DurableObject<LensDo, LensShape>()("LENS_DO", {
+  className: "OddlynewLensServer",
+}) {}
+
+describe("class-form className", () => {
+  it.effect(
+    "a cross-script binding carries the physical class and resolves its namespace id by it",
+    () =>
+      Effect.gen(function* () {
+        const bound: unknown[] = [];
+        const worker = {
+          bind:
+            (strings: TemplateStringsArray, ...args: unknown[]) =>
+            (data: unknown) =>
+              Effect.sync(() => {
+                bound.push({ sid: String.raw(strings, ...args), data });
+              }),
+          durableObjectNamespaces: Output.literal({
+            OddlynewLensServer: "ns-lens",
+          }),
+        };
+
+        // `yield* Worker` resolves the `Self` tag of the Worker type at
+        // runtime; the `Worker` requirement is a type-level marker.
+        const handle = yield* (
+          LensDo.from("host-script") as Effect.Effect<
+            Cloudflare.DurableObject<LensDo>
+          >
+        ).pipe(
+          Effect.provideService(Self(WorkerTypeId), worker as never),
+          Effect.provideService(WorkerEnvironment, undefined as never),
+        );
+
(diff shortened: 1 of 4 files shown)


agcty:
@sam-goodwin a data point on why this one matters to us, in case it helps with prioritising.

We ran our monorepo's test suites against a build of current `main` (with `effect@4.0.0-rc.115`), once without and once with this PR applied, nothing else changed:

| Suite | Failing without this PR | Failing with it |
| --- | --- | --- |
| documents | 272 | 0 |
| Durable Object substrate | 6 | 2 |
| MCP server | 8, plus 7 files that do not load | 4 |
| unrelated resources (control) | 6 | 6 |

All 284 of those failures share one cause: a class-form Durable Object whose physical class name differs from its `env` key. The class form uses the binding name as the class name in two places, the binding's `className` and the `worker.export(...)` registration, so the class is registered under the `env` key, the runtime bridge looks it up by the physical name, finds nothing and destructures `undefined`. This PR changes exactly those two places. The failures that remain with it applied are ours and unrelated. The unchanged control suite shows the local build did not disturb anything else.

So for us this PR is the one thing standing between the next release and being able to take it. The branch is rebased onto current `main` as of today (`e6f6987d7`). The earlier red check was the website docs build failing on an AWS RDS page this PR does not touch, which has since been fixed on `main`; CI is running again now.

Happy to reshape it, split it, or drop the `WorkerProvider` refactor if you would rather take only the two-line `DurableObject.ts` change.

packages/alchemy/test/Cloudflare/Workers/DurableObjectClassName.test.ts
sam-goodwin: This doesn't follow our testing practices. Do no mock anything. Move this to DurableObject.test.ts or Worker.test.ts and make an actual fixture for it.

sam-goodwin (changes requested):

packages/alchemy/test/Cloudflare/Workers/DurableObjectClassName.test.ts
agcty: Thanks, done. The mocked test is gone. `Worker.test.ts` now has a live case, "class-form Durable Object with a distinct className converts in place from the async form", with a real fixture in `fixtures/do-class-name/`. It deploys a host in the async form (`DurableObject("LENS_DO", { className: "LensServer" })`) and stores a value, then converts it to the class form with the same `className`. It asserts the worker updates in place with the same namespace id, that the host binding and a second worker's `.from()` cross-script binding both carry `LensServer`, and that both workers read the stored value back through the object. It passes against a real account (1 passed, 54s). I also rebased onto main.

PR #1509, opened 2026-09-05
feat(kubernetes): add Secret resource with Redacted data

connyay:
- `Kubernetes.Secret` applies a v1/Secret via server-side apply; values stay `Redacted` through plan and state and are base64-encoded only at the API edge
- hoist the shared `isNotFound` helper into `internal/client.ts`
- cover the Secret lifecycle in the gated Deployment E2E suite

### DIFF
--- a/packages/alchemy/src/Kubernetes/Connection.ts
+++ b/packages/alchemy/src/Kubernetes/Connection.ts
@@ -2,12 +2,11 @@
  * The cluster-agnostic Kubernetes connection model.
  *
  * Every `Kubernetes.*` workload (`Deployment`, `Job`, `Manifest`,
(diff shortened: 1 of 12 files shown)


isakgustavsen:
Could this be integrated with pr #1502? It should cover what you need in terms of importing secrets, and if there's anything missing it we could look at it now

connyay:
> Could this be integrated with pr #1502? It should cover what you need in terms of importing secrets, and if there's anything missing it we could look at it now

I don't fully grok what you mean by integrating these PRs. They are related, but not directly overlapping. I think the best would be maybe some doc/example improvements showing how they could be used together. Could be added to either PR after the other one lands.

Did you have something specific in mind?

PR #1516, opened 2026-09-07
feat(pkg): preview package registry verified against GitHub Actions runs

BlankParticle:
Replace `@alchemy.run/pr-package` and the `alchemy-run/actions` publish scripts with `@alchemy.run/pkg`: a Cloudflare Worker registry plus a `pkg` CLI that packs and publishes workspace packages from the job that built them. Every publication is verified against the GitHub Actions run that produced it, so fork pull requests publish with no permissions and no secrets.

```ts
// stacks/pkg.ts
import { PkgRegistry } from "@alchemy.run/pkg/Registry";

const registry = yield* PkgRegistry("Registry", {
  worker: {
    domain: {
      name: "pkg.alchemy.run",
      aliases: ["pkg.distilled.cloud", "📦.alchemy.run", "📦.distilled.cloud"],
    },
  },
  github: {
    appId: Config.String("GH_APP_ID"),
    privateKey: Config.Redacted("GH_APP_PRIVATE_KEY"),
  },
  policy: {
    repos: ["alchemy-run/alchemy", "alchemy-run/distilled"],
    ttl: Duration.weeks(1),
    maxPackageSize: ByteSize.megabytes(100),
  },
});
```

`PkgRegistry` declares `Registry/Worker`, `Registry/Bucket`, `Registry/Database`, and `Registry/Cache` under the namespace it is given. The package is the Worker's bundle entry: the options are encoded into a `PKG_SETTINGS` json binding at deploy time and decoded back inside the isolate through one `RegistryConfig` service, so there is no separate entry file to write.

```sh
pnpm install https://pkg.alchemy.run/alchemy/pr:1516
pnpm install https://pkg.alchemy.run/alchemy/branch:main
pnpm install https://pkg.alchemy.run/alchemy/163c051
```

### Proof of origin

One workflow, `pkg.yml`, runs on `push` and `pull_request` with `contents: read` and nothing else. `pkg pack` packs and prints an artifact name, `pkg-manifest-<sha256 of the manifest>`; `actions/upload-artifact` uploads the manifest under that name; `pkg publish` sends it. Only a job's runtime token can add artifacts to its run, and the App reads the run's artifact list back through the API, so the artifact is GitHub's record that this run vouched for exactly these package hashes.

Requests name the run (`{ repo, runId, attempt }` in the publish body, in the query for uploads) as a lookup hint and carry no credential. The Worker fetches the run through the App, requires it to be in progress, and refuses any manifest whose hash is not vouched for. Every tag comes from the resolved run, never from the manifest: full SHA, short SHA, `branch:<name>`, and `pr:<number>` for the repository's own runs; a fork's run gets only `pr:<number>`, since commit tags are shared by every publisher and a fork can run any commit, including one the repository already published. The pull request is matched by head repository as well as head commit, and the manifest's `head` must be the run's head, so a job that packs the merge commit is refused. Naming another run can only get that run's own manifest accepted. Uploads are verified by R2 against the hash in the URL, so a run cannot pre-plant bytes under a hash another build will produce, and package and group names are constrained to npm's name rules and plain labels before they reach R2 keys or markdown.

### Content-addressed dependencies

`pkg pack` packs in dependency order and rewrites every `workspace:*` dependency on another packed package to that package's immutable tarball URL, `https://<registry>/<name>/-/<sha256>.tgz`. A tarball's bytes depend only on its source and its dependencies' bytes, so identical builds deduplicate across commits, pull requests, and repositories, and a consumer installing several preview packages gets one copy of each shared dependency.

### Registry

- The HTTP contract is an Effect `HttpApi` in `Protocol.ts` (`POST /api/publish`, `PUT /api/tarballs/:name/:sha256`, `GET /api/health`) with `Schema.TaggedError` errors carrying their status. The Worker serves it with `HttpApiBuilder` groups plus an `HttpRouter` catch-all for install URLs; the CLI calls it through `HttpApiClient`, so `MissingTarballs` is a typed catch rather than a status check.
- Storage and GitHub access go through the requirement channel: `Cloudflare.R2.ReadWriteBucket`, `SQL.D1Layer`, `Cloudflare.KV.ReadWriteNamespace`, and a `GitHubApp` service over `@distilled.cloud/github`. The App JWT is signed with `node:crypto`. Installation tokens and resolved runs live in a KV namespace shared by every isolate, through a lookup-through cache in which concurrent misses each run the lookup. It is deliberately not Effect's `Cache`: that one makes concurrent misses join the first caller's fiber, and workerd drops a fiber resumed from another request's context once that request ends, which showed up as uploads hanging with "Worker's code had hung" whenever `pkg publish` fired its four concurrent uploads at a fresh isolate.
- One D1 table, `tags(package, tag, sha256, expires_at, linked_prs)`, with tarballs in R2. Tag rows are merged in one `ON CONFLICT` statement so concurrent publications of a shared tag keep every linked pull request. An hourly sweep extends rows while any linked pull request is open, pins expiry to close time plus `ttl` once they close, keeps rows it could not check alive until the next sweep, then deletes expired rows, unreferenced tarballs, and uploads that never got tagged.
- Publish is idempotent: it either answers `409 MissingTarballs` or writes tags, posts a "Preview packages" check run on the commit, and updates the PR comment with grouped `pnpm install` lines plus publish and expiry times as `<relative-time>`. Groups are `alchemy` for the `alchemy` package, then `@alchemy.run` and `@distilled.cloud`, which are given as `NAME[Collapsed]=<glob>` and render collapsed.
- `--registry` falls back to the `PKG_REGISTRY` env var, then `https://pkg.alchemy.run`.
- The App needs `checks: write`, `pull_requests: write`, and `actions: read`.

### Consumers

- pnpm 11+ needs `blockExoticSubdeps: false` in `pnpm-workspace.yaml`, since the tarballs link dependencies by URL.
- bun needs 1.4 or newer and the `bun add alchemy@<url>` form.

### Also

- `stacks/github.ts` moves to `Cloudflare.ApiToken.AccountApiToken`, drops the Neon provider, and publishes the secrets the workflows actually read: `AWS_ROLE_ARN` becomes a secret instead of a variable, `ANTHROPIC_API_KEY`, `DISCORD_WEBHOOK_URL`, and the unprefixed Cloudflare pair for `cloudflare-tools.yml` are added, `DOPPLER_TOKEN`, `PR_PACKAGE_TOKEN`, and the `AWS_REGION` variable are dropped.
- `packages/pr-package`, `examples/cloudflare-pr-package`, the old stack, and `pr-package.yml` are removed; `check.yml` runs the `pkg` tests instead and `build:pr-packages` becomes `build:pkg`.
- `@distilled.cloud/github` joins the root `tsconfig.json` references; `.pkg` is gitignored.


sam-goodwin:
Can we also rename "@alchemy.run/pr-package" to "@alchemy.run/pkg"?

Issue #563, opened 2026-06-06
Planetscale.PostgresRole replaces on every deploy when inheritedRoles order differs


Gerbuuun:
`Planetscale.PostgresRole` appears to compare `inheritedRoles` order-sensitively, which can cause a no-op deploy to plan a role replacement.

### Repro

Configure a Postgres role with:

```ts
yield* Planetscale.PostgresRole("Role", {
  database,
  branch,
  inheritedRoles: ["pg_write_all_data", "pg_read_all_data"],
});
```

After deploy, PlanetScale/Alchemy state can contain the same roles in a different order:

```ts
["pg_read_all_data", "pg_write_all_data"]
```

A subsequent plan then reports:

```text
[.../Planetscale/Role] replace
```

### Expected

If `inheritedRoles` is semantically set-like, role order should not affect diffing. A no-op deploy should not replace/rotate the role when the desired and observed role sets are identical.

### Actual

The provider compares arrays order-sensitively:

```ts
const newRoles = resolveInheritedRoles(news.inheritedRoles);
const oldRoles = output?.inheritedRoles ?? [];
if (!deepEqual(newRoles, oldRoles)) {
  return { action: "replace" };
}
```

So a different returned order causes a replacement.

### Suggested Fix

Normalize both arrays before comparing:

```ts
const newRoles = [...resolveInheritedRoles(news.inheritedRoles)].sort();
const oldRoles = [...(output?.inheritedRoles ?? [])].sort();

if (!deepEqual(newRoles, oldRoles)) {
  return { action: "replace" };
}
```

### Workaround

Ordering `inheritedRoles` to match PlanetScale's returned order avoids the repeated replacement:

```ts
inheritedRoles: ["pg_read_all_data", "pg_write_all_data"]
```

But that depends on API return ordering and should not be necessary.

PR #1517, opened 2026-09-07
chore: bump distilled to 78a9304 (verbNoun operation names, Railway selection fixes)

Mkassabov:
Bumps `submodules/distilled` to `78a9304` (current distilled `main`) and fixes the type errors that surfaced.

## What changed in distilled

- **Every operation is now named verbNoun** (alchemy-run/distilled#545, #548, #549, #563). GraphQL-derived names like `serviceCreate` / `volumeDelete` and REST-derived ones like `infoIndex` are gone; the generated SDKs expose `createService`, `deleteVolume`, `getIndexInfo`, `queryIndex`, etc. Request/response/error type aliases follow (`ServiceCreateResponse` → `CreateServiceResponse`).
- **Railway** (alchemy-run/distilled#562): drops the admin-only `platformFeatureFlags` and the upstream-removed `supportTierOverride` from generated selections, so CLI-login tokens can resolve workspaces and workspace queries no longer 400. This is the same fix `main` picked up from the `alchemy-pointer` branch, now on distilled `main` proper.
- 49 new SDK packages landed on distilled `main` (PayPal, Plaid, Auth0, Clerk, Datadog, Docker, Slack, …). They join the pnpm workspace via `submodules/distilled/packages/*`, hence the lockfile growth. Nothing in alchemy depends on them yet.

## Fixes in this PR

| Area | Change |
|---|---|
| `src/Railway/**`, `test/Railway/**` | Mechanical rename of 70 operations and response types to their verbNoun form (e.g. `railway.serviceCreate` → `railway.createService`, `volumeInstanceBackupRestore` → `restoreVolumeInstanceBackup`). Behaviour unchanged. |
| `src/Cloudflare/Vectorize/SearchIndexHttpClient.ts` | `infoIndex` → `getIndexInfo`, `queryIndex` stays `queryIndex` (+ response types). |
| `src/Cloudflare/Snippets/SnippetRules.ts` (+ test) | `snippets.listRules` stays `listRules`. Cloudflare's docs publish this route twice; distilled#563 collapses the twins into a single `listRules` carrying the typed `SnippetRulesNotFound`/`Forbidden` errors the provider catches. |
| `src/AWS/AMP/Scraper.ts` | `destination.ampConfiguration` is optional in the regenerated model; use `?.`. |

## Review follow-ups

@sam-goodwin's two naming comments (`listInfo`, `getRule`) were generator bugs, fixed upstream in alchemy-run/distilled#563 and folded into this PR after rebasing onto `main`. #1537 was the stacked version of that fix and is superseded.

## Verification

- `bun tsc -b` — clean (was 399 errors before the fixes).
- `bun lint:providers` — all provider factories resolve.
- `oxfmt` — applied.
- No live tests were run; all changes are renames or an added optional-chain.

### DIFF
--- a/packages/alchemy/src/Cloudflare/Snippets/SnippetRules.ts
+++ b/packages/alchemy/src/Cloudflare/Snippets/SnippetRules.ts
@@ -224,7 +224,7 @@ interface WireRule {
 }
 
 const listObservedRules = (zoneId: string) =>
-  snippets.getRule({ zoneId }).pipe(
+  snippets.listRules({ zoneId }).pipe(
     // A zone that has never had a snippet-rule list 404s — there is simply
     // no rule list, which is equivalent to an empty one.
     Effect.catchTag("SnippetRulesNotFound", () => Effect.succeed([])),
--- a/packages/alchemy/src/Cloudflare/Snippets/SnippetRules.ts
+++ b/packages/alchemy/src/Cloudflare/Snippets/SnippetRules.ts
@@ -224,7 +224,7 @@ interface WireRule {
 }
 
 const listObservedRules = (zoneId: string) =>
-  snippets.listRules({ zoneId }).pipe(
+  snippets.getRule({ zoneId }).pipe(
     // A zone that has never had a snippet-rule list 404s — there is simply
     // no rule list, which is equivalent to an empty one.
     Effect.catchTag("SnippetRulesNotFound", () => Effect.succeed([])),
--- a/packages/alchemy/src/Cloudflare/Vectorize/SearchIndexHttpClient.ts
+++ b/packages/alchemy/src/Cloudflare/Vectorize/SearchIndexHttpClient.ts
@@ -61,13 +61,13 @@ export const makeHttpSearchIndexClient = (
     describe: () =>
       local((name) =>
         vectorize
-          .listInfo({ accountId, indexName: name })
+          .getIndexInfo({ accountId, indexName: name })
           .pipe(Effect.map(toIndexInfo)),
       ),
     query: (vector, options) =>
       local((name) =>
         vectorize
-          .listQuery({
+          .queryIndex({
             accountId,
             indexName: name,
             vector: Array.from(vector),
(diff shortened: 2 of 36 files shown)


packages/alchemy/src/Cloudflare/Snippets/SnippetRules.ts
```diff
@@ -224,7 +224,7 @@ interface WireRule {
 }
 
 const listObservedRules = (zoneId: string) =>
-  snippets.listRules({ zoneId }).pipe(
+  snippets.getRule({ zoneId }).pipe(
```
sam-goodwin: list changed to get and singular?

packages/alchemy/src/Cloudflare/Vectorize/SearchIndexHttpClient.ts
```diff
@@ -61,13 +61,13 @@ export const makeHttpSearchIndexClient = (
     describe: () =>
       local((name) =>
         vectorize
-          .infoIndex({ accountId, indexName: name })
+          .listInfo({ accountId, indexName: name })
```
sam-goodwin: Should this be `listIndexInfo` ? Why was `Index` dropped? `Index` is the actual entity noun here

packages/alchemy/src/Cloudflare/Snippets/SnippetRules.ts
Mkassabov: You are right — `listRules` is the correct name. The rename happened because Cloudflare's docs publish this route twice (`RulesGet` and `RulesList`); the old distilled patch aliased `listRules` → `getRule`, and after the verbNoun change the alias stopped applying so only `getRule` carried the typed `SnippetRulesNotFound`. Fixed upstream in alchemy-run/distilled#563 (single `listRules`, errors attached) and adopted in #1537, stacked on this PR.

packages/alchemy/src/Cloudflare/Vectorize/SearchIndexHttpClient.ts
Mkassabov: Agreed. The generator's Rails `index → list` alias was firing on the trailing token of `InfoIndex`. alchemy-run/distilled#563 makes `index` a noun inside a camel id and names this op `getIndexInfo` (with `queryIndex` for the sibling); adopted in #1537, stacked on this PR.

PR #1521, opened 2026-09-07
fix(cloudflare): forward container memoryMib

danieljvdm:
Forward `memoryMib` to Cloudflare and local container attributes, and suppress the implicit `lite` instance type when a memory override is supplied.

```ts
Cloudflare.Container("App", {
  image: "mendhak/http-https-echo:41",
  vcpu: 1,
  memoryMib: 4096,
});
```

Custom sizing requires at least 1 vCPU and 3 GiB of memory per vCPU for the first four vCPUs.

### DIFF
--- a/packages/alchemy/test/Cloudflare/Container/MemoryMib.test.ts
+++ b/packages/alchemy/test/Cloudflare/Container/MemoryMib.test.ts
@@ -1,0 +1,151 @@
+import { AlchemyContext } from "@/AlchemyContext.ts";
+import { ArtifactStore, createArtifactStore } from "@/Artifacts.ts";
+import { CloudflareEnvironment } from "@/Cloudflare/CloudflareEnvironment.ts";
+import { ContainerPlatform } from "@/Cloudflare/Containers/ContainerPlatform.ts";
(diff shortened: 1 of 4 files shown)


packages/alchemy/test/Cloudflare/Container/MemoryMib.test.ts
sam-goodwin, line 1: Please remove this test and add one in Container.test.ts that actualy tests end to end. Don't mock things

packages/alchemy/test/Cloudflare/Container/MemoryMib.test.ts
```diff
@@ -0,0 +1,151 @@
...
+    },
+  );
+
+  it.effect("preserves memoryMib in local application attributes", () =>
+    Effect.gen(function* () {
+      const provider = yield* ContainerPlatform.Provider;
+      const output = yield* provider.reconcile({
+        ...baseInput,
+        news: { name: "memory-test", image, memoryMib: 6144 },
+      });
```
sam-goodwin, line 142: Don't call reconcilers like this. Follow the already established patterns in the repo

Issue #561, opened 2026-06-06
CLI main imports Cloudflare/Rolldown and crashes Bun for non-Cloudflare stacks


bjorntechTobbe:
Using alchemy@2.0.0-beta.51 on macOS arm64, importing alchemy/Cli/main crashes Bun while loading the Rolldown native binding at node_modules/@distilled.cloud/cloudflare-rolldown-plugin/node_modules/rolldown/node_modules/@rolldown/binding-darwin-arm64/rolldown-binding.darwin-arm64.node. This reproduced with Bun 1.2.20 and 1.3.13. Import isolation: importing alchemy succeeds, importing @finnvid/alchemy-scaleway succeeds, and importing alchemy/Cli/commands/deploy succeeds. Importing alchemy/Cli/main crashes, and importing alchemy/Cloudflare crashes. Impact: bun alchemy deploy crashes before evaluating a Scaleway-only stack because the CLI main module eagerly imports Cloudflare command/provider code. Bypassing the command registry and calling alchemy/Cli/commands/deploy directly allows the Scaleway-only stack to plan and deploy. Expected: non-Cloudflare deploys should not import Cloudflare/Rolldown paths, or Cloudflare command imports should be lazy until the Cloudflare command/provider is actually used.

PR #1535, opened 2026-09-08
feat(digitalocean): Droplet, SshKey, and Cloud Firewall resources

rickyeescobar:
DigitalOcean joins the provider roster: `Droplet`, `SshKey`, and `Firewall` resources, an auth provider for `alchemy login` / `alchemy profile show`, hub and setup docs, and live test suites for all three.

```ts
const key = yield* DigitalOcean.SshKey("deploy-key", { publicKey });
const host = yield* DigitalOcean.Droplet("app", {
  region: "sfo3",
  size: "s-2vcpu-4gb",
  image: "docker-24-04",
  sshKeys: [key.fingerprint],
  replaceAfter: "30 days",
});
yield* DigitalOcean.Firewall("edge", {
  dropletIds: [host.dropletId],
  inboundRules: [{ protocol: "tcp", ports: "22", addresses: ["0.0.0.0/0"] }],
});
```

Built on `@distilled.cloud/digitalocean` from upstream distilled (alchemy-run/distilled#487), plus the response patches in alchemy-run/distilled#561 that type the create envelopes, firewall tags, and server-assigned ids. Until #561 lands, `submodules/distilled` points at the distilled commit `main` pins plus #561's two commits, on the `rickyeescobar/distilled` fork (`alchemy-main-digitalocean`).

### Resources

- `Droplet`: ownership by hashed tag, adoption drift detection, `replaceAfter` age-based replacement, rename and resize through the actions API with bounded polling
- `SshKey`: fingerprint-stable identity, rename in place, replace on key material change
- `Firewall`: set-semantic rules and targets; droplets can be selected by tag so later droplets are covered

### Also on this branch

- `CertRequest` resource: PKCS#10 CSR from a private key, registered in the AWS and Cloudflare provider collections
- `Exec` runs a `destroyCommand` on delete
- Docker: containers take memory, memory-swap, read-only rootfs, no-new-privileges, inline pull registry credentials; network convergence and rename handling on adopted containers
- Cloudflare state store: `decideStateStoreInit` guards (never bootstrap over a live store, never in CI without a deployed store) and `alchemy provider cloudflare state secrets backup|restore`


rickyeescobar:
Merged `main` and resolved the conflicts; the branch is mergeable again.

### Merge resolution

- `Docker.ts`: kept `main`'s publication structure (`RegistryCredentials`, `DockerImagePublicationError`, buildx export mode, hoisted `push`) and re-applied the authenticated `pull` on top. The diff against `main` is additive.
- `ContainerProvider.ts`, `pnpm-lock.yaml`: took `main`. The lockfile is `main`'s plus the one `@distilled.cloud/digitalocean` workspace link.
- `package.json`: re-added the `./DigitalOcean` export, its `publishConfig` entry, the workspace dependency, and the `build:pkg` filter.
- `submodules/distilled`: now the commit `main` pins plus the two commits from alchemy-run/distilled#561 (fork branch `alchemy-main-digitalocean`).
- Effect rc.115 renames (`Config.String`, `Argument.File`, `Flag.Boolean`) and `scope` → `resolveStateStoreScope`.
- `Droplet.test.ts` read a hard-coded `"test"` stage; #1579 made the default `test_$USER`, so it now uses `stack.stage`.

### Isolated docker config

`pull` and `push` share one helper for the temp `DOCKER_CONFIG`:

```ts
const makeIsolatedDockerConfig = Effect.fn("Docker.makeIsolatedDockerConfig")(
  function* (credentials: RegistryCredentials) {
    const configDir = yield* fs.makeTempDirectoryScoped({ prefix: "alchemy-docker-" });
    const currentContext = yield* inheritGlobalContexts(configDir);
    yield* fs.writeFileString(
      path.join(configDir, "config.json"),
      JSON.stringify({
        currentContext,
        auths: { [credentials.server]: { auth: encodeRegistryAuth(credentials) } },
      }),
    );
    return configDir;
  },
);
```

- Docker resolves `--context` and `currentContext` under `DOCKER_CONFIG`, so the isolated config copies the global `contexts/` and `currentContext`. On `main`, `push(ref, credentials, platform, context)` with a named context fails with "context not found"; this fixes that too.
- The contexts are copied, not symlinked: a directory symlink needs elevated privilege on Windows and dangles for a relative `DOCKER_CONFIG`.
- `DOCKER_AUTH_CONFIG` was not an option here because Docker CLI < 28.3 ignores it.
- An empty `DOCKER_CONFIG` is treated as unset. A missing global `config.json` means no context; an unreadable one logs a fixed warning rather than quoting a file that holds `auths`.
- `encodeRegistryAuth` is shared with `registryEnvironment`.

### Tests

- `pull` and `push` are covered offline through the existing fake spawner, asserting the isolated config's `currentContext`, `contexts/`, and `auths`.
- Two live tests remain: a named global context resolves, and the global `currentContext` is honored (skipped when `DOCKER_HOST` or `DOCKER_CONTEXT` is set).

PR #1538, opened 2026-09-08
fix(cli): honor --profile on cloudflare bootstrap

### LINKED ISSUE #252: boostrap cf does not support profiles
`bun alchemy bootstrap cloudflare --force --profile <profile_name>` makes me long in again even though that profile is already configured and works fine for deploy/destroy; therefore it must be using the default profile

Mkassabov:
`alchemy provider cloudflare bootstrap --profile <name>` resolved Cloudflare credentials from the default profile, so a named profile that already worked for deploy/destroy still prompted for login. `--force` only redeploys the state-store worker; it was never supposed to re-authenticate.

`fromProfile` / `fromAuthProvider` require `ConfigProvider`, but bootstrap sibling-merged the `--profile` override instead of providing it into those layers. They kept the CLI's ambient `fromEnv()` and landed on `default`.

The intended composition (not a verbatim excerpt):

```ts
// feed --profile into fromProfile / fromAuthProvider
Layer.provideMerge(cloudflareLayers, config)
```

Closes #252

### DIFF
--- a/packages/alchemy/src/Alchemist/routes/cloudflare.ts
+++ b/packages/alchemy/src/Alchemist/routes/cloudflare.ts
@@ -54,6 +54,17 @@ const services = Effect.fn(function* (target: StateTarget) {
     CloudflareAuth,
     Layer.succeed(AuthProviders, {} satisfies AuthProviders["Service"]),
   );
+  // `provideMerge` (not `mergeAll`) so `fromProfile` / `fromAuthProvider`
+  // actually *see* `--profile`. Sibling-merging ConfigProvider left those
+  // layers requiring it from the outer CLI `fromEnv()` — which is how
+  // `alchemy provider cloudflare bootstrap --profile <name>` ignored the
+  // flag and resolved the default profile (#252).
+  const config = ConfigProvider.layer(
+    withProfileOverride(
+      yield* loadConfigProvider(Option.fromNullishOr(target.envFile)),
+      target.profile,
+    ),
+  );
   return Layer.mergeAll(
     Layer.provideMerge(
       Layer.mergeAll(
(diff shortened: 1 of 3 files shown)


Mkassabov (commented):
Reviewed against AGENTS.md and issue #252. **Verdict: approve** (posted as a comment review — GitHub does not allow a formal approval from the same account that opened the PR).

**Root cause** — confirmed. `CloudflareEnvironment.fromProfile()` / `CloudflareCredentials.fromAuthProvider()` resolve the profile through `resolveProviderConfig` → `ProfileStore.current` → `Config ALCHEMY_PROFILE`, i.e. from whatever `ConfigProvider` is *in their build context*. `services()` previously placed the `withProfileOverride(...)` ConfigProvider as a **sibling** in `Layer.mergeAll`, so those layers took the CLI's ambient `fromEnv()` provider instead and landed on `default` — hence the re-login prompt in #252 even though `--profile <name>` was configured. `.pipe(Layer.provideMerge(config))` feeds the override into the layers that need it while still exporting it downstream (so `bootstrapStateStore` → `CredentialsStore` / `CI` continue to see it). `teardown`, `stateLogs`, and `tailStateLogs` share `services()` and are fixed by the same change. `resolveProfileName` was already correct, which is why `profile` (used for the credentials file path) was right while the account/credentials were wrong — the fix addresses exactly that mismatch, not a symptom.

**Verification**

- `pnpm exec tsc -b` in the branch worktree: clean (exit 0).
- `pnpm test test/Alchemist/cloudflare-bootstrap-profile.test.ts --retry 0` on the branch: 3/3 pass.
- Same test file run against the base commit (`0831bd1ec`, with only `scope` exported under the new name so the file compiles): **2/3 fail** — `expected "bbbb…" to be "aaaa…"` for the `--profile` case and `AuthError: Provider 'Cloudflare' is not configured in profile 'default'` for the no-fallback case. So the test genuinely fails without the fix and pins the regression.

**Conventions** — no `async`/`await`/`node:fs` (temp dir via `FileSystem.FileSystem`, env mutation wrapped in `Effect.sync` with `acquireRelease` restore, `{ exclusive: true }` for the process-global env write), deterministic fixture data, no cloud calls. PR title is conventional-commit, body has no `#`/`##` headings and no test-plan section, includes `Closes #252`.

**Non-blocking nits** (fine to leave as-is or fold into a follow-up):

1. Hermeticity: `resolveStateStoreScope` builds its own `ConfigProvider` via `loadConfigProvider`, which also reads the cwd `.env` (`packages/alchemy/.env`, populated by `pnpm download:env`). If that file ever contains `CLOUDFLARE_API_TOKEN` + `CLOUDFLARE_ACCOUNT_ID`, the environment-credentials path wins and the first assertion would fail spuriously. Passing `envFile` pointing at an empty file inside the temp home (`resolveStateStoreScope({ profile: "staging", envFile })`) would make the test independent of the developer's `.env`.
2. File naming: siblings in `test/Alchemist/` use PascalCase (`Runtime.test.ts`); `CloudflareBootstrapProfile.test.ts` would match.
3. The PR-body snippet `Layer.provideMerge(cloudflareLayers, config)` is illustrative rather than the literal `.pipe(Layer.provideMerge(config))` — harmless, just noting.

Not exercised: a live `alchemy provider cloudflare bootstrap --profile <name>` against a real account (live runs are on hold for this session). The deploy path itself is untouched by this PR; the change is confined to layer wiring, which the unit test covers.

Mkassabov (commented):
Re-checked at `6def782` after the follow-up commit: test renamed to `CloudflareBootstrapProfile.test.ts`, and it now passes an empty `envFile` inside the isolated temp home (via `Path.Path` + `FileSystem`), so it no longer depends on the developer's cwd `.env`. `pnpm exec tsc -b` clean; `pnpm test test/Alchemist/CloudflareBootstrapProfile.test.ts --retry 0` → 3/3 pass. Verdict unchanged: approve.

PR #1539, opened 2026-09-08
fix(cloudflare/email): refuse catch-all Email.Rule and adopt matching rules

### LINKED ISSUE #413: EmailRule provider fails on first deploy when catch-all rule already exists
## Bug

The `EmailRule` provider always calls Cloudflare's `POST /zones/{zone_id}/email/routing/rules` (create) on first deploy because there is no stored `ruleId` in alchemy state. This fails with `Conflict: Invalid rule operation` when a catch-all rule already exists on the zone.

Cloudflare only allows **one** catch-all email routing rule per zone, and the default catch-all created by Email Routing cannot be deleted. This means `EmailRule` with `matchers: [{ type: "all" }]` can never successfully deploy on a zone that already has email routing enabled.

## Reproduction

```typescript
yield* Cloudflare.EmailRule("CatchAll", {
  zone: "example.com",
  matchers: [{ type: "all" }],
  actions: [{ type: "worker", value: ["my-worker"] }],
});
```

Deploy fails with:

```
ERROR: Conflict: Invalid rule operation
  at POST /zones/{zone_id}/email/routing/rules
✗ CatchAll (Cloudflare.EmailRule) fail
```

## Root cause

In `EmailRuleProvider`, the `reconcile` function only attempts `update` when `output?.ruleId` exists (i.e., alchemy previously created the rule). On first deploy there is no stored state, so it falls through to `create`, which conflicts with the existing rule.

```typescript
// EmailRule.ts — reconcile
if (output?.ruleId) {
  // update path — only taken on subsequent deploys
} else {
  // create path — always taken on first deploy, fails if rule exists
  const result = yield* create({ zoneId, ...body });
}
```

## Suggested fix

The `reconcile` function (or `read`) should list existing rules on the zone and adopt a matching one (by matcher shape) instead of unconditionally creating. This would allow alchemy to take over management of a pre-existing catch-all rule on first deploy.

Mkassabov:
`Cloudflare.Email.Rule` with `matchers: [{ type: "all" }]` still POSTs `/zones/{zone_id}/email/routing/rules` on first deploy and 409s (`Conflict: Invalid rule operation`) because the zone catch-all already exists once Email Routing is enabled.

Main already has `Email.CatchAll` (PUT `/rules/catch_all`, restore-on-destroy singleton) and `Rule.list` already filters the catch-all out — but `reconcile` still branched create vs update on `output?.ruleId`.

Fail fast rather than delegating to the catch-all endpoint. CatchAll's lifecycle is different (never physically create/delete; restore the pre-management state on destroy), and `deleteRule` on that id is the same 409. `EmailEventSource` already routes all-matcher subscriptions to CatchAll.

```ts
yield* Cloudflare.Email.Rule("CatchAll", {
  zone: "example.com",
  matchers: [{ type: "all" }],
  actions: [{ type: "drop" }],
});
// CatchAllRuleNotSupported — use Cloudflare.Email.CatchAll
```

Non-catch-all rules now observe → ensure → sync: `read` locates an existing rule by matcher identity and brands it `Unowned`; with `adopt(true)` reconcile updates in place instead of duplicating.

No distilled patch: fail-fast never calls `createRule` with `{ type: "all" }`.

Closes #413

### DIFF
--- a/packages/alchemy/src/Cloudflare/Email/Rule.ts
+++ b/packages/alchemy/src/Cloudflare/Email/Rule.ts
@@ -224,9 +224,13 @@ export const RuleProvider = () =>
         : undefined;
 
       // 2. Fall back to scanning the zone for the same matchers.
-      //    Ownership has already been verified upstream — `read` reports
-      //    existing rules as `Unowned` and the engine gates takeover
-      //    behind the adopt policy before reconcile ever runs.
+      //    `read` brands an existing match `Unowned` so the engine can
+      //    gate takeover behind adopt — but plan skips that probe when
+      //    props are still unresolved (e.g. `zone: routing.zoneId` from
+      //    a sibling created in the same deploy). In that case this scan
+      //    is the AlreadyExists race: Cloudflare rejects a second rule
+      //    with the same literal matchers, so converging on the match is
+      //    the same as catching Conflict and re-listing.
       if (!observed) {
         observed = yield* findByMatchers(zoneId, desired.matchers);
       }
@@ -296,6 +300,8 @@ const isCatchAllRule = (rule: {
 const isCatchAllMatchers = (matchers: Matcher[]): boolean =>
   matchers.length === 1 && matchers[0]?.type === "all";
 
+// `olds.matchers` may still be unresolved Inputs after stripUnresolved;
+// only treat a concrete array as matcher identity for the adoption scan.
 const asMatchers = (value: unknown): Matcher[] | undefined =>
   Array.isArray(value) ? (value as Matcher[]) : undefined;
 
--- a/packages/alchemy/src/Cloudflare/Email/Rule.ts
+++ b/packages/alchemy/src/Cloudflare/Email/Rule.ts
@@ -1,6 +1,9 @@
 import * as emailRouting from "@distilled.cloud/cloudflare/email-routing";
+import * as Data from "effect/Data";
 import * as Effect from "effect/Effect";
+import * as Option from "effect/Option";
 import * as Stream from "effect/Stream";
+import { Unowned } from "../../AdoptPolicy.ts";
 import { isResolved } from "../../Diff.ts";
 import * as Provider from "../../Provider.ts";
 import { Resource } from "../../Resource.ts";
@@ -43,6 +46,10 @@ export type RuleProps = {
   priority?: number;
   /**
    * Matchers that define which inbound emails trigger this rule.
+   *
+   * A sole `{ type: "all" }` matcher is the zone catch-all, which is a
+   * per-zone singleton owned by `Cloudflare.Email.CatchAll` — declaring
+   * it here fails with `CatchAllRuleNotSupported`.
    */
   matchers: Matcher[];
   /**
@@ -68,10 +75,30 @@ export type Rule = Resource<
 >;
 
 /**
+ * Raised when `Cloudflare.Email.Rule` is declared with a sole
+ * `{ type: "all" }` matcher. Cloudflare models the zone catch-all as a
+ * singleton behind `PUT /zones/{zone_id}/email/routing/rules/catch_all`
+ * and rejects creating a second one through the ordinary rule endpoint
+ * (`Conflict: Invalid rule operation`). Use `Cloudflare.Email.CatchAll`.
+ */
+export class CatchAllRuleNotSupported extends Data.TaggedError(
+  "CatchAllRuleNotSupported",
+)<{
+  message: string;
+}> {}
+
+/**
  * A Cloudflare Email Routing rule.
  *
  * Rules forward inbound mail matching `matchers` to the listed actions
  * (forward to a verified destination, drop, or hand off to a Worker).
+ *
+ * Safety: routing rules carry no ownership markers. When there is no
+ * prior state, `read` scans the zone for an existing non-catch-all rule
+ * with the same matchers and reports it as `Unowned`, so the engine
+ * refuses to take it over unless `--adopt` (or `adopt(true)`) is set.
+ * The zone catch-all is excluded from that scan — it is owned by
+ * `Email.CatchAll`.
  * ### Forwarding Mail
  * **Example:** Forward `info@` to a verified destination
  * ```typescript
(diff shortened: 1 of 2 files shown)


Mkassabov (commented):
Reviewed against AGENTS.md (Reconciler doctrine, Typed Error Doctrine, test conventions, PR conventions). Verified from the branch worktree: `pnpm exec tsc -b` clean, `pnpm docs:check-jsdoc` clean. Live `EmailRule.test.ts` run is still pending the shared credentials hold, so this is a comment review — I'll approve once the suite is green.

### What's right

- Root cause is addressed, not the symptom: `reconcile` is now one observe → ensure → sync flow (`observeById` → `findByMatchers` → `createRule` → `updateRule` only on drift) with no `output?.ruleId` create/update branching.
- Refusing the sole `{ type: "all" }` matcher with a typed `CatchAllRuleNotSupported` is the right call given `Email.CatchAll` already owns the singleton with capture-and-restore semantics and `EmailEventSource` already routes all-matchers there. Delegating would have mixed two incompatible lifecycles in one resource.
- `read` now uses typed tags (`EmailRoutingRuleNotFound`/`Forbidden`, `InvalidRoute`) instead of the previous `Effect.catch(() => …)` catch-all, and brands matcher-identity hits `Unowned` — consistent with the other marker-less Cloudflare resources.
- Both new tests would fail on `main` (`main` surfaces `Conflict` instead of `CatchAllRuleNotSupported`; `main` duplicates instead of returning `OwnedBySomeoneElse`/adopting). Deterministic names, capture-and-restore of the catch-all, bounded retries, start+end `stack.destroy()`, out-of-band verification via distilled — all present.
- PR body follows the conventions (no `#`/`##`, no test plan, snippet-first, `Closes #413`).

### Points to address / confirm (none blocking on their own)

1. **(low/medium) `reconcile` step 2 can take over a foreign rule without the adopt gate.** The matcher scan in `reconcile` relies on the plan-time `read` probe having already gated ownership. Plan skips that probe when `news` is unresolved (`!isResolved(news)`), e.g. `zone: routing.zoneId` from an `Email.Routing` created in the same deploy — the exact shape the suite's own list test uses. In that case a pre-existing same-matcher rule is silently updated in place. This matches how other deterministic-identity providers behave, so I'm not blocking on it, but the "Ownership has already been verified upstream" comment overstates the guarantee — please soften it to name the unresolved-props exception (or, if Cloudflare actually rejects duplicate literal matchers, say so and treat the scan as the doctrine's "catch AlreadyExists as a race" step).
2. **(low) Confirm at live run** that `stack.deploy(...).pipe(Effect.flip)` yields the raw `CatchAllRuleNotSupported` instance. If Apply wraps provider failures, reuse the Cause-walking pattern you already have for `OwnedBySomeoneElse` (`findOwnedError`) instead of `toBeInstanceOf` on the flipped value.
3. **(nit)** `asMatchers(value: unknown)` widens `olds.matchers` (already `Matcher[]`) to `unknown` and casts back. If the intent is to survive `stripUnresolved` holes, a one-line comment would make that clear.

Mkassabov (commented):
**Verdict: one change requested** (posted as a comment — GitHub blocks "request changes" from the shared author account).

Follow-up on 14ccde59f: the comment rewrite on the matcher-scan race and the Cause-walking `findCatchAllError` both look good. `pnpm exec tsc -b` still clean.

### Live run — flaky by design, not by luck

Ran from the branch worktree with retries disabled:

```
CLOUDFLARE_TEST_EMAIL_ROUTING=1 env -u CI timeout 300 pnpm test test/Cloudflare/Email/EmailRule.test.ts test/Cloudflare/Email/EmailRuleWorkerTarget.test.ts test/Cloudflare/Email/EmailCatchAll.test.ts --profile alchemy-testing --retry 0
```

- `refuses a sole { type: 'all' } matcher and leaves the catch-all untouched (#413)` → **FAIL** at `EmailRule.test.ts:237` (`expect(after.enabled).toEqual(before.enabled)`; `expected true to equal false`). The provider did the right thing (`CatchAllRuleNotSupported` raised, nothing written).
- Same test run alone (`pnpm test test/Cloudflare/Email/EmailRule.test.ts --retry 0`) → **PASS** 3/3.

The worker-side green run shows the same thing: that test is marked `[retried x2]` in `2026-09-08T18-25-37-pid698329.log`, i.e. it passed only on the third attempt under the default `--retry 2`.

Cause: `describe.sequential` only serialises tests *within* a file. `EmailCatchAll.test.ts` and `EmailRuleWorkerTarget.test.ts` run concurrently in other files and PUT the same zone's `/rules/catch_all` (enable/disable, point at a Worker) between this test's `before` snapshot and its `after` read. The `before`/`after` comparison therefore asserts on a singleton that sibling files legitimately mutate — the assertion is not attributable to the code under test.

### Requested change

Keep the assertion that `CatchAllRuleNotSupported` is raised (that's the #413 regression guard), and make the "left untouched" claim provable without racing siblings. Either:

1. drop the `before`/`after` singleton comparison and instead assert that `stack.deploy` never touched the rule endpoints — e.g. assert the failure is raised *before* any API write by checking `listRules` still contains no non-catch-all row for this stack (or simply that `after.id === before.id`, which is stable regardless of sibling PUTs and proves no second catch-all was created); or
2. if you want to keep the full snapshot comparison, mark the test `{ exclusive: true }` so the runner's whole-process write lock serialises it against the other Email files (AGENTS.md documents this option for tests that touch shared process/global state; the zone catch-all is the same kind of shared singleton).

Option 1 is preferable — it keeps the file parallel-safe and the assertion targeted at what #413 is about. Please re-run the three Email files together with `--retry 0` before marking ready; I'll do the same and approve.

Mkassabov (commented):
**Verdict: APPROVED** (reviewer-a; posted as a comment because GitHub blocks a formal approval from the PR author account, which the reviewer shares).

Approving at cb2099280.

- `pnpm exec tsc -b` and `pnpm docs:check-jsdoc` clean from the branch worktree.
- Live, retries disabled: `CLOUDFLARE_TEST_EMAIL_ROUTING=1 env -u CI timeout 300 pnpm test test/Cloudflare/Email/EmailRule.test.ts test/Cloudflare/Email/EmailRuleWorkerTarget.test.ts test/Cloudflare/Email/EmailCatchAll.test.ts --profile alchemy-testing --retry 0` → 9 passed / 0 failed (`.alchemy/log/test/2026-09-08T18-35-48-pid720946.log`). The #413 refusal test now passes alongside the sibling catch-all files without a retry.
- Root cause fixed per the Reconciler doctrine (observe → ensure → sync, no create/update branch on `output`), typed `CatchAllRuleNotSupported` routes users to `Email.CatchAll`, matcher-identity adoption gated by `Unowned`, and the race caveat is documented at the call site.

PR #1540, opened 2026-09-08
fix(cloudflare): register Worker ExportedHandler methods as listeners

### LINKED ISSUE #414: Worker shape methods for ExportedHandler (email, queue, scheduled) not registered as event listeners

Mkassabov:
`serve()` now registers a listener for every non-fetch `ExportedHandler` method present on the Worker init shape (`email`, `queue`, `scheduled`, …). Those methods are stripped from the RPC shape at runtime and omitted from `RpcAsync` / `RpcWireShape` / `bindWorker` (`RpcMethods<Shape>`), so they are not WorkerEntrypoint RPC methods.

```ts
return {
  fetch: Effect.succeed(HttpServerResponse.text("ok")),
  scheduled: (controller) => record(controller.scheduledTime),
  greet: (name: string) => Effect.succeed(`hello ${name}`),
};
```

Addresses the two notes on the closed [#415](https://github.com/alchemy-run/alchemy/pull/415): tests, and excluding handlers from the Worker RPC interface.

The live cron test is `skipIf(FAST)` because Cloudflare's minimum trigger is one minute (the body waits for the fire after `/reset`). Dispatch is covered in ~4s by the local `cdn-cgi` fixture.

Closes #414

### DIFF
--- a/packages/alchemy/src/Cloudflare/Workers/InferEnv.ts
+++ b/packages/alchemy/src/Cloudflare/Workers/InferEnv.ts
@@ -41,7 +41,7 @@ import type { RateLimitBinding } from "./RateLimitBinding.ts";
 import type { RpcErrorEnvelope, RpcStreamEnvelope } from "./Rpc.ts";
 import type { SecretKeyBinding } from "./SecretKeyBinding.ts";
 import type { VersionMetadataBinding } from "./VersionMetadataBinding.ts";
-import type { ExportedHandlerMethod, Worker } from "./Worker.ts";
+import type { ExportedHandlerMethod, RpcMethods, Worker } from "./Worker.ts";
 import type { WorkerEntrypointBinding } from "./WorkerEntrypoint.ts";
 import type { WorkerLoader as WorkerLoaderResource } from "./WorkerLoader.ts";
 
@@ -168,17 +168,19 @@ export type GetBindingType<T> =
  * where envelopes are decoded for you.
  */
 export type RpcWireShape<Shape> = {
-  [
-    K in keyof Shape as K extends ExportedHandlerMethod ? never : K
-  ]: Shape[K] extends (...args: infer A) => Effect.Effect<infer T, any, any>
+  [K in keyof RpcMethods<Shape>]: RpcMethods<Shape>[K] extends (
+    ...args: infer A
+  ) => Effect.Effect<infer T, any, any>
     ? (...args: A) => Promise<T | RpcErrorEnvelope>
-    : Shape[K] extends (...args: infer A) => Stream.Stream<any, any, any>
+    : RpcMethods<Shape>[K] extends (
+          ...args: infer A
+        ) => Stream.Stream<any, any, any>
       ? (...args: A) => Promise<RpcStreamEnvelope | RpcErrorEnvelope>
-      : Shape[K] extends Effect.Effect<infer T, any, any>
+      : RpcMethods<Shape>[K] extends Effect.Effect<infer T, any, any>
         ? Promise<T | RpcErrorEnvelope>
-        : Shape[K] extends Stream.Stream<any, any, any>
+        : RpcMethods<Shape>[K] extends Stream.Stream<any, any, any>
           ? Promise<RpcStreamEnvelope | RpcErrorEnvelope>
-          : Shape[K] extends (...args: infer A) => infer R
+          : RpcMethods<Shape>[K] extends (...args: infer A) => infer R
             ? (...args: A) => Promise<Awaited<R>>
-            : Promise<Shape[K]>;
+            : Promise<RpcMethods<Shape>[K]>;
 };
--- a/packages/alchemy/src/Cloudflare/Workers/InferEnv.ts
+++ b/packages/alchemy/src/Cloudflare/Workers/InferEnv.ts
@@ -41,7 +41,7 @@ import type { RateLimitBinding } from "./RateLimitBinding.ts";
 import type { RpcErrorEnvelope, RpcStreamEnvelope } from "./Rpc.ts";
 import type { SecretKeyBinding } from "./SecretKeyBinding.ts";
 import type { VersionMetadataBinding } from "./VersionMetadataBinding.ts";
-import type { Worker } from "./Worker.ts";
+import type { ExportedHandlerMethod, Worker } from "./Worker.ts";
 import type { WorkerEntrypointBinding } from "./WorkerEntrypoint.ts";
 import type { WorkerLoader as WorkerLoaderResource } from "./WorkerLoader.ts";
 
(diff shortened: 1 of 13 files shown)


Mkassabov (commented):
**Verdict: changes requested** (posted as a comment — GitHub blocks "request changes" from the PR author account; the reviewer agent shares it).

Reviewed against AGENTS.md and the two notes on #415 (tests + RPC exclusion). Verified from the branch worktree: `pnpm exec tsc -b` clean, `pnpm docs:check-jsdoc` clean, `WorkerRuntimeContext.test.ts` (4 tests) and `WorkerShapeHandlers.local.test.ts` (dev-mode, drives `/cdn-cgi/handler/scheduled` against workerd) pass. The live `WorkerShapeHandlers.test.ts` cron run is pending the shared credentials hold — this is a comment review until that is green.

### What's right

- Root cause fixed where the issue points: `serve()` registers one listener per non-`fetch` `ExportedHandler` method present on the shape, keyed on `isWorkerEvent(event) && event.type === method`, so `dispatch("email"|"queue"|"scheduled")` no longer dies with "No event handler found".
- Runtime RPC exclusion is real: `userShape` is filtered before capture, so the `WorkerEntrypoint` proxy's `built.shape()[prop]` lookup can't reach a handler, and `toRpcAsync` passes handler names straight through.
- Tests would fail on `main`: the unit test dispatches `scheduled`/`queue`/`email` through `ctx.exports` and asserts the shape filter; the types file asserts `RpcAsync`/`RpcWireShape` drop the handler keys; the dev-mode test exercises the actual workerd dispatch path.
- PR body follows the conventions (no `#`/`##`, snippet-first, no test plan, `Closes #414`).

### Requested changes

1. **(medium) `bindWorker` still types handler methods as RPC-callable.** `bindWorker` returns `makeRpcStub<Shape>` where `Shape` is the full `Rpc<Self>` shape, so from an Effect-native Worker `const b = yield* bindWorker(B); b.scheduled(controller)` type-checks — and at runtime the `makeRpcStub` proxy forwards it to the Cloudflare fetcher, which *will* invoke the remote `scheduled` bridge method over JSRPC. The maintainer's ask was to exclude handlers from the Worker's RPC interface; `bindWorker` is that interface on the Effect side. Suggest a single shared type (e.g. `RpcMethods<Shape> = Omit<Shape, ExportedHandlerMethod>`) applied in `bindWorker`'s return type as well as `RpcAsync`/`RpcWireShape`, plus a type assertion for it in `WorkerShapeHandlers.types.ts` (e.g. `Awaited<ReturnType<typeof bindWorker<Shape>>>` lacks `scheduled`). Optionally have `makeRpcStub`'s proxy refuse `isExportedHandlerMethod(prop)` too so the runtime matches the type.

### Non-blocking

2. **(low) Error channel on shape handlers.** The listener casts the user's handler to `Effect<unknown, never, never>`; a typed failure ends up as a rejected promise from the bridge (which is a fine default — queue batches retry, scheduled runs are marked failed). Worth one sentence in the JSDoc/PR body so the contract is explicit, mirroring how `EmailEventSource` documents log-and-die.
3. **(low) Typed handler inputs.** `Main` only declares `fetch`; the fixture types `scheduled` as `{ scheduledTime: number }` by hand. Adding optional typed members to `Main` (`scheduled?: (c: cf.ScheduledController) => …`, `queue?`, `email?`) would give users the real input types and keep them out of the `MainRpc` index signature. Fine as a follow-up.
4. **(nit)** In `WorkerBridge`, `prop in target` already covers every `ExportedHandlerMethods` entry (the constructor assigns them as own props), so the added `isExportedHandlerMethod(prop)` branch there is redundant. Harmless; keep or drop.
5. **(nit)** The live test's poll budget (36 × 5 s) equals its 180 s timeout, same as `CronEventSource.test.ts`, so a slow first fire surfaces as a timeout rather than the assertion. Precedent exists; just noting.

Mkassabov (commented):
**Verdict: APPROVED** (reviewer-a; posted as a comment because GitHub blocks a formal approval from the PR author account, which the reviewer shares).

Approving at b1e346155.

- The blocking item from round 1 is resolved by ba4716552: one `RpcMethods<Shape>` type (index signature stripped, then `Omit<…, ExportedHandlerMethod>`) is shared by `bindWorker`, `RpcAsync`, and `RpcWireShape`, with `KeysEqual` assertions pinning the three surfaces together; `makeRpcStub` refuses handler props at runtime, so the Effect-side stub can no longer invoke `scheduled`/`email`/`queue` over JSRPC. The redundant `WorkerBridge` branch was removed as suggested.
- Verified from the branch worktree: `pnpm exec tsc -b` clean, `pnpm docs:check-jsdoc` clean, `WorkerRuntimeContext.test.ts` (4) + `WorkerShapeHandlers.local.test.ts` (dev-mode workerd `/cdn-cgi/handler/scheduled`) → 5/5 with `--retry 0`.
- Live: `env -u CI timeout 300 pnpm test test/Cloudflare/Workers/WorkerShapeHandlers.test.ts --profile alchemy-testing --retry 0` → PASS (165.6 s; `.alchemy/log/test/2026-09-08T18-41-27-pid727679.log`). An earlier run of mine got 0 fires in 216 s; I re-ran exclusively with an out-of-band `getScriptSchedule` check via distilled confirming `["* * * * *"]` attached to the deployed script, and it fired within ~2 min. The first failure is consistent with Cloudflare cron scheduling latency on a freshly (re)deployed script rather than a dispatch bug — the dev-mode test exercises the same `scheduled()` dispatch deterministically. Stack destroyed and the script confirmed gone afterwards.
- Both maintainer notes on #415 are met: real tests (unit + dev-mode + live), and handler methods excluded from the RPC interface at the type level and at runtime.

PR #1541, opened 2026-09-08
test(cloudflare/workers): pin CORS headers on plain Worker fetch

### LINKED ISSUE #175: HttpMiddleware.cors() doesn't tag non-preflight responses when used inside a Cloudflare.Worker

Mkassabov:
Issue #175 is already fixed on main by #404: the Cloudflare Worker HTTP adapter now routes the handler through `EffectHttp.toHandled`, so `HttpMiddleware.cors()` pre-response handlers run on GET/POST, not only OPTIONS.

HttpApi coverage added in #404 uses `HttpRouter.cors` on a POST. This PR pins the original #175 shape: a plain `Cloudflare.Worker` whose `fetch` is `HttpMiddleware.cors()(handler)`.

```ts
return {
  fetch: HttpMiddleware.cors()(
    HttpServerResponse.json({ message: "world" }),
  ),
};
```

- In-process: `makeRequestEffect` asserts `Access-Control-Allow-Origin` on OPTIONS, GET, and POST
- Live: deploys the fixture Worker and asserts the same headers on the wire

No adapter change. Closes #175.

### DIFF
--- a/packages/alchemy/test/Cloudflare/Workers/Cors.test.ts
+++ b/packages/alchemy/test/Cloudflare/Workers/Cors.test.ts
@@ -1,0 +1,112 @@
+import * as Cloudflare from "@/Cloudflare/index.ts";
+import * as Alchemy from "@/index.ts";
+import * as Test from "@/Test/Alchemy";
+import { expect } from "alchemy-test";
+import * as Effect from "effect/Effect";
+import * as Schedule from "effect/Schedule";
+import * as HttpClient from "effect/unstable/http/HttpClient";
+import * as HttpClientRequest from "effect/unstable/http/HttpClientRequest";
+import type { HttpClientResponse } from "effect/unstable/http/HttpClientResponse";
+import CorsWorker from "./fixtures/cors-worker.ts";
+
+const { test, beforeAll, afterAll, deploy, destroy } = Test.make({
+  providers: Cloudflare.providers(),
+});
+
+const ORIGIN = "https://example.test";
+
+const Stack = Alchemy.Stack(
+  "WorkersCorsStack",
+  { providers: Cloudflare.providers(), state: Cloudflare.state() },
+  Effect.gen(function* () {
+    const worker = yield* CorsWorker;
+    return { url: worker.url.as<string>() };
+  }),
+);
+
+const stack = beforeAll(deploy(Stack));
+afterAll.skipIf(!!process.env.NO_DESTROY)(destroy(Stack));
+
+// Raw HttpClient does not fail on non-2xx, so Effect.retry would not fire
+// through the freshly-deployed workers.dev 404/500 window. Fail non-2xx
+// explicitly so the first request retries until the edge is ready.
+const requestUntilReady = (
+  effect: Effect.Effect<HttpClientResponse, unknown, never>,
+) =>
+  effect.pipe(
+    Effect.flatMap(
+      Effect.fn(function* (res) {
+        return res.status >= 200 && res.status < 300
+          ? res
+          : yield* Effect.fail(
+              new Error(`Worker not ready: ${res.status} ${yield* res.text}`),
+            );
+      }),
+    ),
+    Effect.retry({
+      schedule: Schedule.exponential("500 millis"),
+      times: 10,
+    }),
+  );
+
+test(
+  "HttpMiddleware.cors() adds Access-Control-Allow-Origin on OPTIONS preflight",
+  Effect.gen(function* () {
+    const { url } = yield* stack;
+    const client = yield* HttpClient.HttpClient;
+
+    const res = yield* requestUntilReady(
+      client.execute(
+        HttpClientRequest.make("OPTIONS")(`${url}/hello`).pipe(
+          HttpClientRequest.setHeaders({
+            Origin: ORIGIN,
+            "Access-Control-Request-Method": "GET",
+          }),
+        ),
+      ),
+    );
+    expect(res.status).toBe(204);
+    expect(res.headers["access-control-allow-origin"]).toBe("*");
+  }),
+  { timeout: 120_000 },
+);
+
+test(
+  "HttpMiddleware.cors() adds Access-Control-Allow-Origin on GET responses",
+  Effect.gen(function* () {
+    const { url } = yield* stack;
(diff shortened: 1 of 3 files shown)


Mkassabov (commented):
Reviewed against AGENTS.md and issue #175. **Verdict: changes requested** (posted as a comment review — GitHub does not allow request-changes from the account that opened the PR). One blocking item; the rest is verified and fine.

**Blocking: `pnpm exec tsc -b` fails on this branch** (CI runs `pnpm tsc -b`, so this will be red):

```
packages/alchemy/test/Cloudflare/Workers/HttpServer.cors.test.ts(27,5): error TS2322: Type 'Effect<void, unknown, unknown>' is not assignable to type 'Effect<unknown, unknown, Scope | TestClock | TestConsole>'.
packages/alchemy/test/Cloudflare/Workers/HttpServer.cors.test.ts(41,5): error TS2322: (same)
packages/alchemy/test/Cloudflare/Workers/HttpServer.cors.test.ts(52,5): error TS2322: (same)
```

`makeRequestEffect` is declared `<Req = never>` but its body ends in `}) as any`, so at the call site the yielded value (and the effect's `R`) infers as `unknown`, which `it.effect` rejects. Any of these fixes is fine — the cheapest is to pin the generic at the call site, since the handler has no requirements:

```ts
const response = yield* makeRequestEffect<never>(corsRequest("GET"), corsHandler);
```

or, if you'd rather the test not lean on the `as any`, annotate a typed wrapper once:

```ts
const run = (req: cf.Request): Effect.Effect<Response> =>
  makeRequestEffect<never>(req, corsHandler);
```

(Bun runs the file fine — 3/3 pass at runtime — but the workspace type-check is the gate.)

**Everything else checks out**

- Root cause: agreed that #175 is already fixed on `main` by #404 (`toHandledWebResponse` / `EffectHttp.toHandled` in `HttpServer.ts`), and that the existing HttpApi coverage from #404 exercises `HttpRouter.cors` on an HttpApi route, not the issue's shape (`HttpMiddleware.cors()` around a plain `Cloudflare.Worker` `fetch`). This PR pins that exact shape at two levels, which is the right response to an already-fixed issue.
- The in-process test is a real regression test: I ran it against a copy of the branch with `HttpServer.ts` reverted to the pre-#404 version (`2e3d176^`) — **2/3 fail** with `expected null to be "*"` on GET and POST, OPTIONS still passes (the preflight short-circuit), exactly the #175 symptom. With the current adapter: 3/3.
- Live: `env -u CI timeout 300 pnpm test test/Cloudflare/Workers/Cors.test.ts --profile alchemy-testing --retry 0` from the branch worktree → **3/3 pass**, 17.1s, stack destroyed (log `.alchemy/log/test/2026-09-08T18-27-29-pid702951.log`).
- Conventions: fixture under `fixtures/`, file-based `main: import.meta.url`, `beforeAll(deploy)` / `afterAll.skipIf(NO_DESTROY)(destroy)`, bounded `Effect.retry` (`Schedule.exponential("500 millis")`, `times: 10`) with the non-2xx → failure shim so retry actually fires through the workers.dev warm-up window, deterministic stack name, per-test timeouts ≤ 120s, no `Date.now` polling, no `async`/`node:fs`. PR title is conventional-commit (`test(...)`), body has no `#`/`##` headings and no test-plan section, `Closes #175`.

**Non-blocking nit**: `Effect.promise(() => response.json())` in the in-process test — `Effect.tryPromise` would keep a malformed body in the typed error channel instead of a defect, but for a test asserting a fixed JSON body this is immaterial.

Will approve once `tsc -b` is green.

Mkassabov (commented):
Re-checked at `317bb14`: `pnpm exec tsc -b` is clean (exit 0) and `HttpServer.cors.test.ts` passes 3/3. The `handleCors` wrapper pins `makeRequestEffect<never>` to `Effect.Effect<Response>`, and `Effect.tryPromise` replaces `Effect.promise`. The live `Cors.test.ts` is unchanged since my 3/3 live run (`2026-09-08T18-27-29-pid702951.log`). **Verdict: approve** (comment review — GitHub does not allow a formal approval from the account that opened the PR).

PR #1553, opened 2026-09-09
fix(test): deflake AutoScaling CompleteLifecycleAction binding test

sam-goodwin:
Fix the flaky `CompleteLifecycleAction binding is IAM-wired` test (`Function not ready: 502`).

The fixture ran on Lambda's 3s default timeout while `/complete-bogus` does a live Auto Scaling call. A cold invocation measures ~2.5s alone; under the full suite's load it crosses 3s, Lambda kills it, and the Function URL answers 502. The test then failed outright because `Effect.repeat` only re-runs successes — a non-200 was never retried, unlike the `/health` probe in `beforeAll`.

```ts
// fixtures/lifecycle-handler.ts
timeout: Duration.seconds(30),

// LifecycleHookEventSource.test.ts — retry the failure path too
Effect.retry({ schedule: Schedule.spaced("3 seconds"), times: 10 }),
Effect.repeat({ until: (b) => b.tag !== "AccessDenied" && ..., ... }),
```


sam-goodwin:
Folded into #1552.

PR #1588, opened 2026-09-11
feat(aws/s3): manage blocked encryption types

bjornpagen:
Expose `encryption.blockedEncryptionTypes` on S3 buckets and the S3 state store.

```ts
const bucket = yield* AWS.S3.Bucket("my-bucket", {
  encryption: {
    sseAlgorithm: "AES256",
    blockedEncryptionTypes: ["SSE-C"],
  },
});
```

- Omit the setting to preserve the bucket's observed restrictions, including when changing the default encryption algorithm or KMS key.
- Set `["SSE-C"]` to block new writes encrypted with customer-provided keys.
- Set `[]` to allow SSE-C writes, translated to AWS's `"NONE"` value.
- Compare normalized restrictions alongside encryption defaults so setting-only changes apply and equivalent configurations skip unnecessary writes.

The same configuration is available through `AWS.state({ encryption: { ... } })`.

### DIFF
--- a/packages/alchemy/src/AWS/StateStore/State.ts
+++ b/packages/alchemy/src/AWS/StateStore/State.ts
@@ -478,7 +478,14 @@ const ensureStateBucket = (
       yield* s3.putBucketEncryption({
         Bucket: bucket,
         ServerSideEncryptionConfiguration: {
-          Rules: [desiredEncryptionRule],
+          // Preserve encryption-type blocks managed outside the state store.
+          Rules: [
+            {
+              ...desiredEncryptionRule,
+              BlockedEncryptionTypes:
+                observedEncryption?.BlockedEncryptionTypes,
+            },
+          ],
         },
       });
     }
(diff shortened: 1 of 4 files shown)


packages/alchemy/src/AWS/StateStore/State.ts
```diff
@@ -485,7 +485,14 @@ const ensureStateBucket = (
...
         ServerSideEncryptionConfiguration: {
-          Rules: [desiredEncryptionRule],
+          // Preserve encryption-type blocks managed outside the state store.
+          Rules: [
+            {
+              ...desiredEncryptionRule,
+              BlockedEncryptionTypes:
+                observedEncryption?.BlockedEncryptionTypes,
+            },
+          ],
```
sam-goodwin: This isn't right, this means we can never set this property. Instead, I am re-purposing this PR to expose this property as settable and roll in the right preservation rules.

PR #1589, opened 2026-09-11
fix(aws/rds): enforce desired parameter group values

bjornpagen:
Make RDS parameter groups converge to declared overrides and return values observed from AWS.

```ts
const group = yield* AWS.RDS.DBParameterGroup("DatabaseSettings", {
  family: "postgres16",
  parameters: { work_mem: "8192" },
});
```

Omitting `parameters` is equivalent to `{}`: undeclared user overrides are reset to engine defaults, including during adoption and unchanged-input deployments. Removing an entry or the entire map restores defaults rather than preserving external configuration.

Read back accepted changes within a bounded retry budget, retain explicitly managed engine defaults in outputs, and reconcile tags against live AWS tags. Static parameter values still require an instance reboot to take effect.

### DIFF
--- a/packages/alchemy/src/AWS/RDS/DBParameterGroup.ts
+++ b/packages/alchemy/src/AWS/RDS/DBParameterGroup.ts
@@ -1,4 +1,5 @@
 import * as rds from "@distilled.cloud/aws/rds";
+import * as Data from "effect/Data";
 import * as Effect from "effect/Effect";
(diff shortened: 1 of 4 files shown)


sam-goodwin:
What is the point of this change? Alchemy is not deisgned to be used with "manual overrides" in the dashboard. Either use alchemy to manage it or don't. No in between

PR #1623, opened 2026-09-13
fix(aws)!: preserve explicit security group egress intent

henningpokriefke:
## Summary

- distinguish an explicitly empty SecurityGroup egress array from omitted egress
- remove AWS's automatically created allow-all egress rule when egress is explicitly empty
- retain the default allow-all egress rule when egress is omitted
- add focused mocked provider coverage and a live AWS lifecycle test with optional NO_DESTROY inspection

## Behavior

The optional `egress` property now has three distinct, intentional states:

```typescript
// Omitted: retain the documented AWS default and allow all outbound IPv4 traffic.
SecurityGroup("DefaultEgress", {
  vpcId,
});

// Explicitly empty: do not allow any outbound traffic.
SecurityGroup("NoEgress", {
  vpcId,
  egress: [],
});

// Non-empty: allow only the explicitly configured outbound traffic.
SecurityGroup("HttpsEgress", {
  vpcId,
  egress: [
    {
      ipProtocol: "tcp",
      fromPort: 443,
      toPort: 443,
      cidrIpv4: "0.0.0.0/0",
    },
  ],
});
```

Previously, it was not possible to express a security group with zero egress rules through the `SecurityGroup` resource: `egress: []` fell back to the same allow-all rule as an omitted `egress` property. This change fixes that gap, making `egress: []` reliably converge to no outbound rules.

## Breaking change

BREAKING CHANGE: SecurityGroup with egress: [] now removes AWS's default allow-all egress rule. Previously, both omitted egress and an explicit empty array restored the default rule.

Workloads that passed an empty array while relying on AWS's implicit default egress must now omit egress or configure the desired rule explicitly.


packages/alchemy/src/AWS/EC2/SecurityGroup.ts
```diff
@@ -659,7 +659,7 @@ export const SecurityGroupProvider = () =>
               DryRun: false,
             });
             yield* session.note(`Applied ${news.egress.length} egress rules`);
-          } else {
+          } else if (news.egress === undefined) {
```
sam-goodwin, line 662: Should we check if it's empty too?

packages/alchemy/test/AWS/EC2/SecurityGroupEgress.test.ts
sam-goodwin, line 1: Please don't create new test files. Use the existing ones for the resource.

packages/alchemy/src/AWS/EC2/SecurityGroup.ts
henningpokriefke: The empty case is intentional: so egress: [] leaves the group with no outbound rules. Only omitted egress should restore the default allow-all rule. This gives us the three cases described in the PR body (see below): omitted → default allow-all, empty → no outbound rules, non-empty → only the specified rules. Checking for an empty array here would restore allow-all and undo the fix.

// Omitted: retain the documented AWS default and allow all outbound IPv4 traffic.
SecurityGroup("DefaultEgress", {
  vpcId,
});
// Explicitly empty: do not allow any outbound traffic.
SecurityGroup("NoEgress", {
  vpcId,
  egress: [],
});
// Non-empty: allow only the explicitly configured outbound traffic.
SecurityGroup("HttpsEgress", {
  vpcId,
  egress: [
    {
      ipProtocol: "tcp",
      fromPort: 443,
      toPort: 443,
      cidrIpv4: "0.0.0.0/0",
    },
  ],
});

packages/alchemy/test/AWS/EC2/SecurityGroupEgress.test.ts
henningpokriefke: fixed.

henningpokriefke:
I also removed the non aws live test because its kinda pointless if you have the aws live test. + I made the test more readable. and it ran against real AWS on the commit b8560aa
image

PR #1635, opened 2026-09-14
fix(cli): isolate React renderer from consumer dependencies

### LINKED ISSUE #1634: fix(cli): isolate React renderer from consumer dependencies
## Reproduction

In a Bun consumer that installs Alchemy alongside another React version:

```json
{
  "dependencies": {
    "alchemy": "2.0.0-beta.77",
    "react": "19.3.0-canary-28cd4bb0-20260723"
  }
}
```

Register an auth provider and open its interactive profile flow:

```sh
bun alchemy profile edit \
  --profile default \
  --add Cloudflare \
  --config ./alchemy.run.ts
```

The TUI crashes before authentication starts:

```text
TypeError: null is not an object
(evaluating 'ReactSharedInternals.H.useMemo')
    at ObservingEnvironment
```

## Root cause

The published Bun path executes Alchemy's TypeScript source. Alchemy's CLI hooks can resolve one React installation while `@alchemy.run/sigil` / `react-reconciler` resolve another from the consumer dependency graph. React hook dispatchers are identity-sensitive, so the renderer crashes when those module identities split.

## Expected behavior

The interactive CLI should own one isolated React/Sigil renderer. A user stack's React version and module identity must not affect the CLI, and Alchemy should not expose its internal renderer as an application peer dependency.

## Proposed fix

Bundle the renderer dependency graph into both published CLI entrypoints (`alchemy.js` and `exec.js`) while leaving user-stack dependencies external. Remove React from Alchemy's peer dependencies and add a build-time assertion that neither artifact externalizes the private renderer graph.

Arnav12114:
Bundle the CLI's React, Sigil, reconciler, scheduler, and devtools graph into both published entrypoints while leaving user-stack dependencies external.

- Remove React from Alchemy's consumer peer contract.
- Emit self-contained `alchemy.js` and `exec.js` artifacts without renderer chunks.
- Fail bundling when either artifact externalizes the renderer or omits its private React runtime.

Closes #1634.


BlankParticle:
fixed by #1636, I updated sigil to ship its own react runtime, which is the correct thing to do long term since alchemy itself will never depend on `react` and having a bundled copy will prevent this issue
Thanks for catching this issue 🫡

PR #1644, opened 2026-09-14
fix(aws): replace fixed-IP EC2 instances delete-first

Dawsson:
AWS cannot create a replacement EC2 instance while the existing instance still owns the same explicitly requested primary private IP. The EC2 provider planned these replacements create-first, so AWS rejected the new instance. If that attempt was interrupted after Alchemy removed the old generation, persisted replacement state could contain no output and keep retrying the stale create-first policy.

This change makes replacements delete-first only when the requested fixed private IP matches the current or previously requested address. It also re-plans interrupted replacements with missing output, allowing a later deploy to recover without manual state edits.

Create-first behavior remains unchanged when:

- no fixed private IP is requested;
- the replacement moves to a different private IP.

Update and no-op behavior also remain unchanged.

### Verification

- `bun alchemy-test test/AWS/EC2/Instance.diff.test.ts` passed all 6 cases.
- `bun tsc -b packages/alchemy/tsconfig.json` passed.
- Upstream Check workflow passed.
- A downstream production deployment recovered three interrupted fixed-IP EC2 replacements. Each old instance was deleted before its replacement was created, all three replacements reached `created`, and the following full plan was a no-op.


Dawsson:
@sam-goodwin this is ready for review. It covers the fixed-private-IP create-first conflict and recovery from the interrupted no-output state. The focused test, upstream Check workflow, and a three-node downstream production recovery all pass.

sam-goodwin:
Superseded by [#1667](https://github.com/alchemy-run/alchemy/pull/1667).

GitHub denied maintainer pushes to `withharbor/alchemy`, so I published the follow-up changes in the main repository. The replacement retains @Dawsson's original commits, scopes the IP conflict check to the subnet, removes the unsafe missing-output restart shortcut, and adds real EC2 lifecycle coverage.

Closing this PR in favor of #1667. Thanks @Dawsson for the original fix.

PR #1649, opened 2026-09-15
fix(cloudflare): close request scopes for bodyless responses

### LINKED ISSUE #1648: Cloudflare: a Raw web Response ignores Effect-level status and headers on GET but not on HEAD
## Summary

When a Worker handler answers with a native `Response` (an `HttpServerResponse` whose body is `Raw` with a web `Response`), a status or headers set later through the Effect response APIs, for example by a pre-response handler registered with `HttpEffect.appendPreResponseHandler`, reach a HEAD answer but not a GET answer. The two methods can disagree on the wire for the same handler.

## What happens

`toHandledWebResponse` converts the final `HttpServerResponse` for the client. For a Raw body carrying a web `Response`, the GET path hands that native object to the client as the object it is, so the Effect status and header map are not applied to it. The HEAD path goes through `HttpServerResponse.toWeb` with `withoutBody`, which builds a fresh response from the Effect status and headers, so it reflects what the pre-response handler set.

Reproduction, measured over a real socket with Miniflare:

- handler answers `HttpServerResponse.raw(new Response("onetwo", { status: 202 }))`
- a pre-response handler sets the Effect status to 418
- GET answers 202, HEAD answers 418

## Expected

Either the Effect-level status and headers are applied to the native response on GET as well (a new `Response` over the same body with the merged status and header list, preserving header multiplicity), or the documentation states that for a Raw web response the native object is authoritative and pre-response mutations do not apply, and the HEAD path follows the same rule so both methods agree.

## Context

Related to #1639 (streamed bodies and the event scope): a consumer that owns the event scope for native responses currently answers HEAD from the adapter's own answer to keep GET and HEAD consistent, which is an adapter concern handled outside the adapter. Happy to open a pull request for whichever direction is preferred.

agcty:
Keep response conversion in Effect rather than adding a Cloudflare-specific converter. Use `fromWeb` for native HTTP responses that participate in Effect middleware:

```diff
- HttpServerResponse.raw(nativeResponse)
+ HttpServerResponse.fromWeb(nativeResponse)
```

Only transfer request-scope ownership to a response stream when a body will be sent. HEAD and 204/205/304 responses close their request scopes normally, matching Effect's web handler.

Raw native responses retain Effect's existing passthrough behavior, including WebSocket upgrades. This replaces the original custom-conversion approach; it does not change the `raw(nativeResponse)` GET/HEAD inconsistency reported in #1648.


sam-goodwin:
Thanks for reporting and reproducing this. We revised the approach after comparing it with Effect's existing response APIs.

For ordinary native HTTP responses that need to participate in Effect middleware, `HttpServerResponse.fromWeb` already imports the native status, headers, cookies, and body before middleware runs:

```diff
- HttpServerResponse.raw(nativeResponse)
+ HttpServerResponse.fromWeb(nativeResponse)
```

That lets Effect handle conversion in both directions without Alchemy maintaining a second implementation. We removed the custom adoption/merge/conversion helpers. During review, we also found that the adoption step overwrote an explicit status set before the handler returned: a requested `201` became the native response's `202`.

The remaining production change matches Effect's web handler: transfer request-scope ownership to a stream only when the response will actually send a body. Otherwise HEAD and 204/205/304 responses can leave request finalizers unexecuted.

The fixture now uses `fromWeb` for ordinary HTTP and keeps raw native passthrough separate. We deployed it to real Cloudflare and local workerd and exercised 18 GET/HEAD cases on each, including status/header overrides, cookies, streaming payloads, bodyless responses, and persisted request-finalizer records. A real WebSocket echo checks that upgrades still work. Removing the scope guard reproduced missing HEAD cleanup; restoring it passed. Workspace type-check passed, and Workers coverage finished with 365 passes and 7 TODO across two runs—the four-minute command limit interrupted the last tracing test, which then passed separately.

This is a narrower change than the original proposal: it addresses the middleware use case through `fromWeb`, but **does not change Effect's `raw(nativeResponse)` GET/HEAD inconsistency**. We updated the title and description and removed the claim to close #1648 so that distinction is explicit. The PR is left open for review.

PR #1656, opened 2026-09-15
fix(cloudflare): migrate providers to current SDK contracts

sam-goodwin:
Migrate Cloudflare providers to the current distilled SDK contracts. Depends on [distilled #598](https://github.com/alchemy-run/distilled/pull/598); pins the submodule to its commit.

Containers now use the Worker-created Durable Object namespace rather than creating an unattached application:

```ts
{
  image,
  instanceType,
  environmentVariables,
  maxInstances,
  durableObjects: { namespaceId, className },
  schedulingPolicy: "durable_object",
}
```

- Reconcile observed container state, preserve image caches, and synchronize advanced configuration through PATCH and rollout operations.
- Use TLS LIST's bare-array response and update API Shield multipart, load-balancer, Flagship, IAM, and organization adapters.
- Keep account tag enumeration separate from zone-scoped resources.
- Add lifecycle, serialization, Worker-cycle, and local-runtime regressions, with a typed probe for the observed container API rejection.


packages/alchemy/src/Cloudflare/HostnameTlsSetting/HostnameTlsSetting.ts
```diff
@@ -182,35 +181,30 @@ export const HostnameTlsSettingProvider = () =>
...
+                      ? []
+                      : [
+                          toAttributes(
+                            zone.id,
+                            settingId,
+                            entry.hostname,
+                            entry,
+                          ),
+                        ],
                   ),
```
Mkassabov: dont use the list endpoint; use the get endpoint....

packages/alchemy/src/Cloudflare/Organization/Organization.ts
```diff
@@ -377,8 +378,17 @@ const toAttributes = (
...
+  flags:
+    org.meta?.tenantFlags == null
+      ? undefined
+      : {
+          accountCreation: org.meta.tenantFlags.accountCreation,
+          accountDeletion: org.meta.tenantFlags.accountDeletion,
+          accountMigration: org.meta.tenantFlags.accountMigration,
+          accountMobility: org.meta.tenantFlags.accountMobility,
+          subOrgCreation: org.meta.tenantFlags.subOrgCreation,
+        },
```
Mkassabov, line 391: this is wrong; you shouldn't need to map

Mkassabov, line 391: this looks wrong.... I could be wrong but I'm pretty sure this is wrong

PR #1658, opened 2026-09-16
feat(cloudflare/access): add McpServer resource

dillionverma:
Add `Cloudflare.Access.McpServer`, the AI Controls MCP server resource that the `McpPortal` docs left as "a future resource". Wraps the `access/ai-controls/mcp/servers` create/read/update/delete/list/sync operations with the same observe → ensure → sync reconciler shape as `McpPortal`.

```ts
const docs = yield* Cloudflare.Access.McpServer("CloudflareDocs", {
  hostname: "https://docs.mcp.cloudflare.com/mcp",
  authType: "unauthenticated",
});

const internal = yield* Cloudflare.Access.McpServer("InternalTools", {
  hostname: "https://mcp.internal.example.com/mcp",
  authType: "bearer",
  authCredentials: yield* Config.Redacted("MCP_BEARER_TOKEN"),
  updatedTools: [{ name: "delete_everything", enabled: false }],
});
```

- `hostname` and `authType` are create-only on the API, so reconcile recreates the server under the same id when either changes; a `serverId` change is a replacement
- `authCredentials` / `clientSecret` are write-only `Redacted` props; rotation is detected against the previously deployed value
- A capability sync (tool/prompt discovery) runs after create and after config changes unless `sync: false`; discovery problems land on `status` / `error` and never fail the deploy
- Attributes expose the discovered `tools` / `prompts`, `status`, `authenticationStatus`, and sync timestamps

Depends on alchemy-run/distilled#605 for the typed `McpServerNotFound` / `Forbidden` tags on the server operations; the submodule bump lands once that merges.


dillionverma:
ran tests against the real Cloudflare API, without mocked responses

```text
Tests: 0 failed | 7 passed (1 files, 25.1s)
```

To run the full suite with an entitled Cloudflare profile:

```sh
CLOUDFLARE_TEST_MCP=1 pnpm --package=bun@1.3.13 dlx bun run test \
  test/Cloudflare/Access/McpServer.test.ts --profile testing --retry 0
```

PR #1680, opened 2026-09-17
fix(cloudflare-runtime/vite): relay a worker's non-upgrade response

danielgangl:
A worker that answers a WebSocket handshake with an ordinary HTTP response — `401`, `403`, `404`, a `500` thrown in the handler — had that answer thrown away: the dev server destroyed the client socket, so the browser only saw a reset connection and `WebSocket connection failed`.

```diff
 upstream.on("response", (response) => {
-  // Worker did not accept the upgrade.
-  if (!socket.destroyed) {
-    socket.destroy();
-  }
-  response.resume();
+  // Relay it: the status is the only thing that says why the handshake failed.
+  socket.write(`HTTP/1.1 ${response.statusCode} ...\r\n\r\n`);
+  response.pipe(socket);
 });
```

Hop-by-hop headers are dropped (`response` is already de-chunked, so copying `transfer-encoding` would corrupt the relay) and `connection: close` delimits the body on the hijacked socket.

Found while debugging a React Router RSC app whose agent WebSocket kept reconnecting behind `alchemy dev`. The worker was replying `403`; nothing in the browser, the dev log, or the proxy said so.


BlankParticle:
thanks for your PR
current behaviour is copied from `@cloudflare/vite-plugin`, but I don't see any reason why did they choose to do this.

Other than that, even if we returned the failure you wont see anything in the websocket response/`onerror` since websocket doesn't do http responses, you will only get the status codes in the browser, the body still won't be accessible

But I think better handling is good, I will check back on this once I have some time to review this

Issue #1689, opened 2026-09-17
`2.0.0-beta.78`: every non-interactive deploy dies with `jsxDEV is not a function` — the CLI sets NODE_ENV=production, where sigil's jsxDEV is undefined


Butch78:
Under bun, every `alchemy deploy` in a non-interactive environment (CI) fails in the CLI's own renderer before a single resource is planned:

```
TypeError: jsxDEV_7x81h0kn is not a function. (In 'jsxDEV_7x81h0kn(Status, {
        variant: "info",
        detail: initial.detail,
        children: initial.label
      }, void 0, !1, void 0, this)', 'jsxDEV_7x81h0kn' is undefined)
    at <anonymous> (alchemy/src/Cli/components/view/Runtime.tsx:841:11)
    at state_store.login (alchemy/src/Cloudflare/StateStore/State.ts:872:12)
    at state_store.init (alchemy/src/Telemetry/Metrics.ts:166:12)
    at openStackSessionUncached (alchemy/src/Alchemist/Session.ts:237:11)
    at openStackSession (alchemy/src/Alchemist/routes/stack.ts:103:26)
    at Alchemist.stack.plan (alchemy/src/Cli/commands/deploy.ts:186:34)
```

### Mechanism

1. `alchemy/bin/cli.js:205` sets `process.env.NODE_ENV = "production"` — deliberately, commented *"Set for the CLI's React renderer"*.
2. `@alchemy.run/sigil@0.0.0-alpha.11` (new in beta.78 — alpha.10 shipped no `jsx-dev-runtime` at all) vendors React's `jsx-dev-runtime`, which branches on that variable:
   ```js
   if (process.env.NODE_ENV === "production") module.exports = require_react_jsx_dev_runtime_production();
   else module.exports = require_react_jsx_dev_runtime_development();
   ```
   and React's production build defines `exports.jsxDEV = void 0`.
3. `Runtime.tsx` carries `/** @jsxImportSource @alchemy.run/sigil */` and is transpiled with the dev JSX transform, so it calls the binding that step 2 just made `undefined`.

Confirmed in isolation:

```
$ NODE_ENV=production bun -e 'import {jsxDEV} from "@alchemy.run/sigil/jsx-dev-runtime"; console.log(typeof jsxDEV)'
undefined
$ bun -e 'import {jsxDEV} from "@alchemy.run/sigil/jsx-dev-runtime"; console.log(typeof jsxDEV)'
function
```

### Why only some commands, only in CI

[...]

PR #1691, opened 2026-09-17
fix(cli): leave NODE_ENV alone when the launcher becomes the CLI under bun

Butch78:
`bin/cli.js` sets `NODE_ENV=production` for the CLI's React renderer. On a published install invoked as `bun --bun alchemy`, that assignment lands in the one process that then transpiles the CLI's own `.tsx` — and bun fixed its JSX transform from the environment at startup, before the line ran. The transpiler keeps emitting `jsxDEV` while React's module-level check now answers "production", where `jsxDEV` is `undefined`.

Every non-interactive deploy therefore dies at its first render (#1689):

```
TypeError: jsxDEV_7x81h0kn is not a function
    at src/Cli/components/view/Runtime.tsx:841
    at state_store.login (src/Cloudflare/StateStore/State.ts:872)
```

It is the same disagreement the `--tsconfig-override` a few lines below already guards against in a checkout — *"under a `preserve` tsconfig bun emits `jsxDEV` even in production, which React's production build leaves undefined"* — but a published install reaches neither remedy: it becomes the CLI in-process rather than exec'ing a fresh bun, `files` excludes `tsconfig.json` so there is nothing to point at, and a bun flag cannot be set in-process anyway.

```diff
-process.env.NODE_ENV = "production";
+const transpilesOwnTsxInProcess =
+  typeof globalThis.Bun !== "undefined" && !isDev;
+if (!transpilesOwnTsxInProcess) {
+  process.env.NODE_ENV = "production";
+}
```

The other paths keep the assignment and are unaffected: every bun child is exec'd with it already set, so its transform matches, and node published runs the built `lib/` while the Oxc loader skips `node_modules`, so none of alchemy's `.tsx` is transpiled there at all.

Verified against the published package by importing the component `Runtime.tsx` renders:

```console
$ PROBE_SET_NODE_ENV=1 bun probe.ts   # today's behaviour
TypeError: jsxDEV_7x81h0kn is not a function. (In 'jsxDEV_7x81h0kn(Text, {
$ bun probe.ts                        # this branch
OK: "• hello"
```

`alchemy state list` and `alchemy plan` both still run with the patch applied to a real `node_modules` install.

The cost is React's development build for the CLI renderer on that one path. The alternative shape — always re-exec so a fresh bun starts with the variable set — keeps the production build but gives up "becomes the CLI instead of paying for a second process". Happy to switch to that if you prefer it.

alchemy-run/sigil#1 covers the other half: a module named `jsx-dev-runtime` exporting `undefined` is a landmine for any consumer, whatever set `NODE_ENV`. Either change alone fixes this crash.


sam-goodwin:
I think we fixed this, but @BlankParticle is working on this space right now

PR #1711, opened 2026-09-18
feat(stripe): add catalog props to Product and Price

agcty:
Adds four catalog props that `Stripe.Product` and `Stripe.Price` could not express. They come from a consumer that manages its Stripe catalog as infrastructure: product ids referenced from outside the stack, Stripe Tax on both the product and the price, and marketing features rendered in a pricing table.

`Stripe.Product` gains `id`, `taxCode` and `marketingFeatures`:

```typescript
const product = yield* Stripe.Product("pro-plan", {
  id: "prod_pro_plan",
  name: "Pro Plan",
  taxCode: "txcd_10103001",
  marketingFeatures: [{ name: "Unlimited members" }],
});
```

`Stripe.Price` gains `taxBehavior`:

```typescript
  product,
  currency: "usd",
  unitAmount: 1500,
  taxBehavior: "exclusive",
});
```

### Replacement rules

- `Product.id` is create-only, so it is observed first and replaces the product when it changes. This follows what `Coupon` already does for `couponId`.
- `Price.taxBehavior` is accepted by Stripe as an update only while it is still `unspecified`. That one transition updates in place; any later change replaces the price.

### In place

`taxCode` and `marketingFeatures` update in place, and an empty `marketingFeatures` list clears them by sending `""`, matching how `images` is already cleared.

### Testing

Ran `test/Stripe/Price.test.ts` and `test/Stripe/Product.test.ts` against a real Stripe account with a test-mode key: 9 passed, 0 failed. The run caught one bug that is fixed in the last commit: when a fixed `id` changed, the ownership-tag fallback found the outgoing product during the replacement, so the new id was never created. A fixed `id` now only matches a product with exactly that id. `Coupon` uses the same lookup for `couponId` and has no test for an id change, so it may have the same behaviour; this PR leaves it alone.


sam-goodwin:
Thanks! @Mkassabov looks like we missed some features. We need a better way to audit misses like this with ai.

PR #1732, opened 2026-09-20
feat(better-auth)!: support direct upgrades to 1.7.5

sam-goodwin:
Support direct upgrades from Better Auth 1.6 to 1.7.5+ without rewriting account identities. Adapted from #1251 by @xofromthemoon; upstream restored the 1.6 account schema in 1.7.3, so the earlier issuer backfill is no longer needed.

> **BREAKING CHANGE:** Requires Better Auth 1.7.5+ and matching plugins/adapters. Drizzle consumers must adopt Relations v2 and regenerate their auth schema. Custom secondary storage must implement the required TTL and atomic-operation contracts. Applications using affected OAuth, account-selection, rate-limit, or joins APIs must update their calls as detailed below.

### Existing SQL-layer usage stays the same

```ts
const auth = yield* BetterAuth({
  emailAndPassword: { enabled: true },
});
```

SQL layers with migration support continue applying schema changes during `alchemy deploy`. Table-index changes now trigger migrations, and upstream unsafe-migration diagnostics surface as `BetterAuthMigrationError`.

Aurora Data API supports the new runtime schema validation by casting PostgreSQL's catalog table-kind column to text and preserving array-valued results such as the schema search path. No application configuration changes or validation bypass are required.

A direct 1.6 upgrade adds no `account.issuer` column and does not rewrite account IDs or password hashes. Keep the same auth secret.

### Breaking developer-experience changes

**Dependencies:** use Better Auth 1.7.5 or newer and upgrade its synchronized plugins/adapters together. The matching CLI command for 1.7.5 is:

```sh
npx auth@1.7.5 upgrade
```

Check dependency overrides as well: Better Auth 1.7.5 requires Zod `^4.5.4`. An older forced version causes HTTP 500 responses when parsing `x-forwarded-for`; this PR aligns the workspace override.

**Drizzle:** install `@better-auth/drizzle-adapter` alongside `drizzle-orm`. The layer now uses Relations v2; regenerate the auth schema and compose its `authRelations`:

```diff
-const db = drizzle(pool, { schema });
+const db = drizzle({
+  client: pool,
+  relations: { ...appRelations, ...schema.authRelations },
+});

 Effect.provide(Drizzle(db, { provider: "pg", schema }))
```

The upstream CLI config must export a `betterAuth(...)` instance using `drizzleAdapter` from `@better-auth/drizzle-adapter/relations-v2`, with the same plugins/schema options as the app. Generate through that configured adapter:

```sh
npx auth@1.7.5 generate --config ./auth.cli.ts
```

Passing `--adapter drizzle` bypasses the configured adapter and emits legacy relations. Drizzle users still review and apply migrations through their own tooling; Alchemy does not migrate this layer.

**Custom secondary storage:** `increment` now requires a TTL in seconds. Both `increment` and `getAndDelete` must be atomic; increments must not extend an existing counter's expiry.

```diff
-increment(key: string, ttlSeconds?: number)
+increment(key: string, ttlSeconds: number)
```

**Upstream application APIs:** generic OAuth moves to `signIn.social` / `linkSocial` and `/api/auth/callback/:id`; account selectors use local account-row IDs or explicit `useAccountCookie`; custom rate-limit stores use `consume`; joins move from `experimental.joins` to `advanced.database.joins`. See the [upstream upgrade guide](https://better-auth.com/docs/guides/1-7-upgrade-guide) for plugin-specific changes.

### Schema migration is not identity-data conversion

Back up and rehearse before deployment. Microsoft identifier mappings, legacy OIDC/MCP client records, SCIM reprovisioning, and duplicate device codes may require manual preparation. The wrapper does not infer identity ownership, merge users, or convert legacy plugin data.

The dedicated [1.6 → 1.7.5 upgrade guide](https://github.com/alchemy-run/alchemy/blob/feat/better-auth-1.7.5/website/src/content/docs/better-auth/upgrades/from-1-6-to-1-7.mdx) separates automatic schema changes from those manual steps. Deploy ordering protects the new host; it does not stop old application instances during incompatible database changes.

### Tutorial and integration guides

A [six-part tutorial](https://github.com/alchemy-run/alchemy/tree/feat/better-auth-1.7.5/website/src/content/docs/better-auth/tutorial) builds an Auth service, browser sign-in, Effect HTTP API middleware, and GitHub authentication. Standalone pages cover sign-in providers and database layers. Focused guides cover Alchemy configuration and secrets, Effect HTTP API middleware, secondary-storage layers, and deploy-time migrations.

Run the companion application after installing workspace dependencies:

```sh
cd examples/cloudflare-better-auth
bun run dev
```

Use Alchemy's printed URL; the example does not fix a development port. Better Auth supplies its own OAuth callback error page.


patrikduksin:
👀 comment to sub for pr

PR #1750, opened 2026-09-21
fix(cloudflare): expose ref on Container so another stack can reference it

### LINKED ISSUE #1752: feat(Cloudflare.Container `ref`): Cloudflare.Container is missing ref, so a Container-owning stack cannot be split into a master + slices
Cloudflare.Container is missing ref. ref is useful. add ref. thank you.

slop below

## Summary

`Cloudflare.Container.ref` is missing from both the declared type and the runtime value, so a container cannot be referenced from another stack:

```ts
Cloudflare.Container.ref("App", { stack: "my-master-stack" })

// TS2339: Property 'ref' does not exist on type
// 'ResourceClassLike<ContainerApplication> & { <DOShape, Id, PropsReq>(…): Decl<…> }'
// and at runtime: TypeError: Cloudflare.Container.ref is not a function
```

## Why this matters

Cross-stack references are what let a program be split into a master stack plus per-service slices — the documented "reference another stack" pattern, available on `R2.Bucket`, `D1.Database`, `KV.Namespace` and `Worker`:

```ts
const bucket = Cloudflare.R2.Bucket.ref("Bucket", { stack: "my-master-stack" });
```

A container is the resource for which the usual workaround is unavailable:

- A Durable Object binding has to be declared on a Worker in the stack that owns the container application, so a stack holding a Container is necessarily where every Worker that binds it must live. There is no "declare it in the other stack instead".
- Without a reference, that other stack cannot reach the application at all.

So once a Container is involved, the recommended layout either cannot be expressed, or the Containers all stay in the largest stack — defeating the point of the split.

## Cause

`Cloudflare.Container` is hand-assembled rather than built through `Platform(...)` — every other effect-native resource (`Worker`, `Railway.Service`, `Fly.Service`, `AWS.Lambda.Function`, …) is. The hand-written wrapper `Object.assign`ed only `{ Type }` and never spread the `ResourceClass` that `ContainerPlatform` builds.

`Platform` internally calls `Resource(type, ...)` and spreads the resulting `ResourceClass` onto its return value (`Platform.ts:691-704`: `instance = Object.assign(constructor, resource, …)`), so `ref`, `Type`, `Provider`, `Self`, `Aliases` all come for free — `Resource.ref` lives on `ResourceClass.Service` (`Resource.ts:534-538`). `ContainerPlatform` does get all of this; the static `Cloudflare.Container` value just doesn't.

## Fix

Spread `ContainerPlatform` onto the static `Container` value instead of `Object.assign`-ing `{ Type }` literally. The custom dispatcher survives unchanged because the per-id return values carry Container-specific markers (`~alchemy/Container/Binding`, `~alchemy/Container/ClassName`, `~alchemy/Container/Shape`, `Application`) that `Platform`'s generic constructor doesn't add.

The one-liner `Pick<ResourceClass<ContainerApplication>, "ref"> &` (the form `Worker` uses) does not work alone: type-only. Verified by checking out the pre-PR `Container.ts`, adding the `Pick<>` to the declared intersection, and running `ContainerRef.test.ts` — `tsc --noEmit` passes, the test fails at runtime (`TypeError: Cloudflare.Container.ref is not a function`).

## Scope

Both Container declaration paths verified: image/context (the `.main`/`.image` form) and effect-native `.make()` Layer variant. Three integration tests in `Container.test.ts` (remote image, env threading, request proxying through the DO) pass after the fix.

Cyberistic:
Adds `ref` to cloudflare containers. Fixes #1752.

**Not established:** *why* `Container` omits it where `Worker` includes it. Both are built by `Platform(...)`, and neither spreads `ResourceClass`'s shared `Service`. `Container.ref` apresent in one import path and absent in another, and could not reconcile the two. That is why this patch implements `ref` explicitly rather than assuming it is inherited. Feel free to close or edit if my implementation is wrong. I might be missing something.

slop below:

## What

`Cloudflare.Container.ref` is missing from the declared type, so a container cannot be referenced from another stack:

```ts
Cloudflare.Container.ref("App", { stack: "my-master-stack" })

// before
// TS2339: Property 'ref' does not exist on type
//   'ResourceClassLike<ContainerApplication> & { <DOShape, Id, PropsReq>(…): Decl<…> }'
```

This adds `ref` to both the declared intersection and the runtime object, and a test that exercises it.

## Why it matters

Cross-stack references are what let a program be split into a master stack plus per-service slices — the documented "reference another stack" pattern, available on `R2.Bucket`, `D1.Database`, `KV.Namespace` and `Worker`:

```ts
const bucket = Cloudflare.R2.Bucket.ref("Bucket", { stack: "my-master-stack" });
```

A container is the resource for which the *usual workaround is unavailable*:

- A Durable Object binding has to be declared on a Worker in the stack that owns the container application, so a stack holding a Container is necessarily where every Worker that binds it must live. There is no "declare it in the other stack instead".
- Without a reference, that other stack cannot reach the application at all.

So once a Container is involved, the split either cannot be expressed, or the Containers all stay in the largest stack (defeating the point). That is the practical consequence, and it is why I think this is worth fixing rather than documenting.

## What I verified, and what I did not

I want to be precise about the cause, because my first reading of it did not survive checking.

**Verified:**

- `.ref` is absent from the declared type (the `TS2339` above).
- An explicit implementation makes the reference work end to end — `pnpm test Cloudflare/Containers` passes, and fails (`Container.ref is not a function`) without the source change.
- `Cloudflare.Worker` is the precedent for a hand-assembled resource that does expose `ref`.

## Alternative

`Worker` declares the same member with:

```ts
Pick<ResourceClass<Worker>, "ref"> &
```

If `Container.ref` is *meant* to come from `Platform` and only the type is missing, that one-line form is the better fix and I will rework this patch — I tried it, and in the test environment `Container.ref` was `undefined`, which is why I did not go that way. If you can tell me which of the two is intended, I will match it.

## Scope note

I checked `Cloudflare.Container` as declared for an image/context container (the `.main`/`.image` path). I did not exercise the effect-native `.make()` Layer variant, so if it reaches a different code path a maintainer may want to widen this — happy to adjust.


Cyberistic:
Replaced the original `Object.assign(dispatcher, {Type, ref})` literal with the structural fix: `Object.assign(dispatcher, ContainerPlatform)`. `ContainerPlatform` is `Platform(...)` and already spreads the `ResourceClass` it builds onto its return value, so `ref` (and `Type`, `Provider`, `Self`, `Aliases`) come for free — same shape `Worker` gets by being `Platform(...)` directly.

Verified empirically that the previously-discussed alternative — `Pick<ResourceClass<...>, "ref"> &` alone, the form `Worker` uses — does *not* work here: `tsc --noEmit` passes with it added to the pre-fix declaration, but `Container.ref` is still `undefined` at runtime, so `ContainerRef.test.ts` fails with `TypeError: Cloudflare.Container.ref is not a function`. Type-only fix; the runtime `Container` value needs `ref` to actually be on it, and `Platform(...)` is what attaches it.

`Container.Decl.Any` consumers (StartContainer.ts, Worker.ts, WorkerAsyncBindings.ts, WorkerBinding.ts) still typecheck — no downstream breakage.

